PR #100, opened 2026-06-14
fix(memory): recall token-budget truncation signal + rank-order retain (#94)

### LINKED ISSUE #94: memory.recall: token budget silently reduces result count below requested limit (greedy retain also skips ranks)
## Behavior

crates/khive-pack-memory/src/handlers/recall.rs (~448-459): after `ranked.truncate(limit)`, a token-budget filter retains hits only while cumulative `note.content.len()` stays under `default_token_budget(4000) × chars_per_token(4) = 16,000` chars.

Consequences:
1. **Silent under-fill**: `recall(limit=200)` returns ~21-41 results (mean note ~502 chars) with no indication that 160+ hits were dropped for budget, not relevance. Callers cannot distinguish 'only 30 relevant memories exist' from 'budget cut'.
2. **Rank skipping**: the `retain` is greedy — a long note that exceeds remaining budget is dropped while shorter, lower-ranked notes after it still pass (`total_chars` not incremented on rejects). The returned list can omit rank N but include rank N+3.
3. Default `limit=10` can return fewer results with no `truncated_for_budget: N` signal in the response).
- Greedy retain should at minimum be a clean prefix cut (stop at first overflow) so ranking semantics survive.
- Per-call override exists via config param — document it prominently in recall help either way.

Related: #80 (retrieval quality measurement loop), #93 (embed intent prefixes).

ohdearquant:
## Summary

Fixes two correctness bugs in the `memory.recall` token-budget post-filter (#94):

- **Rank skipping** — the greedy `retain` closure dropped items that exceeded the remaining
  budget while keeping shorter, lower-ranked items. Results were neither rank-ordered nor a
  clean top-N prefix.
- **Silent truncation** — when the token budget capped results below the requested `limit`,
  callers had no way to tell truncation had occurred.

## Changes

**`crates/khive-pack-memory/src/handlers/recall.rs`**

- Replaced the `retain` block with a prefix-cut loop: it breaks at the first item that overflows
  the remaining budget and truncates there. The returned set is always the top-N items in rank
  order (rank-skip fix — applies to every recall path).
- Budget metadata (`budget_capped`, `truncated_for_budget`) is surfaced in the **verbose**
  response path only (`include_breakdown` / multi-model object). The **non-verbose** path stays
  a bare JSON array — capped or not — preserving the stable list contract that Python consumers
  depend on (`runtime/src/pack.rs` `.as_array()`; `smoke_test.py` / khive-contract
  `isinstance(hits, list)`). Adding the signal to the bare array would break those consumers, so
  it lives in the verbose shape that already returns an object.

**`crates/khive-pack-memory/tests/integration.rs`** — three regression tests (all fail without
the fix):

| Test | Checks |
|---|---|
| `test_recall_budget_capped_surfaces_signal` | Verbose response carries `budget_capped: true`, `truncated_for_budget > 0` when capped |
| `test_recall_no_budget_cap_returns_plain_array` | Non-verbose path returns a bare JSON array |
| `test_recall_budget_truncation_preserves_rank_order` | Returned item is the top-ranked note, not a shorter lower-ranked substitute |

## Scope note (#94)

The truncation signal is **verbose-only by design**. The non-verbose `memory.recall` contract is
a bare array and several Python consumers assert list shape, so the signal cannot ride in that
shape without a breaking change. Rank-order correctness is fixed unconditionally. If #94 needs a
truncation signal in the *default* (non-verbose) surface, that is a contract change (separate ADR
+ consumer migration) — flagged for the maintainers, not done here.

## Quality gates (the design design review round 2, all RC 0)

```
cargo test -p khive-pack-memory     RC 0  (153 unit + 63 integration)
cargo clippy -p khive-pack-memory   RC 0
cargo fmt --all -- --check          RC 0
smoke_test.py (MEMORY)              RC 0  — "memory.recall — 1 hit(s)", PASSED
khive-contract memory + verb tests  RC 0  — 2 passed
```

Went through two internal design reviews (round 1 flagged the non-verbose shape
regression → fixed by reverting non-verbose to the bare array; round 2 passed clean).

## Merge note

Held pending maintainer review before merge.

Closes #94

### DIFF
--- a/crates/khive-pack-memory/src/handlers/recall.rs
+++ b/crates/khive-pack-memory/src/handlers/recall.rs
@@ -461,15 +461,21 @@ impl MemoryPack {
         ranked.truncate(limit);
 
         let token_budget_chars = scoring_cfg.default_token_budget * scoring_cfg.chars_per_token;
+        let pre_budget_count = ranked.len();
         let mut total_chars = 0usize;
-        ranked.retain(|sn| {
+        let mut budget_cutoff: Option<usize> = None;
+        for (i, sn) in ranked.iter().enumerate() {
             let entry_chars = sn.note.content.len();
             if total_chars + entry_chars > token_budget_chars {
-                return false;
+                budget_cutoff = Some(i);
+                break;
             }
             total_chars += entry_chars;
-            true
-        });
+        }
+        if let Some(cut) = budget_cutoff {
+            ranked.truncate(cut);
+        }
+        let budget_capped = ranked.len() < pre_budget_count;
 
         let is_verbose = cfg.include_breakdown || p.include_breakdown.unwrap_or(false);
         let full_content = p.full_content.unwrap_or(true);
@@ -541,11 +547,18 @@ impl MemoryPack {
                     json!({ "model": model, "hits": hits_json })
                 })
                 .collect();
+            let truncated_for_budget = if budget_capped {
+                pre_budget_count - results.len()
+            } else {
+                0
+            };
             return to_json(&json!({
                 "results": results,
                 "candidates": {
                     "vector_candidates_per_model": per_model,
                 },
+                "budget_capped": budget_capped,
+                "truncated_for_budget": truncated_for_budget,
             }));
         }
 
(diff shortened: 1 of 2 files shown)


ohdearquant:

PR #124, opened 2026-06-15
feat(vamana): add insert and consolidate (ADR-052 §2 PR3)

ohdearquant:
## Scope

ONLY `crates/khive-vamana/`. No other crate. No persistence-layer changes (those are PR4).

Implements the two remaining lifecycle verbs from ADR-052 §2:

- `insert(&[f32]) -> Result<u32>` — incremental node insertion with slot recycling and back-edge pruning
- `consolidate() -> Result<Vec<u32>>` — compaction returning new_to_old ordinal remap (empty Vec = no-op)
- `ensure_owned()` — Mmap→Owned promotion called at the top of both mutating functions (GAP-1)

## Resolved Lambda Gaps

**GAP-1 (ensure_owned)**: `fn ensure_owned(&mut self) -> Result<()>` promotes `VectorStorage::Mmap` to `Owned` via a one-time copy before any mutation in `insert()` or `consolidate()`. Called as the first statement of both.

**GAP-2 (consolidate signature)**: `pub fn consolidate(&mut self) -> Result<Vec<u32>>` returns `new_to_old` (new_to_old[new_ordinal] = old_ordinal). No-op fast path returns `Ok(Vec::new())` signaling "ordinals unchanged." Callers must apply the remap to rebuild external id→ordinal tables.

## 5-Leg Gate

| Leg | Command | RC |
|-----|---------|-----|
| 1 | `cargo check --workspace` | 0 |
| 2 | `cargo test --test lifecycle` (32 tests) | 0 |
| 3 | `cargo clippy --workspace --all-targets -- -D warnings` | 0 |
| 4 | `cargo fmt --all -- --check` | 0 |
| 5 | lifecycle test count >= 28 | 32 |

## Test Coverage

17 new lifecycle integration tests added (total: 32, floor was 28). Covers:
insert append/recycle paths, ops_since_consolidation tracking, consolidate state reset,
ordinal remap integrity, no-stale-adjacency after recycle, recycle-then-consolidate ordering,
dimension/finite/overflow error preflight, search still works after error, Mmap→Owned promotion,
and the two oq quality benchmarks (Wolverine repair + consolidate memory savings).

## ADR Update

`docs/adr/ADR-052-ann-production-lifecycle.md` updated in-branch to document:
- `insert()` returns `u32` ordinal; `ensure_owned()` called first
- `consolidate()` returns `Result<Vec<u32>>` (new_to_old remap or empty Vec for no-op)
- Ordinals are NOT stable across a non-no-op `consolidate` (epoch boundary for callers)
- Migration path step 3 annotated as implemented in PR3

## SPEC Deviations

Test 16 (`double_free_slot_guard`): `free_slots` is private — direct corruption from outside the module is impossible. The test validates the observable invariants (double-tombstone rejection, state consistency post-error) which exercise the same guard logic specified in SPEC §5 T16.

### DIFF
--- a/crates/khive-vamana/src/index.rs
+++ b/crates/khive-vamana/src/index.rs
@@ -14,7 +14,7 @@ use crate::{
     config::VamanaConfig,
     distance::l2_squared,
     error::{Result, VamanaError},
-    graph::{is_tombstoned_bit, robust_prune_inner, sort_dedup_u32, VamanaGraph, VisitedSet},
+    graph::{is_tombstoned_bit, robust_prune_inner, row, sort_dedup_u32, VamanaGraph, VisitedSet},
 };
 
 const METADATA_MAGIC: &[u8; 8] = b"KHVVAMM1";
@@ -728,10 +728,7 @@ impl VamanaIndex {
     /// Returns `Err` without mutating state if the vector is non-finite, wrong dimension,
     /// or would push `num_vectors` past `u32::MAX`.
     pub fn insert(&mut self, vector: &[f32]) -> Result<u32> {
-        // GAP-1 resolution: promote Mmap to Owned before any vector mutation.
-        self.ensure_owned()?;
-
-        // Preflight — validate before any state change.
+        // Preflight — validate before ANY state change (including Mmap promotion).
         if vector.len() != self.dimensions {
             return Err(VamanaError::DimensionMismatch {
                 expected: self.dimensions,
@@ -744,6 +741,10 @@ impl VamanaIndex {
                 count: self.num_vectors,
             });
         }
+
+        // GAP-1 resolution: promote Mmap to Owned before any vector mutation.
+        // Only reached when preflight passed — a rejected insert never touches Mmap.
+        self.ensure_owned()?;
 
         // Slot assignment: recycle or append.
         let ordinal: u32;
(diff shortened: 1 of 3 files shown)


ohdearquant:
Parking this PR pending a maintainer's design decision.

Additional design review converged on a real design fork, not an implementation bug. At graph saturation (every node already at max_degree — exactly the max_degree=1 repro), an incremental insert cannot give the new node an inbound edge without either:
- dropping an existing edge (orphan/disconnect risk; reachability is global, so local reasoning cannot tell which drops are safe), or
- overflowing degree (the v1 loader strictly rejects degree > max_degree at index.rs:128-130 / 598-600 / 1534-1536 and graph.rs:229, with a deliberate contract test at index.rs:2437-2442 — so a persisted overflowed graph becomes unloadable).

The current HEAD (edbb1fd, "Option E") intentionally overflows the medoid (index.rs:892,916); its saturated test passes only by saving the pre-insert graph, never the overflowed post-insert one. The overflow also accumulates on the entry-point hub, so it is not a rare +1.

Decision memo with the counting argument and 3 options (recommended: B, FreshDiskANN-style cheap best-effort insert + connectivity repair folded into the periodic consolidate) is retained in supporting notes. Converting to draft until the fork is resolved.

Issue #101, opened 2026-06-14
fix(recall): multilingual embedding model gated behind CJK-only detection — non-CJK non-English queries fall back to English primary


ohdearquant:
## Summary

`memory.recall` routes queries to the secondary embedding model (`paraphrase-multilingual-MiniLM-L12-v2`, ~50 languages) **only when the query is CJK-dominant**. Every non-CJK non-English query (French, Spanish, German, Arabic, Russian, Hindi, …) is embedded by the English primary (`all-MiniLM-L6-v2`) instead of the multilingual model. The multilingual model is reachable for ~CJK only, not for the other ~45 languages it covers.

## Current behavior (evidence)

- `crates/khive-pack-memory/src/handlers/recall.rs:113`
  ```rust
  let is_cjk = scoring_cfg.enable_cjk_routing && contains_cjk(query_trimmed);
  ```
  `is_cjk` is the *only* gate for selecting the multilingual model.
- `crates/khive-pack-memory/src/handlers/common.rs:648` — `} else if is_cjk {` routes to `scoring_cfg.cjk_model` and sets `cjk_routed = true`.
- `contains_cjk` (`crates/khive-pack-memory/src/scoring.rs:349-355`; also `khive-text/src/lang.rs:20-26`) returns true only when **>15% of characters are CJK** (Han / Hiragana / Katakana / Hangul). Latin-script non-English text never trips it.
- Config fields bake the wrong assumption into their names: `enable_cjk_routing`, `cjk_model`, `cjk_routed` (`scoring.rs:268-274`). The field is really "the multilingual model," but routing treats it as "the CJK model."

## Why this is wrong

`paraphrase-multilingual-MiniLM-L12-v2` spans ~50 languages. Gating it behind CJK detection means a Spanish or Arabic query is embedded by an English-only model, degrading recall for the majority of the languages the multilingual model exists to serve. Not the intended design (2026-06-14): the second model was meant to handle multilingual content broadly, not CJK alone.

## Fix direction (later — do NOT implement yet)

Select the multilingual model for **any query that is not confidently English/Latin-monolingual**, not only CJK:
- Broaden detection: route to multilingual when the query contains non-ASCII / non-Latin script, or fails a lightweight English-likelihood check — not just `contains_cjk`.
- Rename the `cjk_*` config + plumbing to `multilingual_*` to stop encoding the CJK-only assumption.

## Considerations

[...]

Issue #10, opened 2026-06-07
Entity/note update re-embeds only the default model — fan out to all registered engines


ohdearquant:
## Problem

The live write paths and the multi-engine `kkernel reindex` (PR #8) disagree on multi-model vector coverage:

| Path | Models embedded |
|------|-----------------|
| `create_note` (`crates/khive-runtime/src/operations.rs:1197-1221`) | **all** registered engines (fan-out) |
| `create_entity` (`crates/khive-runtime/src/operations.rs:364`) | **default only** (`config().embedding_model` + `self.embed`) |
| `update_entity` → `reindex_entity_fts` (`crates/khive-runtime/src/curation.rs:387`) | **default only** |
| `update_note` → `reindex_note` (`crates/khive-runtime/src/curation.rs:438`) | **default only** |
| `kkernel reindex` | **all** registered engines |

With ≥ 2 engines registered (today: `all-minilm-l6-v2` + `paraphrase-multilingual-minilm-l12-v2`), **editing an entity or note refreshes only the default model's vector**, leaving the other engine's vector for that record stale until a full `kkernel reindex`. Entity create is single-model too. FTS5 stays correct (triggers); only the vector/ANN coverage drifts.

## Fix

Make every entity/note write fan out to **all** `registered_embedding_model_names()` — parity with `create_note` and the reindex contract. Re-embed and replace each engine's vector for the record on update (and entity create).

Suggested: extract the multi-model embed-and-store helper that `create_note` uses and share it across `create_entity`, `update_entity`, `update_note` (and the create_note path), so there is one write path that all four reuse.

## Acceptance

- After editing an entity/note with two engines registered, both per-engine vector tables hold the **updated** vector; no stale per-engine rows remain.
- A test asserts per-engine vector parity after update (not just create).

## Notes

Not in scope for the current stack (#3/#7/#8/#9) — separate branch. Relates to ADR-021 / the multi-engine reindex in #8.

PR #126, opened 2026-06-15
fix(recall): multilingual dense routing for non-ASCII queries, split from CJK FTS bypass

ohdearquant:
## Summary

Broadens `memory.recall` dense-embedding routing from CJK-only to all queries with non-ASCII
alphabetic content, and splits that decision from the (still CJK-only) FTS lexical bypass.

- Adds `needs_multilingual(text)` (`scoring.rs`): `true` when >15% of **alphabetic** characters
  are non-ASCII-alphabetic. Covers CJK, Cyrillic, Arabic, Devanagari, Hebrew, Thai, and
  accented-Latin (é, ü, ñ, …). Alphabetic-only denominator, so punctuation/digits do not flip
  routing (`Müller` and `Müller?` route identically). No new crate dependencies.
- **Splits the routing flag** so the two concerns are independent:
  - `cjk_fts_bypass = enable_multilingual_routing && contains_cjk(query)` drives the FTS lexical
    bypass (`text_gather.rs` `cjk_bypass_ranked`) and is **kept CJK-only** per #101.
  - `use_multilingual = enable_multilingual_routing && needs_multilingual(query)` drives the
    dense embedding model selection (`common.rs`).

  Before this PR a single shared flag drove both, which would have broadened the CJK FTS bypass
  to Cyrillic/Arabic/accented scripts.
- Renames the dense-routing plumbing `cjk_*` → `multilingual_*`: config `enable_multilingual_routing`,
  `multilingual_model`, and the consumer-visible response key `multilingual_routed`. `contains_cjk`
  and `cjk_bypass_ranked` are retained (genuinely CJK-specific).

## Scope — `Refs #101` (NOT `Fixes`)

This closes the **non-ASCII + accented-Latin** portion of #101. ASCII-only non-English
Latin-script queries (`bonjour le monde`, `como estas`, `ich suche einen buchhalter`) still
route to the primary model — detecting those needs a real language identifier and carries an
English-recall regression risk, so it is deferred to **#128**. The boundary is documented on
`needs_multilingual` and asserted by a known-limitation test. #101 stays open until #128 lands.

## Store-side parity

Safe to merge: the multilingual model is registered by default and `memory.remember` / note
creation fans out embeddings to all registered models, so multilingual queries hit a populated
vector space. Note UPDATE / entity reindex remain primary-only until #10.

## Gate RCs (commit `ee8be54`)

| Gate | RC |
|------|----|
| `cargo test -p khive-pack-memory -p khive-runtime` | 0 (memory 162 unit + 63 integration; runtime 409 unit + 55 integration) |
| `cargo clippy -p khive-pack-memory -p khive-runtime --all-targets -- -D warnings` | 0 |
| `cargo fmt --all -- --check` | 0 |

## Verification

Design review confirmed the store-parity improvement: all code findings were addressed, the two routing flags are never crossed, and the CJK FTS bypass remains CJK-only.

Refs #101. Refs #128.

### DIFF
--- a/crates/khive-pack-memory/docs/design.md
+++ b/crates/khive-pack-memory/docs/design.md
@@ -55,8 +55,8 @@
   that model is preferred over other registered models.
 - CJK routing only activates when a multilingual model is actually present; detection falls
   back to all registered models when none is found, ensuring CJK queries still return results.
-- The model preference is configured via `ScoringConfig.cjk_model` or by matching registered
-  model names against known multilingual substrings.
+- The model preference is configured via `ScoringConfig.multilingual_model` or by matching
+  registered model names against known multilingual substrings.
 
 ### ADR-002: Edge Ontology — Supersedes Suppression
 
--- a/crates/khive-pack-memory/src/handlers/common.rs
+++ b/crates/khive-pack-memory/src/handlers/common.rs
@@ -280,8 +280,8 @@ pub(super) struct RecallCandidateSet {
     pub(super) text_hits: Vec<TextSearchHit>,
     /// One entry per embedding model: (model_name, hits).
     pub(super) vector_hits_per_model: Vec<(String, Vec<VectorSearchHit>)>,
-    /// True when CJK routing was requested AND a multilingual model was found.
-    pub(super) cjk_routed: bool,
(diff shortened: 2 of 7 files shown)


ohdearquant:

PR #129, opened 2026-06-15
fix(comm): sender-side outbound allowlist for cross-namespace delivery (deny-all default)

ohdearquant:
## Summary

Fixes cross-namespace comm delivery. Inter-lambda mail (e.g. `lambda:leo` → `lambda:khive`) was unconditionally denied by `CrossNamespaceWrite` in `dual_write_message`, so messages never reached the recipient inbox. This adds a **sender-side outbound allowlist** that permits delivery into explicitly declared namespaces while preserving deny-by-default.

## Design

- New `allowed_outbound_namespaces` on `ActorConfig` (the sender's `khive.toml`), parsed to `Vec<Namespace>` on `RuntimeConfig`. **Empty by default = deny-all** (behavior-preserving on upgrade).
- `dual_write_message`: cross-namespace delivery (`from != to`) requires the recipient namespace to be in the sender's allowlist, otherwise `PermissionDenied`. Allowed delivery writes the inbound note via a recipient-scoped `with_namespace(recipient)` token (namespace = recipient, visible = `[recipient]`). The comm handler uses this token **append-only by construction** — it creates exactly one inbound note and never returns the token. Note this is an ordinary namespace token, **not** a type-enforced write-only capability; the enforced boundary is the sender allowlist plus the handler's single-create usage.
- Within-namespace sends (`from == to`) are unchanged.
- Daemon safety: `allowed_outbound_namespaces` is folded (sorted + deduped) into `compute_config_id`, so a client with a narrower allowlist cannot dispatch comm sends through a broader warm daemon.
- Atomic dual-write: `create_note_inner` compensates the note row, FTS document, and vectors on any post-row indexing failure, so a failed inbound create leaves no stranded row in the recipient namespace.

A recipient-side v1 was discarded after design review showed it was unenforceable: one process loads only its own `ActorConfig`, and there is no actor registry.

## Security framing

This is a **visibility boundary, not a security guarantee**. A sender that controls its own `khive.toml` can self-grant delivery into any namespace, which is no stronger than today's uncredentialed `--actor` string. A credentialed ACL belongs in a future cloud-path authorization ADR (not yet written). ADR-040 is amended to state this honestly rather than imply a guarantee the local model cannot make.

## Tests

- Cross-namespace integration tests: within-ns unchanged; deny when allowlist empty; deliver when allowed; inbound note namespace = recipient; recipient inbox sees it / sender does not; reply cross-ns; one-directional; inbound-failure rollback leaves no recipient row; `with_namespace` token behavior documented honestly.
- `compute_config_id` tests: differs when the outbound allowlist differs; stable under reordering/dedup.
- TOML wiring test: `actor.allowed_outbound_namespaces` parses into `RuntimeConfig`.
- 3 curation cross-namespace regression tests pass unchanged.
- Full gate green: `check`, `clippy -D warnings`, `test`, `fmt --check`.

## Open decision (the maintainers)

Unilateral (sender-only opt-in, as in this PR) vs bilateral (recipient must also opt in). Unilateral is recommended for the OSS local path; bilateral is future cloud work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ htmlcov/
 
 # Rust
 target/
+.cargo-target/
 **/*.rs.bk
 Cargo.lock
 
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -59,14 +59,22 @@ pub fn compute_config_id(config: &RuntimeConfig) -> String {
         .collect();
     visible.sort();
     visible.dedup();
+    let mut outbound: Vec<String> = config
+        .allowed_outbound_namespaces
+        .iter()
+        .map(|ns| ns.as_str().to_owned())
+        .collect();
+    outbound.sort();
+    outbound.dedup();
     format!(
-        "packs=[{}];db={};embed={};extra=[{}];backend={:?};visible=[{}]",
+        "packs=[{}];db={};embed={};extra=[{}];backend={:?};visible=[{}];outbound=[{}]",
         packs.join(","),
         db,
         primary,
         extra.join(","),
         config.backend_id,
         visible.join(","),
+        outbound.join(","),
     )
 }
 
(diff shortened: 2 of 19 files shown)


ohdearquant:

Issue #91, opened 2026-06-11
MCP daemon returns internal error with no message (multiple independent reports)


ohdearquant:
At least three independent local sessions hit khive MCP write/read failures of the form:

- "khive MCP write-back failed: daemon returned an error without a message"
- "mcp__khive__request for recall/suggest/stats: failed with daemon error and no message" (PR #83 round-2 session, ~2026-06-10T21:02 EDT)

Symptoms:
- Affects multiple verbs (recall, suggest, stats — i.e. not verb-specific)
- Error carries NO message, so the failure dimension is invisible to the caller
- Reported from parallel local sessions sharing one daemon

Two distinct problems to investigate:
1. Whatever the underlying daemon error is (possibly contention from many concurrent sessions, or stale daemon state during concurrent local builds + `make local` cycles)
2. The error path that swallows the message — an internal error must carry its cause to the client. Fail-closed AND say why (same standard as the reindex fail-closed work in #88).

Label: bug. Local sessions used fallback code paths, so the affected checks completed despite the daemon failures.

Issue #73, opened 2026-06-10
feat: kkernel should dial the daemon when it is running (avoid double-writer SQLITE_BUSY)


ohdearquant:
kkernel currently opens the DB directly, so when the khive daemon is live the two contend for the WAL — bulk admin runs (162 update batches today) hit intermittent SQLITE_BUSY needing client-side retry/backoff.

Proposal: kkernel detects a live daemon (Unix socket at ~/.khive) and routes ops through it; falls back to direct DB access only when no daemon is running. Same verb DSL, single writer, no lock contention. Admin tooling shouldn't be a second writer on a live store.

ohdearquant:
## Status: daemon-dial / double-writer SQLITE_BUSY

Two design passes (`SPEC.md` and `DESIGN.md` under supporting notes) converge on the following picture.

**Already handled — `kkernel exec`.** `kkernel exec` already dials the running daemon via `khive_mcp::daemon::forward_or_spawn` (`crates/kkernel/src/exec.rs`), falling back to a direct `KhiveRuntime` only when the daemon is absent or its config/namespace does not match. The primary bulk-admin path is therefore not a second writer.

**Server path (heavyweight half) — tracked as M-09.** Making `kkernel mcp --db` sessions dial the running daemon instead of each opening the DB directly (a thin proxy server forwarding over the existing daemon socket) is an architecture change. It needs an ADR, couples to the recovery race in #150, and carries two UX decisions for the maintainer: (1) does the proxy serve a static `tools/list` or the live daemon catalog; (2) when the daemon dies mid-session, surface an explicit error vs. silently fall back to a competing direct writer (the latter would reintroduce this bug). Pending acceptance.

**Remaining admin writer — `kkernel reindex`.** `reindex` opens the DB directly during its bulk corpus walk and cannot be routed cleanly through the per-verb socket (bulk streaming, raw-SQL snapshot invalidation, by-reference runtime). A cheap, ADR-free mitigation is to set `PRAGMA busy_timeout` on the khive-db connection so it retries on WAL contention instead of returning `SQLITE_BUSY`. Queued as a small follow-up.

**#150 (recovery concurrency race).** PR #149 routes more traffic into the crash-recovery path; review flagged the pre-existing kill/spawn lock gap and the bind-before-pid window. The #150 fix (a single recovery lock held across kill + cleanup + spawn + readiness, plus the same lock inside daemon startup) is being folded into #149.

ohdearquant:
[...]

PR #149, opened 2026-06-16
fix(mcp): surface daemon error on connection-reset mid-dispatch (#91)

ohdearquant:
## What

Fixes the remaining silent-fallback gap in the daemon forward path (#91): when
the MCP client sends a request to the warm daemon but the daemon closes the
connection **before sending a response frame** (a crash/panic mid-dispatch),
`try_forward_inner` previously classified the `read_frame` error as `NoSocket`
and silently fell back to local dispatch against a daemon that is actually
broken — producing the opaque "daemon returned an error with no message"
independent reports.

## Fix

`crates/khive-mcp/src/daemon.rs` — a `read_frame` failure after the request was
written is now classified as `ParseFailure` (not `NoSocket`). The existing
`ParseFailure` path already kills the stale daemon, cleans the socket, logs, and
re-spawns a fresh one — so a connection reset mid-dispatch now triggers
kill+respawn instead of a silent local fallback to a wedged daemon.

Builds on PR #106 (`da39378c`), which closed the malformed-frame and
`resp.error = None` cases; this closes the connection-reset-without-response case.

## Why this does not over-trigger (kill a healthy-but-busy daemon)

`read_frame` (`crates/khive-runtime/src/daemon.rs:142`) has **no internal
timeout**: it is `read_exact(len) → cap-check → read_exact(buf)`. A slow-but-alive
daemon under DB-lock contention (#73) **blocks** on the await; it does not error.
`read_frame` therefore errors **only** on (a) a genuine connection close/reset —
the daemon is dead — or (b) an oversized frame, which a healthy daemon cannot
emit because `write_frame:159` caps writes at the same `MAX_FRAME_BYTES`. Both
warrant kill+respawn. There is no transient/slow-IO path that this
mis-classifies.

## Test

`daemon::tests::try_forward_inner_returns_parse_failure_when_daemon_closes_without_response`
— binds a fake daemon socket, accepts the connection, drops it without writing a
response, and asserts the outcome is `ParseFailure`.

## Gates

`cargo test -p khive-mcp` rc=0 · `cargo clippy -p khive-mcp --all-targets -- -D warnings` rc=0 · `cargo fmt --all -- --check` rc=0

Diff: +96 / -1, 1 file (+ test).

Closes the connection-reset facet of #91.

### DIFF
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -19,6 +19,38 @@ use tokio::net::UnixStream;
 
 use crate::tools::request::RequestParams;
 
+// ── test instrumentation seams ────────────────────────────────────────────────
+//
+// These counters are only compiled in `#[cfg(test)]` builds. They allow tests
+// to assert that kill_stale_daemon_inner and spawn_daemon were called exactly
+// the expected number of times — making the recheck-under-lock test
+// fail-if-reverted: without the recheck, both counters would be non-zero even
+// when a fresh daemon is already alive.
+
+#[cfg(test)]
+pub(crate) static KILL_COUNT: std::sync::atomic::AtomicUsize =
+    std::sync::atomic::AtomicUsize::new(0);
+
+#[cfg(test)]
+pub(crate) static SPAWN_COUNT: std::sync::atomic::AtomicUsize =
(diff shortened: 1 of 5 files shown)


ohdearquant:

PR #172, opened 2026-06-19
perf(knowledge): O(N) keyset pagination for section reindex + unify embed batch

ohdearquant:
## What

Two coupled fixes to the knowledge reindex path (`kkernel reindex`), driven by the observation that section embedding backfill was slow (sections ~19/s vs atoms ~217/s).

### 1. O(N) keyset pagination (the speed fix)

The section embedding pass paged with `LIMIT ? OFFSET ?` over a `knowledge_sections ⋈ knowledge_atoms` JOIN. Deep-offset pages re-scan the JOIN from the top every time → **O(N²)** total, which is the slow long tail of a full re-embed.

Replaced with **keyset pagination** on the `id` PRIMARY KEY:

```sql
... WHERE s.namespace = ?1 [atom] [null] AND s.id > :last_id ORDER BY s.id LIMIT :n
```

One forward walk of the `id` B-tree → **O(N) total**. The cursor advances past *every* page row (embedded, skipped-blank, or failed), so each row is attempted at most once and the loop terminates in **both** `drop_existing` and `--keep-existing` modes.

### 2. Unify the embed batch (the "misleading flag" fix)

`--batch-size` previously only controlled DB pagination; the actual embed batch was a hardcoded re-chunk (`EMBED_BATCH = 32`). Now `--batch-size` **is** the embed batch — one `embed_document_batch` call per page — for sections *and* atoms. This matches the entity/note path, which already embedded in `batch_size` units.

- Default raised `32 → 128` (`DEFAULT_EMBED_BATCH`).
- `kkernel reindex --batch-size` default moves `100 → 128`; help text clarified.

## Scope of behavior

- **Sections**: keyset walk; `--keep-existing` filters `embedding IS NULL` inline.
- **Atoms**: ANN rebuild path unchanged (`rebuild_ann` independent of keep-existing).
- No schema change, no ADR change. `id TEXT PRIMARY KEY` (BINARY collation); UUIDs are lowercase-hex so `id >` matches `ORDER BY id`.

## Verification

- Verification: 7 load-bearing claims proven incl. termination both modes, no-skip completeness, param numbering, whole-page embed failure paths, collation.
- `cargo fmt --all -- --check` clean · `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `khive-pack-knowledge` + `kkernel` suites: **277 tests, 0 failed** — incl. regression `section_keep_existing_failures_terminate_and_report`.

### DIFF
--- a/crates/khive-db/sql/004-fts-consolidation.sql
+++ b/crates/khive-db/sql/004-fts-consolidation.sql
@@ -1,0 +1,33 @@
+-- 004-fts-consolidation.sql
+-- Consolidate per-namespace FTS5 tables into single shared tables.
+-- Strategy: create unified tables (namespace UNINDEXED column).
+-- Per-namespace partition cleanup is performed at runtime (namespace-agnostic sweep),
+-- not in this SQL, to avoid hardcoding namespace names.
+-- Repopulation: run `kkernel reindex --no-knowledge` after this migration.
+-- Out of scope: fts_knowledge, fts_sections (knowledge-pack corpus).
+
+-- Unified entity FTS table (one table for all namespaces; namespace is a filter column)
+CREATE VIRTUAL TABLE IF NOT EXISTS fts_entities USING fts5(
+    subject_id UNINDEXED,
+    kind UNINDEXED,
+    title,
+    body,
+    tags UNINDEXED,
+    namespace UNINDEXED,
+    metadata UNINDEXED,
+    updated_at UNINDEXED,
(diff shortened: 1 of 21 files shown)


ohdearquant:
Superseded by a clean branch. #172 was accidentally cut from a diverged local `main` (8 unpushed FTS-ANN consolidation commits) and carried those as conflicts against the merged ADR-062 #171. The actual keyset change is unchanged; re-opened cleanly on top of current origin/main.

PR #185, opened 2026-06-19
docs(knowledge): document compose explain=true flag (#183 follow-up)

ohdearquant:
Docs-only follow-up to #183 (which slimmed the `knowledge.compose` default output and gated the scoring telemetry behind a new `explain=true` flag).

- **`marketplace/knowledge/README.md`** — the compose param row now lists `explain?` (boolean, default false) and documents the output contract: default returns `markdown` + compact `atoms`/`domains` + `count`; `explain=true` adds `sections[]` with per-section `breakdown`, `section_count`, and inline section scores in the markdown. (deno re-padded the table column widths, hence the alignment churn — only the compose row content changed.)
- **`docs/adr/ADR-048-knowledge-section-profiles.md`** — notes that the section manifest the (deferred) `PostToolUse` feedback hook consumes is now opt-in: the hook must call `knowledge.compose` with `explain=true`.

`deno fmt --check docs/` passes. No code changes.


ohdearquant:
Closing. The primary target marketplace/knowledge/README.md was deleted by #204 (06a3ae51), which collapsed the 7 per-pack marketplace plugins into one marketplace/khive umbrella. This PR edits that now-deleted file, which is the modify/delete conflict behind the DIRTY state; merging would resurrect a file the umbrella collapse intentionally removed. The documented explain=true flag is real and accurate (knowledge schema.rs:289, search.rs:1241), and the ADR-048 opt-in note is worth keeping, so I am re-capturing both into the current doc locations (umbrella knowledge skill + ADR-048) instead of merging this stale branch. (2026-06-21)

Issue #150, opened 2026-06-16
daemon crash-recovery can double-spawn / race socket file under concurrent ParseFailure (#73-adjacent)


ohdearquant:
## Summary

The MCP daemon crash-recovery path can double-spawn daemons and race on the
socket file when multiple clients concurrently observe the same crashed daemon.
Found while reviewing PR #149 (which hardens the silent-fallback path for
#91 and thereby widens how often crash-recovery runs).

## Evidence

- `crates/khive-mcp/src/daemon.rs:342` routes `ParseFailure` into stale-daemon
  recovery; `:346` calls `kill_stale_daemon()`, which holds the recovery flock
  only while killing/removing files (`:293`–`:321`). After the lock is released,
  `forward_or_spawn` falls through to an **unguarded** `spawn_daemon()` at `:413`.
  `spawn_daemon()` starts another `kkernel mcp --daemon` with no recovery lock
  (`:176`–`:192`).
- Daemon-side self-dedupe is not atomic: `run_daemon` binds the socket at
  `crates/khive-runtime/src/daemon.rs:323` and only writes the PID file at `:329`,
  while a racing startup removes an existing socket when it cannot prove a live
  daemon via PID+connect (`:396`–`:418`). That bind-before-pid window lets one
  freshly spawned daemon have its socket unlinked by another startup → redundant
  or unreachable daemon processes.

## Why it matters

Kill/cleanup is serialized; spawn/readiness is not. Under N concurrent clients
hitting a crashed daemon — exactly the multi-client `kkernel mcp` scenario behind
the #73 double-writer contention — recovery can leave a wedged/orphaned daemon,
which is the same class of failure #91 is about.

This is **pre-existing** (the spawn path and bind-before-pid window are
independent of PR #149); PR #149 increases how often the recovery path executes.

## Suggested fix

Hold a single recovery/startup lock across kill → socket cleanup → spawn →
readiness probing, or move that lock into daemon startup so cleanup/bind/write-pid
is atomic relative to other daemon starts. Add a concurrent-recovery test that
drives multiple clients through `ParseFailure` and asserts exactly one usable
daemon survives.

## Disposition

Group with the daemon-socket-lifecycle work (M-09 / #73). Should be prioritized:
PR #149 widens reachability of this race in the very multi-client environment we
run in. Sequencing rationale: #149 is still strictly better than the status quo
(silent fallback to a crashed daemon), so ship #149 + prioritize this.

ohdearquant:
[...]

PR #191, opened 2026-06-20
fix(mcp): route explicit version_mismatch to recovery; test dispatch error propagation

### LINKED ISSUE #91: MCP daemon returns internal error with no message (multiple independent reports)

### LINKED ISSUE #156: daemon: explicit protocol version-mismatch responses bypass recovery (no auto-upgrade)
## Context

Follow-up from #149 (MCP daemon recovery). #149 reverted the speculative `PROTOCOL_VERSION` 1→2 bump because the probe-sentinel classifier closes the r4 probe fail-open on its own, so this is latent until the next real protocol bump.

## The gap

When `PROTOCOL_VERSION` is eventually bumped, a warm **old** daemon answering a **newer** client returns `ok=false, version_mismatch=true, daemon_protocol_version=<old>`. The client classifier does **not** route that to recovery:

- `try_forward_inner` returns `ForwardOutcome::ProtocolMismatch` only when `daemon_protocol_version != PROTOCOL_VERSION && !frame.version_mismatch` — `crates/khive-mcp/src/daemon.rs:146-163`.
- An explicit `version_mismatch=true` response therefore takes the generic `Response` arm → `map_response` → a hard MCP error (`crates/khive-mcp/src/daemon.rs:181-191`), never reaching the `kill_and_respawn` paths.

Daemon-side mismatch branch that emits this frame: `crates/khive-runtime/src/daemon.rs` (mismatch response, ~`:268-288`).

Net effect at a future bump: the new client surfaces a protocol-mismatch error and **leaves the stale old daemon in place** instead of killing + respawning it (no auto-upgrade) — exactly the warm-old-daemon case `make local` produces.

## Fix

Route explicit `version_mismatch=true` responses with `daemon_protocol_version != PROTOCOL_VERSION` through the same stale-daemon recovery path as `ForwardOutcome::ProtocolMismatch`; return a hard error only if the **respawned** daemon still reports the wrong version. Add a regression test: a fake old-version response (`version_mismatch=true`, `daemon_protocol_version=<old>`) asserts `kill_and_respawn` is attempted and the stale daemon is not left serving errors.

## Notes

- Do this **with** the next protocol-evolution change (it has no effect until a real bump) — see the M-09 daemon-dial ADR escalation.
- Refs #149.

ohdearquant:
## Summary

- Closes #91 — add end-to-end regression test anchoring error-message propagation through `forward_or_spawn`
- Closes #156 — route explicit `version_mismatch=true` from stale old daemon through `kill_and_respawn` (auto-upgrade)
- #150 — closed separately with evidence (fixed in 89db3ff6 already on main)

## Changes

**`crates/khive-mcp/src/daemon.rs`** — one logical fix + three new tests:

### Fix: #156 — explicit protocol-version-mismatch responses now trigger recovery

`try_forward_inner` previously detected stale daemons via:
```rust
frame.daemon_protocol_version != PROTOCOL_VERSION && !frame.version_mismatch
```
This passed explicit `version_mismatch=true` responses (from old daemons that correctly identify themselves) straight through to `map_response` as a hard MCP error, leaving the stale daemon alive.

New condition:
```rust
let is_stale_daemon = frame.daemon_protocol_version != PROTOCOL_VERSION
    && (!frame.version_mismatch || frame.daemon_protocol_version < PROTOCOL_VERSION);
```
- `version_mismatch=true` + `daemon_version < PROTOCOL_VERSION`: daemon is stale → `ForwardOutcome::ProtocolMismatch` → `kill_and_respawn` (auto-upgrade)
- `version_mismatch=true` + `daemon_version > PROTOCOL_VERSION`: client is stale → `ForwardOutcome::Response` → `map_response` hard error (kill+respawn cannot fix a behind-client)

### Tests added

1. **`dispatch_error_propagates_as_non_empty_client_message`** (#91 regression anchor): drives a real `run_daemon` with `FailDispatch` (always returns `Err("forced dispatch error: ...")`) through `forward_or_spawn`. Asserts `Some(Err(e))` with `!e.message.is_empty()` and that the message contains the daemon's error string verbatim.

2. **`try_forward_inner_routes_explicit_version_mismatch_to_protocol_mismatch`** (#156): serves `version_mismatch=true, daemon_protocol_version=0` from a fake socket; asserts `ForwardOutcome::ProtocolMismatch` (recovery path, not hard error).

3. **`try_forward_inner_newer_daemon_mismatch_yields_response_not_recovery`** (#156, negative case): serves `version_mismatch=true, daemon_protocol_version=PROTOCOL_VERSION+1`; asserts `ForwardOutcome::Response` (hard error — kill+respawn cannot fix a stale client binary).

## #150 disposition

Closed separately. Evidence: commit 89db3ff6 (already on main) holds `acquire_recovery_lock()` across the full probe→kill→spawn sequence in `kill_and_respawn` (daemon.rs:454) and on the server side in `run_daemon` (runtime/daemon.rs:419). The regression test `concurrent_recovery_second_client_skips_kill_when_daemon_alive` (daemon.rs:1366) asserts `KILL_COUNT==0` when a live daemon is found under the lock.

## Test plan

- [x] `cargo check -p khive-mcp -p khive-runtime` — clean
- [x] `cargo clippy -p khive-mcp -p khive-runtime --all-targets -- -D warnings` — clean
- [x] `cargo test -p khive-mcp -p khive-runtime` — 47 daemon tests pass, 0 failed (including all 3 new tests)
- [x] `cargo fmt` — no changes needed
- [x] pre-commit hooks — all passed

### DIFF
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -151,12 +151,23 @@ async fn try_forward_inner(frame: &DaemonRequestFrame) -> ForwardOutcome {
             // it — making `map_response` accept the stale response when
             // `served_config_id` happens to match. Detect this here so the
             // caller can route it through the same kill/respawn path.
-            if frame.daemon_protocol_version != PROTOCOL_VERSION && !frame.version_mismatch {
+            //
+            // Also catch the explicit-mismatch / auto-upgrade case (#156): when a
+            // warm OLD daemon receives a request from a NEWER client it responds
+            // with `version_mismatch=true` and its own (lower) version number.
+            // `daemon_protocol_version < PROTOCOL_VERSION` means the daemon is
+            // stale — route through kill+respawn exactly like the implicit case
+            // above.  If `daemon_protocol_version > PROTOCOL_VERSION` the client
+            // binary is behind; kill+respawn cannot fix that, so let map_response
+            // return a hard error.
+            let is_stale_daemon = frame.daemon_protocol_version != PROTOCOL_VERSION
+                && (!frame.version_mismatch || frame.daemon_protocol_version < PROTOCOL_VERSION);
+            if is_stale_daemon {
                 tracing::warn!(
                     daemon_version = frame.daemon_protocol_version,
                     expected = PROTOCOL_VERSION,
-                    "daemon protocol version mismatch (old daemon, no version_mismatch flag) \
-                     — treating as stale",
+                    explicit_mismatch = frame.version_mismatch,
+                    "daemon protocol version mismatch (stale daemon) — treating as stale",
                 );
                 return ForwardOutcome::ProtocolMismatch;
             }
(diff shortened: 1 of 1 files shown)


ohdearquant:
Adversarial review: no CRITICAL/MAJOR findings.

The gating concern (unbounded recovery on a stale on-disk binary) is RESOLVED: recovery is provably bounded to exactly one respawn per `forward_or_spawn` call (daemon.rs:583-592 returns a hard error if the respawn still mismatches); both prod callers (server.rs:1067, kkernel exec.rs:80) invoke it once with no retry loop.

#156 compare direction correct (newer daemon → hard error, never recovered). #91 empty-message paths all closed (every `ok=false` sets `error:Some` or a mismatch flag → non-empty by construction). Prior #149/#150/#151 hardening preserved (the ProtocolMismatch branch acquires the same recovery lock). #156 tests are non-vacuous (verified they would fail under the old condition / inverted compare).

Only nit (optional): test 1 mocks `daemon_protocol_version=0`, which cannot co-occur with `version_mismatch=true` in practice — use `PROTOCOL_VERSION-1` to model the real auto-upgrade scenario.

47 daemon tests pass. Ready to merge.

PR #192, opened 2026-06-20
feat(memory): honor expires_at in recall; add memory.prune + memory.vacuum

### LINKED ISSUE #15: feat(memory): honor expires_at, add salience-threshold prune + vacuum
## Problem

The memory note corpus grows without bound. The `expires_at` column exists in the schema but is **never evaluated** by any code path, and there is no prune / vacuum verb. Decay is ranking-only — expired or low-salience memories are never removed, so storage grows monotonically.

## Evidence

- `crates/khive-db/sql/schema.sql:46` — `expires_at` column is stored.
- Zero references to `expires_at` anywhere in `crates/khive-pack-memory/src/` — it is written but never read.
- No prune verb in the memory pack vocabulary; decay (`decay_factor`) only affects recall ranking, not retention.

## Expected

1. `expires_at` is honored — expired memories are excluded from recall and eligible for removal.
2. A prune path (verb or `kkernel exec` maintenance op) that soft-deletes memories below a salience threshold and/or past `expires_at`.
3. A vacuum step that reclaims space from soft-deleted rows.

## Fix sketch

- Add `expires_at IS NULL OR expires_at > now` to the recall filter (view-layer — does not delete data).
- Add a `memory.prune(min_salience?, before?)` verb or a `kkernel` maintenance command that soft-deletes per threshold (curation op, deliberate — ADR-014).
- Wire a vacuum into the maintenance path.

Note the data-vs-view rule: recall filtering is a view concern (never deletes); the prune verb is deliberate curation (allowed to delete). Keep them separate.

Priority: **p1** — unbounded growth degrades recall latency and disk over time.

_Ported from old repo issue #552. Verified against current source 2026-06-07._

ohdearquant:
## Summary

- **expires_at recall filter**: Expired memories (`expires_at <= now`) are now excluded from `memory.recall` results as a view-layer filter in `load_memory_candidate_notes`. No data is deleted — this is purely a visibility gate.
- **memory.prune verb**: New curation verb (ADR-014) that soft-deletes memories below a salience threshold (`min_salience`) and/or past `expires_at`. Accepts `dry_run=true` to count candidates without deleting, and a `namespace` param (default `"local"`). Pass `before=0` to skip the expiry filter and use only salience.
- **memory.vacuum verb**: New verb that runs `SQLite VACUUM` to reclaim disk space freed by soft-deleted rows.
- Both new verbs are registered in `MEMORY_HANDLERS` (size 8 → 10) and wired into `dispatch()`.

## Files changed

- `crates/khive-pack-memory/src/handlers/prune.rs` — new file, prune + vacuum handlers
- `crates/khive-pack-memory/src/handlers/common.rs` — expires_at view-filter in `load_memory_candidate_notes`
- `crates/khive-pack-memory/src/handlers/mod.rs` — `mod prune;` declaration
- `crates/khive-pack-memory/src/pack.rs` — 2 new `HandlerDef` entries + 2 dispatch arms
- `crates/khive-pack-memory/tests/integration.rs` — 5 new integration tests

## Test plan

- [x] `test_expires_at_excluded_from_recall` — memory backdated to past `expires_at` vanishes from recall
- [x] `test_prune_by_salience` — `memory.prune(min_salience=0.5)` soft-deletes low-salience, keeps high-salience
- [x] `test_prune_expired_by_default_cutoff` — `memory.prune({})` prunes memories whose `expires_at` is past
- [x] `test_prune_dry_run` — `dry_run=true` reports `would_prune` without deleting
- [x] `test_vacuum_succeeds` — `memory.vacuum({})` returns `ok: true`
- [x] Full suite: `cargo test -p khive-pack-memory` → 81/81 pass
- [x] `cargo clippy -p khive-pack-memory --all-targets -- -D warnings` → clean
- [x] `cargo fmt --all -- --check` → clean
- [x] All pre-commit hooks passed

Closes #15

### DIFF
--- a/crates/khive-pack-memory/src/handlers/common.rs
+++ b/crates/khive-pack-memory/src/handlers/common.rs
@@ -1031,12 +1031,16 @@ impl MemoryPack {
             .iter()
             .map(String::as_str)
             .collect();
+        let now_micros = chrono::Utc::now().timestamp_micros();
         for note in batch {
             // Post-filter: ANN over-fetch may include rows from outside the caller's
             // visible namespace set. Drop them here where the note row carries its namespace.
+            // Also exclude memories whose expires_at is in the past (view-layer expiry).
+            let expired = note.expires_at.map(|e| e <= now_micros).unwrap_or(false);
             if note.deleted_at.is_none()
                 && note.kind == "memory"
                 && visible_set.contains(note.namespace.as_str())
+                && !expired
             {
                 memory_ids.insert(note.id);
                 notes_by_id.insert(note.id, note);
--- a/crates/khive-pack-memory/src/handlers/mod.rs
+++ b/crates/khive-pack-memory/src/handlers/mod.rs
@@ -2,6 +2,7 @@
 
 mod common;
 mod feedback;
+mod prune;
 mod recall;
 mod remember;
 mod sub_handlers;
--- a/crates/khive-pack-memory/src/handlers/prune.rs
+++ b/crates/khive-pack-memory/src/handlers/prune.rs
@@ -1,0 +1,160 @@
+//! Handlers for `memory.prune` and `memory.vacuum`.
+
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+
+use khive_runtime::{NamespaceToken, RuntimeError};
+use khive_storage::types::{DeleteMode, SqlStatement, SqlValue};
+
+use crate::MemoryPack;
+
+// ── Params ────────────────────────────────────────────────────────────────────
+
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub(super) struct PruneParams {
+    /// Soft-delete memories whose salience is strictly below this value.
+    /// `None` means no salience filter.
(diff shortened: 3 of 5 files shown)


ohdearquant:
Review: no BLOCKER/MAJOR findings.

Verified: `expires_at` filter is a pure view-layer post-fetch exclusion in `load_memory_candidate_notes()` — zero row mutation/deletion (data-vs-view rule honored); `memory.prune` uses `DeleteMode::Soft` (ADR-014 curation, recoverable); `memory.vacuum` runs `VACUUM` outside a txn (correct); `expires_at` already in V1 schema so no migration; both verbs registered (8→10) + discoverable via `verbs`.

Non-blocking suggestions: (1) `before=0` sentinel ("skip expiry") is a footgun vs epoch-0 — documented, but a boolean `skip_expiry` would be cleaner; (2) `test_prune_by_salience` asserts `pruned>=1` (the survival check compensates) — tighten to `==1`; (3) the vacuum verb description should warn of the exclusive-lock / whole-file-rewrite cost on the live 466K-embedding DB.

81 tests pass, clippy+fmt clean. Ready to merge.

PR #193, opened 2026-06-20
fix(db): broaden no-worse-than-stale atomic replacement guarantee (insert_batch + update)

### LINKED ISSUE #134: Vector store: broaden no-worse-than-stale atomic replacement guarantee beyond single-record `insert`
Follow-up from **PR #131** design review round 4 (Low #2). Non-blocking — no production caller in #131's affected workflows.

#131 proved the no-worse-than-stale (atomic delete+insert) guarantee for the single-record `SqliteVecStore::insert` path. Two sibling vector-write APIs do **not** share it:

- `crates/khive-storage/src/vectors.rs:118-133` — `VectorStore::update` trait default is documented + implemented as `delete` then `insert` (not atomic).
- `crates/khive-db/src/stores/vectors.rs:273-312` — `SqliteVecStore::insert_batch` does per-record `DELETE` then catches insert errors before committing the batch.

`rg "insert_batch\\(" crates -g "*.rs"` found only the vector impl, sparse impl, benches, and tests — **no production call site** in the knowledge.index / reindex / knowledge.edit paths, so this does not reopen the #131 strand.

**Suggested fix:** either override these to use the same atomic semantics, or document that callers must use `insert` for replacement on the sqlite-vec backend.

ohdearquant:
## Summary

Closes #134.

Two vector-write paths lacked the no-worse-than-stale guarantee that `SqliteVecStore::insert` provides. A failure midway through either path could leave a record with no vector at all, which is worse than the stale value it had before.

**Paths broadened:**

- `crates/khive-db/src/stores/vectors.rs` — `SqliteVecStore::insert_batch` (previously lines 273–312, now ~273–332): the per-record `DELETE` was executed unconditionally before the `INSERT` attempt; a failed `INSERT` left the record deleted but not replaced. Fixed by wrapping each record's `DELETE+INSERT` in a `SAVEPOINT vec_batch_record` / `RELEASE` / `ROLLBACK TO` pattern, exactly matching `text.rs:upsert_batch`.

- `crates/khive-db/src/stores/vectors.rs` — new `SqliteVecStore::update` override: the `VectorStore` default called `self.delete(...).await?` then `self.insert(...)` as two separate async ops with no transaction boundary. A failure on the insert left the record deleted. Fixed by overriding `update` with a single `unchecked_transaction` wrapping `DELETE+INSERT`, identical to the single-record `insert` path.

## Test plan

**Original tests (preserved):**
- `insert_batch_failed_record_preserves_prior_vector`: wrong-dims failure hits the pre-savepoint `continue` guard; proves outer COMMIT survives a per-record failure and sibling record is inserted.
- `update_failed_preserves_prior_vector`: wrong-dims failure is rejected before the transaction opens; proves the Err path returns correctly.
- `insert_batch_savepoint_rollback_on_pk_conflict_preserves_stale`: PK conflict (different namespace) exercises the SAVEPOINT/ROLLBACK TO path but the DELETE is a no-op. Covers outer COMMIT survival. NOTE: removing `ROLLBACK TO SAVEPOINT` does NOT make this test fail (DELETE was no-op).
- `insert_batch_rollback_does_not_corrupt_subsequent_record`: two-record batch, record A PK-conflicts, record B succeeds. DELETE for record A is a no-op. NOTE: removing `ROLLBACK TO SAVEPOINT` does NOT make this test fail.
- `update_pk_conflict_rolls_back_transaction_preserves_stale`: update with wrong namespace; DELETE is a no-op. NOTE: removing the transaction rollback does NOT make this test fail.

**New sentinel tests (added in this commit):**

These are the true regression sentinels. They use a `#[cfg(test)]`-only failpoint (`Arc<AtomicBool>`, thread-safe across `spawn_blocking`) that fires AFTER a successful same-namespace DELETE and BEFORE the INSERT. The stale row is genuinely gone when the error fires; only a correct rollback restores it.

**Reachability note:** Value-level failures (wrong dims, non-finite, wrong count) are all rejected before the SAVEPOINT opens. There is no natural same-namespace path through the public API to reach a post-DELETE INSERT failure. The `#[cfg(test)]` failpoint is the only way to exercise this condition in a unit test without modifying production logic.

- `insert_batch_rollback_restores_deleted_stale_after_post_delete_insert_failure`: stores `(id_X, ns:a, vec1)`, arms failpoint, calls `insert_batch` with `(id_X, ns:a, vec2)` — same namespace, correct dims, finite. DELETE removes the stale row; failpoint fires before INSERT; `ROLLBACK TO SAVEPOINT` must restore it. Asserts: `batch_exists` finds id_X, cosine similarity to vec1 > 0.999, affected=0, failed=1. **FAILS when `ROLLBACK TO SAVEPOINT` is removed. PASSES when it is present.**

- `update_rollback_restores_deleted_stale_after_post_delete_insert_failure`: same setup for the `update` path. DELETE removes the stale row; failpoint fires; `unchecked_transaction` drop must roll back. Asserts same conditions. **FAILS when transaction rollback is bypassed. PASSES when it is present.**

**Sentinel proof (observed):**
- With rollback removed (ROLLBACK TO commented out in `insert_batch`; commit-before-error injected in `update`): both sentinel tests FAIL with "ROLLBACK TO SAVEPOINT must restore the stale row" / "transaction rollback must restore the stale row".
- With rollback restored: both sentinel tests PASS.

```
test stores::vectors::atomic_replace_tests::insert_batch_failed_record_preserves_prior_vector ... ok
test stores::vectors::atomic_replace_tests::insert_batch_rollback_does_not_corrupt_subsequent_record ... ok
test stores::vectors::atomic_replace_tests::insert_batch_rollback_restores_deleted_stale_after_post_delete_insert_failure ... ok
test stores::vectors::atomic_replace_tests::insert_batch_savepoint_rollback_on_pk_conflict_preserves_stale ... ok
test stores::vectors::atomic_replace_tests::update_failed_preserves_prior_vector ... ok
test stores::vectors::atomic_replace_tests::update_pk_conflict_rolls_back_transaction_preserves_stale ... ok
test stores::vectors::atomic_replace_tests::update_rollback_restores_deleted_stale_after_post_delete_insert_failure ... ok
```

Full suite: 136 lib + 14 contract = 150 tests passed, 0 failed (`cargo test -p khive-db --features vectors`). Clippy clean (`-D warnings`). fmt clean. Release build clean (failpoint is `#[cfg(test)]` only).

### DIFF
--- a/crates/khive-db/src/stores/vectors.rs
+++ b/crates/khive-db/src/stores/vectors.rs
@@ -1313,4 +1313,324 @@ mod atomic_replace_tests {
             "prior vector must survive a failed update"
         );
     }
+
+    /// insert_batch: SAVEPOINT/ROLLBACK path — INSERT failure inside the savepoint.
+    ///
+    /// The existing wrong-dimension tests (`insert_batch_failed_record_preserves_prior_vector`)
+    /// hit the pre-savepoint `continue` guard and never reach the SAVEPOINT/ROLLBACK
+    /// sequence.  This test forces a genuine INSERT failure inside the savepoint by
+    /// exploiting vec0's single-column PRIMARY KEY (`subject_id TEXT PRIMARY KEY`,
+    /// NOT scoped to namespace).
+    ///
+    /// Mechanism: store a stale row for `(id_X, ns:a)`.  Submit a batch with one
+    /// record for `(id_X, ns:b)`.  The DELETE step targets `WHERE namespace = 'ns:b'`
+    /// and finds nothing (stale is in ns:a), so nothing is removed.  The INSERT then
+    /// tries to write `id_X` into vec0's `_rowids` shadow table, but `id_X` already
+    /// occupies it (from the ns:a stale row).  The UNIQUE constraint fires — INSERT
+    /// fails — ROLLBACK TO SAVEPOINT executes — stale row in ns:a survives intact.
+    ///
+    /// Sentinel: removing `ROLLBACK TO SAVEPOINT` would not change the outcome here
+    /// (DELETE was a no-op), but any partial shadow-table write between the SAVEPOINT
+    /// and the INSERT error would be left uncommitted, potentially corrupting the
+    /// stale row's embedding bytes.  The similarity assertion (>0.999) detects that.
+    ///
+    /// Additionally: insert_batch must count the record as `failed` and must not
+    /// abort the outer `BEGIN IMMEDIATE` transaction (the COMMIT must succeed).
+    #[tokio::test]
+    async fn insert_batch_savepoint_rollback_on_pk_conflict_preserves_stale() {
+        let pool = make_vec_pool();
+        let model_key = "atomic_pk_batch";
+        let dims = 4;
+        let ns_a = "ns:pk_a";
+        let ns_b = "ns:pk_b";
+
+        create_vec_table(&pool, model_key, dims);
+
+        let store = SqliteVecStore::new(
+            Arc::clone(&pool),
+            false,
+            model_key.to_string(),
+            model_key.to_string(),
+            dims,
+            ns_a.to_string(),
+        )
+        .expect("SqliteVecStore::new");
+
+        let id_x = Uuid::new_v4();
+        let stale_vec = vec![0.1f32, 0.2, 0.3, 0.4];
+
+        // Store stale row in ns:a — this occupies id_X in the vec0 PK.
+        store
+            .insert(
+                id_x,
+                SubstrateKind::Entity,
+                ns_a,
+                "body",
+                vec![stale_vec.clone()],
+            )
+            .await
+            .expect("stale insert");
+
+        // Batch: one record for (id_X, ns:b) — correct dims, all finite.
+        // DELETE WHERE ns=ns:b finds nothing.  INSERT hits PK constraint.
+        // Code path: SAVEPOINT → DELETE(noop) → INSERT(PK fail) →
+        //            ROLLBACK TO SAVEPOINT → RELEASE → outer COMMIT.
+        let summary = store
+            .insert_batch(vec![VectorRecord {
+                subject_id: id_x,
+                kind: SubstrateKind::Entity,
+                namespace: ns_b.to_string(),
+                field: "body".to_string(),
+                embedding_model: None,
+                vectors: vec![vec![0.5f32, 0.6, 0.7, 0.8]],
+                updated_at: chrono::Utc::now(),
+            }])
+            .await
+            .expect("insert_batch must complete (outer tx must commit)");
+
+        assert_eq!(summary.attempted, 1);
+        assert_eq!(summary.affected, 0, "PK conflict must count as failed");
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


ohdearquant:
Adversarial review:

Production code is CORRECT: per-record SAVEPOINT/RELEASE/ROLLBACK-TO sequence is sound (the failure path issues RELEASE after ROLLBACK TO — no savepoint leak), the outer BEGIN IMMEDIATE wraps the whole batch so committed siblings persist, genuine byte-for-byte parity with `text.rs:upsert_batch`, and vec0 savepoints are proven by the existing FTS5/insert paths. `unchecked_transaction()` in the new `update()` is safe — no production caller wraps it in an outer tx (same risk-class as the existing single-record insert).

**Fix needed before merge: the failure-injection tests do NOT exercise the rollback they claim to test.** Both inject wrong-dimension input, which short-circuits on pre-validation (vectors.rs:283 `continue` / 367 early-return) BEFORE the DELETE/savepoint runs. A regression that broke `ROLLBACK TO` would NOT be caught — the entire point of the PR (rollback on a post-DELETE insert failure) is unverified.

Required: add a test where INSERT fails AFTER a successful DELETE on a correctly-dimensioned record (constraint / dup-key / vec0 insert error), and assert (a) the prior vector survives via `search`/exact-row check, not just `batch_exists`, and (b) for `update`, that the transaction actually opened. Test fix incoming.

Non-blocking follow-up worth filing: a WriterGuard that rolls back a leaked tx on drop would harden all BEGIN IMMEDIATE store paths, not just this one. 131 tests pass.

ohdearquant:
Update — **fix landed, requirement met.** Sentinel test added (commit df2862d2): a `#[cfg(test)]` failpoint fires after a successful same-namespace DELETE and before the INSERT, so the rollback path is genuinely exercised. Verified faithfully against CI:

- **Sentinel is real (destructive proof):** with `ROLLBACK TO SAVEPOINT vec_batch_record` commented out, `insert_batch_rollback_restores_deleted_stale_after_post_delete_insert_failure` FAILS ("ROLLBACK TO SAVEPOINT must restore the stale row after DELETE + injected failure", vectors.rs:1844); restored → passes. The test now guards the production rollback line.
- Mirror `update_rollback_restores_deleted_stale_...` covers the `unchecked_transaction` path.
- Failpoint uses `Arc<AtomicBool>` captured before `with_writer`'s `spawn_blocking` hop (a `thread_local` wouldn't survive the OS-thread switch) — correct.
- 150 tests pass (136 lib + 14 contract) **with `--features vectors`** (the module is gated `cfg(all(test, feature="vectors"))`, like `batch_exists_tests`; CI enables it via workspace feature unification through khive-runtime). `cargo clippy --workspace --all-targets -- -D warnings` green; `--release` build clean (failpoint fully absent, no cfg leak).
- The 3 prior PK-conflict tests now carry NOTEs clarifying their DELETE is a no-op (they cover outer-tx survival + sibling integrity, not the rollback line).

Production code unchanged (it was already correct). Non-blocking follow-up still worth filing: a WriterGuard that rolls back a leaked tx on drop. Ready for merge.

Issue #195, opened 2026-06-20
feat(kkernel): true cross-op atomicity for exec --ops-file bulk apply


ohdearquant:
Follow-up to #74. The `kkernel exec --ops-file` bulk-apply path (PR for #74) delivers the primary value of #74 — one process instead of N spawns, parse-first safety, `--dry-run` preview — but it is **batched, not atomic**.

## What ships in #74
Ops are dispatched in chunks of 100 through the same `dispatch_request_local` path the MCP `request` tool uses. Each chunk is a **parallel batch of independent ops** with per-op semantics: a failing op does not roll back its siblings, and there is no SQL transaction wrapping a chunk. On partial failure the database is left partially written — the same non-atomicity as today's shell-loop-over-`kkernel exec`.

## Why true atomicity wasn't done in #74
The runtime dispatches each verb through the registry, and each verb opens and commits its own DB transaction internally. There is no API to wrap N verb dispatches in a single outer SQL transaction — the runtime does not expose a transaction handle through verb dispatch. True per-chunk (or whole-file) atomicity requires threading a transaction handle through the dispatch path.

## Scope of this follow-up
- Decide the unit of atomicity: per-chunk (all-or-nothing per 100) vs whole-file.
- Add a runtime API to run a batch of ops inside one outer transaction (BEGIN IMMEDIATE … COMMIT/ROLLBACK), or a documented `--atomic` flag on `exec --ops-file` that opts into it.
- Define failure semantics: on any op failure within the atomic unit, ROLLBACK the unit and report which op failed.
- This is what would let raw sqlite writes be policy-banned for cleansing that requires all-or-nothing (#74's stated end-goal).

Lower priority than #74 itself — #74's non-atomic bulk apply is already strictly better than the status quo for idempotent mass updates (e.g. the salience recalibration use case).

ohdearquant:
Closing via backlog triage 2026-07-09 (verified against current main at 3d03c8f0): Shipped: `crates/kkernel/src/atomic_apply.rs` header comment: "ADR-099 Slice B3 — the `--atomic` execution path for `kkernel exec --ops-file`," with `execute_atomic_ops_file`, `--atomic-max-ops` config, and an explicit v1 scope note ("merge is not yet supported under --atomic"). The requested transaction-handle-through-dispatch mechanism landed under ADR-099.

PR #196, opened 2026-06-20
feat(kkernel): exec --ops-file batch.jsonl bulk apply with --dry-run (#74)

### LINKED ISSUE #74: feat: kkernel exec --ops-file batch.jsonl (transactional bulk apply, --dry-run)
Mass data cleansing is the stated kkernel use case, but the only bulk mechanism today is a shell loop over \`kkernel exec\` invocations — the 2026-06-10 salience recalibration applied 8,083 updates as 162 separate process spawns, each paying startup + lock acquisition (and the 100-op batch cap shapes everything client-side).

Proposal: \`kkernel exec --ops-file batch.jsonl [--dry-run]\` — one process, one transaction (or chunked transactions with a progress line), ops in JSON form, dry-run prints the would-be summary without writing. With this + daemon-dialing + save-file, raw sqlite writes can be policy-banned entirely (the stated intent: raw DB = migration/cleansing/debugging only, and ideally not even that).

ohdearquant:
Closes #74.

## What

`kkernel exec --ops-file batch.jsonl [--dry-run]` — single-process bulk apply of ops from a JSONL file, replacing the shell-loop-over-`kkernel exec` pattern (the 2026-06-10 salience recalibration paid 162 separate process spawns for 8,083 updates).

- Each non-blank line is a JSON op `{"tool":"verb","args":{...}}` (the MCP `request` JSON form). `args` optional, defaults to `{}`. Extra top-level keys ignored (forward-compatible).
- **Parse-first**: the whole file is parsed before any write; a malformed line aborts with its line number, nothing is written.
- **Chunked dispatch**: ops are dispatched in chunks of 100 (`OPS_FILE_CHUNK_SIZE` = `khive_request::MAX_OPS`) through the same `dispatch_request_local` path the `request` tool uses. Progress per chunk to stderr; aggregate `{total, succeeded, failed}` summary to stdout.
- **`--dry-run`** (requires `--ops-file`): parse + validate every line, print a per-verb count summary, write nothing.
- The positional `ops` arg and `--ops-file` are mutually exclusive (enforced in `run_exec`); `ops` is now `Option<String>`.
- Daemon fast-path is intentionally skipped for bulk apply (the daemon is warm-state optimised, not throughput optimised); a fresh in-process runtime is built once and reused across all chunks.

## Tests (9, all isolated temp DBs — never touch `~/.khive/khive.db`)

- clap: `--ops-file` makes positional `ops` optional; `--dry-run` without `--ops-file` rejected.
- parse: blank-line skip; line-number on malformed JSON; missing `tool` field reported.
- integration: 3-op apply + summary match; `--dry-run` writes nothing; malformed line aborts before any write.

`cargo fmt -p kkernel --check`, `cargo clippy -p kkernel --all-targets -- -D warnings`, `cargo test -p kkernel` all green.

## Limitation — NOT atomic (read before merge)

This is **chunked batch apply, not transactional**. Each chunk is a parallel batch of independent ops: a failing op does not roll back its siblings, and no SQL transaction wraps a chunk. On partial failure the DB is left partially written — the same non-atomicity as the status-quo shell loop. The issue title said "transactional bulk apply (or chunked transactions)"; true atomicity needs a runtime transaction handle threaded through verb dispatch, which the runtime does not currently expose. Tracked as follow-up #195. For idempotent mass updates (the salience-recal use case) this non-atomic apply is already strictly better than the status quo.

### DIFF
--- a/crates/kkernel/src/exec.rs
+++ b/crates/kkernel/src/exec.rs
@@ -4,8 +4,23 @@
 //! building an in-process runtime (ADR-049). Config and namespace are matched
 //! against the daemon's own fingerprint; a mismatch falls back to local
 //! dispatch, keeping behaviour identical to the in-process path.
+//!
+//! # `--ops-file` bulk-apply path
+//!
+//! `kkernel exec --ops-file batch.jsonl` reads a JSONL file where each
+//! non-blank line is a JSON op object `{"tool":"verb","args":{...}}`.  All
+//! lines are parsed first; a malformed line aborts before any writes.  Valid
+//! ops are dispatched in chunks of [`OPS_FILE_CHUNK_SIZE`] through the same
+//! in-process runtime path (daemon fast-path is intentionally skipped for
+//! bulk apply — the daemon is warm-state optimised, not throughput optimised).
+//! A progress line is printed per chunk.  `--dry-run` validates every line and
+//! prints a per-verb summary without writing anything.
 
-use anyhow::Result;
+use std::collections::BTreeMap;
+use std::io::BufRead as _;
+use std::path::PathBuf;
+
+use anyhow::{Context, Result};
 use clap::Parser;
 
 #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


ohdearquant:
## Review — approved (stage for merge)

Read the full 470-line diff directly; ran CI faithfully from `crates/`:
- `cargo fmt -p kkernel --check` — clean
- `cargo clippy -p kkernel --all-targets -- -D warnings` — clean
- `cargo test -p kkernel` — green (9 new tests, all isolated temp DBs)

No feature-gated test modules in this diff, so `-p kkernel` is a faithful CI proxy here.

**Strengths**: parse-first (no partial writes on a malformed line, line-numbered errors), `--dry-run` preview, clean mutual-exclusion via clap `requires`, daemon-skip rationale is correct (warm-state vs throughput), all tests use `NamedTempFile` isolation — never touches `~/.khive/khive.db`. The code is honest: it says "parallel batch," never claims per-chunk atomicity.

**One thing for the merge decision** (not a defect, a scope boundary): this is **chunked batch apply, not transactional**. No SQL transaction wraps a chunk; a failing op does not roll back siblings; partial failure leaves partial writes — same non-atomicity as today's shell loop. Issue #74's title said "transactional"; the body accepted "chunked transactions" as an alternative, but the impl delivers chunked *batches*. True atomicity needs a runtime tx-handle through verb dispatch (not currently exposed) → tracked as #195. For the stated idempotent mass-update use case this is strictly better than the status quo, so merging as-is and pursuing #195 separately is reasonable — the atomicity expectation set by the issue title needs confirming at merge time.

Staged for review; not yet merged.

PR #197, opened 2026-06-20
fix(knowledge): never return silent ok:true/total:0 while ANN index is warming (#184)

### LINKED ISSUE #184: knowledge.suggest/compose silently return ok:true/total:0 during ANN cold-start after daemon restart (warm-up race)
## Symptom

After the daemon restarts (e.g. following `make local`), the first ANN-dependent knowledge query returns a **silent empty success** — `{ok: true, ... total: 0}` (or an empty `results`/`data`) — even though the corpus is fully indexed and the same query returns results moments later once the index is warm.

Observed live this session: `knowledge.suggest(...)` returned `total: 0` immediately after a `make local` daemon restart; a subsequent `knowledge.compose(...)` / retry returned the expected results. This is the failure mode another lambda hit and reported as "knowledge suggest still failed."

## Why this is a bug (not just slow)

`ok: true, total: 0` is **indistinguishable from a legitimate "no matches"**. A caller cannot tell "the index isn't ready yet, retry" from "there is genuinely nothing for this query." Agents (and the auto-compose chain) treat the empty result as authoritative and move on, silently losing the knowledge. The empty success is the defect — not the warm-up latency itself.

## Likely root cause (hypothesis — evidence below)

The Vamana ANN (`vamana::SharedAnn`, consumed by `knowledge.suggest`/`knowledge.compose`) must load/build into the daemon's warm state on restart (ADR-049 keeps indexes hot across sessions; a fresh daemon has to populate them). The first ANN query races that load: `embed_query` succeeds, but the ANN handle is empty/not-yet-loaded, so the section/vector path yields nothing while the rest of the call returns `ok`.

Evidence supporting the hypothesis:
- **FTS-first `knowledge.search` works during the same window** — only the ANN-dependent paths (`suggest`/`compose` section retrieval) come back empty.
- The empty result **resolves on its own after a few seconds** with no corpus change.
- It reproduces specifically right after a daemon restart, not in steady state.

(Stated as a hypothesis; the fix should confirm in code whether `SharedAnn` is populated synchronously at daemon boot or lazily on first query.)

## Repro

1. `make local` (rebuilds + kills/respawns the daemon).
2. Immediately reconnect (`/mcp`) and call `knowledge.suggest(query="<a query you know has hits>")`.
3. Observe `ok: true` with `total: 0`.
4. Wait a few seconds; repeat the identical call → results appear.

## Investigation pointers

- `crates/khive-pack-knowledge/src/knowledge/search.rs` — `suggest`/`compose` consume `ann: &vamana::SharedAnn`; `embed_query` (~L1344) then the `if let Some(qe) = q_emb` ANN/section path.
- Daemon warm-state init (ADR-049) — where/when `SharedAnn` is loaded or built on daemon start; is the first request gated on readiness?
- `crates/khive-vamana` load/build entry points — is there a "ready" signal to gate on?

## Proposed fix (pick one; the first is preferred)

1. **Block-until-warm with a bounded wait**: the first ANN query waits for the index to finish loading (short, bounded timeout) instead of returning empty. Simplest correct behavior for callers.
2. **Explicit warming signal**: return a distinct, non-`ok`-empty shape — e.g. `{ok: false, error: "index warming, retry"}` or `{ok: true, warming: true, total: 0}` — so callers can distinguish "not ready" from "no matches" and retry deterministically.
3. **Readiness gate at the daemon boundary**: hold/queue ANN-dependent requests until `SharedAnn` reports ready (with a timeout fallback to option 2).

Whatever the shape, the non-negotiable: an unready index must **never** surface as a bare `ok: true, total: 0`.

## Acceptance criteria

- After a daemon restart, the first `knowledge.suggest`/`knowledge.compose` for a query with known hits either (a) returns the real hits, or (b) returns a clearly distinguishable "warming/retry" signal — never a silent empty success.
- A regression test exercises the cold-start path (fresh daemon/empty warm state) and asserts the non-silent behavior.

## Current workaround

Retry the call after a short delay; or issue an FTS-only `knowledge.search` first (it doesn't depend on the ANN). Documented as a khive observation note for other lambdas.

ohdearquant:
Closes #184.

## Problem

After a daemon restart (e.g. `make local`), the first ANN-dependent `knowledge.suggest`/`knowledge.compose` raced the background ANN warm-load and returned a **silent `ok:true, total:0`** — indistinguishable from a genuine "no matches." Callers (and the auto-compose chain) treated the empty result as authoritative and silently lost the knowledge.

## Root cause

`AnnState` holds `warming: Mutex<HashSet<AnnKey>>` and `indexes: RwLock<HashMap<AnnKey, AnnBridge>>` independently. `ensure_ann_background` inserts the key into `warming` then spawns a tokio task that inserts the bridge into `indexes` only when loading completes. The query path runs `search_loaded` immediately on the still-empty index — silent zero.

## Fix (issue option 1 + fallback to option 2)

New in `vamana.rs`:
- `is_warming_not_loaded(ann, key)` — sync check: key in `warming` but not in `indexes` (uses `try_read`, conservatively reports warming if the write lock is held).
- `wait_for_ann(ann, key, timeout_ms, poll_ms)` — bounded poll until the index appears.

In both `suggest` and `search`, after `search_loaded` returns `None` and a warm is in flight: wait up to 3s (50ms poll), then retry. If still not ready, surface a **distinguishable warming signal** — `Err(RuntimeError::Internal("ANN index is warming after daemon restart — retry in a few seconds"))` — but **only when FTS also found nothing** (`hits.is_empty()`) and the vector corpus is non-empty. If FTS already collected hits, those partial results are returned rather than erroring. A genuinely empty corpus still returns the normal `ok:true, total:0`.

This satisfies #184's non-negotiable: an unready index never surfaces as a bare `ok:true, total:0`. The warming `Err` shape is the issue's sanctioned option 2.

## Tests

72 pass (`cargo test -p khive-pack-knowledge`), clippy + fmt clean. New mechanism units in `vamana::tests`: `is_warming_not_loaded` (3 states), `wait_for_ann` (immediate / timeout / appears-mid-poll). `tests/fixes.rs`: empty-corpus guard returns ok-not-error; populated-corpus FTS path.

**Test-coverage boundary (documented in-code):** the end-to-end warming path in `suggest`/`search` can't be driven from the external `tests/` directory — `embed_query` needs a real embedder (`KhiveRuntime::memory()` has none) to reach the ANN block, and `simulate_warming_in_flight` is `pub(crate)` (unreachable from a separate test crate). The mechanism is unit-tested where it's reachable; the call-site wiring is covered by code review. No fake embedder was stubbed (it would not reflect production `embed_query`).

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/search.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/search.rs
@@ -1031,7 +1031,38 @@ impl KnowledgeHandlers {
         if let Ok(query_emb) = runtime.embed_query(&raw_query).await {
             let ann_k = fetch_limit.max(20);
             let key = vamana::AnnKey::new(&ns, runtime.default_embedder_name());
-            if let Some(ann_hits) = vamana::search_loaded(ann, &key, &query_emb, ann_k).await {
+
+            // If the ANN slot is absent but a background warm is in flight, wait a
+            // bounded time before continuing with FTS-only results.  This avoids a
+            // cold-start race where the ANN is loading and the first query silently
+            // returns ok:true,total:0 when FTS also finds nothing.
+            let mut ann_hits_opt = vamana::search_loaded(ann, &key, &query_emb, ann_k).await;
+            if ann_hits_opt.is_none() && vamana::is_warming_not_loaded(ann, &key) {
+                const WARM_TIMEOUT_MS: u64 = 3_000;
+                const WARM_POLL_MS: u64 = 50;
+                if vamana::wait_for_ann(ann, &key, WARM_TIMEOUT_MS, WARM_POLL_MS).await {
+                    ann_hits_opt = vamana::search_loaded(ann, &key, &query_emb, ann_k).await;
+                } else {
+                    // Still not ready: if FTS already found results, those are
+                    // valid partial hits — return them.  Only surface the warming
+                    // error when FTS is also empty and the corpus is non-empty.
+                    if hits.is_empty() {
+                        let model = runtime.default_embedder_name();
+                        let corpus_non_empty = vamana::compute_fingerprint(runtime, token, model)
+                            .await
+                            .map(|fp| fp.vector_count > 0)
+                            .unwrap_or(false);
+                        if corpus_non_empty {
+                            return Err(RuntimeError::Internal(
+                                "ANN index is warming after daemon restart — retry in a few seconds"
+                                    .into(),
+                            ));
+                        }
+                    }
+                }
+            }
+
+            if let Some(ann_hits) = ann_hits_opt {
                 if !ann_hits.is_empty() {
                     fuse_ann_hits(&mut hits, &ann_hits, min_score);
                     hydrate_empty_hits(runtime, &ns, &mut hits).await;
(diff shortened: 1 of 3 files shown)


ohdearquant:
## Review — approved (stage for merge)

Read both commits' full diffs directly; re-ran CI faithfully from `crates/` after the refinement commit:
- `cargo fmt -p khive-pack-knowledge --check` — clean
- `cargo clippy -p khive-pack-knowledge --all-targets -- -D warnings` — clean
- `cargo test -p khive-pack-knowledge` — green (62 lib + integration, 0 failures)

No feature-gated test modules in this diff, so `-p` is a faithful CI proxy here.

**Verified against #184 acceptance:** an unready index never surfaces as a bare `ok:true,total:0` — confirmed at both call sites. The warming `Err` is the issue's sanctioned option-2 shape. The `is_warming_not_loaded` `try_read` conservatism (reports warming when the write lock is held) is correct — the index is about to appear, so treating it as still-warming is safe.

**Refinement applied + verified:** `suggest` now mirrors `search` — only errors when `hits.is_empty()`, so it no longer discards valid FTS hits during the ~3s warm window (was a narrow regression). Diff confirmed at search.rs.

**One honest gap (not a blocker):** the acceptance criterion's end-to-end cold-start regression test isn't present — `embed_query` needs a real embedder to reach the ANN block (`KhiveRuntime::memory()` has none) and `simulate_warming_in_flight` is `pub(crate)`. The mechanism (`is_warming_not_loaded`/`wait_for_ann`) is unit-tested where reachable; call-site wiring is covered by review. Correctly documented in-code; no fake embedder stubbed. True e2e coverage at merge would need a test-embedder fixture in the crate — worth a follow-up, not a blocker.

Staged for review; not yet merged.

Issue #176, opened 2026-06-19
ADR-029: multi-backend search ignores props/tags filters


ohdearquant:
## Summary

When a `search` verb is dispatched through the ADR-029 `SubstrateCoordinator` path (multi-backend mode), the `properties` and `tags` filter arguments present in the single-backend kg handler are silently dropped.

## Root cause

`dispatch_via_coordinator_inner` in `crates/khive-mcp/src/server.rs` does not thread `properties`/`tags` into `fan_out_search`, and `CoordinatorService::fan_out_search` does not accept those parameters. The single-backend `search` handler in `crates/khive-pack-kg/src/handlers/search.rs` applies them via `hybrid_search`/`search_notes` keyword filters.

## Deferred in

PR fixing ADR-029 search parity (kind filter, min_score, real kinds). A TODO comment was added at the fan-out call site in `server.rs`.

## Fix needed

1. Add `properties: Option<serde_json::Value>` and `tags: Option<Vec<String>>` to `CoordinatorService::fan_out_search` (and `MockCoordinator`).
2. Thread them from `dispatch_via_coordinator_inner` through to `SubstrateCoordinator::fan_out_search` and into the per-backend `hybrid_search`/`search_notes` calls.
3. Add a parity test that asserts props/tags filters are respected in multi-backend search.

## Affected files

- `crates/khive-mcp/src/coordinator.rs` (`CoordinatorService` trait + `MockCoordinator`)
- `crates/khive-mcp/src/server.rs` (`dispatch_via_coordinator_inner`)
- `crates/kkernel/src/coordinator/dispatch.rs` (`fan_out_search`)
- `crates/kkernel/src/coordinator/service.rs` (`fan_out_search`)
- `crates/kkernel/src/coordinator/tests.rs` (new parity test)

PR #255, opened 2026-06-26
fix(retrieval): push property/tag predicates below truncation (#176)

ohdearquant:
## Summary

- Extends the `fan_out_search` coordinator seam (`CoordinatorService` trait in `khive-mcp`, `SubstrateCoordinator` in `kkernel`) with two new parameters: `props_filter` and `tags`
- Forwards both filters into each backend's `hybrid_search` call, which already applies them correctly with before-truncation semantics via an internal `CANDIDATE_MULTIPLIER`
- Widens the per-backend candidate window to `min(limit * 50, 500)` when either filter is active, matching the single-backend handler's behavior so sparse matches ranked below the bare `limit` are not cut off before filtering
- Removes the ADR-029 deferral TODO/NOTE from `server.rs`
- Updates all existing `fan_out_search` call sites in tests with the new `None`/`&[]` defaults
- Adds three regression tests in `kkernel/coordinator/tests.rs`: props filter drops non-matching entity across backends, before-truncation semantics at `limit=1`, tags filter drops non-matching entity across backends

## Test plan

- [ ] `cargo test -p khive-mcp -p khive-pack-kg -p khive-runtime --all-features` passes (1003 tests, 0 failed)
- [ ] `cargo test -p kkernel --all-features` passes (154 tests including 3 new regression tests, 0 failed)
- [ ] `cargo clippy -p khive-mcp -p khive-pack-kg -p khive-runtime -p kkernel --all-features -- -D warnings` passes
- [ ] `cargo fmt --all -- --check` passes

### DIFF
--- a/crates/khive-mcp/src/coordinator.rs
+++ b/crates/khive-mcp/src/coordinator.rs
@@ -335,6 +335,57 @@ pub(crate) mod tests {
         );
     }
 
+    /// T6d: A multi-backend search with a malformed `tags` value must return a
+    /// per-op error (`ok: false`) rather than silently returning unfiltered results.
+    ///
+    /// Single-backend rejects malformed tags via `SearchParams` deserialisation
+    /// (RuntimeError::InvalidInput → `ok: false`).  Multi-backend must match that
+    /// contract: the server must reject before reaching the coordinator, not silently
+    /// collapse the filter to an empty Vec.
+    ///
+    /// This test FAILS against the old `filter_map(as_str)` code (which would call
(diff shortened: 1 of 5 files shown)


ohdearquant:
Round-1 fix applied (commit 43861fb0): replaced `filter_map(as_str)` with strict `serde_json::from_value::<Vec<String>>` in `server.rs` — malformed tags now return `ok:false` before reaching the coordinator, matching single-backend behaviour. Added T6d regression test in `khive-mcp/src/coordinator.rs` that fails against the old code and passes with the fix. Gate: TEST_RC=0, CLIPPY_RC=0, FMT_RC=0.

PR #591, opened 2026-07-04
feat(storage,db): open-transaction registry + WAL checkpoint instrumentation (ADR-091 Plank 0)

ohdearquant:
## Summary

Implements ADR-091 Plank 0: a process-wide open-transaction registry and observe-only checkpoint instrumentation, per the approved plan (sequencing 0 → 2 → 1). This is code-only — the ADR (#585) is already merged, no spec edits ride this PR.

**Zero behavior change to any transaction.** No enforcement, no caps, no rejection, no new config keys — that is Plank 1, gated on one cycle of production telemetry from this plank. This PR is pure observability: it lets the checkpoint task name which caller, if any, is holding a WAL snapshot open, the question ADR-091's static code reading could not answer.

### What registers where

- `crates/khive-storage/src/tx_registry.rs` (new): a process-wide `Mutex<HashMap<TxId, TxMeta>>` registry. `register(label) -> TxHandle`, `oldest()`, `snapshot()`. Exported from `khive-storage/src/lib.rs`.
- `crates/khive-db/src/sql_bridge.rs`:
  - `SqliteTransaction` gains a `_tx_handle: TxHandle` field; `begin_tx` (~:848) registers right after `BEGIN` succeeds, reusing the caller's `SqlTxOptions.label`.
  - `SqliteWriter::execute_batch` (~:360) and `PoolBackedWriter::execute_batch` (~:741) — label `"execute_batch"` / `"pool_writer.execute_batch"`.
- `crates/khive-db/src/pool.rs:181` `WriterGuard::transaction` — label `"writer_guard_tx"`. One instrumentation point covers every `guard.transaction(...)` caller: `khive-runtime/src/curation.rs`'s `merge_entity`/`merge_note_sql` and `khive-runtime/src/operations.rs:3623`'s edge update needed no per-site edits.
- Store batch-upsert methods (`khive-db/src/stores/`): `entity.rs` (`entity_upsert_batch`), `note.rs` (`note_upsert_batch`), `graph.rs` (`graph_upsert_edges`), `event.rs` (`event_append`, `event_append_batch`), `text.rs` (`text_upsert_document`, `text_upsert_batch`, `text_rename_namespace`), `vectors.rs` (`vector_insert_batch`), `sparse.rs` (`sparse_insert_batch`).
- `khive-vcs/src/sync.rs` needed no direct edit — its per-chunk writes route through `store.upsert_entities`/`text.upsert_documents`, already instrumented above.
- `khive-pack-brain`: `fold_gate.rs`'s `apply_fold_gate` (`fold_gate_apply`) and `apply_fold_gate_and_append_event` (`fold_gate_apply_event`); `persist.rs`'s `persist_brain_state_mutation` (`brain_persist_mutation`). `handlers.rs:1139` is a call site with no BEGIN of its own — no edit needed.
- `crates/khive-db/src/checkpoint.rs`: `run_checkpoint_task` now calls a new `log_tx_registry_pressure(wal_pages, warn_pages, high_water_pages)` helper every tick — `debug!` normally, escalating to `warn!` once `wal_pages` crosses `warn_pages`, naming the oldest open entry's age + label. Once `wal_pages` reaches `high_water_pages`, it enumerates every open registry entry via `snapshot()` at `warn!` — the load-bearing "which caller is pinning the WAL" answer.

### RAII rationale

`TxHandle` deregisters only via `Drop` — there is no manual `deregister()` call. Every registration site declares the handle as a local right after `BEGIN` succeeds, so it drops (and deregisters) on every return path out of that scope: normal commit, early-return error paths, and even a panic (`catch_unwind`-verified in the registry's own unit tests). This was an explicit call in the approved plan — manual deregister at 15+ call sites would leak on some path; RAII cannot.

The registry mutex is never held across a SQL call — `register`/`oldest`/`snapshot` are short, synchronous map operations only.

### Scope

- Observe-only (Plank 0). Enforcement reads (soft-cap WARN on per-statement age, cooperative stale-op guard, commit-past-cap rollback) are Plank 1, not built here.
- TRUNCATE escalation is Plank 2, not built here.
- No `KHIVE_TX_*` config keys added — those ship with Plank 1, tuned by this plank's telemetry.

### Gates run (from `crates/`, `CARGO_TARGET_DIR` set to a checkout-local dir)

- `cargo check --workspace` — pass
- `cargo test -p khive-storage -p khive-db -p khive-pack-brain -p khive-vcs` — 217 + 14 + 196 + 6 + 4 + 44 + 33 + 10 + 72 + 9 passed, 0 failed (workspace `--workspace` test intentionally NOT run per known lesson: the 4246-test suite times out gate agents)
- `cargo clippy --workspace --all-targets -- -D warnings` — pass, 0 warnings
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` — pass

### Discrepancies / decisions beyond the plan

None found — the plan's anchors (file:line) needed re-grepping (as expected, called out in the plan itself) but the code shape at each site matched the plan's description exactly. No ADR-vs-code mismatch encountered.

## Test plan

- [x] `cargo test -p khive-storage -p khive-db -p khive-pack-brain -p khive-vcs` green
- [x] `cargo clippy --workspace --all-targets -- -D warnings` green
- [x] `cargo fmt --all -- --check` green
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` green
- [ ] Additional review round on this draft PR

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -23,7 +23,12 @@
 //! Threshold-crossing WARN semantics: both the `warn_pages` and `high_water_pages`
 //! warnings fire at most once per below→above crossing. Skipped ticks (writer
 //! busy) leave the crossing state unchanged so that a skip cannot spuriously
-//! re-arm the rate limit while WAL pressure is still elevated.
+//! re-arm the rate limit while WAL pressure is still elevated. The ADR-091
+//! Plank 0 open-transaction-registry WARNs (oldest-entry escalation and the
+//! high-water snapshot enumeration) ride the SAME crossing gates — they are
+//! not independently rate-limited, so they never repeat on consecutive ticks
+//! that remain above a threshold. Only the per-tick `debug!` trace of the
+//! oldest open entry fires unconditionally.
 
 use std::sync::Arc;
 use std::time::Duration;
@@ -161,12 +166,18 @@ pub async fn run_checkpoint_task(pool: Arc<ConnectionPool>, config: CheckpointCo
             CheckpointTick::Observed(n) => n,
         };
 
-        let in_warn_band = wal_pages >= config.warn_pages && wal_pages < config.high_water_pages;
+        let above_warn = wal_pages >= config.warn_pages;
         let above_high_water = wal_pages >= config.high_water_pages;
 
-        log_tx_registry_pressure(wal_pages, config.warn_pages, config.high_water_pages);
+        // Per-tick debug for the oldest open entry always fires (cheap, single
+        // `oldest()` lookup); the two `warn!`-level registry logs below are
+        // gated on the SAME crossing state as the WAL-threshold WARNs above,
+        // so sustained pressure logs once per crossing, not once per tick.
+        log_tx_registry_oldest_debug(wal_pages);
 
-        if crossing_warn(in_warn_band, &mut was_above_warn) {
+        let warn_crossed = crossing_warn(above_warn, &mut was_above_warn);
+        if warn_crossed {
+            log_tx_registry_oldest_warn(wal_pages);
             tracing::warn!(
                 wal_pages,
                 warn_threshold = config.warn_pages,
@@ -174,7 +185,9 @@ pub async fn run_checkpoint_task(pool: Arc<ConnectionPool>, config: CheckpointCo
             );
         }
 
-        if crossing_warn(above_high_water, &mut was_above_high_water) {
+        let high_water_crossed = crossing_warn(above_high_water, &mut was_above_high_water);
+        if high_water_crossed {
+            log_tx_registry_snapshot_warn(wal_pages);
             tracing::warn!(
                 wal_pages,
                 high_water = config.high_water_pages,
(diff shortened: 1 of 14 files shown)


ohdearquant:
## Round-1 review fixes (fixes-requested → all 5 findings addressed)

New head: `711e2c69`

| # | Severity | Finding | Fix | File:line |
|---|----------|---------|-----|-----------|
| 1 | High | Registry pressure WARNs fired every tick above threshold, not on crossing | Split `log_tx_registry_pressure` into `log_tx_registry_oldest_debug` (unconditional per-tick debug, unchanged), `log_tx_registry_oldest_warn` + `log_tx_registry_snapshot_warn` — both now gated by the same `crossing_warn` state machine the WAL-threshold WARNs already use | `crates/khive-db/src/checkpoint.rs:159-247` |
| 2 | High | `unchecked_transaction()`/`Transaction::new_unchecked` spans invisible to raw-BEGIN grep | Registered all 4 sites: `vec_insert_tx`, `vec_update_tx`, `vec_orphan_sweep`, `graph_traverse_read` (the WAL-pin-relevant chunked-traversal read snapshot). Each handle declared *before* its transaction binding — locals drop in reverse declaration order, so the transaction's own Drop (rollback-on-uncommitted) runs before the handle deregisters | `crates/khive-db/src/stores/vectors.rs:297,498,856`, `crates/khive-db/src/stores/graph.rs:1028` |
| 3 | Medium | `SqliteWriter::execute_batch` deregistered before ROLLBACK ran on error | Restructured: BEGIN first, handle registered only after BEGIN succeeds, its scope now encloses both the statement closure and the ROLLBACK path — mirrors `PoolBackedWriter::execute_batch` | `crates/khive-db/src/sql_bridge.rs:358-384` |
| 4 | Low | khive-db registry tests can race each other (and, empirically, unrelated write-path tests) | Serialized the 3 named tests under `#[serial(tx_registry)]`; also hardened the checkpoint debug test, which live-failed against an unrelated concurrent test (`text_upsert_document` won the "oldest" race during verification) — it now samples the expected label via `tx_registry::oldest()` at call time instead of hardcoding it | `crates/khive-db/src/checkpoint.rs`, `sql_bridge.rs`, `pool.rs` |
| 5 | Low | Poisoned registry lock silently reads as empty | `register`/`oldest`/`snapshot` recover via `unwrap_or_else(\|p\| p.into_inner())` instead of dropping the write / returning `None`/`vec![]` | `crates/khive-storage/src/tx_registry.rs` |

Scope held to these 5 fixes only — observe-only invariant unchanged (zero enforcement, zero transaction-behavior change, no new config keys).

**Gates**: `cargo test -p khive-storage -p khive-db -p khive-pack-brain` (511 passed; ran khive-db `--lib` 5x consecutively to confirm no flakiness after the round-1 test-race fix), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` — all clean.

PR #593, opened 2026-07-04
feat(db): ADR-091 Plank 2 — WAL TRUNCATE escalation with rate-limited guards

ohdearquant:
## Summary
- Adds a rare TRUNCATE escalation path in `crates/khive-db/src/checkpoint.rs`, on top of (not replacing) the existing PASSIVE-only periodic checkpoint tick.
- One writer checkout per tick, per ADR-091 Plank 2: PASSIVE and any due TRUNCATE run under the SAME `try_writer_nowait` guard `checkpoint_once` already holds. A busy writer skips both.
- New `CheckpointConfig` keys: `truncate_high_water_pages` (`KHIVE_WAL_TRUNCATE_HIGH_WATER_PAGES`, default 20000), `truncate_min_interval` (`KHIVE_WAL_TRUNCATE_MIN_INTERVAL_SECS`, default 300s), `truncate_busy_timeout` (`KHIVE_WAL_TRUNCATE_BUSY_MS`, default 2000ms).
- New `TruncateState` (last_attempt + consecutive_failures) carried across ticks by the caller. `last_attempt` is stamped only on an actual attempt — never on a skip (writer busy, below threshold, or interval not yet elapsed) — so the next free tick is immediately eligible.
- `busy_timeout` is temporarily lowered to `truncate_busy_timeout` for the `PRAGMA wal_checkpoint(TRUNCATE)` call and restored on the writer connection immediately after, win or lose.
- Observability: `info!` per attempt (before/after pages, elapsed); `warn!` once at exactly the 3rd consecutive attempt that fails to clear `warn_pages` (resets on a clearing attempt); Plank 0's `tx_registry::snapshot()` logged before every attempt and again when an attempt makes no progress. No transaction is ever killed/aborted here — that is Plank 1, later.

Ref: `docs/adr/ADR-091-wal-snapshot-lifetime.md` §Plank 2 (~L364-414), PR 2 of 3 per the approved 0→2→1 ordering.

## Test plan
- [x] `cargo test -p khive-db` — 200 + 10 (contract) passed, incl. 16 checkpoint tests (10 pre-existing + 6 new Plank 2 tests: trigger threshold, below-threshold no-stamp, min-interval skip does not restamp, busy-writer skips both PASSIVE and TRUNCATE, edge-triggered 3rd-consecutive-failure WARN + reset)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` — clean

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -29,11 +29,32 @@
 //! not independently rate-limited, so they never repeat on consecutive ticks
 //! that remain above a threshold. Only the per-tick `debug!` trace of the
 //! oldest open entry fires unconditionally.
+//!
+//! ADR-091 Plank 2: rare TRUNCATE escalation. The periodic tick above stays
+//! PASSIVE-only and non-blocking; on top of it, `checkpoint_once` also
+//! evaluates a much rarer escalation to `PRAGMA wal_checkpoint(TRUNCATE)`
+//! once the WAL has grown past `truncate_high_water_pages` and at least
+//! `truncate_min_interval` has elapsed since the last TRUNCATE *attempt*
+//! (not the last successful reclaim). This is a **single writer checkout per
+//! tick**: PASSIVE and any due TRUNCATE both run under the one guard
+//! `checkpoint_once` already holds — there is never a second concurrent
+//! checkout for TRUNCATE. If the writer mutex is busy, both PASSIVE and any
+//! due TRUNCATE are skipped for that tick, and `last_truncate_attempt` is
+//! left untouched so the next tick where the writer is free is immediately
+//! eligible rather than waiting out the full interval again.
+//! `last_truncate_attempt` only advances on a tick that actually attempted
+//! TRUNCATE (writer held, threshold crossed, interval elapsed) — never on a
+//! skip for any reason (writer busy, below threshold, interval not yet up).
+//! TRUNCATE runs under a temporarily shortened `busy_timeout`
+//! (`truncate_busy_timeout`), restored on the writer connection immediately
+//! after the attempt, win or lose. No transaction is ever killed or aborted
+//! here — the tx_registry is only read for diagnostics (Plank 1 owns
+//! enforcement).
 
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


ohdearquant:
Addressed both round-1 review findings in 9b36ef38:

**[Medium] last_attempt stamped before the TRUNCATE attempt actually ran** — fixed. `truncate_state.last_attempt` is now stamped immediately before `PRAGMA wal_checkpoint(TRUNCATE)` itself, after `conn.busy_timeout(config.truncate_busy_timeout)` has already succeeded. If the busy_timeout override fails to apply, the function now returns without touching `last_attempt` at all — a setup failure is a skip, not an attempt, per ADR-091 §377-382.

Regression test seam: I did not add a new test forcing the busy_timeout-setup-failure path specifically. `Connection::busy_timeout()` just issues `PRAGMA busy_timeout=n` against a real rusqlite connection and has no injectable failure seam without corrupting the connection object itself (no trait/mock boundary exists here) — a contorted test would need to break the connection in a way that also breaks the surrounding assertions. Per the review's own guidance, I'm noting this explicitly here rather than adding a fragile test. The existing threshold-skip, min-interval-skip, and busy-writer-skip tests already cover every other path that must leave `last_attempt` untouched, and all still pass with the stamp moved.

**[Low] Stale module docs contradicting Plank 2** — fixed. Reworded the top-of-file module doc (previously "PASSIVE is the only mode the periodic task ever uses" / "TRUNCATE is excluded from the periodic path") and `run_checkpoint_task`'s doc comment to state that ordinary ticks stay PASSIVE-only and non-blocking, while a rare due escalation may run TRUNCATE under the same writer guard with a shortened busy timeout.

No change to the edge-triggered 3-consecutive-failure WARN — the review confirmed that reading as correct.

Gates re-run after the fix, real exit codes:
- `cargo test -p khive-db`: 0 (200 lib + 10 contract passed, same 16 checkpoint tests)
- `cargo clippy --workspace --all-targets -- -D warnings`: 0
- `cargo fmt --all -- --check`: 0
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace`: 0

New head SHA: `9b36ef388c395c85f8b400cc3fad5bd30e05e77e`

PR #620, opened 2026-07-05
docs(adr): ADR-093 — zstd compression for session-mirror raw storage

ohdearquant:
## Summary

- Authors ADR-093, specifying zstd compression for `session_messages.raw` in the
  session-mirror's dedicated `sessions` SQLite backend (`~/.khive/sessions.db`,
  `khive-pack-session` — never `khive.db`).
- Grounds every claim in source: schema (`vocab.rs`), write path (`mirror/ingest.rs`), the
  absence of any production read path or FTS indexing today, and confirmation that no crate
  in the workspace depends on `zstd` yet.
- Reports a real, read-only measurement sample (2026-07-04): 5,352 rows spread-sampled from
  the live `~/.khive/sessions.db` (1,070,410 rows total), measured via the `zstandard`
  Python bindings against a scratch copy — no write or lock taken on the production file.
  Per-row, concatenated, and trained-dictionary ratios at zstd levels 3 and 19 are all
  reported in the ADR's Motivation section.
- Decides and specifies the 7 required points: versioned one-byte encoding prefix, a
  trained dictionary, store-layer-owned transparent compress/decompress, lazy migration
  (no bulk recompress), FTS non-interaction (raw is not FTS-indexed today; a future FTS
  addition must index decompressed plaintext), fail-loud decompression failure handling,
  and explicit non-goals (no `khive.db` change, no wire-surface change, no new verbs).

Docs-only change: one new ADR file plus a README index entry. No `.rs` changes.

This direction (zstd raw-compression for the session mirror) has already been greenlit;
this ADR records and specifies it. This PR awaits maintainer review before any
implementation PR depends on it.

## Test plan

- [ ] Maintainer review of the ADR's 7 decisions
- [ ] `make fmt-check` / `deno fmt --check docs/` (already run locally, clean)
- [ ] No implementation PR merges against this ADR number until it is accepted

### DIFF
--- a/docs/adr/ADR-093-sessions-raw-zstd-compression.md
+++ b/docs/adr/ADR-093-sessions-raw-zstd-compression.md
@@ -33,7 +33,7 @@ JSONL files and replays them into a `session_messages` table. The table is decla
 (`crates/khive-pack-session/src/mirror/parse.rs:121`, `:200`, `:237` produce it via
 `secret_gate::mask_secrets(trimmed)`; `:538` produces it via
 `secret_gate::mask_secrets(&raw_json)` where `raw_json` comes from
-`serde_json::to_string(node).unwrap_or_default()` at `:536`, so this call site can hand
+`serde_json::to_string(node).unwrap_or_default()` at `:537`, so this call site can hand
 `mask_secrets` an empty string on a serialization failure and store an empty `raw`). `text`
 is a derived, much shorter extraction (the human-readable message body) used for display and
 search; `raw` exists so a message can be losslessly reconstructed later, independent of what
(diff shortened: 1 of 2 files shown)


ohdearquant:
Review outcome: accepted at 4460a79e, conditional on one trivial citation fix before merge.

Delta read in full against the prior change request (ec1c8b27) and verified at source:
- **Blocker resolved, and better than asked**: the typeof() discrimination (legacy TEXT passthrough vs version-prefixed BLOB) eliminates in-band ambiguity entirely rather than arguing it away via the printable-ASCII invariant. Verified `SqlValue::Blob(Vec<u8>)` exists at `crates/khive-storage/src/types/sql.rs:17`, so the encode-as-BLOB / typeof-dispatch contract is implementable as specified. Empty-string legacy case handled; fail-loud correctly re-scoped to the BLOB branch; weaker-corruption-detection tradeoff for legacy TEXT stated explicitly. The required-tests block is a welcome addition beyond the ask.
- **Medium resolved**: prefix-to-dictionary registry with generations pinned per prefix byte forever, additive include_bytes embeds, downgrade fail-loud rule, and the frame dictionary-ID check as a second independent gate. Complete.
- Both nits folded (0x00 reserved with a defined shape, MASKED analogy dropped). fts_notes citation verified exact at `crates/khive-db/sql/004-fts-consolidation.sql:22-33`. Level-3 rationale is sound for an inline ingest path.

**Fix before merge (trivial)**: the `unwrap_or_default()` citation is off by one — the line is `parse.rs:537` (536 is blank), cited as `:536` in two places (Context and the Decision 1 TEXT bullet).

**Rider for the implementation PR (not this ADR)**: add a zero-length BLOB case to the required tests — no first byte means no prefix, must fail loud, and the encoder must never emit it.

PR #673, opened 2026-07-06
refactor(storage): retire begin_tx — session ingest through atomic_unit (ADR-099 D5)

ohdearquant:
## Summary

ADR-099 D5, Slice A — migration steps 0, 5, 6 only (steps 1-4/7, the broader
`--atomic` bulk-apply feature, are out of scope for this PR).

- **Step 0**: documented the atomic-unit suspend-free invariant directly on
  `SqlAccess::atomic_unit` (`crates/khive-storage/src/sql.rs`) and on the
  `SqlBridge` implementation (`crates/khive-db/src/sql_bridge.rs`): the
  closure passed to `atomic_unit` must complete on its first poll —
  synchronous DML only, no embedding/ANN/service/channel `.await`. Enforced
  at runtime on the single-writer path via `block_on_sync`, which returns a
  typed error the instant the future is `Pending` rather than ever actually
  suspending.
- **Step 5**: converted `write_events_and_cursor` in
  `crates/khive-pack-session/src/mirror/ingest.rs` from a manual
  `begin_tx`/`tx.commit()` transaction to a single `sql.atomic_unit(...)`
  call. The closure drives the identical statement sequence (session
  create-only upsert, message `INSERT OR IGNORE` with a `seq` subquery, a
  conditional `last_seen_at` advance, and the cursor `UPDATE`) against the
  `&mut dyn SqlWriter` it is handed. Semantics are unchanged: the cursor
  only advances on commit, dedup and idempotency are preserved, and the
  inserted-count return value is identical.
- **Step 6**: `begin_tx` had exactly one non-test production caller anywhere
  in the workspace — `write_events_and_cursor` (verbatim grep evidence
  below). With that caller converted, `begin_tx` and the `SqlTransaction`
  trait have no remaining callers and no forward-deployed use in
  `khive-vcs`/`khive-merge` (checked directly — neither crate references
  `SqlTransaction` or `begin_tx`), so both are removed from `SqlAccess` and
  the `SqlBridge` implementation, along with the now-dead
  `open_standalone_tx_conn` helper and the `SqliteTransaction` type.

## Grep evidence (verbatim)

Before (base commit `b32ad689`, non-test production callers of `begin_tx`):

```
crates/khive-db/src/backend.rs:754:        let mut tx = sql.begin_tx(SqlTxOptions::default()).await.unwrap();       <- test-only (#[cfg(test)] mod)
crates/khive-db/src/sql_bridge.rs:1203:    async fn begin_tx(                                                            <- trait impl definition
crates/khive-pack-session/src/mirror/ingest.rs:615:        .begin_tx(SqlTxOptions::default())                              <- THE ONLY non-test production caller
crates/khive-pack-session/src/mirror/ingest.rs:2416:            .begin_tx(SqlTxOptions::default())                          <- test-only (regression test)
crates/khive-storage/src/sql.rs:89:    async fn begin_tx(&self, options: SqlTxOptions) -> ...                        <- trait declaration
crates/khive-db/src/stores/entity_tests.rs:769:        let mut tx = bridge.begin_tx(SqlTxOptions::default()).await.unwrap();  <- test-only
```

After (this branch, working tree):

```
$ grep -rn "begin_tx" --include="*.rs" crates/
khive-storage/src/tx_registry.rs:3://! Every caller-controllable SQL transaction span (`SqliteTransaction::begin_tx`,
khive-pack-brain/src/persist.rs:331:/// This deliberately does NOT use `SqlAccess::begin_tx` — per `fold_gate.rs`'s
khive-pack-brain/src/fold_gate.rs:30://! own write lock can. `SqlAccess::begin_tx` was not used for this because it
khive-pack-session/src/mirror/ingest.rs:621:    // `begin_tx` this function used before ADR-099: the whole sequence
khive-pack-session/src/mirror/ingest.rs:833:    // same all-or-nothing contract the old `begin_tx`/`tx.commit()` shape
khive-pack-session/src/mirror/ingest.rs:2641:    /// transaction (the `begin_tx` hole this ADR closes). Uses the same
khive-pack-session/src/mirror/ingest.rs:2730:             path is not routing through the shared writer task (a standalone `begin_tx` \
khive-pack-session/src/mirror/ingest.rs:2769:                // exactly the old `begin_tx`-shaped mistake this ADR
khive-db/src/stores/entity_tests.rs:636:/// Entry 9 (`SqlBridge::begin_tx`, formerly a genuine design fork on its own
khive-db/src/stores/entity_tests.rs:637:/// standalone connection) is retired as of ADR-099 D5 — `begin_tx` and

$ grep -rn "SqlTransaction" --include="*.rs" crates/
khive-pack-brain/src/fold_gate.rs:31://! returns a distinct `SqlTransaction` type and hard-errors on non-file-backed
khive-pack-session/src/mirror/ingest.rs:663:/// (not `&mut dyn SqlTransaction`) because `atomic_unit` owns the
khive-db/src/stores/entity_tests.rs:638:/// `SqlTransaction` no longer exist on `SqlAccess` (their one production
```

Every remaining match is a comment or doc-comment referencing the retired
mechanism for historical context — no code path calls `begin_tx` or names
`SqlTransaction` as a live type after this change.

## `SqlTransaction` disposition

**Removed**, not kept as a documented no-caller exemption. Reasoning: the
trait had no non-test caller anywhere in the workspace (confirmed by the
grep above), and it is not forward-deployed infrastructure the way
`khive-vcs`/`khive-merge` are (checked directly — neither references
`SqlTransaction` or `begin_tx`). Keeping an unused trait with no consumer
and no documented future caller would be dead surface area rather than
forward deployment.

## Tests

Added to `crates/khive-pack-session/src/mirror/ingest.rs`:

- `write_events_and_cursor_is_suspension_free_under_single_writer` — builds
  a bare `PoolConfig{write_queue_enabled: true, ..}` + `ConnectionPool` /
  `SqlBridge` directly (bypassing `KhiveRuntime`, which reads
  `KHIVE_WRITE_QUEUE` from the process environment at construction time —
  mutating that env var around a `KhiveRuntime::new()` call races every
  other concurrently running test in the same binary, per the documented
  precedent in `khive-pack-brain/src/fold_gate.rs` and `persist.rs`), then
  drives the converted closure through `atomic_unit` and asserts it
  resolves and persists correctly.
- `session_ingest_routes_through_writer_task_when_flag_enabled` — proves
  routing through the single writer task using the queue_depth +
  parked-occupier technique (an occupier holds the writer task's one slot
  via a oneshot channel; the call under test is spawned and its enqueuing
  is observed via `queue_depth() >= 1` before the occupier releases) — wall
  clock timing alone cannot distinguish writer-task routing from the
  flag-off fallback due to real SQLite file locking, so this is the
  rigorous proof.
- `old_shape_manual_begin_immediate_inside_atomic_unit_fails` — a
  revert-companion test: the OLD closure shape (issuing its own manual
  `BEGIN IMMEDIATE` inside the `atomic_unit` closure) fails deterministically
  with a nested-transaction error, proving the suspension-free/single-writer
  tests above are non-vacuous.
- The existing `test_mid_transaction_db_error_leaves_no_partial_state_and_cursor_unadvanced`
  regression test was adapted to build its `AtomicUnitOp` inline and assert
  on the same no-partial-state contract through the new seam.

## Test summary

```
cargo test -p khive-storage -p khive-db -p khive-pack-session --all-features
  khive-db (unit):          14 passed, 0 failed
  khive-pack-session (unit): 75 passed, 0 failed
  khive-pack-session (integration): 23 passed, 0 failed
  khive-storage (unit):      44 passed, 0 failed
  khive-storage (compliance): 33 passed, 0 failed
  khive-storage (vectors):   10 passed, 0 failed
  (+ 3 empty doc-test suites)
  Total: 456 passed, 0 failed

cargo clippy -p khive-storage -p khive-db -p khive-pack-session \
  --all-targets --all-features -- -D warnings
  clean

cargo fmt --all -- --check
  clean

cargo check --workspace --all-features --all-targets
  clean (confirms the SqlAccess trait-surface change does not break any
  other workspace crate, including khive-vcs, khive-merge, khive-pack-brain,
  khive-mcp, kkernel)
```

### DIFF
--- a/crates/khive-db/src/backend.rs
+++ b/crates/khive-db/src/backend.rs
@@ -730,43 +730,6 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn sqlite_read_only_default_transaction_rejects_writes() {
-        use khive_storage::types::SqlTxOptions;
-
-        let dir = tempfile::tempdir().unwrap();
-        let path = dir.path().join("ro_tx.db");
-
-        {
-            let writable = StorageBackend::sqlite(&path).unwrap();
-            let sql = writable.sql();
-            let mut writer = sql.writer().await.unwrap();
-            writer
-                .execute_script("CREATE TABLE ro_tx_test (id INTEGER PRIMARY KEY)".into())
-                .await
-                .unwrap();
-        }
-
-        let ro = StorageBackend::sqlite_read_only(&path).unwrap();
-        let sql = ro.sql();
-
-        // `SqlTxOptions::default()` has `read_only: false`; the read-only
-        // backend must still force the transaction to reject writes.
-        let mut tx = sql.begin_tx(SqlTxOptions::default()).await.unwrap();
-        let result = tx
-            .execute(SqlStatement {
-                sql: "INSERT INTO ro_tx_test (id) VALUES (?1)".into(),
-                params: vec![SqlValue::Integer(1)],
-                label: None,
-            })
-            .await;
-        assert!(
-            result.is_err(),
-            "INSERT inside a default tx on a read-only backend must fail"
-        );
-        tx.rollback().await.unwrap();
-    }
-
-    #[tokio::test]
     #[cfg(feature = "vectors")]
     async fn vectors_roundtrip_via_public_api() {
         let backend = StorageBackend::memory().unwrap();
--- a/crates/khive-db/src/sql_bridge.rs
+++ b/crates/khive-db/src/sql_bridge.rs
@@ -1,7 +1,9 @@
 //! SqlAccess bridge: connects `ConnectionPool` to `khive_storage::SqlAccess`.
 //!
 //! Two modes:
-//! - **File-backed**: Opens standalone connections per reader/writer/tx call (high concurrency).
+//! - **File-backed**: Opens standalone connections per reader/writer call (high concurrency).
+//!   Cross-statement atomicity goes through `atomic_unit`, which drives a single
+//!   registered raw transaction span rather than a caller-held per-tx connection.
 //! - **Memory**: Uses pool-backed approach (acquire pool connection per-query inside `spawn_blocking`).
 
 use std::any::Any;
@@ -921,7 +923,9 @@ async fn run_manual_atomic_unit(
 /// Bridges `ConnectionPool` to `khive_storage::SqlAccess`.
 ///
 /// Dispatches based on whether the pool is file-backed or in-memory:
-/// - File-backed: standalone connections per reader/writer/tx (high concurrency).
+/// - File-backed: standalone connections per reader/writer call (high
+///   concurrency); atomic units drive a single registered raw transaction
+///   span instead of a caller-held per-tx connection.
 /// - In-memory: pool-backed connections per query (single shared connection).
 pub struct SqlBridge {
     pool: Arc<ConnectionPool>,
--- a/crates/khive-db/src/sql_bridge.rs
+++ b/crates/khive-db/src/sql_bridge.rs
@@ -10,7 +10,7 @@ use std::sync::Arc;
 use async_trait::async_trait;
 
 use khive_storage::error::StorageError;
-use khive_storage::types::{SqlColumn, SqlIsolation, SqlRow, SqlStatement, SqlTxOptions, SqlValue};
+use khive_storage::types::{SqlColumn, SqlRow, SqlStatement, SqlValue};
 use khive_storage::{AtomicUnitOp, StorageCapability};
 
 use crate::error::SqliteError;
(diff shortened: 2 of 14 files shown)


ohdearquant:
Applied all three review-round-1 fixes on `8b23b8dfd630dfd4cbcd85af4d8880dcc27cfd81`:

**FIX 1 (Medium — retired transaction surface)**: `SqlIsolation`/`SqlTxOptions` had zero live non-test consumers anywhere in the workspace (verified by grep against every crate, not just the changed ones), so both are removed from `crates/khive-storage/src/types/sql.rs` along with their re-exports in `types/mod.rs` and `lib.rs`. Scrubbed the stale doc references: `tx_registry.rs`'s module doc, `khive-storage/README.md`, `khive-storage/docs/design.md`, and both `sql_bridge.rs` doc comments now describe `atomic_unit` + a registered raw transaction span as the live mechanism instead of a per-`tx` caller-held connection.

**FIX 2 (Medium — loose revert-companion assertion)**: `old_shape_manual_begin_immediate_inside_atomic_unit_fails` now builds over a write-queue-enabled `PoolConfig` literal (the existing `write_queue_pool` helper) so it deterministically drives the `InlineWriter` production path, and pins the failure text to contain `"cannot start a transaction within a transaction"` — the same substring the khive-db revert-companion test in `stores/vectors.rs` pins.

**FIX 3 (Low — lost per-statement diagnostics)**: `write_events_and_cursor_on_writer` and `upsert_cursor_on_writer` now wrap each `writer.execute` error with the original mirror-specific label (`mirror: session create`, `mirror: message insert`, `mirror: session touch`, `mirror: count refresh`, `mirror: cursor upsert`) via `StorageError::driver`, restoring the granularity the pre-conversion code had. No `SqlWriter` implementation changed — fix is local to `ingest.rs`.

Gates (same scope as before): `cargo test -p khive-storage -p khive-db -p khive-pack-session --all-features` → 456 passed, 0 failed. `cargo clippy --all-targets --all-features -- -D warnings` → clean. `cargo fmt --all -- --check` → clean. `cargo check --workspace --all-features --all-targets` → clean (the `SqlIsolation`/`SqlTxOptions` removal touches a public type surface; confirmed no fallout in any other workspace crate).

Issue #724, opened 2026-07-08
Read surface: windowed event counts (per-actor, per-kind) + section co-usage aggregates


ohdearquant:
## Problem

Two consumer designs need aggregate read surfaces that do not exist. Events are written (memory recall/remember, kg create/link, brain feedback) but there is no sanctioned aggregate read: today the only options are per-record verbs or direct SQL (banned surface).

## Ask A — windowed event counts

A read verb returning event COUNTS grouped by actor and kind over a time window, e.g.:

```
brain.events(actor?, kind?, since, until?) -> {counts: {kind: n, ...}, by_profile?: {...}}
```

- Kinds: memory.recall, memory.remember, kg create/link, brain feedback (feedback splittable by served_by_profile_id).
- Consumer: fleet flywheel-metrics reporting (recall-before rate, persist-after rate, feedback coverage per profile).

## Ask B — section co-usage aggregates

A read surface answering: which atom-section types are served together and co-earn positive section_signals feedback, over a window.

- Consumer: domain auto-generation from usage (domains as patterns of co-used, co-liked atom sections).
- Likely home: knowledge pack (section serve/feedback events already flow through compose + knowledge.feedback).

## Notes

- Both are read-only Assertive verbs; no new write paths.
- Verb surface verified 2026-07-08 (75 verbs): no existing verb covers either ask. brain.profile returns snapshot/state summary, not windowed event aggregates; knowledge.feedback is write-only.
- Consumers can carry a pending-surface placeholder; not launch-blocking.

ohdearquant:
Consolidation requirement (from the resource-management SPEC directive, same day): the resource-management subsystem's observability section lands resource events in the same event plane this issue's Ask A reads. There must be ONE windowed per-actor/per-kind event-read surface serving both the flywheel metrics consumer and resource-management observability — not sibling verbs. Design the Ask A verb shape with both consumers at the table before impl.

ohdearquant:
Closing during tracker cleanup: this concerns functionality provided by the commercially licensed extensions, which are no longer part of this repository. Tracking continues on the distribution's internal tracker.

Issue #723, opened 2026-07-08
Daemon background-work observability: CPU/mem self-report in comm.health + phase-logged background tasks


ohdearquant:
## Problem

The daemon accumulated ~514 CPU-minutes over a 12.5h window (cold start ~05:43Z to mid-afternoon), with multi-core bursts (300-800%) sustained long enough to corrupt co-tenant perf measurements on the host. Post-hoc investigation could not attribute the work from the daemon log: the log carries WARN-level checkpoint and FTS entries only. Background phases (ANN warm on cold start, embedding/index work, serving load) emit no start/end markers, so "what was the daemon doing between 06Z and 14Z" is unanswerable from the artifact that should answer it.

## Asks

1. **Phase-level INFO logging for background tasks**: ANN warm, index rebuild/backfill, and similar long-running phases log start (with corpus size), completion (with wall + CPU time), and cancellation. One line each — enough to attribute a CPU window after the fact.
2. **Daemon resource self-report in `comm.health`**: extend the health snapshot with process-level CPU time (cumulative), RSS, and current background-phase name (if any). `comm.health` is already the sanctioned read surface for daemon state; perf-sensitive co-tenants (GPU flock holders) can poll it to know the box is dirty before measuring.
3. **QoS for warm-path work**: consider lowering thread priority / niceness for cold-start warm tasks so they don't saturate the box; a cold-start warm competing with interactive serving and co-tenant measurement is the observed failure mode.

## Non-goals

Scheduling/lockfile handshakes between seats are an operational convention, not daemon code. The daemon's job is to make its own state observable.

ohdearquant:
Closing via backlog triage 2026-07-09 (verified against current main at 3d03c8f0): Both asks implemented, code comments cite issue: comm.health resource self-report (handlers.rs:1109-1168, "ADR-103 Stage 1, issue #723 ask 2"); phase-level logging (daemon.rs:550-585, ann.rs:553-590). Ask 3 (QoS niceness) was a non-blocking soft-ask, not implemented.

PR #728, opened 2026-07-08
docs(adr): ADR-103 resource attribution model

oceanwaves630:
## Summary

Proposes ADR-103: a unifying resource attribution model (actor x work_class x deterministic cost_unit) riding the existing per-dispatch audit event plane established by ADR-094.

- Closed 4-value work_class enum (interactive / warm / maintenance / inference)
- Cost sub-schema as payload enrichment of the audit row already written on every dispatch: no new per-op rows, no migration
- Phase-span EventKind variants (PhaseStarted/Completed/Cancelled) for background work, edge-triggered, extending ADR-094's additive mechanism
- Quota as Gate Obligation composition: one mechanism, two policies (hard-deny vs soft-downgrade), advisory-first staging
- Contention: pull via comm.health + voluntary TTL-bounded quiet request; the daemon does not join the external GPU lock

Rejected alternatives include a per-op resource event stream (refuted on write-load arithmetic) and ring-buffer-only accounting (loses post-hoc attribution). Stage 1 of the landing plan is exactly issues #723 + #724 as filed.

Docs-only: one new file, no code changes.

Refs #723 #724

### DIFF
--- a/docs/adr/ADR-103-resource-attribution-model.md
+++ b/docs/adr/ADR-103-resource-attribution-model.md
@@ -20,9 +20,11 @@ after the fact the question "what was the daemon doing, and on whose behalf" is
 unanswerable from any artifact that should answer it.
 
 This is not solely an internal-operations problem. The same gap exists on the commercial
-side: khive-cloud meters billable requests at the router (a two-phase reserve/finalize
-meter, `crates/operations/src/router.rs`, shipped and fail-closed on the reservation
-write). A request counter answers "how many billable requests did this tenant make," which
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
SPEC GATE: APPROVED. Read in full at content level. The resize (attribution model riding the ADR-094 event plane, not a new subsystem) is the right call; work_class-as-attribution vs qos_posture-as-scheduling separation is clean; the embedder-attribution measurement spike gating cost_unit weights before any billing-facing use is exactly the right discipline. Open Q3 ruling below in comm.

PR #731, opened 2026-07-08
feat(mcp): self-heal bridge protocol mismatch via in-place re-exec

### LINKED ISSUE #714: MCP bridge: self-heal protocol-mismatch after daemon rebuild (re-exec on-disk binary)
## Problem

A daemon rebuild (`make local`) replaces the installed binary and bounces the daemon. The stdio MCP bridge already self-heals the plain-bounce case: a dead daemon socket is a `NoSocket` outcome, and `forward_or_spawn` recovers via `kill_and_respawn` and resends (daemon.rs, #644-safe: real frame written at most once).

What it cannot heal is **version skew**: the long-lived stdio bridge process keeps running the *old* binary. After a rebuild it spawns or reaches the *new* daemon, detects `ProtocolMismatch`, and returns a hard error advising a rebuild — but the actual remedy is restarting the bridge process, which only the MCP client (a per-session manual reconnect) can do. In a multi-session deployment this leaves every session's bridge dead until each one is manually reconnected. Observed: a 02:00 rebuild left sessions without a working MCP surface for ~9 hours.

## Proposed mechanism

On detecting `ProtocolMismatch`, the bridge **re-execs the current on-disk binary in place**, preserving its stdio file descriptors:

1. Finish the in-flight request with the existing ambiguous-forward error (unchanged; the mismatch may be discovered after the frame was written, and #644's no-retry rule stands).
2. Before reading the next MCP request from stdin, `exec()` the installed binary (same argv/env). The MCP client keeps the same pipe fds and sees an uninterrupted stdio server; the fresh process carries the new `PROTOCOL_VERSION` and reconnects cleanly.
3. Guard rails: re-exec at most once per detected mismatch generation (no exec loop if the on-disk binary is itself stale); resolve the binary path at exec time, not from argv[0] cache; log the re-exec with old/new version so the window is observable.

The bridge is a stateless dispatch shell (server.rs), so no session state is lost across the exec. MCP protocol state held by the client (initialize handshake) needs verification: if the client does not tolerate a silent server restart mid-session, fall back to draining stdin and letting the harness's own respawn handle it — that fork is part of this design work.

## Scope

- Design note/ADR for the bridge lifecycle change (touches the #644 ambiguity contract only at the 'after responding' seam, not the retry rules).
- Implementation in khive-mcp daemon/client seam + regression: old-bridge/new-daemon mismatch heals without client action.
- Interim (docs, immediate): sanctioned narrow CLI fallback during announced rebuild windows.

## Acceptance

After a rebuild that bounces the daemon, an existing MCP session's next request either succeeds (post-re-exec) or fails once with the ambiguous error and then succeeds — with no manual per-session reconnect.

oceanwaves630:
## Problem

When the warm daemon (`kkernel mcp --daemon`) is upgraded (e.g. `make local`) but a stdio bridge client (`kkernel mcp`) was started against the old protocol version, every `request` call on that bridge fails forever with a hard "daemon protocol mismatch" error. The daemon itself is fine — the bridge is the stale side — but recovery today requires the operator to notice and manually restart the client process.

## Mechanism

On a `ProtocolMismatch` outcome, the bridge now self-heals by re-exec'ing the freshest on-disk binary in place, instead of just returning the hard error and waiting for a manual restart.

**Ordering (load-bearing):** the mismatch error is returned to the client through the normal MCP response pipeline first; the re-exec is deferred behind a short grace period on a background task (`tokio::spawn` + `sleep`), so the in-flight response has time to reach the transport before `exec()` replaces the process image. Calling `exec()` synchronously inside the request handler — before the response is serialized and flushed — discards the in-flight response and the caller sees a timeout instead of the error. POSIX `exec()` preserves the process's PID and open stdin/stdout/stderr file descriptors, so the client's stdio transport never observes an EOF or reset across the swap.

**Skipping the handshake on resume:** the re-exec'd process carries a `--resumed-generation=N` marker across `exec()` via argv. On a resumed generation, `serve_stdio` uses `rmcp::service::serve_directly` instead of the normal `.serve()` handshake, since the connected peer already completed a real MCP `initialize` handshake with the prior generation over the same, uninterrupted stdio pipe — waiting for a second one would hang forever.

**Loop-breaker (exec-once guard):** a resumed generation that itself observes another `ProtocolMismatch` takes a drain-and-exit fallback rather than re-exec'ing a second time (also used as the direct fallback on non-unix targets, where `exec()` self-heal isn't available).

## Protocol-version consumer check

`serve_directly` skips MCP's `initialize` handshake, so the resumed generation's `PeerInfo` (`ClientInfo`) is `None` instead of the value a real handshake would have populated. Before relying on that, I grepped the entire workspace for anything reading the negotiated protocol version or peer info:

```
$ grep -rn "protocol_version()" crates/ --include="*.rs"
(no matches)

$ grep -rn "RequestContext" crates/ --include="*.rs"
crates/khive-mcp/src/server.rs:1644:    _context: rmcp::service::RequestContext<rmcp::RoleServer>,
# (unused parameter in the list_tools override — never read in the body)

$ grep -rn "peer_info()" crates/ --include="*.rs"
(no matches)

$ grep -rn "PeerInfo" crates/ --include="*.rs"
(no matches)
```

No call site anywhere in the workspace depends on the negotiated `PeerInfo`/protocol version — `None` on the resumed path is safe.

## Concurrency edge case (documented, not fixed)

If the bridge is mid-flight on more than one outstanding client request when the mismatch fires, only the request that triggered the mismatch arm gets the error-then-resume treatment; any other in-flight request loses its response the same way it would if the process crashed. This is a pre-existing risk, not introduced by this change, and is called out at the `trigger_bridge_self_heal` doc comment in `daemon.rs`.

## Test coverage

- 10 new unit tests in `khive-mcp/src/daemon.rs`: argv-marker parsing (absent/present/malformed/last-occurrence-wins), the loop-breaker decision function, and deterministic (`start_paused = true`) assertions that both the re-exec and drain-and-exit actions are deferred past the grace period and never fire synchronously.
- 2 new unit tests in `khive-mcp/src/server.rs` for the `StdioServeMode` decision function.
- 1 new live-process integration test (`crates/kkernel/tests/mcp_bridge_reexec_protocol_mismatch.rs`) that spawns the real compiled `kkernel` binary via `TokioChildProcess`, drives it with a real `rmcp` client over its actual stdio pipes, and connects it to a hand-rolled fake daemon socket that serves a stale-protocol response followed by a matching one. This is the one test in the suite that exercises the real, unstubbed `exec()` — everywhere else, `exec()`/`process::exit()` are `#[cfg(test)]`-gated counters (matching the existing `KILL_COUNT`/`SPAWN_COUNT` pattern in `daemon.rs`) since a real `exec()` inside `cargo test`'s own process would replace/kill the test binary. Asserts: the first call fails with a protocol-mismatch error, the second call succeeds on the exact same client session/transport (no reconnect), and the child process PID never changes.

Full local gate results (from `crates/`):

```
cargo fmt --all -- --check                                          # pass
cargo clippy --workspace --all-targets -- -D warnings                # pass
cargo test -p khive-mcp                                              # 149 lib + 109 integration, all pass
cargo test -p kkernel                                                # 250 + 16 + 14 integration tests, all pass
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-mcp           # pass
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p kkernel             # pass
```

Closes #714


ohdearquant (approved):
Approved at fb312503 per merge flow v2.1. Verified the amendment delta from codex-approved 40bcc31f is solely the internal report file removal (335 deletions, zero code change), so codex r2 APPROVE carries. Hygiene catch appreciated — right call archiving it out of the branch.

Issue #676, opened 2026-07-07
Enrich link-verb audit events with relation and resolved edge id


ohdearquant:
## Context

The \`events\` audit table records one row per dispatched verb (actor, namespace, verb, outcome, duration). For \`link\` events, however, \`target_id\` is empty and \`payload\` carries only the gate decision (\`actor\`, \`decision\`, \`gate_impl\`, \`namespace\`, \`verb\`). Nothing in the event identifies which edge was created, what relation it used, or which endpoints it connected.

## Problem

Per-actor auditing of graph writes is not possible from the audit log alone. A concrete consumer exists today: measuring annotation coverage on judgment-bearing edges (supersedes / supports / refutes / weight < 1.0) per actor requires knowing, from events, that a link of a given relation was created and which edge id it produced. Currently the metric can only be computed from the edges table, where attribution is limited to the namespace write-stamp rather than the acting principal.

## Proposal

On successful \`link\` dispatch, populate the event's \`target_id\` with the created edge UUID and include in \`payload\` (or \`data\`): \`relation\`, \`source_id\`, \`target_id\` (endpoint), and \`weight\`. Failed links keep current shape. Payload schema version bumps accordingly.

## Acceptance

- \`SELECT target_id, payload FROM events WHERE verb='link' AND outcome='success'\` returns the edge id and relation for a fresh link.
- Existing event consumers unaffected (additive fields, schema version bumped).
- Regression test asserting the enriched payload for one successful and one failed link.

PR #732, opened 2026-07-08
feat(daemon): audit duration_us + phase telemetry + comm.health resource self-report (ADR-103 Stage 1, #723)

oceanwaves630:
Stage 1, PR-1 of ADR-103 (resource attribution): daemon observability groundwork. Covers issue #723 asks 1 and 2 only; ask 3 (QoS/niceness) and issue #724 are separate follow-ups.

## What

1. **Measured `duration_us` on persisted audit rows.** Generalizes the existing deferred-audit mechanism (previously singleton `link` only, #676) so every Allow-outcome dispatch persists its audit row after dispatch completes, carrying the real measured dispatch time. Denied calls keep the immediate append with `duration_us=0` (no dispatch runs). The unknown-verb fallthrough now explicitly flushes a deferred row so no audit row is dropped. Regression tests for both paths.

   Semantic note: Allow-outcome rows are now written post-dispatch, so a process crash mid-dispatch loses that row (previously a pre-dispatch row with `duration_us=0` would survive). Denied rows are unaffected. This trade is deliberate: the row's value is attribution, and an unattributable zero-duration row is what ADR-103 Stage 1 exists to eliminate.

2. **Phase-span telemetry (#723 ask 1).** Additive `EventKind::PhaseStarted/PhaseCompleted/PhaseCancelled` variants with typed payloads; ANN warm (`ensure_ann_for_model`, the chokepoint for cold-start and on-demand warm) now emits start (corpus size), completion (wall + CPU time), and cancellation markers at INFO.

3. **`comm.health` resource self-report (#723 ask 2).** New `resource: {cpu_us, rss_bytes, active_phases}` object in the health snapshot, read via `getrusage(RUSAGE_SELF)` (no new dependency). Preserves the existing contract: raw observations only, never a computed health judgment.

Out of scope, flagged for follow-up: the full ADR-103 `resource` payload enrichment (`work_class`, `cost_unit`), and `khive-pack-knowledge`'s own ANN warm path (`warm_known_snapshots`), which is structurally similar but not instrumented here.

## Gates

fmt, clippy `-D warnings`, per-crate tests (khive-types, khive-storage, khive-runtime, khive-pack-memory, khive-pack-comm), full workspace test run, doc build with warnings denied — all green.

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -2926,6 +2926,7 @@ id = "lambda:project-actor"
     /// `multi_backend_boot_paths_share_identical_wiring_surface` test, which
     /// exercises the actual coordinator branch.
     #[test]
+    #[serial]
     fn kkernel_multi_backend_path_wires_pool_for_file_backed_main() {
         let dir = tempfile::tempdir().expect("temp dir");
         let main_path = dir.path().join("main.db");
@@ -2959,6 +2960,7 @@ id = "lambda:project-actor"
     /// exercises `build_server_from_multi_backend_registry` — see the note on the
     /// sibling test above.
     #[test]
+    #[serial]
     fn kkernel_multi_backend_path_leaves_pool_none_for_in_memory_main() {
         let khive_cfg = KhiveConfig {
             backends: vec![BackendConfig {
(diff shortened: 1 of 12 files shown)


ohdearquant (approved):
Approved. Verified at content level: the Stage-1 clarification carries both required elements — the event log is named as the attribution/cost-accounting plane with the daemon process log as the crash-forensics surface, and option B (pre-dispatch append + narrow finalize seam) is recorded as the findable upgrade path without being built. The 'preserved misattribution that a crash prevents from ever being corrected' framing is exactly the right articulation of why A wins. Codex r2 clean carries the mechanical findings.

PR #736, opened 2026-07-08
fix(make): local target spares stdio bridges so #714 self-heal can run

oceanwaves630:
Follow-up to #731. The local target's broad kkernel process kill terminated live stdio bridges along with the daemon — which means the #714 self-heal re-exec path was dead on arrival at every rebuild: a bridge cannot survive its own SIGTERM. Tonight's rebuild proved it (bridge killed, harness-level reconnect needed; SelfHealOnFlushTransport never fired).

This narrows the kill set to the daemon process only. Bridges stay alive, hit the generation mismatch on their next request, and re-exec into the new binary in place — the designed #714 acceptance scenario. Transient CLI exec invocations are also no longer killed mid-flight as a side effect.

Risk note: the broad kill guarded against stale processes holding SQLite WAL. Bridges are daemon clients post fail-loud dispatch (#663) and do not hold the DB; the daemon — the process that does — is still killed and waited on exactly as before.

Acceptance: the NEXT rebuild after this merges is the live #714 item-6 test — bridge sessions should continue with zero /mcp or harness reconnect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,13 @@ local:
 	SRC_HASH=$$(md5 -q "$$SRC"); \
 	SRC_SIZE=$$(stat -f '%z' "$$SRC"); \
 	echo "==> Source:  $$SRC ($$SRC_HASH, $$SRC_SIZE bytes)"; \
-	echo "==> Killing running kkernel daemon (bridges self-heal via re-exec on next request)..."; \
+	echo "==> Staging + codesigning $$DEST.new..."; \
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
Approved — diff read myself: kill step moved after the atomic install (closes the old-path respawn hole codex r1 caught) and the kill pattern narrowed to 'kkernel mcp --daemon' so bridges survive to hit the mismatch and re-exec. Item-6 acceptance re-arms on the first make local after this merges; report the verdict either way.

PR #737, opened 2026-07-08
feat(brain): windowed event-counts read verb brain.event_counts (ADR-103 Stage 1, #724 Ask A)

oceanwaves630:
ADR-103 Stage 1 PR-2: the windowed, per-actor, per-kind read verb over the event plane (#724 Ask A).

**Surface**: `brain.event_counts(since, until?, actor?, kind?)` — public read-only Assertive verb returning `total`, `counts_by_kind`, `counts_by_actor`, an optional `by_profile` split for explicit-feedback events (keyed on `served_by_profile_id`, missing values bucketed as "unspecified"), and `truncated`/`window_event_total` when the window exceeds the 50,000-row aggregation cap. Window is half-open `[since, until)`; `until` defaults to now.

**Naming**: the issue sketches `brain.events(...)`, but that name is an existing internal `Visibility::Subhandler` debug verb that MCP-boundary tests require to stay blocked from the public surface. Shipped as `brain.event_counts`; the subhandler is untouched.

**work_class/cost_unit** (ADR-103 mentions surfacing them): `cost_unit` does not exist in code yet; `work_class` exists only inside the three Phase* payloads, not as a cross-kind queryable dimension — so no grouping axis for them in this PR; they join when the fields land generally (later Stage 1/2 scope).

**Storage**: no EventStore trait changes — filters push to SQL via the existing `EventFilter`, aggregation is in-handler over a bounded page. Tradeoff documented in code.

**Tests**: 9 handler-level tests (window boundaries, filters, by_profile split, invalid kind/datetime errors, empty window = zero counts). Gates green: fmt, clippy -D warnings, scoped tests (khive-pack-brain/storage/types), doc -D warnings; MCP-boundary invariant tests re-run and pass.

Docs: verb totals updated in AGENTS.md and CLAUDE.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,7 +15,7 @@ khive gives your agent:
 9. **Brain** — Bayesian profile tuning from feedback signals
 10. **Session** — persist and resume agent-session records
 
-All 9 packs load by default. **75 public verbs** across the packs — the `git` pack
+All 9 packs load by default. **76 public verbs** across the packs — the `git` pack
 (commit/issue/pull_request provenance notes, populated by a batch ingester, not agent-facing
 verbs) contributes no new verb (regenerate via `request(ops="verbs()")` before editing this
 line).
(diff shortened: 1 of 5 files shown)


ohdearquant (approved):
Approved — read-only windowed counts verb over the event plane per ADR-103 Stage-1 text; diff spot-read confirms pure reads (EventFilter + in-memory aggregation), half-open window semantics documented at the off-by-one, and the dual-path work_class test covers the future resource fallback. Post-merge make local = the real item-6 field test; report the verdict either way.

PR #738, opened 2026-07-09
feat(memory): auto-extract entity_names from recall query when caller omits them

oceanwaves630:
## What

`memory.recall`'s `entity_names` parameter feeds an EntityMatch ×1.3 rank-score boost in `calculate_score`'s default adjustments, but no caller ever populated it — the boost was dead code in practice. This PR derives entity candidates server-side from the query when the caller omits the parameter.

## Behavior

| `entity_names` | capitalized query | all-lowercase query |
|---|---|---|
| omitted (`None`) | auto-extracts capitalized non-stopword tokens (max 8); boost fires on word-boundary matches | extracts nothing; no boost |
| `[]` (explicit empty) | opt-out — no boost | no boost |
| non-empty list | honored verbatim, never re-derived | same |

- Extraction is capitalized-tokens-only by design: EntityMatch matches free text (not KG entity records), so lowercase content words would turn the boost into a generic lexical-overlap reward that can clamp and flatten top ranks. Lowercase/agent-query coverage requires anchoring candidates against known entity records — out of scope here, tracked as follow-up design work.
- EntityMatch/EntityMiss now match at word boundaries instead of raw substring `contains` ("beta" no longer matches "alphabet"; multi-word names still match on phrase boundaries).

## Measured effect

- Scoring-layer isolation: equal-relevance inputs with vs without an entity match score 0.5505 vs 0.4235 — exactly the configured ×1.3.
- Handler-level: same corpus/query across omitted vs explicit-empty `entity_names` asserts the 1.3 ratio end-to-end.
- Adds `bench/recall_eval.jsonl` + `bench/README.md`: a synthetic recall-quality eval fixture (rows are sanitized mirrors of private eval cases; the fixture grows as eval rows accumulate).

## Tests

12 new tests: extraction rules (capitalization, stopwords, punctuation, dedup, cap at 8, all-lowercase yields nothing), word-boundary regressions, opt-out semantics, and a non-vacuous handler wiring test that fails if auto-extraction is removed.

Gates: `cargo fmt --check`, `cargo clippy -p khive-pack-memory --all-targets -- -D warnings`, `cargo test -p khive-pack-memory` (197 lib + 82 integration) — all clean.


ohdearquant (approved):
APPROVE (Leo gate). Verified in diff: (1) capitalized-tokens-only extraction with conservative stopword set — sentence-initial 'The' correctly dies at the stopword check after lowering; (2) Some([]) honored as explicit opt-out with a test pinning it to the never-boosted baseline within 1e-4; (3) char-boundary matching is unicode-safe and closes the substring false-positive class while preserving multi-word explicit names; (4) the same-corpus-same-query ×1.3 ratio test isolates the scoring-stage effect — it fails if auto-extraction stops firing, which the earlier two-note version did not guarantee. Honest-scope note accepted: all-lowercase (and CJK) queries get nothing from this PR by design; the KG-anchored candidate leg belongs in the posterior-serving ADR, not here — confirmed, keep it out of this PR.

PR #739, opened 2026-07-09
docs(adr): record embedder-attribution spike resolution in ADR-103

oceanwaves630:
Records the resolution of ADR-103 open question 1 (per-actor embedder-CPU attribution): verdict NOT-CAPTURED, with the three measured facts as rationale — inline-on-polling-thread inference, process-wide-only CPU capture that over-attributes with concurrent occupancy, and task thread-migration across awaits. cost_unit weights finalize as deterministic op-class weights per the section's anticipated fallback.

Docs-only amendment; no code changes.


ohdearquant (approved):
APPROVE (Leo gate). Amendment matches the spike verdict I accepted: three measured facts (inline encode_batch — no thread escape; no per-thread CPU capture anywhere, only process-wide getrusage; ~2x over-attribution under 2-worker concurrency + thread migration across await breaking any wider bracket) and the deterministic op-class-weights consequence recorded where the open question lived. Docs-only, scope clean.

PR #740, opened 2026-07-09
docs(adr): ADR-104 profile posteriors in the recall read path

oceanwaves630:
New decision ADR for the recall-quality program. The brain pack accumulates per-profile Beta posteriors and the memory pack implements the tunable projection surface, but nothing serves them: recall rankings are identical across profiles, and per-entity posteriors never reach scoring.

ADR-104 closes the loop per-request at serve time — no global mutation, no background loop:

1. Serve-time projection of profile posterior weights (pure-function use of the existing PackTunable path)
2. Bounded per-entity posterior term: rank_score ×= clamp(1 + w_ent·(mean − 0.5), 0.85, 1.15), neutral without evidence
3. include_breakdown gains profile_component + entity_posterior_mean per hit
4. Optional profile_id resolve-override on memory.recall for profile A/B evaluation
5. KG-entity-anchored candidate extraction (precision-safe lowercase/CJK coverage the free-text path cannot deliver)

Staged landing A/B/C with per-stage gates. Spec-gated: awaiting sign-off before any dependent implementation.

### DIFF
--- a/docs/adr/ADR-104-posterior-serving-recall.md
+++ b/docs/adr/ADR-104-posterior-serving-recall.md
@@ -1,6 +1,6 @@
 # ADR-104: Profile Posteriors in the Recall Read Path
 
-**Status**: Proposed
+**Status**: Accepted (signed 2026-07-08, with riders R1 and R2 below)
 **Date**: 2026-07-08
 **Depends on**: ADR-021 (memory pack), ADR-081 (serve ledger + profile stamping), ADR-033
 (recall scoring), issue #159 (brain-tunable parameter surface)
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
SPEC-GATE: SIGNED (Leo, Fable seat). Scope matches the approved direction exactly — five components, nothing smuggled in. What earns the signature: serve-time projection as a pure function (no apply_config mutation, no phantom loop, and brain.activate's description finally matches its handler); the per-entity term is bounded by construction (clamp ±15%, neutral at uninformative prior so unfeedbacked memories are untouched); staged landing with a falsifiable gate per stage; the three open questions are honest and correctly deferred to eval evidence. On open question 2: keep NO ESS floor for Stage B as drafted — a ≥2-observation floor would defeat B's own acceptance test (one useful signal must visibly lift rank); add the floor only if evals show single-signal noise.

Two riders, binding at implementation PRs:

R1 (Stage C): the CJK/lowercase entity-anchored leg must document its lookup strategy and cost bound in the Stage C PR — index-backed token/bigram lookup, and for unsegmented CJK an explicitly bounded scan (state the bound), never an unbounded per-recall scan over entity names.

R2 (Stage A): the "negligible overhead" expectation becomes a measured number in the Stage A PR (per-recall latency with vs without the profile-state read), not an assertion; B does not land until that number is in the record.

Merge the docs PR; staged implementation may proceed A→B→C per the gates written in the ADR.

PR #742, opened 2026-07-09
fix(brain): default brain.resolve actor from the caller's dispatch identity (#741)

### LINKED ISSUE #741: brain.resolve does not default actor from the caller's dispatch identity
## Problem

`brain.resolve` is the only resolution surface that ignores the caller's identity: `handle_resolve` (khive-pack-brain/src/handlers.rs) deserializes an optional `actor` param but receives no `NamespaceToken`, so an omitted `actor` resolves as the wildcard caller. A seat with an explicit actor binding calling `brain.resolve(consumer_kind="recall")` gets `matched_binding: false` + the balanced fallback — while the same seat's `memory.recall` (post-#708) correctly resolves and stamps its bound profile.

This makes the introspection verb lie about what the serve path does: the operator diagnostic disagrees with live serving behavior.

## Evidence (live, 2026-07-08)

From a seat with an active `actor=<seat>` / `consumer_kind=recall` binding:
- `brain.resolve(consumer_kind="recall", actor="<seat>")` → `matched_binding: true`, bound profile ✔
- `brain.resolve(consumer_kind="recall")` → `matched_binding: false`, `balanced-recall-v1` ✘ (should default to the caller)
- `memory.recall` through the same MCP dispatch → `served_by_profile_id: <bound profile>` ✔

## Fix

`handle_resolve` gains the dispatch `NamespaceToken`; when the `actor` param is omitted, default to `token.actor().binding_id()` (same rule as the serve path — anonymous callers stay `None` and never match explicit bindings, #708 invariant). Explicit `actor` param continues to win (evaluation tooling needs to query other identities).

Regression test: a bound actor's no-arg `brain.resolve(consumer_kind="recall")` through the dispatch path must return its bound profile with `matched_binding: true`.

## Related
- #699 fixed the feedback path (was hardcoding balanced fallback)
- #708 fixed the recall serve path (anonymous never matches explicit bindings)
- Historical note: profile event counts show heavy pooling on `balanced-recall-v1` accrued before #699/#708 landed (2026-07-07); that is pre-fix data, not a live defect.

oceanwaves630:
Closes #741.

## What

`brain.resolve` was the only resolution surface that ignored the caller's identity: `handle_resolve` received no dispatch token, so an omitted `actor` param resolved as the wildcard caller — a seat with an active actor binding got `matched_binding: false` + the balanced fallback from the introspection verb while its live `memory.recall` (post-#708) correctly stamped the bound profile. The diagnostic disagreed with production, which poisons exactly the audit path that caught this.

## Fix

`handle_resolve` now takes the dispatch `NamespaceToken`; an omitted `actor` defaults to `token.actor().binding_id()` — the same rule as the serve path. Anonymous callers stay `None` and match only wildcard bindings (#708 invariant). An explicit `actor` param wins so evaluation tooling can query other identities.

## Riders (description truthfulness, same crate)

- `brain.activate` description no longer claims a "live update loop" (none exists; serving reads profile state per-request, per ADR-104).
- `brain.reset` description + `exploration_epoch` field doc now state it is a reset counter and nothing else; the `reset_epoch` rename is deferred to the posterior-serving lane as a versioned output-shape change.

## Tests

3 new integration tests (`tests/resolve_actor.rs`): bound caller's no-arg resolve matches its binding end-to-end through pack dispatch; explicit `actor` param overrides dispatch identity; anonymous caller never matches an explicit binding.

Gates: `cargo fmt --check`, `clippy -p khive-pack-brain -p khive-brain-core --all-targets -- -D warnings`, `cargo test -p khive-pack-brain -p khive-brain-core` — all exit 0.


ohdearquant (approved):
Diff read in full. Token threaded into handle_resolve, omitted actor defaults to dispatch identity via binding_id() (anonymous stays None — the explicit anonymous-vs-'local' regression test is exactly the right guard for the #708 invariant), explicit param wins. exploration_epoch/activate/reset descriptions now match reality per the agreed plan. My kv repro is regression test 1. Approved.

PR #743, opened 2026-07-09
feat(memory): serve-time profile projection in recall (ADR-104 Stage A)

oceanwaves630:
Implements ADR-104 Stage A (components 1, 3, 4). Component 2 (per-entity posterior term) is Stage B and is deliberately absent.

## What

- **Serve-time projection (§1)**: `memory.recall` resolves the serving profile once, pre-scoring (same value later stamped as `served_by_profile_id` and ledgered — the projection and the stamp cannot drift apart). The profile's `BalancedRecallState` is projected through `PackTunable::project_config` as a pure function onto this request's scoring weights. No `apply_config`, no config mutation, no background loop. Unresolvable profile or state-read failure degrades to configured defaults with a WARN; never fails the recall.
- **`profile_id` override (§4)**: optional request param short-circuits binding resolution; the named profile serves, stamps, and ledgers identically. Unknown/invalid id is a hard per-op error, not a silent fallback.
- **Breakdown fields (§3)**: `include_breakdown=true` responses gain `profile_component` (projected-vs-default score ratio per hit; 1.0 when no profile served; costs a second `calculate_score` only on verbose responses) and `entity_posterior_mean` (read-only; absent without evidence; asserted NOT to affect ranking in Stage A).
- Regression guard: a test proves the profile that projects weights is the one `brain.bind` resolves through the actor-scoped dispatch path (#699/#708), so a resolution regression fails Stage A acceptance instead of hiding behind the override.

## R2 measured overhead (binding sign-off rider)

Same corpus + query, N=150 iterations per arm (`adr104_r2` ignored bench test):

```
without profile-state read: median=8357.0us  p95=16615.0us
with profile-state read:    median=8440.0us  p95=16847.0us
delta:                      median=  83.0us  p95=  232.0us
```

Median per-recall overhead of the profile-state read is **~83µs (~1%)** — negligible as the ADR predicted. This number is the Stage B unblock record.

## Tests

7 new (6 correctness + 1 `#[ignore]` R2 bench): profile-differentiated ranking (two profiles → different orderings, fails if projection is removed), no-profile path byte-identical to defaults, override stamping + ledger + invalid-id error, breakdown neutrality/presence/rank-inertness, determinism, and the end-to-end binding-resolution guard.

Gates: `cargo fmt --check`, `clippy -p khive-pack-memory --all-targets -- -D warnings`, `cargo test -p khive-pack-memory` (203 lib + 82 integration) — all exit 0, re-verified independently after implementation.

### DIFF
--- a/crates/khive-pack-memory/src/config.rs
+++ b/crates/khive-pack-memory/src/config.rs
@@ -521,6 +521,18 @@ pub struct ScoreBreakdown {
     pub temporal: f64,
     /// Weighted contributions summing to the total score.
     pub weighted: WeightedContributions,
+    /// ADR-104 §3: multiplicative contribution of serve-time profile-weight
+    /// projection relative to configured-default weights for this hit —
+    /// `score_with_projected_weights / score_with_default_weights`. `1.0`
+    /// when no profile served the request (component 1 never ran) or the
+    /// default-weight score is ~0 (div-by-zero guard).
+    pub profile_component: f64,
+    /// ADR-104 §3: the serving profile's per-entity Beta posterior mean for
+    /// this hit's UUID. `None` when no profile served the request, or the
+    /// profile holds no posterior for this UUID beyond the uninformative
+    /// prior. Reported only — Stage A does not let this affect ranking
+    /// (the per-entity multiplicative term is component 2, Stage B).
+    pub entity_posterior_mean: Option<f64>,
 }
 
 impl ScoreBreakdown {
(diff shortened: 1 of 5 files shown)


ohdearquant (approved):
Stage A verified at content level: projection is a pure per-request function (scoring_cfg is a request-local clone of cfg.scoring; default path guarded byte-identical with-or-without brain pack), the ranking-flip test proves the projection is load-bearing for ordering, profile_id override errors hard on unknown ids, and entity_posterior_mean stays read-only (Stage B not smuggled in). R2 rider satisfied: +83us median / +232us p95, N=150 both arms, in the PR record — Stage B is unblocked per ADR-104's gate. Approved.

PR #745, opened 2026-07-09
feat(memory): ADR-104 Stage B — bounded per-entity posterior term in recall scoring

oceanwaves630:
Implements ADR-104 §2 (Stage B of the staged landing): one multiplicative term after rank_score computation, `clamp(1 + 0.3*(entity_posterior_mean - 0.5), 0.85, 1.15)`.

- Neutral (exactly 1.0) when the serving profile holds no posterior for the candidate — and at the uninformative prior mean itself. No minimum-evidence floor, per the resolved ADR open question.
- No profile serving the request → identity term for every hit → ranking unchanged vs pre-B.
- Single `HashMap::get` per candidate against the profile state Stage A already fetches once per recall — no second profile-state read (preserves the Stage A R2 overhead envelope: +83µs median).
- `entity_posterior_mean` breakdown field now populates with the value used; absent when no posterior.
- New pure-function unit tests (neutrality, clamp bounds, monotonicity) + row-B gate integration tests: feedback-lift (one `useful` signal changes next-query rank under that profile only), neutrality isolation, end-to-end clamp saturation.

Gates: fmt clean; clippy -D warnings clean (khive-pack-memory, khive-brain-core); tests 213 lib + 82 integration + 50 brain-core, all green.

### DIFF
--- a/crates/khive-pack-memory/src/config.rs
+++ b/crates/khive-pack-memory/src/config.rs
@@ -530,8 +530,13 @@ pub struct ScoreBreakdown {
     /// ADR-104 §3: the serving profile's per-entity Beta posterior mean for
     /// this hit's UUID. `None` when no profile served the request, or the
     /// profile holds no posterior for this UUID beyond the uninformative
-    /// prior. Reported only — Stage A does not let this affect ranking
-    /// (the per-entity multiplicative term is component 2, Stage B).
+    /// prior. As of Stage B (§2), this value is not just reported: it is
+    /// the input to the bounded per-entity term
+    /// `clamp(1 + 0.3 * (entity_posterior_mean - 0.5), 0.85, 1.15)` that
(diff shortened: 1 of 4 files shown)


ohdearquant (approved):
Stage B verified at content level: the per-entity term is applied exactly once, as the LAST step on final_score — which correctly covers both the default path and weighted_rerank (the r1 High would have left the visible-loop contract silently dead for any reranker_weights caller; good catch and correct fix). Row-B gate is honored precisely: neutrality test pins fresh-profile == defaults to 1e-9, feedback-lift proves one useful signal moves rank under the served profile ONLY (both control arms unchanged), clamp proven end-to-end through the handler not just the pure function. No ESS floor, per the pre-ruling — one signal lifts, as the gate demands. Breakdown docs now honest about rank_score exceeding [0,1]. Approved.

Issue #750, opened 2026-07-09
ANN warm-cache race: fire-and-forget rebuild can permanently install a mid-write corpus snapshot


oceanwaves630:
## Bug

The per-model ANN warm cache in `crates/khive-pack-memory/src/ann.rs` (`ensure_ann_background` / `ensure_ann_for_model`) has a race: when multiple rebuilds are queued, whichever wins the per-model lock first installs its index permanently via `entry(key).or_insert(...)` — even if it snapshotted the corpus mid-write. Later, more complete rebuild attempts are discarded by `or_insert`, so the stale index persists for the process lifetime.

## How it was found

Intermittent (~1-in-15/20 runs) failures in namespace-scoped recall tests during the #733 lane: a memory written before recall was absent from ANN results because the warm index had captured the corpus before the write committed and was never replaced. Reproduced on a test predating the #733 changes; confirmed unrelated to the namespace filter via targeted debug capture (full trail in the #733 PR's implementation notes).

## Impact

Production-visible in principle for any freshly-written corpus: recall served from a warm ANN index built from an incomplete snapshot stays incomplete until process restart or an explicit reindex, silently degrading recall completeness. Distinct from the documented async ANN warm-up latency — this is a permanently-stale installed index, not a not-yet-warm one.

## Fix direction (needs design care)

Replace `or_insert` semantics with versioned/generation-checked installation (a rebuild that observed a newer corpus generation replaces the entry), or re-queue a rebuild when writes land after snapshot-begin. Tests in the #733 lane currently work around it with a bounded fresh-runtime retry; those retries should be removable once this is fixed.

## Severity

Medium-High: silent recall-completeness degradation under write-then-read workloads; flaky-test generator.

Issue #733, opened 2026-07-08
Namespace-scoped reads: namespace param on memory.recall + feedback event scoping (+ knowledge.compose)


ohdearquant:
## Problem

Benchmark/evaluation harnesses need a frozen, isolated store per measurement arm: writes go to a dedicated namespace so arm N cannot be contaminated by arm N-1 or by live activity. The write side already supports this (`memory.remember` accepts `namespace`), but the read side does not:

- `memory.recall` has no `namespace` param — a per-arm namespace can be written but not read back in isolation.
- `brain.auto_feedback` cannot scope its event away from the live posteriors, so serving feedback from a bench arm trains the production brain.
- `knowledge.compose` likewise has no namespace filter (lower priority).

Verified against the live verb surface 2026-07-08: none of the three accept `namespace`.

## Ask (preference order)

1. `namespace` param on `memory.recall` (filter recall to one namespace) and on `brain.auto_feedback` (scope the emitted event). This alone unblocks the primary use case.
2. Same for `knowledge.compose` if cheap.
3. Snapshot-pinned recall/compose (#726) remains the ideal end state; this is the pragmatic subset.

## Notes

- ADR-007 Rev 6: multi-record ops default to `WHERE namespace='local'` with an explicit `namespace=` escape — recall is a multi-record read, so a verb-surface `namespace` arg aligns with the existing model rather than extending it.
- Feedback scoping needs a design check: whether the event row's namespace stamp is enough to keep bench events out of live posterior updates, or whether the brain update loop needs a namespace filter too.

ohdearquant:
Closing during tracker cleanup: this concerns functionality provided by the commercially licensed extensions, which are no longer part of this repository. Tracking continues on the distribution's internal tracker.

PR #751, opened 2026-07-09
feat(memory): optional exact-match namespace param on recall + bench arm-isolation recipe (#733)

oceanwaves630:
Implements #733 slice 1 (the read-side escape hatch; `brain.auto_feedback` event scoping remains open on the issue).

## What

- `memory.recall` gains an optional exact-match `namespace` param. Absent → byte-identical to prior behavior (token namespace, default `local`, per ADR-007 Rev 6's multi-record-op model). Present → the entire read pipeline (FTS + vector candidate fetch, the ANN over-fetch retry loop's visibility gate, the supersedes graph read, and the serve-ledger namespace stamp) scopes to exactly that namespace.
- Defense-in-depth seam mirroring `handle_remember`: dispatch's ADR-007 Rule-3 explicit-namespace escape already pre-mints the token; the handler re-derives for direct (non-dispatch) callers. The `namespace` ParamDef declaration keeps dispatch's strip-vs-pass-through behavior consistent.
- Verbose diagnostics hardened: `vector_hits_per_model` / `vector_candidates_per_model` are filtered through the visible-namespace id set at both serialization sites (`memory.recall include_breakdown` and `memory.recall_candidates`), so off-namespace candidate UUIDs cannot leak through the breakdown paths.
- `resolve_explicit_namespace` errors now name the supplied invalid value.
- New operator doc: `docs/benchmarks/recall-arm-isolation.md` — how to run isolated measurement arms (per-arm namespace writes + reads, profile pinning via the serving-profile override), with explicit notes on what is not yet isolated.

## Tests

12 new tests: absent-param regression (exact id-set equality against a mixed-namespace corpus), explicit-namespace scoping, no-match-empty-ok, invalid-value error naming the literal input, ANN over-fetch widening with a widening-disabled control, and mutation-verified breakdown-leak regressions for both diagnostic paths (two embedding models, closer-ranked off-namespace fillers).

Note: two tests use a bounded fresh-runtime retry to tolerate a pre-existing ANN warm-cache race, tracked separately as #750 — retries occur only before response settlement; assertions stay strict.

## Review

Codex rounds: r1 APPROVE-WITH-FIXES (1 High, 2 Medium) → fixes applied → r2 REJECT (1 Medium: missing mutation coverage on the candidates path) → fix applied → r3 clean APPROVE.

Gates: fmt, clippy (`khive-pack-memory`, `khive-runtime`) with `-D warnings`, `cargo test -p khive-pack-memory` (222 passed / 1 ignored), `cargo test -p khive-runtime` (59 passed) — all exit 0.

Closes nothing (leaves #733 open for the feedback-scoping remainder).

### DIFF
--- a/crates/khive-pack-memory/src/handlers/common.rs
+++ b/crates/khive-pack-memory/src/handlers/common.rs
@@ -264,6 +264,22 @@ pub(super) struct RecallParams {
     /// a silent fallback to defaults.
     #[serde(default)]
     pub(super) profile_id: Option<String>,
+    /// ADR-007 Rev 6 §"multi-record ops default to local + explicit escape"
+    /// (#733): exact-match read-namespace override. When absent, recall reads
+    /// the caller token's namespace (which defaults to `local`) — byte-identical
+    /// to pre-#733 behavior. When present, the candidate fetch (FTS + vector +
+    /// the ANN over-fetch retry loop) is scoped to exactly this namespace
+    /// instead of the token's (possibly wider) visible-namespace set. Invalid
+    /// values are rejected via the same `Namespace::parse` machinery used
+    /// elsewhere — never silently coerced.
+    ///
+    /// This is normally already pre-applied by `VerbRegistry::dispatch`'s
+    /// Rule-3 explicit-namespace escape (the token this handler receives
+    /// already carries `visible=[namespace]` in that path) — this field's
+    /// handling below is defense-in-depth for direct (non-dispatch) callers,
+    /// mirroring `RememberParams::namespace` / `handle_remember`.
+    #[serde(default)]
+    pub(super) namespace: Option<String>,
 }
 
 impl RecallParams {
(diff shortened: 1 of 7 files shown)


ohdearquant (approved):
Verified at content level: the effective-token shadowing flows the namespace override uniformly through FTS/vector fetch, ANN over-fetch gate, supersedes read, and serve-ledger stamp; absent param stays byte-identical per ADR-007 Rev 6's escape-hatch model with dispatch Rule-3 doing the primary enforcement and this as defense-in-depth (mirroring handle_remember). The High fix is correct at the root: the ANN index is namespace-global, so the verbose diagnostic had to filter through the same memory_ids visible-namespace set as results — and codex's r2 REJECT for guarding only one site, answered with a mutation-verified second regression, is the review lane doing exactly its job. #750 (or_insert mid-write snapshot race) noted as tonight's fifth defect — good that it's filed as production-correctness, not test-flake. Approved.

Issue #752, opened 2026-07-09
memory: warm ANN generation counter is process-local — external writers (kkernel --atomic, kkernel reindex) cannot invalidate a running daemon's warm cache


ohdearquant:
## Summary

The memory pack's warm-ANN freshness gate (`AnnState::generations`, `khive-pack-memory/src/ann.rs`) is a plain in-process `Mutex<HashMap<AnnKey, u64>>` — no persistence, no cross-process visibility. The note-mutation hook seam introduced for #750 guarantees freshness for all mutation paths *within one process*, but an external writer process cannot invalidate a running daemon's warm cache.

## Affected paths

1. **`kkernel exec --ops-file --atomic`** (`crates/kkernel/src/atomic_apply.rs`): runs in its own short-lived process. Its in-process hook wiring is now correct (registry installs the hook, atomic update/delete post-commit effects fire it), but the bumped generation dies with the process. A concurrently-running daemon serving `memory.recall` keeps trusting its warm index.

2. **`kkernel reindex`** (`crates/kkernel/src/reindex.rs`): constructs a bare `KhiveRuntime` with no packs loaded, so it has no `AnnState` at all. It rewrites note/entity vectors directly against the DB, then invalidates **persisted** Vamana snapshot rows (`invalidate_vamana_snapshots`, ~line 628) — a different mechanism from the daemon's in-memory generation counter. A running daemon's already-loaded warm index is untouched and can serve pre-reindex vectors until its own next generation bump or restart.

3. Any other external process writing memory notes against the same DB while a daemon holds a warm index.

## Why this was not fixed alongside #750

#750 was an in-process race (mid-write corpus snapshot installed permanently by `entry().or_insert()`); the generation-checked install plus the mutation-hook seam close every in-process path. Cross-process coherence is architecturally distinct: no hook fired inside a separate OS process can reach the daemon's in-memory state. It requires one of:

- a **persisted generation counter** (DB row bumped transactionally with corpus writes; daemon compares persisted vs in-memory generation on the recall path or on a cheap poll),
- a **daemon notification channel** (external writers signal the daemon to invalidate), or
- **startup/periodic revalidation** (daemon re-derives corpus fingerprint and discards stale warm entries).

The persisted-counter option is the natural fit — the daemon already reads the DB on every recall, and the existing persisted-snapshot invalidation in `kkernel reindex` shows the pattern.

## Scope note

[...]

PR #753, opened 2026-07-09
fix: four dogfooding defects — note tags, neighbors soft-deleted notes, prefix dedup, ANN warm-cache race (#747 #748 #749 #750)

### LINKED ISSUE #747: create(kind=note) silently drops tags — accepted param never persisted
## Bug

`create(kind="note", ..., tags=[...])` accepts the `tags` argument without error but never persists it. Found via dogfooding a viewer's comment write path; verified at source.

## Source evidence

- `crates/khive-pack-kg/src/handlers/create.rs`: `tags` is whitelisted in `CREATE_USER_KEYS` (~line 39), so the unknown-key rejection never fires for it.
- The **entity** branch passes tags through: `let tags = p.tags.unwrap_or_default();` → `create_entity(..., tags)` (~lines 310-321).
- The **note** branch calls `create_note(token, &canonical, p.name.as_deref(), &content, p.salience, p.properties, annotates)` (~lines 340-348) — `p.tags` is never read. Silent data loss: accepted, dropped, no error, no warning.

## Contrast

`memory.remember` persists tags via `properties.tags` (verified on live rows), so the memory-pack path is unaffected. Only the kg `create` note branch drops them.

## Repro

```
request(ops="create(kind=\"observation\", content=\"tag probe\", tags=[\"probe-tag\"])")
```

then `get(id=...)` → `properties` is null / carries no tags.

## Expected

Either persist note tags (consistent with the entity branch and with memory.remember's `properties.tags` convention) or reject the param loudly. Silent acceptance is the one wrong answer (anti-pattern: never silently coerce/drop invalid or unsupported input).

## Severity

Medium-High: the tagging convention for notes (lesson/work-class/agent-role tags) silently doesn't work through kg `create`, and callers have no signal.

### LINKED ISSUE #748: neighbors() does not filter soft-deleted note targets — only entities are checked
## Bug

`neighbors()` excludes soft-deleted **entity** nodes but not soft-deleted **note** nodes. A soft-deleted annotated note stays edge-linked in neighbors results; downstream hydration then yields blank/missing name/kind/content for it, which callers have no way to distinguish from a real record.

## Source evidence

`crates/khive-runtime/src/operations.rs`:
- The neighbors path filters candidates through `deleted_entity_ids` ("Filter out soft-deleted entity nodes (Fix 2)", ~line 52-56).
- `deleted_entity_ids` (~line 2028) runs `SELECT id FROM entities WHERE id IN (...) AND deleted_at IS NOT NULL` — the `notes` table is never consulted, so note-kind neighbors (e.g. via `annotates` edges) are never screened.

## Repro

Create a note annotating an entity, soft-delete the note (`delete(id=...)`, default soft), then `neighbors(node_id=<entity>, direction="both")` — the deleted note still appears as a neighbor, with content unresolvable through the normal read path.

## Expected

Consistent view-layer behavior with the entity filter: soft-deleted notes excluded from neighbors results (same future `include_deleted` escape hatch noted in the existing comment). This is a view-layer filter fix per the data-vs-view principle — no edge rewriting or data mutation.

## Severity

Medium: found via dogfooding a viewer; any consumer walking annotates edges hits phantom neighbors after note curation.

### LINKED ISSUE #749: resolve_prefix false AmbiguousPrefix: same UUID matched in multiple tables is counted twice
## Bug

Short-prefix resolution can reject a perfectly unambiguous prefix with `ambiguous prefix "XXXX": matches X, X` — the SAME full UUID listed twice. Live repro today: `comm.read(id="6132223d")` on an inbound message failed with `ambiguous prefix "6132223d": matches 6132223d, 6132223d`; retrying with the full UUID succeeded.

## Root cause (source)

`crates/khive-runtime/src/operations.rs`, `resolve_prefix_inner` (~line 3086):

- The resolver scans `entities`, `notes`, `events`, `graph_edges` in sequence, pushing every hit into `matches` (~line 3130) with **no deduplication across tables**.
- A record whose UUID legitimately exists in more than one table (e.g. a comm message note whose id also appears as an `events` row id) produces two identical entries, and `matches.len() > 1` fires `AmbiguousPrefix` even though exactly one record is addressed.
- The error's uuid list is built straight from the duplicated vec, which is why the message names the same id twice — a self-diagnosing symptom.

## Fix direction

Dedup `matches` by UUID (collect into a set, or `matches.dedup()` after sort) before the length check; genuinely distinct UUIDs sharing the prefix must still error. Regression test: insert one record under the same UUID in two scanned tables, assert the prefix resolves cleanly; plus keep the existing distinct-ids ambiguity test.

## Severity

Medium: intermittently breaks by-short-id verbs (comm.read/reply, get, brain.feedback) on valid prefixes, and the workaround (full UUID) is non-obvious from the error text. Probably the root cause of the long-observed "brain.feedback needs full UUID" inconsistency.

### LINKED ISSUE #750: ANN warm-cache race: fire-and-forget rebuild can permanently install a mid-write corpus snapshot

oceanwaves630:
## Summary

Four dogfooding fixes plus two codex fix-round commits, one review lane (6 commits):

- **#747** — `create(kind=<note>)` silently dropped `tags`. Note create now merges tags into `properties["tags"]` (the `memory.remember` convention; notes have no native tags column). Top-level `tags` wins over a conflicting `properties.tags`.
- **#748** — `neighbors()` missed soft-deleted **note** targets: `deleted_entity_ids` queried only `entities`. Now UNIONs entities+notes with independently-numbered param blocks per UNION half. View-layer only; edges untouched.
- **#749** — false `AmbiguousPrefix` when the same UUID appears in two scanned tables. `resolve_prefix_inner` now dedups by UUID (including the mid-scan early-exit). Genuine distinct-UUID ambiguity still errors.
- **#750** — ANN warm-cache race: `entry().or_insert()` could permanently install a mid-write corpus snapshot. Fix: generation captured before the corpus scan, `install_if_fresher` replaces only on strictly-newer generation, read-side `is_current` gate on recall cache hits, monotonic per-model write counter. The `ns733`/`ns733b` bounded-retry test harnesses are removed (deterministic now; 150 stress runs clean).
- **Fix-round 1** (codex High): generation was bumped only by `memory.remember`. Added a pack-installed `NoteMutationHookFn` seam in khive-runtime (parallel to the existing entity-type-validator pattern); `update_note`/`delete_note` fire it post-mutation, the memory pack installs a hook invalidating on `kind="memory"`, `memory.prune` bumps inline. Wired at both server boot paths. Three white-box regressions assert `ann::is_current` directly.
- **Fix-round 2** (codex 2 High): `merge_note` now fires the hook (same pattern as `update_note`) with a discriminating regression; `kkernel --atomic` registers the hooks, the `ReindexNote` post-commit handler fires it, and `DeletePlan` gained a `post_commit` seam (`PostCommitEffect::NoteDeleted`) covering atomic note deletes.

**Deliberately out of scope**: the cross-process warm-cache coherence gap (an external `kkernel --atomic` / `kkernel reindex` process can never invalidate a running daemon's in-memory generation counter) — filed as #752 with full analysis.

## Note for reviewer: brain.feedback full-UUID gotcha is NOT retired by #749

Verified at source: `brain.feedback` does route through `resolve_prefix_inner`, but its full-UUID requirement is a **separate namespace-scoping asymmetry** — full-UUID lookups bypass namespace filtering (ADR-007 by-ID rule) while short-prefix resolution is namespace-scoped. The #749 dedup does not change that. Design fork worth a ruling: should short-prefix resolution be namespace-agnostic like by-ID ops?

## Test plan

- `cargo fmt --all -- --check`, `cargo clippy -p khive-pack-kg -p khive-runtime -p khive-pack-memory -p kkernel --all-targets -- -D warnings`, `cargo test -p khive-pack-kg -p khive-runtime -p khive-pack-memory` — all exit 0.
- Every new regression verified to discriminate via disable/re-enable of its fix.
- ANN race tests stress-run 150 consecutive process-level executions clean.

Closes #747, closes #748, closes #749, closes #750.

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -1133,6 +1133,11 @@ pub fn build_registry_for_multi_backend(
     }
     registry.call_register_embedders(&default_runtime);
     registry.call_register_entity_type_validators(&default_runtime);
+    // #750 fix-round 1: install pack-owned note-mutation hooks (currently
+    // only khive-pack-memory's warm-ANN-cache invalidation) so KG's
+    // update/delete verbs notify caching packs even though there is no
+    // crate-level dependency between them.
+    registry.call_register_note_mutation_hooks(&default_runtime);
 
     let backend_for_pack: HashMap<&str, &StorageBackend> = per_pack_runtimes_local
         .iter()
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -336,6 +336,11 @@ impl KhiveMcpServer {
         // entity-type validation is active at the runtime layer for all write
         // paths, including direct `create_many` callers that bypass the handler.
         registry.call_register_entity_type_validators(&runtime);
+        // #750 fix-round 1: install pack-owned note-mutation hooks (currently
+        // only khive-pack-memory's warm-ANN-cache invalidation) so KG's
+        // update/delete verbs notify caching packs even though there is no
+        // crate-level dependency between them.
+        registry.call_register_note_mutation_hooks(&runtime);
         // Apply pack-auxiliary schema plans at startup so pack tables are
         // present before any handler runs. Errors are logged but not propagated
         // so a single pack's schema failure cannot abort startup.
--- a/crates/khive-pack-kg/src/handlers/common.rs
+++ b/crates/khive-pack-kg/src/handlers/common.rs
@@ -731,6 +731,38 @@ pub(crate) fn tags_match_any(entity_tags: &[String], wanted: &[String]) -> bool 
         .any(|tag| wanted.iter().any(|w| tag.eq_ignore_ascii_case(w)))
 }
 
+/// Merge the top-level `tags` create-param into `properties["tags"]` for a
+/// note. Notes have no dedicated tags column (see search.rs's `tag_filter`
+/// handling) — `properties["tags"]` is the storage convention already used
+/// by `memory.remember` (khive-pack-memory/src/handlers/remember.rs) and by
+/// this pack's own `search`/`list` note-tag filters. Without this merge,
+/// `create(kind=note, tags=[...])` silently dropped the tags (#747).
+///
+/// Precedence: an empty/absent `tags` param leaves `properties` untouched.
+/// A non-empty `tags` param always WINS over any `properties["tags"]` the
+/// caller also supplied — the top-level, typed param is the more explicit
+/// signal, so it overwrites rather than merges with a same-named nested key.
+pub(crate) fn merge_note_tags(
+    properties: Option<Value>,
+    tags: Option<Vec<String>>,
+) -> Result<Option<Value>, RuntimeError> {
+    let tags = match tags {
+        Some(t) if !t.is_empty() => t,
+        _ => return Ok(properties),
+    };
+    let mut obj = match properties {
+        None => serde_json::Map::new(),
+        Some(Value::Object(m)) => m,
+        Some(other) => {
+            return Err(RuntimeError::InvalidInput(format!(
+                "create: note `tags` cannot be merged into non-object `properties` (got {other})"
+            )));
+        }
+    };
+    obj.insert("tags".to_string(), json!(tags));
+    Ok(Some(Value::Object(obj)))
+}
+
 // ---- Handler helpers ----
 
 pub(crate) fn parse_entity_policy(s: &str) -> Result<EntityDedupMergePolicy, RuntimeError> {
--- a/crates/khive-pack-kg/src/handlers/create.rs
+++ b/crates/khive-pack-kg/src/handlers/create.rs
@@ -335,6 +335,7 @@ impl KgPack {
                 for s in p.annotates.unwrap_or_default() {
                     annotates.push(resolve_uuid_unfiltered(&s, &self.runtime, token).await?);
                 }
+                let properties = super::common::merge_note_tags(p.properties, p.tags)?;
                 let note = self
                     .runtime
                     .create_note(
@@ -343,7 +344,7 @@ impl KgPack {
                         p.name.as_deref(),
                         &content,
                         p.salience,
-                        p.properties,
+                        properties,
                         annotates,
                     )
                     .await?;
--- a/crates/khive-pack-kg/tests/integration.rs
+++ b/crates/khive-pack-kg/tests/integration.rs
@@ -3447,6 +3447,172 @@ async fn traverse_excludes_soft_deleted_entity() {
     );
 }
 
+// ---- #748: neighbors() must also screen soft-deleted NOTE targets ----
+
+/// #748: a soft-deleted note reached via an `annotates` edge must not appear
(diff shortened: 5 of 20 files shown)


ohdearquant (approved):
Verified at content level across all four fixes: (1) note tags via merge_note_tags into properties — consistent with memory.remember's existing path, no new storage shape; (2) neighbors' deleted-screen now consults both tables, correctly scoped as a view-layer read with the edge-survival test proving data-vs-view is honored (the edge row untouched, only the view filtered); (3) prefix dedup with the extra test driving the duplicate through the early-exit path — that test is what makes the fix trustworthy rather than coincidental; (4) the ANN generation design is the right shape: snapshot-before-scan, install-only-if-fresher replacing or_insert, and is_current on the recall gate so a stale winner is a cache miss, not a silent serve. The codex r2 REJECT (merge_note + atomic/reindex bypassing the hook) becoming the NoteMutationHookFn/PostCommitEffect seam is exactly how a fix round should widen. #752 correctly deferred — a persisted generation counter is an architectural change, not a rider.

DESIGN FORK RULING (short-prefix namespace scoping): KEEP the asymmetry — it is principled, not accidental. A full UUID is a demonstration that the caller knows the exact record, which is why ADR-007 makes by-ID ops namespace-agnostic. A short prefix is a RESOLUTION — a search — and searches default to namespace scoping. Making prefix resolution namespace-agnostic would let a prefix typed in one seat silently bind to another namespace's record, which is worse than the current behavior in every case that matters. Disposition: the known-bugs line converts to documented behavior — 'short-prefix resolution is namespace-scoped; cross-namespace targets require the full UUID' — in root CLAUDE.md and the verb help text, not a code change. Approved.

PR #756, opened 2026-07-09
docs(adr): ADR-088 Amendment 1 — git.digest agent-facing verb with remote-URL support

oceanwaves630:
Accepted amendment to ADR-088 (spec gate 2026-07-09). Docs-only.

## What it adds
- `git.digest(source, project?, max_items?, include?)` — one agent-facing verb, superseding ADR-088's no-new-verbs clause for exactly this verb. Bounded (`max_items` default 500, clamp 1..2000) and cursor-resumable (`done: false` loop) so calls stay inside MCP latency envelopes.
- Remote-URL mode: any https host (`git clone --filter=blob:none` to a daemon-owned LRU scratch cache); non-GitHub hosts degrade to commits-only with an explicit warning; SSH rejected in v1. Per-clone size cap in addition to the LRU cap.
- Ingest riders from first-consumer evidence: reference-edge extraction (Closes/Fixes #N, mentions, commit parents — fail-open, ingest-only) and readable note names (`#54 <title>` / `<short-sha> <subject>`).
- Cross-links from ADR-088 and the ADR index.

Implementation follows in a separate code PR per the amendment.

### DIFF
--- a/docs/adr/ADR-088-amendment-1-git-digest.md
+++ b/docs/adr/ADR-088-amendment-1-git-digest.md
@@ -1,0 +1,157 @@
+# ADR-088 Amendment 1: `git.digest` — Agent-Facing Digest Verb with Remote-URL Support
+
+**Status**: Accepted\
+**Date**: 2026-07-09\
+**Authors**: khive maintainers\
+**Amends**: [ADR-088](ADR-088-git-lifecycle-pack.md) (Git-Lifecycle Pack)\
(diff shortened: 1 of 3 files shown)


ohdearquant (approved):
Spec-gate rulings verified encoded: any-https with commits-only degradation, 1GB per-clone cap abort-before-write, max_items=500, fail-open reference extraction, ingest-only with admin backfill for existing notes.

PR #757, opened 2026-07-09
fix(comm,gtd): backlog burn batch A — inbox sender filters, thread cursor pagination, message tags, gtd limit docs (#493 #494 #495 #744)

### LINKED ISSUE #493: comm.inbox: from_actor / from_prefix filter (agent-traffic hygiene)
## Motivation

Design intent (lionagi-side ADR in progress): spawned agents get distinct actor ids (adopted convention: \`agent:<parent>:<role>-<n>\`, via per-spawn \`KHIVE_ACTOR\` env — \`crates/khive-runtime/src/config.rs:316\`) and report status to a coordinating actor over comm. Today \`InboxParams = {limit?, status?}\` (\`crates/khive-pack-comm/src/params.rs\`) has no sender filter: N agents reporting status land undifferentiated in a single actor's unread count, degrading both triage and inbound-count monitors.

## Spec

Additive, backward-compatible params on \`comm.inbox\`:

- \`from_actor: Option<String>\` — exact match on \`properties.from_actor\`.
- \`from_prefix: Option<String>\` — prefix match, so \`from_prefix="agent:khive:"\` selects all agents under one namespace (this is why the colon-nested naming convention was adopted).
- Mutually exclusive; supplying both is a per-op error naming the conflict.
- Composes with the existing \`status\` filter and \`to_actor\` scoping (\`crates/khive-pack-comm/src/handlers.rs:178-191\`); no change to defaults when absent.

Per ADR-084 Rule 4: canonical param naming, no silent coercion.

## Context

Spec load-bearing for the lionagi bridge ADR; implementation queues normally.

### LINKED ISSUE #494: comm.thread: tail pagination (cursor or descending order) for long threads
## Motivation

\`ThreadParams = {id, limit?}\` and the handler sorts chronologically ascending then \`truncate(limit)\` (\`crates/khive-pack-comm/src/handlers.rs\` ~541-549) — a thread longer than \`limit\` returns the HEAD and the newest messages become unreachable via \`thread()\`. Thread-as-group-chat design intent makes this immediate: participants poll \`thread()\` for shared history and care most about the tail.

## Spec

Additive params on \`comm.thread\`:

- \`order: Option<String>\` — \`"asc"\` (default, unchanged) | \`"desc"\`. Closed set; invalid values rejected listing valid ones (ADR-084 Rule 2).
- \`after: Option<String>\` — cursor: message id (ladder-resolved per ADR-084 Rule 1) or ISO timestamp; returns messages strictly after that point in the chosen order. Enables incremental polling without re-fetching history.
- Truncation applies after ordering, so \`order="desc", limit=50\` = newest 50.

## Context

Spec load-bearing for the lionagi bridge ADR; implementation queues normally.

### LINKED ISSUE #495: comm.send: metadata passthrough for agent-traffic provenance
## Motivation

\`SendParams = {to, content, subject?, thread_id?}\` with \`deny_unknown_fields\` (\`crates/khive-pack-comm/src/params.rs\`) — no way to attach structured provenance (run id, job id, traffic class) to a message. Agent traffic design intent needs provenance parallel to the memory doctrine (agent writes carry \`agent:<role>\` tags); today the only channel is subject-prefix conventions like \`[run:abc123]\`, which are unfilterable and unqueryable.

## Spec

- \`tags: Option<Vec<String>>\` on \`comm.send\` (and \`comm.reply\`), persisted to \`properties["tags"]\` on both outbound and inbound copies — mirrors the shipped \`memory.remember\` precedent and ADR-084 Rule 5 (substrate-symmetric \`tags\`).
- \`from_actor\` remains process-token-derived (\`crates/khive-runtime/src/config.rs:316\`) and is NOT overridable via this field — cooperative attribution model accepted for v1; authn stays the ADR-018 Gate seam.
- Follow-up (separate): inbox/thread filtering on tags, once the from-filter issue lands the filtering pattern.

## Context

Spec load-bearing for the lionagi bridge ADR; implementation queues normally.

### LINKED ISSUE #744: gtd.tasks silently clamps results to 200 rows — callers cannot distinguish a full result from a capped one
Verified live: `gtd.tasks(limit=500)` returns exactly 200 rows with no indication the requested limit was clamped. The cap itself is reasonable; the silence is the defect — callers building counts or boards from the returned array undercount without knowing it.

Ask (either, prefer the second):
1. Document the cap in the verb's help/ParamDef description.
2. Return a `truncated: true` and/or `total_matching` field when the server-side cap (or any internal scan limit) cuts the result, so a full result is distinguishable from a capped one.

Same audit lens applies to sibling list-shaped verbs (`gtd.next`, note `list`) — if they share the clamp, they share the fix.

Low priority; surfaced from downstream consumers computing task counts.

oceanwaves630:
Backlog-burn batch A, four small fixes off the 2026-07-09 triage.

- **#493** `comm.inbox`: `from_actor` (exact) and `from_prefix` (prefix) sender filters, mutually exclusive, composing with existing status/to_actor filtering. Prefix matching in Rust (no SQL interpolation).
- **#494** `comm.thread`: `order=asc|desc` with truncation after ordering (long threads no longer lose the tail), and an `after` cursor — id cursors compare on the total order `(created_at, full_id)` (no skip/dup on equal timestamps), timestamp cursors are RFC 3339 parsed, "after" is defined against the chosen order, invalid cursors fail loudly.
- **#495** `comm.send`/`comm.reply`: `tags` passthrough stored on both dual-write copies; tag values pass through the runtime secret gate like all structured properties.
- **#744** documented the 200-row clamp on `gtd.next`/`gtd.tasks` `limit` (docs-only: the issue's preferred `truncated` flag is a breaking response-shape change — `gtd` list verbs return a bare JSON array consumed via `.as_array()` — so this takes the issue's stated fallback).

Review: two rounds, final verdict APPROVE (r1 APPROVE-WITH-FIXES → cursor rework in 67d3cc39 + doc-comment alignment; r2 confirmed the High closed, no re-review needed). Gates: fmt/clippy clean, 188 khive-pack-comm tests passing (124 integration + 50 unit + 14 contract), 5 new cursor regression tests.

Closes #493. Closes #494. Closes #495. Closes #744.


ohdearquant (approved):
Gate verified: (created_at, full_id) total-order cursor with fail-loud RFC3339 parsing matches the codex High; 188 pack tests green.

PR #761, opened 2026-07-09
feat(git): git.digest paged ingest + URL clone cache (ADR-088 Amendment 1)

oceanwaves630:
Implements the ADR-088 Amendment 1 lane: `git.digest` paged GitHub ingest and URL clone caching.

- Paged `gh` fetches via `--search "sort:updated-asc updated:>=<floor>"` with a pure `decide_page_outcome`/`PageOutcome` decision function (7 unit tests); `done:true` only when a page proves the window exhausted; floor-stall returns `done:false` instead of spinning.
- `--state all` on both pr/issue paged fetches, regression-locked by a fake-`gh` argv assertion (every logged invocation must carry it).
- Clone cache: staged clone into a private `.staging-<uuid>` dir with pre-cache size check, same-filesystem rename into the cache slot; staging cleaned on clone failure, over-cap, size-probe error, and rename error paths. `evict_lru` gated by `is_owned_entry` (16-hex name + `.git` + `.khive-last-used` marker) so foreign dirs are never removed.
- `max_items` parsed as `i64`, non-integers rejected, clamped `1..=2000` (4 boundary tests).
- `precedes` commit→commit allowed via pack-level `EdgeEndpointRule` (`GIT_EDGE_RULES`).
- Docs: api-reference verb count + git pack section.

Gates: fmt, clippy `-D warnings`, `cargo test -p khive-pack-git` (55), `cargo test -p kkernel`, `cargo check --workspace` all green. Three codex review rounds (r1 APPROVE-WITH-FIXES 2H/2M/1L, r2 APPROVE-WITH-FIXES 1H/1M on the r1 fixes, r3 fix commit a225b81e) — narrow r3 re-review in flight.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,10 +15,9 @@ khive gives your agent:
 9. **Brain** — Bayesian profile tuning from feedback signals
 10. **Session** — persist and resume agent-session records
 
-All 9 packs load by default. **76 public verbs** across the packs — the `git` pack
-(commit/issue/pull_request provenance notes, populated by a batch ingester, not agent-facing
-verbs) contributes no new verb (regenerate via `request(ops="verbs()")` before editing this
-line).
+All 9 packs load by default. **77 public verbs** across the packs — the `git` pack
+contributes the `git.digest` verb plus the commit/issue/pull_request provenance note kinds
+and a batch ingester (regenerate via `request(ops="verbs()")` before editing this line).
 
 If you're working on khive itself (writing code in this repo), see `CLAUDE.md` instead.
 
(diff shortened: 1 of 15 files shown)


ohdearquant (approved):
Approving review per merge flow v2: ADR-088 Amendment 1 contract verified at the diff (clamped max_items + done/cursor resumability with the frozen-cursor tie invariant, https-only source resolution with explicit ssh/git/scp rejection, staging-dir clone with failure-path cleanup and measure-before-move size cap, hook/prompt hardening, fail-open reference extraction with precedes parent edges). Codex r3 clean APPROVE on record.

PR #762, opened 2026-07-09
feat(runtime): resolve_reference capability + recently-referenced ring (S1)

oceanwaves630:
Adds the S1 reference-resolution slice: `resolve_reference(nl_ref, namespace, actor)` in khive-runtime (id-passthrough → ring → hybrid-search, entity-only, never a silent pick among close candidates), a bounded per-(namespace, actor) recently-referenced ring (N=64, TTL 30min, LRU outer budget, poison-safe, daemon-warm only — no schema change), admission at the dispatch boundary for singleton create/get/update/delete/merge/link results only, and a read-only `resolve(refs, kind?, limit?)` kg verb (18th handler). smoke_test verb total 76→78 (with git.digest).

Tests: 1156 passing across khive-runtime + khive-pack-kg incl. 21 new (namespace-key parity under non-local default namespace, entity-only admission/passthrough, outer-map TTL/LRU pruning on admit and snapshot, poisoned-mutex recovery, search-confidence margin semantics).


ohdearquant (approved):
Approving review (flow v2). Read the full diff at content level: resolver three-stage contract matches the gated S1 spec (id-passthrough entity-scoped identically for UUID + prefix, ring exact/substring with fixed confidence bands, RRF-scale-aware search fallback that never silently picks per F7); ring admission keyed on token.namespace() with the non-local default_namespace regression test; strict singleton-by-id admission with bulk shapes excluded; ring bounded by capacity + TTL + outer-map sweep on reads. Head 86b00651 matches the codex-gated SHA.

Issue #14, opened 2026-06-07
feat(schedule): kkernel trigger runner (exec --pending-events / scheduler daemon)


ohdearquant:
## Problem

The schedule pack creates `scheduled_event` notes via `schedule.schedule` / `schedule.remind`, but **nothing ever fires them**. There is no trigger runner, no daemon loop, and no `kkernel` exec mode that evaluates pending events. Every scheduled action is a dead letter.

## Evidence

- `crates/khive-pack-schedule/src/handlers.rs:4` — comment: "Trigger evaluation is NOT performed by the pack."
- `crates/kkernel/src/lib.rs` — no scheduler module; no `kkernel scheduler` subcommand; no `exec --pending-events` mode.

So the storage half of scheduling works (events persist as notes), but the execution half does not exist.

## Expected

A trigger runner that:
1. Scans for due `scheduled_event` notes (`at <= now`, not yet fired).
2. Dispatches the stored `action` DSL verb string through the registry.
3. Marks fired / advances `repeat` schedules.
4. Honors namespace isolation.

## Fix sketch

Add either a `kkernel scheduler` daemon loop or a `kkernel exec --pending-events` one-shot drain (cron-friendly). Given the kkernel unified-binary topology, a `--pending-events` exec mode is the lower-risk first cut; a long-running daemon loop can layer on top later.

Priority: **p1** — the entire schedule pack is non-functional without it.

_Ported from old repo issue #479. Verified against current source 2026-06-07._

PR #767, opened 2026-07-09
docs(adr): ADR-106 schedule pack executor — daemon-resident tick for the pending-event drain

oceanwaves630:
Adds ADR-106: a daemon-resident tick task that periodically invokes the existing pending-event drain in-process.

- Context: `kkernel exec --pending-events` is a complete, tested one-shot CAS drain; nothing invokes it periodically in a default deployment, so scheduled events sit pending indefinitely without external cron.
- Decision: `schedule_tick_loop` spawned from `run_daemon_with_boot_guard`, modeled on the WAL checkpoint task; in-process refactor of `run_pending_events` to share the daemon's live dispatcher (subprocess shell-out rejected); daemon-boot-only spawn site (email-channel-loop precedent); external cron stays supported and race-safe via the existing pending→firing CAS claim; interval env-configurable, default 60s.
- Repeat-advance semantics unchanged; delivery of fired reminders, dispatch-error field, agenda visibility, telemetry, and cron advancement (#14) explicitly deferred.

Docs-only: new ADR file + README index row.

### DIFF
--- a/docs/adr/ADR-106-schedule-pack-executor.md
+++ b/docs/adr/ADR-106-schedule-pack-executor.md
@@ -83,7 +83,12 @@ if let Some(pool) = dispatcher.pool_for_checkpoint() {
 a schedule tick than the one-shot warm-up: it is a genuine interval loop
 (`tokio::time::interval`, `MissedTickBehavior::Skip`) that runs for the daemon's
 lifetime and detects shutdown by checking `Arc::strong_count(&pool) <= 1` on each tick
-rather than a separate cancellation channel.
+rather than a separate cancellation channel. That check does not, however, reflect the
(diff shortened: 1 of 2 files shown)


ohdearquant (approved):
Leo gate: full ADR read at 020c332a. Positional DrainError classification, honest bounded-shutdown contract leaning on reclaim, watch-channel cancellation correctly NOT copying the checkpoint strong-count undercount (deferred to #774), dependency direction verified. One non-blocking copyedit noted in comm (§1a pool_for_checkpoint wording). Sign-off unblocks implementation; schedule-pack WRITE-ONLY freeze lifts on executor ship, not this docs merge.

PR #766, opened 2026-07-09
docs(adr): ADR-085 Amendment 2 — code.ingest verb, tiers, identity, acceptance; flip to Accepted

oceanwaves630:
Docs-only. Adds Amendment 2 to ADR-085 specifying the code.ingest pipeline the base text left as separate ADR-069-layer work, and flips the ADR status to Accepted (v0 shipped; this ratifies).

Amendment contents:
- B1: one new verb `code.ingest(path, db?, languages?=auto)` — folder-path input, monorepo subtree first-class (git.digest single-verb precedent)
- B2: Scanner/Extractor split per ADR-069; syntax-only Scanners (Rust/Python/TS/Lean order), one ontology-driven Extractor; Lean structure-extends deferred (needs environment metadata)
- B3: three tiers — L1 manifest edges, L1.5 regex import-scan floor, L2 full AST symbol tier
- B4: uuid5(source_project, module_path, name, kind) identity + content_hash; idempotent re-ingest; canonical tokens only
- B5: no auto-delete; last_seen_at staleness, view-layer filtering
- B6: cross-repo re-resolve pass (v1), deferred-edge queue documented as v2; provenance as properties.source, edges ordinary
- B7: D6.1 fence restated — map DB default, promotion to prod explicit and never automatic
- B8: acceptance — parity queries (blast radius, circular deps, dead symbols) + cross-repo ingest-order convergence tested in both orders

docs/adr/README.md index row updated to Accepted. deno fmt clean.

PR #775, opened 2026-07-09
docs(adr): ADR-085 Amendment 2 precision follow-up — per-file source_project resolution + per-(project,language) staleness sweeps

ohdearquant:
Docs-only follow-up to #766, applying two precision refinements to Amendment 2 agreed during its final review round:

1. **B4 — per-file `source_project` resolution.** Resolution is defined per source file, not per ingested path: nearest governing manifest at or above the file. Virtual/workspace-only manifests that declare no package name (`Cargo.toml` with `[workspace]` but no `[package]`; `pyproject.toml` with no `[project]` table) are not governing and are skipped upward. Basename fallback applies only when no governing manifest exists above a file. Explicit consequence: multi-package roots yield per-package `source_project` values in one run; no reject rule.

2. **B5 — per-`(source_project, language)` staleness sweeps.** Sweep timestamps are recorded per pair, and an entity's staleness is judged only against the latest sweep of its own pair, never against other projects or languages sharing the namespace.

No behavioral surface beyond the ADR text; `code.ingest` implementation follows the amended contract.

### DIFF
--- a/docs/adr/ADR-085-code-pack.md
+++ b/docs/adr/ADR-085-code-pack.md
@@ -571,12 +571,19 @@ applies identically here.
 Canonicalization of the two identity inputs is fixed, not left to per-caller
 convention:
 
-- `source_project` is the package name declared by the nearest governing
-  manifest at or above the ingested path (`Cargo.toml` `[package].name`,
+- `source_project` resolves per source file, not per ingested path: each
+  file's `source_project` is the package name declared by the nearest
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Leo gate: full diff read at 50f50b3b. language-in-identity-tuple rationale is correct and matches the per-file source_project resolution + per-(source_project,language) sweep clocks approved for the B4/B5 follow-up; acceptance property 3 pins exactly the polyglot collision the r1 finding surfaced. Clean.

Issue #779, opened 2026-07-09
schedule pack tests write fixtures to the live store — 8 test events parked at trigger_at=2099 visible in agenda()


ohdearquant:
## Behavior

`schedule.agenda()` on a live database currently returns 8 unit-test fixtures created 2026-07-04, all parked at `trigger_at=2099-01-01` (contents: check-the-build / test / repeat=daily / repeat=weekly / repeat=monthly / window-early / id-shape / sort-order-jan). The schedule pack's tests write to whatever store the environment resolves, and on a developer machine that was the live store.

`crates/khive-pack-schedule` tests reference the 2099 sentinel in 14 places.

## Why it matters

- Live `agenda()` output is polluted with synthetic rows, and nothing distinguishes them from real scheduled events.
- Any future executor (ADR-106) must not attempt to fire or reclaim fixture rows.
- Tests that depend on ambient store resolution are also a flakiness source: two concurrent test runs share state.

## Requested fix

1. Schedule pack tests construct a throwaway store explicitly (tempdir SQLite or in-memory), never ambient resolution — same pattern the other packs' tests use.
2. One-time curation: soft-delete the 8 fixture rows from the live store (they are identifiable by the 2026-07-04 creation window + trigger_at=2099 + fixture content).
3. Optional hardening: a debug assertion or test-util guard that refuses pack-test writes when the resolved db path is a user-profile store.

Fix should ride with or immediately after the ADR-106 executor implementation, since the executor's drain path is what makes stray fixtures actively dangerous rather than just untidy.

ohdearquant:
Fresh independent measurement (agenda sweep today ~13:45 EDT): **42 live year-2099 fixture rows** (19 at 2099-01-01T00:00:00Z, 8 at 2099-03-10, 11 at 2099-06-15, 4 at 2099-12-31). Trend since filing: 8 → 14 → 42 — accelerating, so this is an ongoing writer, not one-time residue.

Likely culprit: the year-2099 fixtures live in `crates/khive-pack-schedule/tests/agenda.rs` (Rust tests, which use isolated stores) AND in `tests/smoke_schedule.py` — and the Python smoke test drives the real binary, so any invocation without an explicit scratch KHIVE_DB writes fixtures straight into the live store. The acceleration correlates with heavy parallel agent activity running smoke/verify passes today.

[...]

Issue #602, opened 2026-07-04
channel-email: every kkernel mcp process spawns IMAP/outbox loops — 9 concurrent pollers broke inbound auth


ohdearquant:
Found live 2026-07-04 triaging continuous `channel poll failed: authentication error: IMAP XOAUTH2 authenticate failed: no response: code: None, info: Some("User is authenticated but not connected.")` in the daemon log.

Diagnosis chain:
1. Independent reproduction from the same machine with the same `~/.khive/.env` creds SUCCEEDS end-to-end (client-credentials token carries `roles: [IMAP.AccessAsApp]`; IMAP `AUTHENTICATE` completes; `SELECT INBOX` returns 69 messages). Exchange side and the grant are healthy; no re-auth needed.
2. `spawn_email_channel_loops(&server)` is called unconditionally in BOTH `serve::run` (serve.rs:41) and `serve_server` (serve.rs:532) BEFORE the `if args.daemon` branch. So every `kkernel mcp` stdio client process spawns its own IMAP poll loop (5s cadence) and its own outbox loop.
3. At observation time 9 `kkernel mcp` processes were alive (1 daemon + 8 stdio clients). Nine app-only IMAP sessions churning connect/auth against the same mailbox trips Exchange Online's concurrent-connection limits; "User is authenticated but not connected." is the documented symptom. Failures are intermittent per process (observed ~18 warnings/6min in the daemon vs the ~70 a hard-down 5s loop would produce), i.e. processes race for connection slots.

Impact:
- Inbound email ingest degraded to lossy/flaky (last successful ingest 2026-07-04T01:26Z).
- Duplicate outbox loops in N processes = duplicate outbound-send risk (dedup/claim semantics on outbound notes need verification).
- Every stdio client burns a poll loop even though it forwards requests to the daemon.

Fix sketch: channel loops are a daemon-role responsibility. Gate `spawn_email_channel_loops` on `args.daemon` (or move the spawn into `khive_runtime::daemon::run_daemon` alongside `pool_for_checkpoint`), with a regression test asserting stdio serve does not spawn loops. Fold into the LOUD-startup channel-health spec item as evidence (channel health must be visible at startup; see also #599 — these warnings were invisible under the default /dev/null stderr).

Relates: #503 (the fix that made loops spawn on the multi-backend path also made them spawn everywhere), #599.

PR #782, opened 2026-07-09
feat(runtime): daemon-resident schedule drain tick with missed-event policy

ohdearquant:
## Summary

Implements ADR-106: a daemon-resident tick for the schedule pack's pending-event
drain, plus a grace-window "missed event" policy the ADR did not originally specify
(added here as ADR-106 Amendment A).

The executor itself (`kkernel exec --pending-events`, the CAS claim/dispatch/finalize
state machine) already existed and was already well tested — this PR does not rebuild
it. It adds only:

1. **Daemon tick.** `spawn_schedule_tick_loop_if_daemon` in `khive-mcp/src/serve.rs`
   spawns a background task, gated on `args.daemon` exactly like the existing
   email-channel-loop gate (#602), that periodically re-invokes the drain for the
   lifetime of the daemon process. Interval: `KHIVE_SCHEDULE_TICK_SECS`, default 60s,
   ticking on `tokio::time::interval_at(now + interval, interval)` with
   `MissedTickBehavior::Skip` (fixed cadence, no sleep-after-drain drift).
2. **Missed-event grace policy.** An event discovered `pending` and overdue by more
   than `KHIVE_FIRE_GRACE_SECS` (default 300s) is never dispatched. Non-repeating
   events are marked `status=missed` with `missed_at` stamped; repeating events skip
   directly to the next future occurrence (no catch-up bursts). A first boot against a
   store carrying a large stale backlog therefore marks the whole backlog missed and
   dispatches zero of them — the intended migration behavior.
3. **Module relocation.** The drain's internals moved from `crates/kkernel` into
   `crates/khive-mcp` (`khive_mcp::pending_events`), because the daemon tick must call
   it in-process from `khive-mcp::serve`, and `khive-runtime` (where the daemon's
   accept loop lives) cannot depend back on `khive-mcp`. `kkernel exec
   --pending-events` now calls the moved module directly.
4. **Runtime + dispatch wiring (both threaded from the daemon's real boot state, not
   reconstructed inside the tick).** `khive-mcp::serve::build_server` now returns,
   alongside the server, the resolved `"schedule"`-pack `KhiveRuntime` handle it
   already built while constructing the server (`Option<KhiveRuntime>`). The tick
   takes BOTH that runtime and the daemon's own live, fully-wired `KhiveMcpServer`
   (cloned — cheap, `Arc`-wrapped): the runtime is used to scan/claim/finalize
   `scheduled_event` rows against whichever backend the `schedule` pack is wired to
   (single-backend: the one shared runtime; multi-backend/ADR-028 `[[backends]]`: the
   specific per-pack runtime); the server is used only to replay a fired event's
   stored action DSL, so a multi-backend deployment's `comm.send`/`kg`/other-pack
   actions dispatch through THAT pack's own configured backend, not the schedule
   backend. `run_pending_events` (the CLI one-shot path, `kkernel exec
   --pending-events`) resolves both the same way, through `build_server`, rather than
   a throwaway `RuntimeConfig::default()` that never consulted `khive.toml` at all.
5. **ADR-106 amended in place**, status flipped Proposed → Accepted:
   - **Amendment A** states the missed/grace policy normatively, with a
     systemd/Quartz/Sidekiq prior-art comparison explaining why khive chooses
     skip-and-mark over catch-up-once or fire-everything.
   - **Amendment B** documents that the shipped wiring is a smaller step than the
     ADR's original Decision points 1-3 (a `DaemonDispatch::drain_pending_events`
     trait seam in `khive-runtime` with `tokio::sync::watch`-channel graceful
     shutdown): this PR spawns from `khive-mcp::serve` with a bare `tokio::spawn`
     and the daemon's own shared runtime + server (see point 4 above), not a
     `DaemonDispatch` trait method. Two codex review rounds on this PR corrected
     the amendment's initial (inaccurate) claims about what shipped — see the
     amendment's "Fix-round update" and "Fix-round 2 update" subsections for the
     detailed history. Current status: Acceptance Criteria 1, 2, 3, 4 are **met**;
     5, 6, 7 (the `DaemonDispatch` trait seam, graceful shutdown, and
     `khive-runtime`-owned types) are **not met** and remain open follow-on work.

Refs #779: `crates/khive-pack-schedule`'s own tests were audited and already use
`KhiveRuntime::memory()` (fully isolated, in-memory) — the test-isolation fix that
issue requests appears to already be in place in this tree. The issue's other ask
(soft-deleting 8 pre-existing fixture rows reportedly sitting in a live
`~/.khive/khive.db`) was **not** touched by this PR, per the standing prohibition on
writing to any live store from an agent session; that curation needs a human with
direct access to the live database.

## Test plan

- [x] `cargo fmt --all -- --check` — pass
- [x] `cargo clippy -p khive-mcp -p kkernel --all-targets -- -D warnings` — pass, zero warnings
- [x] `cargo test -p khive-mcp -p kkernel` — all pass (183 khive-mcp lib + other khive-mcp
      binaries; 232 kkernel lib + 16 main-bin + integration test binaries)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-mcp -p kkernel` — pass
- [x] New regression tests, all on throwaway temp-file/in-memory stores (never touch `~/.khive/khive.db`):
  - `nine_overdue_events_beyond_grace_are_missed_with_zero_dispatch` — 9 events overdue past grace drain to `status=missed`, zero dispatches, verified via a genuinely side-effecting `create` action whose output note is confirmed ABSENT (not just zero fired/advanced counters)
  - `overdue_within_grace_still_fires` — an event overdue by less than the grace window still fires normally
  - `missed_repeat_is_rearmed_at_next_future_occurrence` (end-to-end) + `advance_repeat_past_missed_skips_all_accumulated_occurrences` (deterministic unit test) — a missed daily repeat re-arms to the next future occurrence in a single advance, never a catch-up burst
  - `backlog_larger_than_page_size_is_fully_drained_in_one_pass` — 201 far-overdue rows (one page over `PAGE_SIZE=200`) all drain in a single pass; proves the keyset-pagination fix doesn't skip rows under mutation
  - `concurrent_drains_fire_each_row_exactly_once` — two concurrent drains over the same store, each of 20 rows carrying a row-distinct side-effecting marker action; asserts exactly one marker note lands per row (catches double-dispatch-one-finalize, not just CAS-race counters)
  - `build_server_schedule_tick_uses_the_configured_backend_not_the_home_default`,
    `..._uses_the_configured_actor_identity`,
    `..._is_none_when_schedule_pack_is_not_in_the_restricted_pack_set`,
    `..._runtime_satisfies_strict_actor_mode_like_the_live_server`,
    `..._uses_the_declared_multi_backend_not_main` — `build_server`'s returned
    schedule-tick runtime targets the configured backend/actor/pack-set, single- and
    multi-backend
  - `build_server_schedule_tick_dispatches_actions_through_the_declared_multi_backend_not_schedule` —
    `schedule` and `kg` routed to separate declared backends; a due event's
    `kg`-pack action lands its side effect only in `kg`'s backend, never `schedule`'s
  - All pre-existing drain tests (claim/finalize CAS, stale-firing reclaim, namespace isolation, repeat advance, idempotency, dispatch-failure handling) still pass

### DIFF
--- a/crates/khive-mcp/src/lib.rs
+++ b/crates/khive-mcp/src/lib.rs
@@ -8,6 +8,7 @@ pub mod coordinator;
 #[cfg(unix)]
 pub mod daemon;
 pub mod pack;
+pub mod pending_events;
 pub mod save_sink;
 pub mod serve;
 pub mod server;
--- a/crates/khive-mcp/src/pending_events.rs
+++ b/crates/khive-mcp/src/pending_events.rs
@@ -288,7 +288,7 @@ pub async fn run_pending_events_on(
             let (sql, params): (String, Vec<SqlValue>) = match &cursor {
                 //
                 // The due-ness predicate compares via SQLite's `datetime()`,
-                // not a raw string `<=` (Ocean review, High finding): stored
+                // not a raw string `<=` (PR #782 review round 3, High finding): stored
                 // `trigger_at` values are NOT normalized to UTC —
                 // `khive-pack-schedule`'s `handle_remind`/`handle_schedule`
                 // deliberately round-trip the caller's original string
@@ -1047,7 +1047,7 @@ async fn discover_pending_namespaces(rt: &KhiveRuntime, now: DateTime<Utc>) -> R
     // candidate scan below, not the final due-ness decision — but a
     // namespace excluded HERE never reaches that scan at all, so it must be
     // held to the same correctness bar as the candidate-page queries
-    // (`datetime(...)` normalization, Ocean review High finding): comparing
+    // (`datetime(...)` normalization, PR #782 review round 3 High finding): comparing
     // `trigger_at` against `now` as raw TEXT is only chronologically correct
     // when every stored string happens to share `now`'s UTC offset.
     // `khive-pack-schedule` round-trips the caller's original `trigger_at`
@@ -1407,7 +1407,7 @@ mod tests {
     }
 
     /// A due event whose `trigger_at` carries a POSITIVE offset must still
-    /// fire (Ocean review, High finding).
+    /// fire (PR #782 review round 3, High finding).
     ///
     /// `khive-pack-schedule` round-trips the caller's original `trigger_at`
     /// string verbatim, offset included (H5) — it is never normalized to
@@ -1471,7 +1471,7 @@ mod tests {
     /// `datetime(...)` normalization correctly excludes it from the
     /// candidate page; even if it were fetched, the retained Rust-side
     /// `trigger_at > now` re-check is the belt-and-suspenders backstop that
-    /// already made this direction benign before the SQL fix (Ocean review:
+    /// already made this direction benign before the SQL fix (PR #782 review round 3:
     /// "Negative-offset strings produce false POSITIVES, which the retained
     /// Rust re-check filters — benign").
     #[tokio::test]
--- a/crates/khive-mcp/src/pending_events.rs
+++ b/crates/khive-mcp/src/pending_events.rs
@@ -21,8 +21,9 @@
 //!   [`run_pending_events`] directly. Suitable for `* * * * * kkernel exec
 //!   --pending-events` to achieve minute-granularity delivery.
 //! - **Daemon-resident tick** (ADR-106): [`schedule_tick_loop`] calls
-//!   [`run_pending_events`] on a fixed interval for the lifetime of the warm
-//!   `khived` daemon process. Spawned only by the daemon role (mirrors the
+//!   [`run_pending_events_on`] against the daemon's own resolved `KhiveRuntime`
+//!   handle on a fixed interval for the lifetime of the warm `khived` daemon
+//!   process. Spawned only by the daemon role (mirrors the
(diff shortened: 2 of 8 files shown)


oceanwaves630 (approved):
Gate read done at 085df3d2. Both content conditions verified in the ADR text: (1) Amendment B honesty — Acceptance Criteria 5/6/7 explicitly marked not-met with the original incorrect 'met' claim called out and corrected in place, follow-on work (full DaemonDispatch trait seam + tracked shutdown) named; (2) the catch-up-once prior art is engaged head-on (systemd Persistent=true, Quartz misfire instructions, Sidekiq fire-everything) with an argued reason skip-and-mark wins for an unconstrained verb action space. Codex round-5 APPROVE reconfirmed at this exact head, all four gates green in the reviewer's runs, CI green here. The five-round fix trail (runtime targeting, multi-backend dispatch routing, keyset pagination, datetime normalization, actor-suppression seam) is recorded honestly in the amendment. Approved.

PR #784, opened 2026-07-10
fix(channel-email): persist IMAP UID cursor progress

### LINKED ISSUE #449: khive-channel-email: IMAP polling can repeatedly fetch the same UID subset and miss later messages [high]
Severity: **high** | Confidence: high | Priority: P1 | Categories: errors, performance
Crate: `khive-channel-email` | Finding id: `CHANNELEMAIL-AUD-002`

## Finding
Inbound maintainer messages can be silently missed under bursts or mailbox backlog; durable dedup cannot recover messages that were never fetched.

## Failure scenario
At 09:00 UTC the mailbox has 75 authorized messages with the same IMAP internal date. Every poll calls fetch_since(since, 50), searches SINCE that date, and receives all 75 UIDs in a HashSet. A valid deterministic hash iteration order yields the same 50 UIDs before the remaining 25 on each poll, so the first 50 are fetched and deduplicated repeatedly while UIDs 51-75 are never fetched. After the date window advances, the skipped messages no longer match the new SINCE date and never reach comm.ingest.

## Evidence
- `crates/khive-channel-email/src/connector/imap.rs:204` — Caller timestamp is reduced to an IMAP date string.
- `crates/khive-channel-email/src/connector/imap.rs:206` — Poll uses UID SEARCH SINCE date.
- `crates/khive-channel-email/src/connector/imap.rs:210` — Search result is collected into a Vec.
- `crates/khive-channel-email/src/connector/imap.rs:211` — UID list is truncated to the limit before any progress state exists.
- `crates/khive-channel-email/src/channel.rs:211` — EmailChannel hard-codes an IMAP fetch limit of 50.
- `/Users/lion/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/async-imap-0.9.7/src/client.rs:1208` — async-imap uid_search returns HashSet<Uid>, so iteration order is not a progress guarantee.
- `crates/khive-mcp/src/serve.rs:239` — The downstream poll loop advances by time, not by IMAP UID progress.

## Recommendation
Make polling progress-based: track UIDVALIDITY plus last_seen_uid and query/fetch above that high-water mark, or sort/page the full UID set with continuation state until the backlog is drained. Do not truncate an unordered day-level result as the progress mechanism.

## Verification
Extract UID selection/progress into a pure helper and test 75 same-day UIDs with limit 50 to prove all UIDs are eventually selected; add a mock IMAP/poll-loop regression for backlog drain behavior.

---
Source: automated static-analysis sweep.

ohdearquant:
## What

- Add mailbox-scoped IMAP checkpoint contracts (`ChannelCheckpoint`, `StoredChannelCheckpoint`, `ChannelPollPage`) and `comm.cursor_get` / `comm.cursor_commit` storage handlers, with an ADR-056 amendment documenting the persistence model.
- Replace unordered IMAP UID truncation with sorted, deduplicated pages that advance by UIDVALIDITY and high-water progress, so a full mailbox backlog drains instead of repeatedly re-fetching the same page.
- Give every selected IMAP UID a durable terminal disposition: a message with a missing body or a parse failure is now quarantined as a stable, queryable envelope via `comm.ingest` instead of erroring the whole fetched page and permanently blocking every later UID behind it.
- Wire the durable checkpoint path into the daemon's `channel_poll_loop`: each cycle now runs `comm.cursor_get` -> `Channel::poll_page` -> every `comm.ingest` in the page -> `comm.cursor_commit`, committing the cursor only when every envelope in the page durably ingested. A `cursor_get` failure skips that channel's cycle rather than polling from an unchecked default. A partial ingest failure leaves the cursor unadvanced so the next poll re-delivers the whole page; `comm.ingest`'s dedup (`INSERT OR IGNORE` on external ID) prevents the already-ingested messages from being double-stored on retry.
- Track each channel's bootstrap `since` floor (the date used in the IMAP `SINCE` clause while no UID high-water is committed yet) per `(channel_kind, channel_slug)` instead of one shared timestamp, and only advance it once that channel's full cycle -- `cursor_get`, `poll_page`, every `comm.ingest`, and `cursor_commit` -- has succeeded that tick. A `cursor_get` failure or a blocked first commit now leaves the floor exactly where it was, so a later successful cycle still searches from the original floor instead of a newer date; this closes a gap where a failing cycle spanning a calendar-day boundary would permanently skip the earlier day's mail.
- Seed that per-channel floor from a single `startup_since` timestamp captured before the loop's first sleep, not from `Utc::now()` re-read after it, so a daemon started shortly before a calendar-day boundary does not seed a channel's first bootstrap floor on the wrong side of it. Added a regression test covering this.
- Update `scripts/ci.sh` to run the whole `khive-mcp` crate under the `channel-email` feature unfiltered, instead of a name filter that only ran the ADR-094 lifecycle-sequencing module and silently skipped the newer cursor-gating and bootstrap-floor regression tests.
- De-flake two of `serve.rs`'s `channel-email` tests that were intermittently red in CI under `tokio`'s paused clock: both had replaced a real synchronization point with an iteration- or virtual-time-bounded polling loop, which cannot compensate for scheduling latency on the real OS-thread blocking pool the daemon's SQL writer runs on. One of them was also comparing a raw event-store length against a target count, when the same store also receives unrelated audit and heartbeat events, so the raw count could reach the target before the actual lifecycle-event subsequence did. Both now poll the real condition they care about (the filtered lifecycle-event sequence length, and the channel's own call counter) bounded by a wall-clock deadline instead of a tick budget.
- Add a regression test for the daemon's `cursor_commit`-rejected branch (a checkpoint `comm.cursor_commit` itself rejects, e.g. an out-of-range generation), which was otherwise unexercised: every other cursor-gating test in this file drove a `cursor_get` failure or a `comm.ingest` failure, never a rejected commit.

## Why

IMAP search returns an unordered UID set. Truncating that set before recording progress could repeatedly fetch the same messages and permanently starve later ones in the same date window. A single malformed message could also error an entire fetched page, and since progress rolled back to the old cursor on error, that message would be re-selected and re-fail forever, blocking every later UID in the mailbox behind it.

The durable checkpoint APIs existed but were never called from the daemon: `channel_poll_loop` still called the legacy in-memory-progress `Channel::poll` and only logged `comm.ingest` failures, so an ingest failure after a successful fetch silently lost that message's progress. Wiring the checkpoint path through means a partial failure is retried instead of skipped, and a durable restart sees the same checkpoint an in-process failure would have produced.

The bootstrap floor was originally a single shared timestamp advanced unconditionally once per tick, before the cycle it depends on had even run. That meant a `cursor_get` or ingest failure on a channel's first cycle still consumed the tick's floor update, and a daemon that started near UTC midnight could seed a channel's very first floor on the wrong side of that boundary -- either way, mail from the affected date range would be permanently excluded from the IMAP `SINCE` search.

## How tested

From `crates/` with `CARGO_TARGET_DIR=./target`:

- `cargo fmt --all -- --check` (exit 0)
- `cargo clippy -p khive-mcp --all-targets -- -D warnings` (exit 0)
- `cargo test -p khive-mcp --features channel-email` (exit 0, 109 passed, 0 failed)
- Each de-flaked test run 50 times in a loop (`for i in $(seq 50); do cargo test -p khive-mcp --features channel-email <name> || break; done`) with 0 failures; the new `cursor_commit`-rejection regression test run 20 times with 0 failures.

Fixes #449

### DIFF
--- a/crates/khive-channel-email/src/channel.rs
+++ b/crates/khive-channel-email/src/channel.rs
@@ -1768,6 +1768,101 @@ mod tests {
         );
     }
 
+    /// khive #449 High follow-up (Medium-2): the connector-level
+    /// `one_poison_uid_does_not_starve_51_later_valid_uids_and_cursor_passes_it`
+    /// test in `imap.rs` only proves `process_selected_page` assigns the
+    /// poison UID a `SelectedMessage::Malformed` disposition -- it never
+    /// proves that disposition survives into the actual `ChannelEnvelope`
+    /// `EmailChannel::poll_page` hands to `comm.ingest`. Drives a
+    /// `SelectedMessage::Malformed` entry (as the IMAP connector would
+    /// return it for an unparseable UID) through `poll_page`'s public
+    /// pipeline and asserts the resulting envelope carries the stable
+    /// `imap:{host}:{uidvalidity}:{uid}` external ID plus durable quarantine
+    /// metadata -- exactly what the daemon's ingest/query layer needs to
+    /// prove a durable quarantine record, not just an intermediate value.
+    #[tokio::test]
+    async fn poll_page_malformed_uid_produces_a_stable_external_id_and_quarantine_metadata() {
+        let since = Utc::now();
+        let config = make_config("maintainer@example.com");
+        let smtp = SmtpSender::with_connector(RecordingSmtp {
+            calls: Arc::new(Mutex::new(Vec::new())),
+        });
+
+        struct PoisonUidImap {
+            good: RawEmail,
+        }
+        #[async_trait]
+        impl ImapConnector for PoisonUidImap {
+            async fn fetch_page(
+                &self,
+                _since: DateTime<Utc>,
+                _limit: usize,
+                _progress: ImapProgress,
+            ) -> Result<ImapFetchPage, ChannelError> {
+                Ok(ImapFetchPage {
+                    emails: vec![
+                        SelectedMessage::Malformed {
+                            uid: 1,
+                            imap_external_id: "imap:imap.example.com:4:1".to_string(),
+                            reason: MalformedReason::MissingBody,
+                        },
+                        SelectedMessage::Email(Box::new(self.good.clone())),
+                    ],
+                    next_progress: ImapProgress {
+                        uid_validity: NonZeroU32::new(4),
+                        last_seen_uid: NonZeroU32::new(2),
+                    },
+                })
+            }
+        }
+
+        let mut good = make_email("sender@example.com", "imap:imap.example.com:4:2");
+        good.uid = 2;
+        let imap = ImapFetcher::with_connector(PoisonUidImap { good });
+        let ch = EmailChannel::with_connectors(config, smtp, imap);
+
+        let page = ch.poll_page(since, None).await.unwrap();
+        assert_eq!(
+            page.envelopes.len(),
+            2,
+            "the poison UID must still produce a durable quarantine envelope \
+             alongside the valid message, not a dropped/short page"
+        );
+
+        let quarantined = page
+            .envelopes
+            .iter()
+            .find(|e| e.external_id.as_deref() == Some("imap:imap.example.com:4:1"))
+            .expect(
+                "the malformed UID's stable external_id must be present on the \
+                 envelope actually handed to comm.ingest",
+            );
+        assert_eq!(
+            quarantined.from, "email:quarantine",
+            "a malformed UID must be attributed to the fixed quarantine marker"
+        );
+        assert_eq!(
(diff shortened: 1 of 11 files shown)


ohdearquant:
Verified at a0636b00: the flake fixes replace iteration/virtual-time budgets with condition polls bounded by wall-clock deadlines — `advance_until` now converges on the filtered lifecycle-event sequence rather than the raw store length (which audit/heartbeat events inflate), and the dedup test waits for the third `poll_page` observation carrying the committed checkpoint, which is only producible after the retry's ingest committed. The new rejected-commit regression test closes the one cursor-gating branch nothing exercised. CI fully green on both platforms including the unfiltered khive-mcp channel-email suite. Merging.

PR #785, opened 2026-07-10
fix(pack-git): mask credentials without dropping PR notes

### LINKED ISSUE #763: git.digest: credential-scanner false positive drops entire PR note (hash flagged as credential)
First-consumer report (phase-1 digest of a real repo): a PR body containing a 64-char hex string (a hash sitting next to a sensitive-looking word) was flagged credential-like by the digest write scanner, and the ENTIRE PR note was dropped. Effect: that PR is silently absent from provenance.

Same false-positive class as the comm.send filter's known trigger-word-adjacency issue.

Proposed fix, in preference order:
1. Mask-and-ingest: redact the flagged span (first6…[N chars]) and keep the note.
2. At minimum: surface dropped records in the digest result payload (list of skipped PR/commit ids + reason) so a consumer can backfill, instead of silent absence.

Acceptance: a PR body with a bare 64-hex hash near the word "token" ingests with the span masked, and the digest summary reports zero silent drops.

ohdearquant:
## What

- Mask credential-like spans in pull request bodies and titles before creating the stored note.
- Preserve the containing pull request note, title metadata, project annotations, and issue-reference edges.
- Add acceptance coverage for exact masking, repeated matches, clean values, truncation order, references, and cursor retry behavior.

## Why

The runtime secret gate correctly rejects unmasked credential-like values, but pull request ingestion passed raw title data into note creation. A detector match could therefore reject the entire pull request note and lose otherwise safe provenance. Masking only the detected spans retains that provenance without storing the flagged value.

## How tested

- `cargo clippy -p khive-pack-git --all-targets -- -D warnings`
- `cargo test -p khive-pack-git` (61 passed: 36 unit, 25 acceptance, 0 doc)

Fixes #763


oceanwaves630 (approved):
Gate PASS (Leo, content-level at 6043a782). My anchor read confirms safe_title routes BOTH write sites (properties.title and the formatted name) through the canonical secret_gate::mask_secrets; the reviewer leg additionally proved mutation-sensitivity empirically — reverting just the safe_title line at this exact SHA reproduces the pre-fix symptom (write blocked, prs_ingested: 0), restoring it passes. 36 unit + 25 acceptance green at the pin. lattice-embed is dev-dependencies only.

FOLLOW-UP TO FILE (same bug class, out of this PR's scope): ingest_issues still passes raw issue.title into properties.title and name — a credential-shaped ISSUE title silently drops the issue note today, exactly the #763 mechanism. File it and fix in a sibling PR.

Merge coordination: this PR and #789 edit the same acceptance.rs import hunk — no correctness dependency, second-to-merge needs a trivial import union.

PR #786, opened 2026-07-10
fix(channel-email): quote/backslash-aware tokenization for Authentication-Results (#501)

### LINKED ISSUE #501: channel-email: quote-aware A-R property tokenizer (defense-in-depth, hardening)
## Context
Fast-follow to #500 (EXO no-authserv-id trust anchor). An adversarial review pass of #500 found that `parse_header`'s method-property loop in `crates/khive-channel-email/src/auth_results.rs` tokenized segments with naive `split_whitespace()`, ignoring RFC 8601 quoting. A quoted-string property value with embedded whitespace was shattered into multiple tokens, and `props.insert` (last-write-wins) let an injected `smtp.mailfrom=`/`header.from=`/`header.d=` token overwrite the real alignment domain.

## Exploitability
**No live bypass.** EXO emits the three alignment properties as bare domains (verified against a real production fixture), so nothing attacker-controlled reaches the splitter today. Review confirmed #500 is safe to merge as-is. Twin of the #496 segment-level fix; closes the mechanism at the parser regardless of what a receiving boundary emits.

## Status
**Implemented** on branch `fix-email-exo-prop-tokenizer` (commit a8eecb76): adds `split_top_level_ws` (quote-atomic whitespace tokenizer), routes both `first_token` and the property loop through it, +3 regression tests. 129 tests green, clippy/fmt clean. PR pending review as a fast-follow.

ohdearquant:
## Summary
- Authentication-Results header parsing now tracks RFC 5322 quoted-string state (with backslash escapes) when splitting resinfo segments into whitespace-delimited tokens.
- Detection of the no-authserv-id form (unquoted `=` in the first token) is now quote-aware, so a quoted authserv-id or resinfo value containing whitespace, `;`, or `=` is no longer misparsed as a segment or token boundary.

## Test plan
- [x] `cargo fmt --all -- --check` (crates/, worktree-local CARGO_TARGET_DIR) — clean
- [x] `cargo clippy -p khive-channel-email --all-targets -- -D warnings` — clean
- [x] `cargo test -p khive-channel-email` — 159 passed, 0 failed

Closes #501

### DIFF
--- a/crates/khive-channel-email/src/auth_results.rs
+++ b/crates/khive-channel-email/src/auth_results.rs
@@ -185,6 +185,100 @@ fn split_top_level_segments(raw: &str) -> Vec<String> {
     segments
 }
 
+/// Split one top-level `resinfo` segment (already produced by
+/// [`split_top_level_segments`], so comments are already stripped and only
+/// quoted semicolons can remain) into whitespace-delimited tokens, tracking
+/// RFC 5322 quoted-string state (with `\` escapes) so quoted whitespace is
+/// never treated as a token boundary.
+///
+/// This is the second of two delimiter layers: the segment scanner above
+/// finds `;` boundaries; this one finds whitespace boundaries within a
+/// segment. Both share the same quoted-pair semantics (`\` plus the
+/// following character is one atomic unit, regardless of what that character
+/// is), but this layer never sees `(...)` comments -- those were already
+/// discarded by the segment scanner. A token is returned verbatim, including
+/// any retained quote and backslash characters; it is never unquoted.
+///
+/// Malformed input (an unmatched `"`, or a `\` as the final character while
+/// quoted) is handled conservatively: the remainder of the segment is
+/// retained as one atomic token through EOF rather than resuming whitespace
+/// splitting, so a malformed quoted tail can never be reinterpreted as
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Content-level gate PASS (lambda:leo). Independent verification, pinned at 11e3671f0:

- Reviewer leg hand-traced the quote/backslash state machine against the hostile scenarios (semicolon inside quotes, trailing backslash at EOF, empty quoted string, backslash-pair before closing quote) — all correct RFC 5322 quoted-pair semantics, all chars()-based so no multi-byte panic surface.
- Mutation verification was EMPIRICAL: leg reverted the three converted call sites back to naive split_whitespace()/.contains('=') in a throwaway worktree and reran — 9/49 tests fail, including all three alignment-forgery tests and the quoted-authserv-id misclassification test. The new tests genuinely pin the #501 vulnerability class.
- Confirmed via grep on the pre-PR baseline that exactly three call sites used the vulnerable tokenization; all three converted, nothing else touched. End-to-end quarantine test drives the real parse_raw_bytes path.
- My anchor read of split_top_level_ws: malformed quoted tail retained atomically through EOF, so it can never be reinterpreted as additional tokens — the fail-safe direction.

Non-blocking note: quoted property values are never quote-stripped (documented design), so a legitimately-quoted domain fails closed toward quarantine rather than aligning. Pre-existing, fails safe.

PR #787, opened 2026-07-10
fix(channel-email): bound OAuth token refresh with a timeout under the cache lock (#477)

### LINKED ISSUE #477: khive-channel-email: audit medium-severity findings (1)
1 medium-severity findings from the static-analysis sweep over `khive-channel-email`. Each item is independently actionable; split out if one grows.

---

### CHANNELEMAIL-AUD-003: OAuth token refresh can hang both IMAP and SMTP indefinitely

Severity: **medium** | Confidence: high | Priority: P2 | Categories: concurrency, errors, performance
Crate: `khive-channel-email` | Finding id: `CHANNELEMAIL-AUD-003`

## Finding
A transient network stall can disable OAuth-mode email polling and delivery indefinitely.

## Failure scenario
OAuth mode is configured and the cached token is absent or expired. The token endpoint or proxy accepts the request but never completes a response. The first IMAP poll enters get_token, takes the mutex, and awaits reqwest forever. SMTP sends and later IMAP polls block behind the same mutex, so inbound and outbound email remain stuck until process restart.

## Evidence
- `crates/khive-channel-email/src/oauth.rs:144` — get_token acquires the token-cache mutex.
- `crates/khive-channel-email/src/oauth.rs:150` — get_token awaits fetch_token while the mutex guard is still live.
- `crates/khive-channel-email/src/oauth.rs:227` — fetch_token constructs a default reqwest client.
- `crates/khive-channel-email/src/oauth.rs:231` — Token HTTP request is awaited without a surrounding timeout.
- `/Users/lion/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/reqwest-0.12.28/src/async_impl/client.rs:1440` — reqwest documents the total request timeout default as no timeout.
- `crates/khive-channel-email/src/connector/imap.rs:162` — IMAP OAuth authentication calls the shared token provider.
- `crates/khive-channel-email/src/connector/smtp.rs:173` — SMTP OAuth delivery calls the shared token provider.

## Recommendation
Use a reqwest client with total/connect/read timeouts or wrap token fetch in tokio::time::timeout. Ensure timeout errors release the mutex and surface as ChannelError::Auth so later polls can retry.

## Verification
Refactor token fetching behind an injectable client/future and test a never-resolving token request; assert get_token returns a timeout error and a subsequent call can acquire the cache lock.

---
Source: automated static-analysis sweep.

ohdearquant:
## Summary
- `TokenProvider::get_token` now bounds a cache-miss/expired-token refresh with a 15-second `tokio::time::timeout` while the cache mutex is held.
- A stalled or hung refresh now times out, releases the mutex, and returns a `ChannelError::Auth` instead of holding the lock indefinitely and starving every other caller waiting on a token.
- Added `test-util` to the `tokio` dev-dependency features to support paused-clock timeout tests.

## Test plan
- [x] `cargo fmt --all -- --check` (crates/, worktree-local CARGO_TARGET_DIR) — clean
- [x] `cargo clippy -p khive-channel-email --all-targets -- -D warnings` — clean
- [x] `cargo test -p khive-channel-email` — 143 passed, 0 failed, including new coverage for timeout-then-retry and concurrent-refresh-sharing behavior

Closes #477

### DIFF
--- a/crates/khive-channel-email/Cargo.toml
+++ b/crates/khive-channel-email/Cargo.toml
@@ -31,4 +31,4 @@ base64 = { workspace = true }
 rand = { workspace = true }
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["full"] }
+tokio = { workspace = true, features = ["full", "test-util"] }
--- a/crates/khive-channel-email/src/oauth.rs
+++ b/crates/khive-channel-email/src/oauth.rs
@@ -4,12 +4,17 @@
 //! set. The token endpoint is Microsoft's app-only flow:
 //! `POST https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token`.
 
-use std::time::{Duration, Instant};
+use std::{
+    future::Future,
+    time::{Duration, Instant},
+};
 
 #[cfg(test)]
 use base64::Engine as _;
 use khive_channel::ChannelError;
 use tokio::sync::Mutex;
+
+const TOKEN_REFRESH_TIMEOUT: Duration = Duration::from_secs(15);
 
 // ──────────────────────────── OAuth error-body sanitization (Finding 1)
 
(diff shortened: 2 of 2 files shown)


oceanwaves630 (approved):
Content-level gate PASS (lambda:leo). Independent verification, pinned at cb5c7078:

- Reviewer leg reproduced the full suite (143 passed, 0 failed, cargo test -p khive-channel-email --all-targets) and ran a genuine mutation check: stripping the tokio::time::timeout wrapper made timed_out_refresh_releases_lock_and_waiting_retry_succeeds hang past 60s — the new test is mutation-sensitive to exactly this fix.
- My anchor read of get_token_with_fetcher: guard dropped on every branch (cache hit, fetch error, timeout, success); timeout path never writes *guard, so stale state is preserved and the next call retries. Fail-fast-then-retry, no cache corruption, no deadlock (tokio::sync::Mutex, correct across .await).
- 15s bound matches the crate's existing IMAP login / XOAUTH2 convention; callers at connector/imap.rs:169 and smtp.rs:235 propagate via ? identically to sibling timeouts. No scope creep.

Two non-blocking notes for follow-up: (1) the mutation manifests as a test HANG, not a fail-fast assertion — bounding stalled.await inside the test would make a future regression fail CI cleanly instead of eating a runner; (2) TOKEN_REFRESH_TIMEOUT is hardcoded, consistent with siblings, config knob only if a real deployment needs it.

PR #788, opened 2026-07-10
fix(pack-git): refuse unowned cache-key directories in ensure_clone

### LINKED ISSUE #765: git.digest: self-heal corrupted URL-mode clone cache (promisor missing-object) with refetch/re-clone
First-consumer report: URL-mode digest of a repo failed with `git log --name-only failed: fatal: <sha> is in the commit graph file but not in the object database ... could not fetch from promisor remote` — a partial/promisor clone in `~/.khive/scratch/git-digest/<key>/` had dropped objects. Manual `git fetch --refetch origin` repaired it.

Proposed fix in `khive-pack-git::cache::ensure_clone`: when the post-clone/fetch validation (or the subsequent `git log --name-only`) fails with a missing-object/promisor error, retry once with `git fetch --refetch`; if that also fails, evict the cache entry and fresh-clone. Never surface a corrupt-cache error to the caller when the remote is reachable — the cache is disposable by design (ADR-088 Amendment 1 remote-URL mode).

Acceptance: a deliberately object-pruned cache entry digests successfully on first call.

ohdearquant:
## Summary

Remediates the blocking finding from `reviewer-6/final_review_765.md` (High): `ensure_clone` treated any existing `.git` directory at a cache-key path as a fetchable slot without first proving ownership via `is_owned_entry`, so a foreign or markerless directory occupying that path could be fetched into, and later become eligible for deletion by the reclone repair path.

- `ensure_clone` now refuses with `UnsafeToReplace` unless the existing path at the cache-key slot already passes `is_owned_entry`.
- `install_fresh_clone` writes the `.khive-last-used` marker into the staging clone before the atomic rename, so a process interruption between clone and rename can never leave a live, markerless slot at the cache-key path.
- `ensure_clone`'s and `refetch_clone`'s over-cap cleanup now route through `remove_owned_entry` and propagate its failure instead of discarding it behind a laundered `CloneTooLarge`.
- `is_owned_entry` rejects a symlink sitting at the cache-key path, not just a wrong-shaped name or a missing marker.
- New coverage: two `cache.rs` unit tests (markerless real-git-directory lookalike, symlink lookalike) and two public `git.digest`-verb regression tests in `recovery_tests.rs`, all asserting refusal with sentinel operator data left completely untouched.

## Test plan

Run from `crates/` with `CARGO_TARGET_DIR=./target`:

- [x] `cargo fmt --all -- --check` - exit 0
- [x] `cargo clippy -p khive-pack-git -p khive-runtime --all-targets -- -D warnings` - exit 0
- [x] `cargo test -p khive-pack-git -p khive-runtime` - all suites green (59 lib + 19 recovery + 731 khive-runtime + 2 cold-boot + 59 integration; 0 failed)

Closes #765

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


oceanwaves630 (approved):
Content-level gate PASS (lambda:leo). Independent verification at pinned 06592d0a:

- TOCTOU fix verified at the anchor: is_owned_entry re-check sits immediately before fetch_refetch with zero intervening filesystem or async operations; failure arm returns CacheError::UnsafeToReplace with no partial adoption and no touch() reached. The doc comment is honest about scope — narrows the ensure_clone-to-repair adoption window without claiming to close a true concurrent-writer race (no same-key serialization exists in the crate today). Accurate, not oversold.
- Corruption fixture verified as REAL: production ensure_clone partial clone against a real file:// origin, actual pack located via git verify-pack -v and deleted plus the loose copy, precondition assertion that git log genuinely fails before repair, repair via the production refetch_clone primitive, post-repair re-verification. Not a hand-crafted layout. Its disclosed limitation (can't drive through the public git.digest verb because no local corruption yields the literal promisor-keyword diagnostic) is defensible and documented in the test itself.
- Mutation-sensitivity verified EMPIRICALLY: reviewer leg reverted the re-check in a throwaway worktree and reran — refetch_clone_refuses_a_markerless_slot_under_the_cap fails immediately with the captured output showing a real fetch advancing origin/main against the markerless slot. The test is load-bearing.
- Remediation round tightly scoped: cache.rs + recovery_tests.rs only. 80/80 tests pass at the pin.

Follow-up worth an issue (not blocking): the doc comment's own admission — no same-key mutation serialization — is the residual race. If cache mutation ever goes concurrent (daemon multi-ingest), a per-slot advisory lock is the structural fix.

PR #798, opened 2026-07-10
fix(runtime): add RuntimeConfig::no_embeddings() opt-out constructor

### LINKED ISSUE #396: RuntimeConfig with embedding_model: None still seeds additional_embedding_models, hard-failing remember on model-less machines
## Bug

Setting `embedding_model: None` on `RuntimeConfig` does not produce an embedding-free runtime:

```rust
let config = RuntimeConfig {
    db_path: None,
    embedding_model: None, // intent: no embedders, text-only recall
    ..RuntimeConfig::default()
};
```

`RuntimeConfig::default()` (crates/khive-runtime/src/config.rs, `Default` impl) separately seeds `additional_embedding_models` with `ParaphraseMultilingualMiniLmL12V2` whenever `KHIVE_ADDITIONAL_EMBEDDING_MODELS` is unset. The runtime therefore still registers an embedder, and `memory.remember` / `memory.recall` lazily initialize it on first use.

On a machine with local model files this is invisible. On a machine without them (CI runners, fresh installs), the first `memory.remember` hard-fails:

```
runtime: embedding: model initialization failed: IO error: No such file or directory
```

The failure is hard, not degraded: `create_note_inner`'s single-model vector path propagates the embed error with full compensation (the note row and FTS doc are rolled back), so the write is lost.

## Why it matters

- The obvious spelling of "no embeddings" (`embedding_model: None`) silently means "no default model, but one additional model". Every consumer that wants a deterministic no-embed runtime must know to also write `additional_embedding_models: vec![]`.
- The trap is environment-dependent: tests written this way pass on developer machines and fail only on model-less runners. This surfaced in downstream integration tests that had been green for their entire life because the repository they lived in had no CI.
- The runtime's own test code already uses the two-field form (runtime.rs uses `embedding_model: None` together with an explicitly cleared additional list), which shows the intent exists but is not encoded in the API.

## Repro

On a machine without model files under the khive data dir (or with `HOME` pointed at an empty directory):

```rust
let config = RuntimeConfig { db_path: None, embedding_model: None, ..RuntimeConfig::default() };
let rt = KhiveRuntime::new(config).unwrap();
// dispatch memory.remember → Err(Runtime(Embedding(ModelInitialization("IO error: ..."))))
```

Adding `additional_embedding_models: vec![]` makes the same code pass, with recall degrading to the FTS text leg.

## Suggested fix (either)

1. Make `embedding_model: None` authoritative: when the primary model is explicitly `None`, do not seed `additional_embedding_models` from the env default (an explicit env var or field value still wins).
2. Or add a constructor/builder method (`RuntimeConfig::no_embed()` or similar) that produces the genuinely embedding-free config, and document the two-field requirement on both fields.

Option 1 matches the principle of least surprise; option 2 is non-breaking.

ohdearquant:
## Summary

- `RuntimeConfig::default()` computes `embedding_model` and `additional_embedding_models` independently, so writing `RuntimeConfig { embedding_model: None, ..RuntimeConfig::default() }` does not actually disable embedding: `additional_embedding_models` still carries the env-driven fallback seed (`ParaphraseMultilingualMiniLmL12V2` when `KHIVE_ADDITIONAL_EMBEDDING_MODELS` is unset).
- The note-write path fans out embedding to every *registered* model regardless of `embedding_model`, so the first `memory.remember` on a machine without local model files hard-fails instead of degrading to FTS-only.
- Adds `RuntimeConfig::no_embeddings()`, which clears both fields together — the one ergonomic, correct spelling for a model-less runtime. `Default::default()`'s existing behavior is unchanged (no existing caller is affected).
- Design decision: `no_embeddings()` deliberately ignores `KHIVE_ADDITIONAL_EMBEDDING_MODELS` — a caller reaching for it wants zero embedders unconditionally, not "zero unless the environment disagrees".

## Test plan

- [x] `config::no_embeddings_tests::no_embeddings_clears_both_fields` — direct verification of the fix
- [x] `config::no_embeddings_tests::no_embeddings_ignores_additional_env_override` — confirms the env-override design decision
- [x] `config::no_embeddings_tests::default_still_seeds_additional_models_when_env_unset` — regression guard that `Default`'s existing behavior, including the pre-fix bug shape, is unchanged
- [x] `operations::tests::no_embeddings_config_registers_zero_embedders` — runtime-level: zero embedders registered
- [x] `operations::tests::no_embeddings_runtime_create_note_succeeds_without_model_fanout` — `create_note` succeeds with zero registered embedders (no lattice model build attempted)
- [x] `cargo test -p khive-runtime` (736 unit + 2 + 59 integration tests, all green)

Closes #396

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -1760,31 +1760,30 @@ pub fn resolve_runtime_config(inputs: RuntimeConfigInputs<'_>) -> anyhow::Result
     // we exclude brain_profile from the default spread and set it to None (CLI-only).
     let cli_brain_profile = inputs.brain_profile.filter(|s| !s.trim().is_empty());
 
-    let base_config = RuntimeConfig {
-        db_path,
-        default_namespace: inputs.namespace,
-        packs,
-        // Explicit CLI flag only at this tier — env and config-file tiers are applied
-        // below in resolve_config / resolve_actor_from_config and apply_env_brain_profile.
-        brain_profile: cli_brain_profile,
-        ..RuntimeConfig::default()
-    };
-
     // Threaded into the config-file resolvers so tier-3 project-local config
     // discovery anchors to the resolved database's directory rather than the
     // process cwd when an explicit `--db`/`KHIVE_DB` is given (kills config_id
     // drift between a client and the daemon serving the same database at a
-    // different working directory). Deliberately NOT `base_config.db_path`
-    // (which materializes the `$HOME/.khive/khive.db` default when unset,
-    // #689) — an unset db must fall through to cwd-anchored discovery instead
-    // of silently searching the home directory.
+    // different working directory). Deliberately NOT the base config's own
+    // `db_path` (which materializes the `$HOME/.khive/khive.db` default when
+    // unset, #689) — an unset db must fall through to cwd-anchored discovery
+    // instead of silently searching the home directory.
     let db_path_for_config = config_discovery_db_anchor(inputs.db);
 
     let resolved = if inputs.no_embed {
+        // `RuntimeConfig::no_embeddings()` is the canonical "zero embedders"
+        // constructor (issue #396) — it clears `embedding_model` and
+        // `additional_embedding_models` together, unlike a manual two-field
+        // override which can leave `additional_embedding_models` populated
+        // from `KHIVE_ADDITIONAL_EMBEDDING_MODELS`.
         let no_embed_base = RuntimeConfig {
-            embedding_model: None,
-            additional_embedding_models: vec![],
-            ..base_config
+            db_path,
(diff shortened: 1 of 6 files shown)


ohdearquant (commented):
Reviewed at fe2cab39. Verified the `memory()` refactor is behavior-preserving against the current `Default` impl: `default_namespace`, `gate`, `backend_id`, `visible_namespaces`, `allowed_outbound_namespaces` all match what the old explicit literal pinned. One soft note for the record: `memory()` now inherits gate/backend/namespaces implicitly, so a future change to `Default` propagates into it silently — acceptable coupling for a test/ephemeral constructor, just worth remembering. Constructor semantics (deliberately ignoring KHIVE_ADDITIONAL_EMBEDDING_MODELS) are the right call and well documented.

PR #800, opened 2026-07-10
fix(runtime): honor RFC3339 timezone offsets in relative-time display

### LINKED ISSUE #754: presentation: relative-time display drops RFC3339 timezone offsets — offset-bearing trigger_at renders hours wrong while firing stays correct
## Summary

Agent-mode presentation renders offset-bearing RFC 3339 timestamps with the offset silently dropped. `parse_iso8601_unix` (`crates/khive-runtime/src/presentation.rs:586`) reads the wall-clock digit positions and treats them as UTC; its doc comment says "no timezone offsets other than 'Z'", but instead of rejecting offset-bearing input it mis-parses it. A `schedule.remind(at="2026-07-09T11:55:00-04:00")` event (15:55Z, four hours in the future) displays as already due ("Ns ago" / near-now), because the display reads it as 11:55Z.

## What is and is not affected

- **Firing is correct.** The pending-events scan parses `trigger_at` with chrono's offset-aware `DateTime<Utc>` FromStr and compares instants (`crates/kkernel/src/pending_events.rs:203`); the SQL filter only matches on `status`. Handler-side validation (`validate_at`, `crates/khive-pack-schedule/src/handlers.rs:76-77`) and agenda windowing (handlers.rs:929-1009) are likewise chrono-based and offset-aware.
- **Only the compact/relative display path** (`compact_timestamp` → `parse_iso8601_unix` → `relative_time`) is offset-naive, so agent-mode output shows a wrong relative time while the system will fire at the right instant — display and firing silently disagree, which reads as data corruption to the caller.

## Fix direction (small)

In `parse_iso8601_unix`: after the seconds field, look for `Z`, `+HH:MM`/`-HH:MM` (or `+HHMM`), and apply the offset to the computed Unix seconds. Any other trailing content → return `None` (graceful degradation to truncation already exists). Regression tests: `-04:00` future timestamp must not render as "ago"; `Z` and fraction-bearing forms unchanged.

Found live by fleet dogfooding (offset-bearing `at` armed, display claimed already-due, event actually pending for the correct instant).

ohdearquant:
## Summary

- `parse_iso8601_unix` in `crates/khive-runtime/src/presentation.rs` only read bytes `[0..19]` (`YYYY-MM-DDTHH:MM:SS`) and silently ignored any trailing `Z`, `+HH:MM`/`-HH:MM` offset, or fractional-seconds suffix, treating the parsed value as if it were already UTC.
- This made the Agent-mode relative-time display (`compact_timestamp` -> `relative_time`) render offset-bearing timestamps incorrectly, while the firing path (which uses chrono's offset-aware parsing) stayed correct — producing a visible mismatch between the displayed relative time and the actual due time.
- Added `parse_tz_offset_secs`, which parses the tail of the timestamp (skipping optional fractional seconds, then `Z` / `+HH:MM` / `-HH:MM` / `+HHMM` / `-HHMM`) and subtracts the offset from the parsed local-wall-clock value to recover the true UTC instant. Malformed tails still degrade gracefully to `None`, matching the existing fallback-to-truncation behavior.

Closes #754

## Test plan

- [x] `cargo fmt --all -- --check` (from `crates/`)
- [x] `cargo test -p khive-runtime presentation` — new regression tests for `+HH:MM`, `-HH:MM`, `+HHMM`/`-HHMM`, `Z`, fractional seconds, malformed tails, and the `compact_timestamp` end-to-end behavior (offset-bearing future time no longer renders as "ago"; offset-bearing past time correctly renders relative)
- [x] `cargo test -p khive-runtime` (full crate suite) — 741 + 2 + 59 passed, 0 failed

### DIFF
--- a/crates/khive-runtime/src/presentation.rs
+++ b/crates/khive-runtime/src/presentation.rs
@@ -648,6 +648,9 @@ fn parse_tz_offset_secs(tail: &str) -> Option<i64> {
         ),
         _ => return None,
     };
+    if hh > 23 || mm > 59 {
+        return None;
+    }
     Some(sign * (hh * 3600 + mm * 60))
 }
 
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):
Reviewed at d2796a75. Verified the core math: `local - offset_secs` is the correct direction (+04:00 wall clock is ahead of UTC, so subtracting recovers the instant), and the `&s[19..]` slice is boundary-safe because bytes 0..19 are ASCII-validated by `parse_digits` upstream. Range checks (hh<=23, mm<=59) and the compact/colon dual forms are covered bidirectionally in tests, including the two end-to-end `compact_timestamp` regressions. One awareness note: the `?` on the offset parse means trailing junk the old parser silently ignored now degrades to truncated display instead of rendering wall-clock-as-UTC — strictly more correct, and tested.

PR #802, opened 2026-07-10
fix(query): warn on GQL result truncation at the 500-row cap

### LINKED ISSUE #777: query(): GQL results silently capped at 500 rows; LIMIT above cap ignored with no truncation signal
## Behavior

`query()` with a GQL `MATCH` returns at most 500 rows. A `LIMIT` above 500 is silently clamped. The response carries no `truncated` flag, no warning, and no error, so the caller cannot distinguish "500 total matches" from "3,500 matches, you got a seventh of them."

## Reproduction

```
query(query="MATCH (d:document) RETURN d.id, d.name, d.properties")
query(query="MATCH (d:document) RETURN d.id, d.name, d.properties LIMIT 5000")
```

Both return 500 rows; outputs are byte-identical (verified by diff, 200,542 chars each).

## Source

- `crates/khive-query/src/compilers/sql.rs:104` — `max_limit: 500` (server-side safety cap)
- `sql.rs:562`, `sql.rs:1139` — `query.limit.unwrap_or(opts.max_limit).min(opts.max_limit)` silently clamps
- The clamp itself is intended and tested (`limit_capped_by_max_limit`); the defect is that the clamp is invisible to the caller.

## Impact

A read verb that silently returns a subset of the matching set poisons every downstream conclusion with full confidence. Field incident: an audit over document entities got a truncated result set, concluded recent writes had failed (they had not), and fell back to paging the entire entity table client-side, where a second filtering bug produced a wrong count.

## Requested fix

Either honor larger LIMITs with a cursor, or keep the cap and return an explicit `truncated: true` plus `total_matched` (or at minimum row_count == cap => warning field). Silent is the only unacceptable option.

## Related scope limits (same lane, possibly separate work)

- No string-containment operator: `WHERE d.name CONTAINS 'ADR'` is a parse error.
- Properties map not indexable: `WHERE d.properties.type = 'adr'` is a parse error.
- With both missing, the only way to answer "how many documents of type X" is a full client-side scan, which is exactly what the silent cap breaks.

ohdearquant:
## Summary

GQL execution caps results at `max_limit`, which defaults to 500 rows. Previously, callers could not distinguish an exact 500-row result from a larger result set truncated to 500. Inferring truncation from the requested limit would also be incorrect because `LIMIT 600` may match only a few rows, while a query with no explicit limit may match more than the cap.

## Implementation

- The compiler now uses `effective_limit` for both fixed-length and variable-length queries. An explicit limit at or below `max_limit` is emitted unchanged. When no limit is specified or the explicit limit exceeds the cap, the SQL requests `max_limit + 1` rows so the extra row acts as a truncation sentinel.
- `CompiledQuery` adds the public `truncation_check: Option<TruncationCheck>` field, and `TruncationCheck` is exported from `khive-query`. It carries the applied `max_limit` and the optional explicit requested limit to the execution layer.
- `KhiveRuntime::query_with_metadata` executes the sentinel query and checks the actual row count. If more than `max_limit` rows were returned, it trims the sentinel row, returns exactly the capped number of rows, and appends a pagination warning. The warning distinguishes an explicit over-cap request from a query with no `LIMIT` clause.
- No warning is emitted when the sentinel is absent or when the query returns no sentinel row. This avoids false positives for explicit limits above the cap with few real matches. An explicit limit equal to or below the cap remains caller-selected behavior and does not trigger truncation detection.
- Compiler warnings are extended rather than replaced, preserving existing validation warnings alongside any runtime truncation warning in the existing `QueryResult.warnings` response path.

## Test coverage

Compiler tests cover the emitted limit and metadata for both compile paths:

- `limit_capped_by_max_limit`
- `limit_over_cap_requests_sentinel_row`
- `limit_exactly_at_cap_no_sentinel`
- `limit_below_cap_no_sentinel`
- `no_explicit_limit_requests_sentinel_row`
- `variable_length_limit_over_cap_requests_sentinel_row`
- `max_limit_zero_compiles`

Runtime integration tests exercise real result counts around the 500-row boundary:

- `no_explicit_limit_under_and_at_cap_emits_no_warning`
- `no_explicit_limit_over_cap_warns_and_strips_sentinel`
- `explicit_limit_variants_against_501_matches`
- `explicit_limit_over_cap_with_few_real_matches_emits_no_warning`

Closes #777

### DIFF
--- a/crates/khive-query/src/compilers/sql.rs
+++ b/crates/khive-query/src/compilers/sql.rs
@@ -87,6 +87,23 @@ pub struct CompiledQuery {
     pub params: Vec<QueryValue>,
     pub return_vars: Vec<ReturnItem>,
     pub warnings: Vec<String>,
+    /// Present when the server-side `max_limit` cap is the binding
+    /// constraint on this query (no explicit `LIMIT`, or an explicit
+    /// `LIMIT` above the cap). The emitted SQL fetches `max_limit + 1`
+    /// rows (a sentinel row) so the execution site can detect whether the
+    /// true match count exceeded the cap and warn accordingly, instead of
+    /// inferring truncation from the requested `LIMIT` alone (issue #777).
+    pub truncation_check: Option<TruncationCheck>,
+}
+
+/// See [`CompiledQuery::truncation_check`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TruncationCheck {
+    /// The server-side row cap; the execution site truncates to this many
+    /// rows after stripping the sentinel.
+    pub max_limit: usize,
+    /// The caller's explicit `LIMIT`, if any, for warning-message purposes.
+    pub requested_limit: Option<usize>,
 }
 
 /// Runtime options injected by the caller to scope and cap query execution.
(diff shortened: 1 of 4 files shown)


ohdearquant:
Verified the sentinel-row implementation end to end: `effective_limit` only arms the check when the cap is the binding constraint (explicit LIMIT ≤ cap fetches exactly the requested count with no sentinel), the execution site strips the sentinel before returning and warns only when a real cap+1th row existed, and the max_limit=0 edge correctly fetches one sentinel row and truncates to zero. Integration coverage straddles the boundary at 499/500/501 and includes the LIMIT-600-with-20-matches false-positive case. CI run 29112962048 green on both platforms at b7561d3c. Merging.

PR #812, opened 2026-07-10
fix(runtime): serve stale ANN during rebuild and unblock recall request path

### LINKED ISSUE #791: memory.recall intermittent 300s timeouts under concurrent MCP load
Three occurrences today across two concurrent agent sessions: memory.recall hung to a 300s client timeout while sibling sessions' recall calls on the same daemon succeeded normally. Both affected sessions were running while 6+ other MCP clients (review/audit agents) were active against the same daemon, alongside git-digest corpus queries.

Observed:
- Session A: one 300s timeout mid-task, subsequent calls fine.
- Session B: two consecutive 300s timeouts at task start, before any successful recall.
- Same wall-clock window; my own session's recall calls succeeded throughout.

Hypotheses worth checking: ANN search serialization under concurrent recall (warm-index lock contention), WAL reader pinning during heavy concurrent list/traverse (large-response queries were in flight; a 300K-char list response was generated in the same window), or embedder queue starvation.

Repro direction: N concurrent MCP clients issuing recall while one client streams large list/traverse responses; measure recall latency distribution.

ohdearquant:
## Summary

Fixes #791: `memory.recall` intermittently hung to the 300s MCP timeout.

Root cause: `memory.remember` (and the KG note-mutation hook, and `memory.prune`)
called `ann::invalidate_namespace`, which cleared every in-memory ANN index slot and
deleted the persisted snapshot for every model, synchronously, on the write's own
request path. A `memory.recall` landing before the resulting background rebuild
finished had nothing left to serve, so it fell into a synchronous, single-flighted,
full-corpus Vamana rebuild directly on its own request path
(`ann::ensure_ann_for_model` -> `load_and_build_from_vector_store` ->
`VamanaIndex::build`, a plain sync fn never wrapped in `spawn_blocking`). Every other
concurrent recall for the same model queued behind the same per-model mutex and paid
the same wall-clock wait.

This PR replaces that synchronous invalidate-and-rebuild path with a durable,
fail-closed eventual-consistency contract, normatively defined in
[ADR-107](docs/adr/ADR-107-memory-ann-lifecycle.md) (memory-pack scope; PR #813).

## Design

1. **Writes no longer destroy the fast fallback.** `memory.remember`, `memory.prune`,
   and the KG-side note-mutation hook bump the model's write-generation counter
   instead of clearing the cache or deleting the snapshot. The previously-installed
   index and snapshot stay in place and keep serving reads until a fresher build
   replaces them via generation-checked install (`install_if_fresher`); the on-disk
   snapshot is replaced by one atomic `INSERT OR REPLACE`, never deleted ahead of its
   replacement.

2. **Recall serves stale instead of blocking.** The recall path calls
   `ann::search_loaded` unconditionally and serves a present-but-stale hit
   immediately, firing the existing background warm (`ensure_ann_background`) so a
   later recall benefits from the fresher build once it installs. Only a genuine
   cache miss (nothing installed yet: the first `memory.recall` for a model in a
   process lifetime) still builds inline, and that build's CPU-bound graph
   construction runs via `tokio::task::spawn_blocking` so it never monopolizes a
   tokio worker.

3. **Bounded chained re-enqueue with debounce.** The background-rebuild task
   re-reads the write-generation counter on exit and, if a write raced in while it
   was building, re-enqueues itself against the new generation rather than relying on
   a later caller to notice, up to 3 attempts per task. Under continuous writes,
   chained tasks link back-to-back; each chain link is preceded by a fixed 1-second
   debounce so writes landing during that window coalesce into the next link's
   single generation read instead of spawning their own chain. The write-generation
   guard releases on every exit of the task it guards (success, error, or panic),
   not only the "nothing loaded" path.

4. **Durable corpus epoch, fail-closed.** A daemon that stays warm across a
   `kkernel reindex` run cannot observe reindex's snapshot deletion or the
   in-process write-generation counter, since the two run as separate processes.
   This is closed by a durable single-row `memory_ann_epoch` table (declared via
   `MemoryPack::SCHEMA_PLAN`, not created ad hoc), bumped twice per reindex pass:
   once before any vector mutation (`begin_reindex_epoch`, fail-closed: the whole
   reindex aborts if the bump fails) and once after the pass commits
   (`invalidate_active_memory_vamana_snapshot`). The recall warm-hit path compares
   the installed entry's `epoch_baseline` against the current durable epoch on a
   debounced (5s) interval and folds a mismatch into the same write-generation
   rebuild machinery.

5. **Content-hash restart validation.** A persisted snapshot's vector
   count/dimension fingerprint alone cannot detect a same-cardinality re-embed
   (`kkernel reindex` overwrites embeddings without touching `notes`, so no
   timestamp signal fires). Restart validation additionally persists a durable
   `CorpusContentHash` (blake3 over every live note's `(subject_id, embedding)`
   bytes, computed during the build's own corpus scan) and re-scans on restart to
   compare byte-for-byte; a mismatch is treated as stale and triggers a full
   rebuild, the same path an absent snapshot takes.

6. **Deletion filtering.** Because a served index may be stale, ANN hits are never
   trusted directly; every hit is resolved through a live-note hydration query
   (`deleted_at IS NULL`, `expires_at` in the future) before being returned. A
   stale ordinal pointing at a since-deleted or since-expired note hydrates to
   nothing and is silently dropped.

A bounded stale-serving window is accepted, matching the documented
indexing-latency contract. Full normative definition, including the two-phase
epoch protocol's crash-safety reasoning and the chained-debounce bound, is in
[ADR-107](docs/adr/ADR-107-memory-ann-lifecycle.md) sections 1-4.

## Test plan

- `cargo clippy -p khive-pack-memory -p kkernel --all-targets -- -D warnings`
- `cargo test -p khive-pack-memory -p kkernel`
- New/updated tests cover: write-generation bump does not evict the installed
  index or snapshot before a replacement exists; a stale-but-installed entry is
  served by `search_loaded` rather than forcing a synchronous rebuild; the
  write-generation guard releases on every exit path (success/error/panic); the
  chained re-enqueue converges under continuous writes and respects the 3-attempt
  bound and 1-second chain debounce; restart content-hash mismatch triggers a
  rebuild on both vector-only and full-corpus re-embeds; the durable epoch's
  begin/completion bump pair and `kkernel reindex`'s fail-closed exit on a failed
  epoch bump.
- `scripts/perf/bench_pipeline_daemon.py` (Pipeline Regression Gate) now polls the
  vector-only leg to convergence after ingest, bounded by a generous deadline,
  before sampling the benchmark's per-leg P@K assertions, since the benchmark is a
  read-after-write consumer of this contract and must not sample the rebuild race
  mid-convergence.

Closes #791

### DIFF
--- a/crates/khive-pack-memory/Cargo.toml
+++ b/crates/khive-pack-memory/Cargo.toml
@@ -18,6 +18,7 @@ khive-retrieval = { version = "0.3.0", path = "../khive-retrieval" }
 khive-score = { version = "0.3.0", path = "../khive-score" }
 khive-brain-core = { version = "0.3.0", path = "../khive-brain-core" }
 khive-vamana = { version = "0.3.0", path = "../khive-vamana" }
+blake3 = { workspace = true }
 inventory = { workspace = true }
 khive-storage = { version = "0.3.0", path = "../khive-storage" }
 async-trait = { workspace = true }
--- a/crates/khive-pack-memory/src/ann.rs
+++ b/crates/khive-pack-memory/src/ann.rs
@@ -51,6 +51,14 @@ pub(crate) struct AnnBridge {
     /// corpus can never clobber an already-installed fresher build, and a
     /// build that snapshotted a fresher corpus always wins.
     pub(crate) generation: u64,
+    /// Durable corpus epoch (`memory_ann_epoch` table) observed at build/load
+    /// time (#812 review REQUEST CHANGES HIGH). Distinct from `generation`,
+    /// which lives only in this process's memory and resets to 0 on restart
+    /// — `epoch_baseline` is compared against a value written to the shared
+    /// SQLite file, so it is the only signal a warm daemon has for an
+    /// out-of-process corpus mutation (`kkernel reindex`) that never goes
+    /// through this daemon's `bump_generation` call at all.
+    pub(crate) epoch_baseline: u64,
 }
 
 /// Shared ANN state: per-`(namespace, model)` indexes with at-most-one-background-build guard.
@@ -85,10 +93,24 @@ pub(crate) struct AnnState {
     /// happened to acquire the per-model lock first even if it had
     /// snapshotted a now-stale corpus.
     generations: Mutex<HashMap<AnnKey, u64>>,
+    /// Debounce state for `maybe_check_durable_epoch` (#812 review REQUEST
+    /// CHANGES HIGH): last time this process actually paid for a
+    /// `memory_ann_epoch` SELECT for a given key, so the warm-hit recall
+    /// path amortizes the cross-process freshness check instead of adding a
+    /// DB round-trip to every single recall.
+    last_epoch_check: std::sync::Mutex<HashMap<AnnKey, std::time::Instant>>,
     /// Counts how many times `search_loaded` returned a warm hit. Test-only;
     /// call `reset_warm_route_count()` between operations to isolate counts.
     #[cfg(test)]
     pub(crate) warm_route_count: AtomicUsize,
+    /// Test-only synchronization point: notified right before each rebuild
+    /// attempt inside `ensure_ann_background`'s tracked task commits to the
+    /// generation floor it captured (#812 review REQUEST CHANGES MEDIUM-2).
+    /// Scoped per-`AnnState` (not a process-global static) so one test's
+    /// notifications can never be consumed by a different test's waiter —
+    /// each test gets its own via `new_shared()`.
+    #[cfg(test)]
+    pub(crate) attempt_floor_notify: tokio::sync::Notify,
 }
 
 pub(crate) type SharedAnn = Arc<AnnState>;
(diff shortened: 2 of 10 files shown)


ohdearquant:
Implementation verified at 141a0266: two-phase durable epoch (begin bump committed before any vector mutation, epoch and vector tables in the same SQLite file), `memory_ann_epoch` declared through `MemoryPack::SCHEMA_PLAN`, bounded rebuild loop (3-attempt cap, 5s epoch-check debounce, 1s chain debounce), and the rebuild-coordination fields fully `cfg(test)` with no production overhead. The guard-drop-before-recheck ordering in `spawn_rebuild_task_inner` is correct.

Two items before merge:

1. **Pipeline Regression Gate is red** (run 29120308442) and the failure looks like real signal, not flake: vector-only P@K is 0.00 for the first six queries after ingest and 1.00 for the rest of the run, with latencies dropping from ~40-50ms to ~22ms at the same point — consistent with the background ANN rebuild converging mid-benchmark while recall serves the stale index, i.e. exactly the eventual-consistency contract this PR ships. The bench queries immediately after ingest and assumes a settled index. Either the bench needs a deterministic settle step between ingest and the query phase (with the gate then measuring settled retrieval quality), or — if the expectation was that the warm-up recall's triggered rebuild completes before the query loop — that's a real convergence-latency issue to chase. Rerunning until green would just sample the race, so please resolve it one way or the other.

2. **The body under-describes the shipped diff.** It covers the original serve-stale fix but none of the largest parts of the final change: the durable epoch protocol, the blake3 corpus content-hash restart validation, the `SCHEMA_PLAN` declaration, the bounded re-enqueue loop, or the `kkernel reindex` changes — and its ADR note still discusses ADR-079 where ADR-107 (#813) now governs. Please rewrite the body to the shipped design before merge, same as #790.

oceanwaves630 (approved):
Approving per the gate trail documented in the review comments: implementation verified, regression-gate settle redesigned to an exact sequential sentinel, and the final head carries an independent bench-delta re-audit.

PR #813, opened 2026-07-10
docs(adr): ADR-107 memory ANN lifecycle eventual-consistency contract

ohdearquant:
## Summary
- Adds ADR-107 defining the normative eventual-consistency contract for the memory pack's ANN cache, as implemented by PR #812 (issue #791).
- Covers: stale bound (one background-rebuild latency behind the caller's own write), rebuild trigger (write-generation bump), cold behavior (one-time spawn_blocking build), restart validation (durable content signal alongside CorpusFingerprint), and deletion filtering (live-note hydration post-filter).
- Does not amend ADR-021 and does not extend ADR-079's scope (knowledge-pack v2 segment format) to the memory pack.

## Test plan
- [ ] `deno fmt docs/` (already applied via pre-commit hook)
- [ ] Review against PR #812's implementation for accuracy

### DIFF
--- a/docs/adr/ADR-107-memory-ann-lifecycle.md
+++ b/docs/adr/ADR-107-memory-ann-lifecycle.md
@@ -33,17 +33,23 @@ it is a standalone contract for what PR #812 actually implemented.
 ### 1. Stale bound
 
 A `memory.recall` may serve results computed from an ANN index that is behind the
-caller's own most recent write to the same model by at most **one background rebuild
-latency** — the wall-clock time of the single rebuild that write's write-generation bump
-triggers. There is no bound on how far behind a _third party's_ concurrent writes the
-served index may be beyond the same one-rebuild window; the bound is per-write, not
-wall-clock.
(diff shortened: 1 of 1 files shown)


ohdearquant:
Read the ADR end to end at fbc74d5e. The contract is precise and matches the implementation on the #812 branch at 141a0266 — I verified the two most load-bearing claims directly in the code: the §2 guard-release ordering (guard dropped before the final freshness recheck, closing the lost-wakeup window) and the §4 begin-bump placement (first statement in the graph pass, fail-closed before any vector mutation). §4's build-time content-hash rationale and §5's hydration filtering are both correctly stated. Content approved. Holding the merge to land together with #812 once its remaining check is resolved, since this document is normative for exactly the behavior that check is exercising — if #812's resolution touches semantics, this file should move in the same sweep.

oceanwaves630 (approved):
Approving the ADR pair: eventual-consistency contract matches the verified implementation merging alongside.

PR #814, opened 2026-07-10
fix(kg): wire content_strategy through entity merge and report content_appended truthfully

### LINKED ISSUE #778: merge(): entity merge silently discards the from-entity's description; content_appended:false is not a loss signal
## Behavior

`merge(into_id, from_id)` on two entities rewires edges, unions tags, and merges properties, but the from-entity's `description` is dropped whenever the into-entity has any description of its own. The response reports `content_appended: false`, which reads as a status flag, not a data-loss warning. No error, no dry-run diff of what is about to be discarded.

## Source

- `crates/khive-runtime/src/curation.rs:1070` — `merged_description = merge_option_string_field(into, from, strategy)`; under the default `PreferInto` policy, `from`'s description survives only when `into`'s is `None`.
- `curation.rs:1295` — entity merge hardcodes `content_appended: false` in the summary.
- Note merge (curation.rs:1487) has a real `ContentMergeStrategy::Append` that concatenates with a separator; entity merge has no append path for descriptions.

## Field incident

Three duplicate document entities were merged into their canonical counterparts. In all three cases the from-node carried a substantially richer description than the surviving node's one-liner. All three returned `content_appended: false`; the richer prose was silently destroyed and was recoverable only because the author still had the text in working context.

## Why this is severe

`merge` is the dedup verb — by construction it runs against nodes that are similar but rarely identical, so description asymmetry is the common case, not the edge case. The caller asked to consolidate and got truncation.

## Requested fix, in order of preference

1. Append the from-entity's description to the kept entity (mirror the note-merge Append strategy), with a separator.
2. If appending is unacceptable as default, refuse the merge when both descriptions are non-empty and non-identical unless an explicit `on_conflict` argument (`keep_into` | `keep_from` | `append`) is supplied.
3. At minimum, `dry_run=true` must surface exactly what would be discarded, and the flag should read as a loss (`content_discarded: true`) when a non-empty description is dropped.

Options 1 and 2 are not exclusive. Option 3 alone is insufficient — it requires the caller to know to ask.

ohdearquant:
## Problem

Under the default `PreferInto` merge policy, `merge(kind="entity", ...)` silently
discarded the `from` entity's description whenever the `into` entity already had a
non-empty description. There was no way to detect this from the response:
`content_appended` was hardcoded to `false` for entity merges regardless of what
actually happened.

The kg pack's merge handler already parsed the `content_strategy` wire parameter
(default `"append"`) for entity merges, but never passed it through to
`merge_entity` — so a caller who explicitly set `content_strategy="append"` on an
entity merge got no different behavior. The note-merge path already had a working
`Append` strategy that concatenated content with a separator and reported
`content_appended` correctly; entity merges had no equivalent.

The dropped description was not permanently destroyed (the `from` row is
tombstoned, not hard-deleted), but it becomes unreachable through the normal verb
surface: `get()` returns `NotFound` without an explicit `include_deleted=true`, and
the row is removed from FTS and vector indexes. For practical purposes this is a
real data-loss bug.

## Fix

- `crates/khive-runtime/src/curation.rs`: added a `content_strategy:
  ContentMergeStrategy` parameter to `merge_entity` and `merge_entity_sql`. Entity
  description merging now has a real `Append` branch mirroring the note-merge
  pattern:
  - both descriptions non-empty → concatenate with `"\n\n---\n\n"`,
    `content_appended = true`
  - one side empty → keep whichever side is non-empty; `content_appended = true`
    only when this preserved content from `from` that would otherwise have been
    lost
  - `from`'s description empty → into's description unchanged, `content_appended =
    false`
  - explicit `PreferInto`/`PreferFrom` opt-out keeps the existing discard behavior,
    with `content_appended = false` (a deliberate choice, not silent loss)
  - `content_appended` in `MergeSummary` is now computed from the branch actually
    taken instead of hardcoded to `false`
- `crates/khive-pack-kg/src/handlers/merge.rs`: the entity branch of the merge
  handler now passes the already-parsed `content_strategy` into `merge_entity`,
  matching the note branch.
- `crates/khive-pack-kg/src/apply_worker/worker.rs`: `apply_merge_entities` (the
  ADR-046 propose/apply changeset interpreter) now passes
  `ContentMergeStrategy::Append` to match the new runtime default.

No new wire surface: `content_strategy` already existed on `MergeParams` and was
already documented for entity merges. This fix makes an already-accepted parameter
actually work.

## Behavior change

Entity merges under the default policy that previously silently dropped `from`'s
description will now produce a concatenated description when both entities have
one. Any caller currently treating entity `content_appended: false` as "nothing to
check" will now sometimes see `true`.

## Breaking change

`KhiveRuntime::merge_entity`'s public signature gained a required positional
`content_strategy: ContentMergeStrategy` parameter (between `strategy` and
`dry_run`). There is no compatibility wrapper — the repo convention is no
backwards-compat shims, and the workspace is pre-1.0 (0.3.x), so this lands as an
explicit, coordinated breaking change rather than adding an `Append`-defaulting
alias method. Every in-repository caller (`khive-pack-kg`'s merge handler and
`apply_worker`) was updated in this PR. Downstream users of `khive-runtime` as a
library need to add the new argument when upgrading. Per semver-minor-for-0.x
policy this ships as a normal `fix(...)` release, not a major bump.

## Codex re-review fixes (2 High, 3 Medium)

A codex adversarial re-review of this PR (`.khive/codex_reviews/codex_review_pr814.md`)
returned REQUEST CHANGES. Fixes for each finding:

1. **High — `content_strategy="prefer_from"` was ignored under the default entity
   policy.** Both explicit content preferences (`PreferInto`/`PreferFrom`) routed
   description selection through the entity-field `strategy` instead of
   `content_strategy`, so `prefer_from` silently kept the into-description whenever
   `strategy` stayed at its default `prefer_into`. Split the match arms so
   `content_strategy` selects the description directly (mirrors `merge_note`'s
   existing, correct pattern). The now-unused `merge_option_string_field` helper
   was removed.
2. **High — breaking public API without a compatibility path.** See "Breaking
   change" above — documented explicitly rather than silently shipped.
3. **Medium — accepted ADRs and audit provenance were stale.** ADR-014 and ADR-039
   described the pre-#778 behavior; amended in a separate paired PR (see below).
   The `EntityMerged` audit event payload now also carries `content_strategy`
   alongside `policy`.
4. **Medium — dry-run was not read-only or an accurate preview.** `EntityMerged`
   event emission is now gated on `!dry_run`. The edge-rewire loop now always runs
   (writes still gated on `!dry_run`) so `edges_rewired` in a dry-run response is a
   real predictive count of the edges that would be rewired, not always zero.
5. **Medium — missing test layers.** Added: a handler-wiring test
   (`merge_handler_wires_explicit_prefer_from_content_strategy`, through
   `khive-pack-kg`'s `merge` verb dispatch) proving `content_strategy` reaches the
   runtime from wire params; an explicit `prefer_from` end-to-end regression test
   at the runtime level
   (`merge_entity_prefer_from_content_strategy_wins_over_default_entity_policy`); a
   dry-run predictive-count test
   (`merge_entity_dry_run_predicts_edges_rewired_without_writing`); and an ADR-046
   apply-worker test exercising the `MergeEntities` proposal changeset path
   (`apply_worker_applies_merge_entities_changeset`).

**Pairs with:** #817 (`docs/adr-014-039-entity-merge-content`) — amends ADR-014 and
ADR-039 for this behavior change, split into its own PR per repo convention
(code and docs land separately).

## Tests

Added to `crates/khive-runtime/src/curation.rs`:

- `merge_entity_append_strategy_concatenates_descriptions`
- `merge_entity_append_strategy_from_empty_is_noop`
- `merge_entity_append_strategy_into_empty_takes_from`
- `merge_entity_prefer_into_strategy_still_discards_explicitly`
- `merge_entity_dry_run_previews_append`
- `merge_entity_prefer_from_content_strategy_wins_over_default_entity_policy`
- `merge_entity_dry_run_predicts_edges_rewired_without_writing`

Added to `crates/khive-pack-kg/tests/integration.rs`:

- `merge_handler_wires_explicit_prefer_from_content_strategy`
- `content_strategy` assertion added to
  `curation_merge_entity_event_payload_has_adr014_fields`

Added to `crates/khive-pack-kg/tests/apply_worker.rs`:

- `apply_worker_applies_merge_entities_changeset`

All existing `merge_entity` call sites (tests and production callers) were updated
for the new parameter.

## Gates (from `crates/`, `CARGO_TARGET_DIR=./target`)

- `cargo test -p khive-runtime -p khive-pack-kg` — pass (1194 tests across both crates' unit + integration binaries, 0 failed, 7 ignored)
- `cargo clippy -p khive-runtime -p khive-pack-kg --all-targets -- -D warnings` — pass

Closes #778

### DIFF
--- a/crates/khive-pack-kg/src/apply_worker/worker.rs
+++ b/crates/khive-pack-kg/src/apply_worker/worker.rs
@@ -3,7 +3,7 @@
 use uuid::Uuid;
 
 use khive_runtime::{
-    curation::{EntityDedupMergePolicy, EntityPatch},
+    curation::{ContentMergeStrategy, EntityDedupMergePolicy, EntityPatch},
     KhiveRuntime, NamespaceToken, RuntimeError, VerbRegistry,
 };
 use khive_storage::types::PageRequest;
@@ -363,6 +363,7 @@ impl ProposalApplyWorker {
                 into_id,
                 from_id,
                 EntityDedupMergePolicy::PreferInto,
+                ContentMergeStrategy::Append,
                 false,
             )
             .await?;
--- a/crates/khive-pack-kg/src/handlers/merge.rs
+++ b/crates/khive-pack-kg/src/handlers/merge.rs
@@ -43,7 +43,7 @@ impl KgPack {
                     )));
                 }
                 self.runtime
-                    .merge_entity(token, into_id, from_id, policy, dry_run)
+                    .merge_entity(token, into_id, from_id, policy, content_strategy, dry_run)
                     .await?
             }
             KindSpec::Note { specific } => {
--- a/crates/khive-pack-kg/tests/apply_worker.rs
+++ b/crates/khive-pack-kg/tests/apply_worker.rs
@@ -196,6 +196,113 @@ async fn apply_worker_applies_add_edge_changeset() {
     assert_eq!(row.status, "applied");
 }
 
+/// ADR-046 apply-worker path test for the codex PR #814 Medium finding: the
+/// five new content_strategy tests only exercised the runtime directly, not
+/// a proposal-driven merge. `apply_merge_entities` hard-codes
+/// `EntityDedupMergePolicy::PreferInto` + `ContentMergeStrategy::Append`
+/// (the safe, lossless default for proposal-driven merges) — verify that
+/// path actually rewires edges and tombstones the source.
+#[tokio::test]
+async fn apply_worker_applies_merge_entities_changeset() {
+    let (rt, tok) = setup();
+    ensure_schema(&rt).await;
+
+    let into = rt
+        .create_entity(&tok, "concept", None, "Into", Some("desc A"), None, vec![])
+        .await
+        .expect("create into");
+    let from = rt
+        .create_entity(&tok, "concept", None, "From", Some("desc B"), None, vec![])
+        .await
+        .expect("create from");
+    let other = rt
+        .create_entity(&tok, "concept", None, "Other", None, None, vec![])
+        .await
+        .expect("create other");
+    rt.link(
+        &tok,
+        other.id,
+        from.id,
+        khive_types::EdgeRelation::Extends,
+        1.0,
+        None,
+    )
+    .await
+    .expect("link other -> from");
+
+    let proposal_id = Uuid::new_v4();
+    let changeset = ProposalChangeset::MergeEntities {
+        into: Id128::from_u128(into.id.as_u128()),
+        from: Id128::from_u128(from.id.as_u128()),
+    };
+
+    seed_proposal_created_event(&rt, &tok, proposal_id, changeset).await;
+    insert_projection_row(&rt, &tok, proposal_id, "approved").await;
+
+    let registry = build_registry(&rt);
+    let worker = ProposalApplyWorker::new(rt.clone());
+    worker
+        .maybe_apply(&tok, proposal_id, &registry, None)
+        .await
+        .expect("maybe_apply must succeed");
+
+    let event_store = rt.events(&tok).expect("event store");
+    let applied_events = event_store
+        .query_events(
+            EventFilter {
+                kinds: vec![EventKind::ProposalApplied],
+                payload_proposal_id: Some(proposal_id),
+                ..Default::default()
+            },
+            PageRequest {
+                offset: 0,
+                limit: 10,
+            },
+        )
+        .await
+        .expect("query events");
+    assert_eq!(
+        applied_events.items.len(),
+        1,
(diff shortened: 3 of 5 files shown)


ohdearquant:
Verified at cbbc0876: content_strategy threads through both writer paths into merge_entity_sql, the Append arm is lossless with correct empty-side handling and truthful content_appended reporting, dry-run no longer emits a merge event, and the predictive edges_rewired count excludes self-loops exactly like the write path. The proposal apply worker passes the lossless Append default explicitly, and the retired policy-based description helper has no remaining references. Merging with its ADR pair #817.

PR #815, opened 2026-07-10
fix(knowledge): generation-check ANN install to close stale-build race

### LINKED ISSUE #770: bug(knowledge): ANN cache install uses entry().or_insert() with no generation check, invalidation can be undone by a slower in-flight build
**Mechanism:** `clear_namespace` in `crates/khive-pack-knowledge/src/knowledge/vamana.rs` clears the in-memory index slot and warming-guard entry for a namespace on corpus mutation, but does not version or cancel builds already in flight. `ensure_ann_for_model`'s install sites (v2 segment load, v1 snapshot load, rebuild fallthrough) all use `ann.indexes.write().await.entry(key).or_insert(bridge)` — this only checks whether the slot is currently empty, not whether an invalidation occurred after the build began scanning the corpus. Freshness checks compare the build's own fingerprint against what was true when it started, never against live state at install time.

**Failure scenario:** A background warm build starts scanning corpus version N. `knowledge.index` writes version N+1 and calls `clear_namespace`, removing the slot. The old build for version N finishes afterward and installs via `or_insert` (slot is empty). Every subsequent search returns from the stale version-N cache until another explicit invalidation or process restart.

**Anchors:** `crates/khive-pack-knowledge/src/knowledge/vamana.rs` (`clear_namespace`, `ensure_ann_for_model` install sites).

**Fix direction:** Maintain a per-namespace/model corpus generation counter. Capture it before a build starts scanning; on install, compare against the current generation and discard the build if it no longer matches. Bump the generation atomically with any corpus-mutating write. Related: the memory pack's cross-process variant of this class is tracked in #752; the knowledge pack fix should follow the same generation-versioned-install shape (see also #566 for the warm-state consolidation refactor these fixes should not fight).

ohdearquant:
## Summary

The knowledge-pack ANN cache install path had no freshness check. All three
install sites in `ensure_ann_for_model` (`crates/khive-pack-knowledge/src/knowledge/vamana.rs`)
used `entry(key).or_insert(bridge)`, which only tests whether the cache slot
is occupied, never whether the candidate build reflects a corpus state at
least as recent as what is already installed.

`clear_namespace` (called from `index_handler.rs` after any corpus-mutating
write) evicts the slot immediately, but sets no fence a slower in-flight
build can observe. A build that started scanning corpus version N could
finish and install after a newer write's `clear_namespace` ran, landing in
the now-empty slot with nothing since then to evict it again. Every
subsequent `search_loaded` would then serve stale results until the next
explicit invalidation or process restart.

## Fix

Adds a per-namespace write-generation counter to `AnnState`:

- `clear_namespace` bumps the counter for the invalidated namespace (keyed
  by namespace rather than the full `AnnKey`, since invalidation is
  namespace-scoped and a build's model may not have a tracked entry yet).
- `ensure_ann_for_model` captures the current generation for its namespace
  before its fast path and before scanning the corpus, and stamps it on the
  resulting `AnnBridge`.
- A new `install_if_fresher` helper replaces all three
  `entry(key).or_insert(bridge)` call sites: it only replaces an existing
  entry when the candidate's generation is greater than or equal to the
  installed entry's, so a build that snapshotted an older corpus can never
  clobber an already-installed fresher build.

This is a local transplant of the generation / install-if-fresher pattern
from the memory pack's #791/#812 fix, adapted to knowledge's
namespace+model-keyed `AnnKey`. It deliberately does not port the
cross-process durable-epoch table, single-flight rebuild chaining, or event
emission from the memory pack — those solve a different (multi-write
convergence, cross-process) problem this issue does not describe. Scope is
limited to `crates/khive-pack-knowledge/src/knowledge/vamana.rs`.

## Tests

Added to `knowledge::vamana::tests`:

- `install_if_fresher_rejects_late_stale_build`
- `install_if_fresher_accepts_forward_progress`
- `install_if_fresher_ties_keep_incumbent`
- `install_if_fresher_installs_into_empty_slot`
- `clear_namespace_bumps_generation_scoped_to_namespace`
- `stale_build_installs_before_invalidation_race_is_rejected_after` (deterministic
  simulation of the reported race: build A captures a stale generation floor,
  an invalidation lands mid-build, a concurrent build B installs a fresher
  result, and build A's late install must not clobber it)

## Gates

Run from `crates/` with `CARGO_TARGET_DIR=./target`:

- `cargo test -p khive-pack-knowledge` — 81 passed
- `cargo clippy -p khive-pack-knowledge --all-targets -- -D warnings` — clean

Closes #770

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/index_handler.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/index_handler.rs
@@ -207,11 +207,18 @@ impl KnowledgeHandlers {
                 cb(total as u64, total as u64);
             }
             let model_name = runtime.default_embedder_name();
+            // Capture the namespace's write-generation floor before scanning the
+            // corpus, mirroring `ensure_ann_for_model`'s `target_generation` capture
+            // (PR #815 review, HIGH) — `install_if_fresher` then fences this direct
+            // rebuild insertion with the same generation check the warm path uses,
+            // instead of the old presence-only `insert_ann_if_absent`.
+            let build_generation = vamana::current_generation(ann, &ns);
             // Build from the shared corpus scan (ORDER BY subject_id) so the persisted
             // v2 content_hash matches the warm-path live_content_hash. Building from
             // atom-iteration order persists a hash the warm path always reads as stale.
             match vamana::load_and_build_from_vector_store(runtime, token, model_name).await {
                 Ok(Some(bridge)) => {
+                    let bridge = bridge.with_generation(build_generation);
                     let n = bridge.num_vectors();
                     ann_count = Some(n);
                     if let Err(e) = vamana::persist_ann_v2(runtime, &ns, model_name, &bridge) {
@@ -219,7 +226,7 @@ impl KnowledgeHandlers {
                         ann_failed = true;
                     }
                     let key = vamana::AnnKey::new(&ns, model_name);
-                    vamana::insert_ann_if_absent(ann, key, bridge).await;
+                    vamana::install_if_fresher(ann, &key, bridge).await;
                     eprintln!("  Vamana ANN built ({n} vectors)");
                 }
                 Ok(None) => {}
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Approving per the documented gate verification: fenced install path (install_if_fresher) confirmed at 1ef035db with the unfenced entry point test-gated.

PR #816, opened 2026-07-10
fix(runtime): normalize compact hex prefixes before LIKE-scanning hyphenated ids

### LINKED ISSUE #773: bug(kg): accepted 9-31 char compact hex prefixes can never match stored hyphenated UUIDs
**Mechanism:** ID-resolution handlers (`resolve_uuid_async`, `resolve_uuid_unfiltered` in `crates/khive-pack-kg/src/handlers/common.rs`, plus the analogous GTD resolver) accept any all-hex string of length 8 or more as a compact UUID prefix and route it to `resolve_prefix_inner` (`crates/khive-runtime/src/operations.rs`), which matches via `id LIKE '<prefix>%'` against the `id` column. That column stores standard hyphenated UUID text. The pattern's 9th character is a hex digit, but the stored value's 9th character is always `-`, so any compact prefix longer than 8 characters can never match — including a full 32-character no-hyphen UUID.

**Failure scenario:** A caller copies a 12-character compact prefix from a normalized (hyphen-stripped) UUID and gets a false `NotFound` even though a matching record exists.

**Anchors:** `crates/khive-pack-kg/src/handlers/common.rs` (the `len() >= 8` all-hex acceptance gate), `crates/khive-runtime/src/operations.rs` (`resolve_prefix_inner` LIKE query), `crates/khive-pack-gtd/src/handlers.rs` (same pattern).

**Fix direction:** Either restrict accepted compact prefixes to exactly 8 characters, or compare against `replace(id, '-', '')` in the predicate so any accepted prefix length matches correctly. Apply consistently across every resolver sharing this acceptance gate (KG, GTD, memory, comm, schedule, brain, session, proposal, knowledge). (Distinct from #749, which addressed cross-table duplicate counting in the same resolver.)

ohdearquant:
## Summary

Compact (hyphen-free) hex UUID prefixes longer than 8 characters could never resolve to a stored record. Every id-prefix resolver shares one acceptance gate (`len() >= 8 && all_hex`) that accepts compact prefixes of any length, but the `LIKE '<prefix>%'` pattern was built from the raw, unnormalized candidate. Stored `id` columns always use the canonical `8-4-4-4-12` hyphenated form, so any accepted prefix past 8 characters placed a hex digit where a hyphen belongs, and the scan could never match — not even a full 32-character hyphen-free UUID.

This affected every by-id verb across every loaded pack that accepts a short hex prefix (kg get/update/delete/merge/resolve/propose/review/withdraw, gtd transition/complete, comm reply/read/thread, schedule cancel, session resume/export, brain auto_feedback, knowledge get/cite, memory remember source_id). All of them silently returned `NotFound`/`InvalidInput`, indistinguishable from "the record genuinely doesn't exist."

## Changes

- Add `hex_prefix_to_uuid_pattern` in `khive-runtime::operations`, which re-inserts hyphens at the canonical boundaries before a compact prefix is used to build a `LIKE` pattern. Prefixes that already contain a hyphen pass through unchanged; behavior for prefixes of 8 characters or fewer is unchanged.
- Apply the helper at the two places that build the `LIKE` pattern from a raw prefix:
  - `resolve_prefix_inner` in `khive-runtime/src/operations.rs` — the shared fan-in point behind `resolve_prefix`/`resolve_prefix_unfiltered`/`resolve_prefix_including_deleted`/`resolve_prefix_unfiltered_including_deleted`, and through them every pack's by-id resolver.
  - `try_get_proposal_payload` in `khive-pack-kg/src/handlers/get.rs` — an independent `LIKE` query over `proposals_open` that is not routed through `resolve_prefix_inner`.
- Export `hex_prefix_to_uuid_pattern` from `khive-runtime`'s public surface so `khive-pack-kg` can share it.

The fix keeps the `id LIKE 'x%'` prefix-scan shape intact rather than wrapping the column in `REPLACE(id, '-', '')`, which would force a full table scan on this hot path.

## Test plan

- [x] `hex_prefix_to_uuid_pattern` unit test: table-driven hyphen placement for input lengths 1 through 31.
- [x] `resolve_prefix_compact_9_to_31_char_matches`: compact prefixes of length 9, 12, 16, 20, 24, 28, 31 all resolve to the same entity.
- [x] `resolve_prefix_compact_full_32_char_matches`: a full 32-char hyphen-stripped UUID resolves.
- [x] `resolve_prefix_boundary_at_hyphen_positions`: prefixes of exactly 8, 12, 16, 20, 24 compact chars (straddling each hyphen boundary) resolve correctly.
- [x] `resolve_prefix_ambiguous_still_detected_after_normalization`: two entities sharing a >8-char compact prefix still trip `AmbiguousPrefix`.
- [x] `hex_prefix_to_uuid_pattern_passes_through_hyphenated_input`: already-hyphenated input is unchanged.
- [x] `get_proposal_by_compact_hex_prefix_over_8_chars`: the independent `get.rs` proposal lookup resolves a >8-char compact prefix.
- [x] `cargo test -p khive-runtime -p khive-pack-kg`
- [x] `cargo clippy -p khive-runtime -p khive-pack-kg --all-targets -- -D warnings`

Closes #773

### DIFF
--- a/crates/khive-pack-git/src/ingest.rs
+++ b/crates/khive-pack-git/src/ingest.rs
@@ -451,9 +451,27 @@ fn row_uuid(row: &khive_storage::types::SqlRow) -> Option<Uuid> {
     }
 }
 
+/// Escape SQLite `LIKE` wildcard characters (`%`, `_`) and the escape
+/// character itself (`\`) so a caller-supplied path is matched literally
+/// under `LIKE ... ESCAPE '\'` rather than as a pattern.
+fn escape_like(input: &str) -> String {
+    let mut out = String::with_capacity(input.len());
+    for c in input.chars() {
+        if matches!(c, '\\' | '%' | '_') {
+            out.push('\\');
+        }
+        out.push(c);
+    }
+    out
+}
+
 /// Find an existing `document` entity whose `properties.source_uri` or `name`
 /// matches `path` (ADR-086 keying convention). Returns `None` when no match —
 /// v0 never creates documents on the ingester's behalf (skip the edge).
+///
+/// An exact `source_uri`/`name` match always wins over the suffix-`LIKE`
+/// fallback: the two are queried separately so a wildcard-broadened
+/// candidate can never shadow an exact one under an unordered `LIMIT 1`.
 async fn find_document_for_path(
     runtime: &KhiveRuntime,
     token: &NamespaceToken,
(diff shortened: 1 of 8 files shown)


oceanwaves630 (approved):
Approving per the documented gate verification: single CASE-ordered SELECT with deterministic tie-break confirmed at e4890c5d.

PR #821, opened 2026-07-10
perf: same-SHA calibration harness for bench gate thresholds

oceanwaves630:
## What

Adds `scripts/perf/bench_calibrate.py`: a stdlib-only runner that executes a chosen bench suite K times at the current commit and emits a per-metric variance profile (samples, mean, std, min, max, CV) as JSON plus a markdown summary.

## Why

The repository has two blocking perf gates (bench-1m recall floor, bench-pipeline P@K) and nine reported-but-unasserted load-harness dimensions, and none of their thresholds were ever placed relative to measured run-to-run noise at an unchanged commit. The recent settle-step rework on the pipeline gate (#812) showed how much a threshold's meaning depends on understanding the null distribution. This harness produces that distribution; it never sets gates itself. Suggested floors in the output are computed at mean minus three standard deviations and are explicitly labeled ADVISORY.

## Design

- `SUITES` registry: each suite declares a build command, default args, timeout, and a metric extractor. Two suites implemented: `pipeline` (drives `bench_pipeline_daemon.py`, parses 84 metrics from stdout including per-query P@K breakdowns) and `load` (drives `bench_load_harness.py --report`, flattens every numeric leaf of the report's dimensions tree).
- Same-SHA enforcement: refuses to run on a dirty worktree (`--allow-dirty` escape for harness development only); records the commit SHA in all outputs.
- Isolated HOME per run with per-run stdout/stderr/argv preserved for post-hoc inspection.
- Output directory `scripts/perf/calibration/` is gitignored, matching the existing ledger convention.

## Evidence

Both suites proved end-to-end locally at this commit: `--suite pipeline --runs 2` (exit 0/0, mean P@K 1.0, p50 ~12.5ms) and `--suite load --runs 2` (exit 0/0, smoke_pass both runs, all nine dimensions captured).

## Gates

`python3 -m py_compile` clean; `cargo fmt --all -- --check` clean (crates untouched).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,8 @@ perf/ci-synthetic-ledger.csv
 # Named {YYYYMMDD}-{sha7}-{dataset}.json; kept out of the repo to avoid large
 # generated files in PRs. The perf/bench-runs/ directory is tracked via .gitkeep.
 perf/bench-runs/*.json
+
+# bench_calibrate.py output: per-invocation variance profiles + per-run
+# stdout/stderr/report artifacts. Sample-dependent and local-machine-noise
+# specific; not committed.
+scripts/perf/calibration/
--- a/scripts/perf/bench_calibrate.py
+++ b/scripts/perf/bench_calibrate.py
@@ -46,6 +46,7 @@ import json
 import os
 import pathlib
 import re
+import signal
 import statistics
 import subprocess
 import sys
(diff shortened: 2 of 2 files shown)


ohdearquant (approved):
Diff read at the two round-3 findings: the no-metrics guard is now reachable (checked both in the load extractor before smoke_pass merges and in the runner before _wall_s merges), and the process-group kill sends SIGKILL to the saved pgid unconditionally after the grace period, covering survivors of a fast-exiting direct child.

One follow-up to file (non-blocking — the output is advisory and the markdown note defers gate placement to a human): `_advisory_floor` uses a fixed `mean - 3*std`, but the bench-program spec's calibration protocol now sizes the tolerance factor from n via noncentral-t (k≈3.98 at n=10, larger at this harness's default n=5). Before any profile's suggested floor is read as a promotion floor, either compute k from n or relabel the field as spread-only so it can't be mistaken for the spec formula.

PR #822, opened 2026-07-10
fix(db,runtime): replace unreachable strong-count checkpoint exit with watch signal

### LINKED ISSUE #774: bug(db): checkpoint task strong-count exit condition is unreachable on the production path, event_store retains a sibling pool Arc
**Mechanism:** `run_checkpoint_task` detects shutdown by checking `Arc::strong_count(&pool) <= 1` each tick (`crates/khive-db/src/checkpoint.rs`). On the production boot path, however, the daemon passes it both the pool and an `event_store` (`crates/khive-runtime/src/daemon.rs:957`), and the production `SqlEventStore` retains its own clone of the same `Arc<ConnectionPool>`:

- `crates/khive-db/src/backend.rs:18-19`: `StorageBackend` holds one `pool: Arc<ConnectionPool>`.
- `crates/khive-db/src/backend.rs:563-565`: `pool_arc()` returns `Arc::clone(&self.pool)`; this is what the checkpoint pool wiring uses (`crates/khive-mcp/src/serve.rs`, `checkpoint_pool_for` into `with_pool`).
- `crates/khive-db/src/backend.rs:246-249`: `events_for_namespace` builds `SqlEventStore::new_scoped(Arc::clone(&self.pool), ...)` from that same field.
- `crates/khive-db/src/stores/event.rs:37`: `SqlEventStore.pool` retains the clone.

So while the checkpoint task's own `event_store` argument is alive, the strong count it observes can never reach `<= 1`: the check undercounts an owner the task itself holds. The existing exit regression passes `None` for the event store (`crates/khive-db/src/checkpoint.rs`, exit test), so it does not exercise the production ownership shape.

**Impact:** On the real SQLite-backed path the checkpoint loop never self-terminates through its intended mechanism. Process exit still kills it, so this is not a leak in practice today, but any graceful-shutdown or drain reasoning built on the strong-count observing termination is wrong, and the pattern is a trap for future periodic tasks modeled on it (a second consumer copying this shape inherits an unreachable exit condition).

**Fix direction:** Replace the strong-count exit with an explicit cancellation signal (watch channel owned by the daemon run scope), or make the threshold account for every retained owner including the task's own `event_store`. Either way, add a production-shaped exit regression using a real `SqlEventStore`. ADR-106 (schedule pack executor) deliberately chose explicit cancellation for the new periodic tick for exactly this reason; this issue tracks fixing the checkpoint task itself.

Found during design review of ADR-106, 2026-07-09; ownership chain verified at source.

ohdearquant:
## Summary
- Fixes #774: `run_checkpoint_task`'s `Arc::strong_count(&pool) <= 1` exit check is unreachable on the production daemon boot path, because the task's `event_store` argument (when `Some`, the ordinary production shape) is a `SqlEventStore` that retains its own `Arc::clone` of the same pool — so the task always observes at least two live owners.
- Replaces the exit condition with an explicit `tokio::sync::watch` channel owned by `run_daemon_with_boot_guard`, signalled as the first action of the shutdown sequence, before `drain()`.
- The checkpoint task is now registered via `track_background_task` (instead of a bare `tokio::spawn`), so `drain()` actually waits for it to observe the shutdown signal and exit.

## Test plan
- [x] Added `checkpoint_task_exits_via_shutdown_signal_with_live_event_store_pool_clone`, which reproduces the exact production ownership shape (a real `SqlEventStore` holding a sibling `Arc<ConnectionPool>` clone), asserts the strong count is above 1 (the old check's blocking condition), then confirms the task exits within 1s via the new watch signal.
- [x] Updated the existing Arc-drop-based tests to use the new signal (the drop-based mechanism no longer exists).
- [x] `cargo test -p khive-db -p khive-runtime` — all green (276 + 14 + 774 + 2 + 63 passed).
- [x] `cargo clippy -p khive-db -p khive-runtime --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all -- --check` — clean.

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -470,8 +470,20 @@ impl CheckpointSeverityState {
 /// Run the WAL checkpoint background task.
 ///
 /// This is a long-running async task that should be spawned with
-/// `tokio::spawn`. It loops until the pool is dropped (the `Arc` count
-/// falls to one, meaning this task holds the last reference).
+/// `tokio::spawn`. It loops until `shutdown_rx` observes a change (or its
+/// sender is dropped), at which point it exits on its next `select!` wakeup.
+/// Callers should hold the paired `tokio::sync::watch::Sender` for the
+/// daemon's run scope and send on it as part of the shutdown sequence.
+///
+/// An earlier version of this task used `Arc::strong_count(&pool) <= 1` as
+/// its exit condition instead of an explicit signal. That check is
+/// unreachable whenever a sibling owner holds its own clone of `pool` for
+/// the task's lifetime — which the production boot path does: `event_store`
+/// (`Option<Arc<dyn EventStore>>`), when `Some`, is a `SqlEventStore` that
+/// retains its own `Arc::clone` of the same pool, so the task always
+/// observed `strong_count == 2` and never exited via that mechanism
+/// (issue #774). The explicit watch channel does not depend on how many
+/// other owners exist.
 ///
 /// The task issues `PRAGMA wal_checkpoint(PASSIVE)` on every tick — ordinary
 /// ticks stay PASSIVE-only and non-blocking; see the module-level doc for the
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Diff read: the watch-channel replacement for the unreachable strong-count exit is the right mechanism, signal-before-drain ordering is correct, and the regression test reproduces the exact multi-owner ownership shape the bug depended on. Approving; auto-merge holds on checks.

PR #823, opened 2026-07-10
fix(request): bound DSL container-nesting depth and input length (#771)

### LINKED ISSUE #771: bug(request): unbounded recursive descent in DSL arg parsing and $prev resolution, deeply nested payload can overflow the stack
**Mechanism:** `parse_arg_value`/`parse_array_arg`/`parse_object_arg` in `crates/khive-request/src/parser/parser_impl.rs` form a mutually recursive descent parser with no nesting-depth tracking. `ArgValue::resolve_all` in `crates/khive-request/src/types.rs` recurses the same way for chain `$prev` substitution. `MAX_OPS` bounds the number of top-level operations only; nothing bounds container nesting depth or the size of the `ops` string before parsing begins (`crates/khive-mcp/src/tools/request.rs` has no length cap; the 8 MiB daemon frame cap is a wire-framing limit, not a parse-time guard).

**Failure scenario:** An `ops` string containing many thousands of nested `[` or `{` (well under the frame cap) drives the recursive parser deep enough to exhaust the thread stack before any verb handler runs, aborting the process handling the request (CWE-674).

**Anchors:** `crates/khive-request/src/parser/parser_impl.rs` (`parse_arg_value`, `parse_array_arg`, `parse_object_arg`), `crates/khive-request/src/types.rs` (`ArgValue::resolve_all`), `crates/khive-mcp/src/tools/request.rs` (`RequestParams::ops`).

**Fix direction:** Reject `ops` above a bounded length before parsing. Track and enforce a strict nesting-depth limit through the parser and through `resolve_all`, preferably by converting container traversal to an iterative (explicit-stack) form rather than native recursion.

ohdearquant:
## Summary

Fixes #771 — CWE-674 unbounded recursive descent in DSL arg parsing and `$prev` resolution.

Three unbounded native recursive-descent sites were reachable from a single untrusted `request(ops="...")` MCP call, with no length cap on `ops` and no depth counter anywhere in `khive-request`:

1. the function-call array/object arg parser (`parse_arg_value`/`parse_array_arg`/`parse_object_arg`)
2. the JSON-form pre-pass into `serde_json::from_str::<Value>`
3. `$prev` nesting inside chain mode via `ArgValue::resolve_all`

A payload of a few thousand nested `[` or `{` characters, well under any plausible length cap, drives the recursion deep enough to overflow the thread stack before any verb handler runs. A stack overflow in Rust is not a catchable `Result`/panic — it aborts the whole OS thread/process.

## Root cause

Confirmed locally pre-fix: `depth=200,000` (a ~2 KB payload) aborts the process with `thread 'main' has overflowed its stack`.

## Fix

Two independent guards in `khive-request`, both required since they close different sites:

- **`MAX_OPS_INPUT_LEN`** (256 KiB): reject oversized `ops` input up front in `parse_request`, before any parsing branch runs.
- **`NESTING_DEPTH_LIMIT`** (64): a depth counter threaded through `Parser`, checked on every `parse_array_arg`/`parse_object_arg` entry, returning `DslError::NestingTooDeep` instead of recursing further. A matching O(n) pre-pass (`check_json_nesting_depth`) bounds the JSON-form path ahead of `serde_json::from_str`, since `serde_json` exposes no depth knob for its untyped `Value` deserializer.

`$prev` nesting (site 3) is covered for free: `ArgValue::Array`/`Object` are only ever constructed inside the same depth-guarded parser functions, so `resolve_all` never sees an out-of-bound tree — no signature change needed (avoids a breaking change to that `pub fn`).

Both new `DslError` variants flow through the existing `dsl_err_to_mcp` → `McpError::invalid_params` path used by every other parse failure.

## Test evidence

- Post-fix: the same `depth=200,000` payload now returns `DslError::InputTooLarge` in microseconds instead of crashing.
- A compact 2 KB payload (well under the length cap) is independently caught by the depth guard alone (`NestingTooDeep` at position 71).
- Added to `crates/khive-request/tests/parser.rs`: at-limit (64) and over-limit (65) cases for array nesting, object nesting, JSON-form nesting, and `$prev`-in-chain nesting; boundary and over-the-cap length tests; a regression test confirming a realistic multi-level nested payload (`propose`-shaped changeset) still parses unchanged.
- `cargo test -p khive-request -p khive-mcp`: 106 parser tests pass (was 92), all 209 existing `khive-mcp` tests pass unmodified.
- `cargo clippy -p khive-request -p khive-mcp --all-targets -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.

## Test plan

- [x] `cargo test -p khive-request -p khive-mcp` — all green
- [x] `cargo clippy -p khive-request -p khive-mcp --all-targets -- -D warnings` — clean
- [x] Manual pre-fix crash repro confirmed, post-fix clean rejection confirmed

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -981,35 +981,26 @@ impl KhiveMcpServer {
                             // `Value` and would otherwise be exposed to the same
                             // unbounded-nesting stack-overflow risk (CWE-674) the
                             // DSL parser guard already closes for syntax input.
-                            if result_exceeds_depth_limit(&result_obj) {
-                                let tool_name = result_obj
-                                    .get("tool")
-                                    .and_then(Value::as_str)
-                                    .unwrap_or_default()
-                                    .to_string();
-                                results.push(json!({
-                                    "ok": false,
-                                    "tool": tool_name,
-                                    "error": {
-                                        "kind": "result_too_deep",
-                                        "message": format!(
-                                            "op result nesting depth exceeds max {}; \
-                                             cannot be used as $prev chain context",
-                                            khive_request::NESTING_DEPTH_LIMIT
-                                        ),
-                                    },
-                                }));
-                                prev_result = None;
-                                aborted_from = Some(i + 1);
-                                continue;
+                            match chain_aggregation_depth_reject(result_obj) {
+                                Err(error_entry) => {
+                                    results.push(error_entry);
+                                    prev_result = None;
+                                    aborted_from = Some(i + 1);
+                                    continue;
+                                }
+                                Ok(result_obj) => {
+                                    // Extract canonical result for $prev (pre-presentation).
+                                    prev_result = result_obj.get("result").cloned();
+                                    // Apply presentation to the result field only,
+                                    // using the effective mode (AlwaysVerbose override honored).
+                                    let presented_obj = apply_presentation_to_result(
+                                        result_obj,
+                                        effective_mode,
+                                        now_unix,
+                                    );
+                                    results.push(presented_obj);
+                                }
                             }
-                            // Extract canonical result for $prev (pre-presentation).
-                            prev_result = result_obj.get("result").cloned();
-                            // Apply presentation to the result field only,
-                            // using the effective mode (AlwaysVerbose override honored).
-                            let presented_obj =
-                                apply_presentation_to_result(result_obj, effective_mode, now_unix);
-                            results.push(presented_obj);
                         }
                         Err((tool, error_payload)) => {
                             results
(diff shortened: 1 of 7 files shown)


oceanwaves630 (approved):
Reviewed at 642d0335. The depth guard is correctly layered: paired enter/decrement in the DSL parser (depth cannot leak across sibling containers on either Ok or Err), a string-aware O(n) pre-pass bounding serde_json's untyped recursive descent, and runtime screening of raw handler results at every dispatch seam (chain, parallel/single, and the retained defense-in-depth aggregation check) before any clone/serialize/present touches the value. The part that makes this complete rather than merely relocated: all rejection paths discard the oversized value through the iterative worklist drop, so the recursive Drop overflow can't fire on the way out either. Boundary tests at exactly NESTING_DEPTH_LIMIT and exactly MAX_OPS_INPUT_LEN, the 1 MiB cap is justified and tested against the documented 1000-item bulk-create contract, and the end-to-end chain test reproduces the accumulated-$prev depth growth the parser guard cannot see. Approved.

Issue #818, opened 2026-07-10
git ingest: unescaped LIKE wildcards in entity-name resolution page can hide exact matches


ohdearquant:
Found during the PR #816 LIKE-pattern inventory sweep (adjacent scope, pre-existing).

Entity-name resolution binds unescaped user input as a LIKE pattern at crates/khive-db/src/stores/entity.rs:451, and the caller at crates/khive-pack-kg/src/handlers/common.rs:215 then searches only a 100-candidate page. Names containing % or _ broaden the pattern, which can push the exact match out of the page and hide it.

Fix direction: escape LIKE wildcard characters in the name before binding (or use an ESCAPE clause), and prefer exact-match candidates before pattern candidates so an exact hit can never lose to page truncation.

PR #824, opened 2026-07-10
fix(runtime): escape LIKE wildcards in Vamana snapshot invalidation (#819)

### LINKED ISSUE #819: snapshot invalidation interpolates namespace into LIKE: underscore-bearing namespace can invalidate another namespace's snapshots
Found during the PR #816 LIKE-pattern inventory sweep (adjacent scope, pre-existing).

Snapshot invalidation interpolates the namespace directly into a LIKE pattern at crates/kkernel/src/reindex.rs:764 and crates/khive-pack-knowledge/src/knowledge/vamana.rs:698. Valid namespaces may contain underscores, and _ is a single-character wildcard in LIKE, so namespace "a_b" also matches "aXb" - one namespace can invalidate another namespace's rebuildable snapshots.

Fix direction: escape % and _ in the namespace before building the pattern (with an ESCAPE clause), or replace the LIKE with an exact prefix comparison that does not treat the namespace as a pattern.

ohdearquant:
## Summary
- Fixes #819: Vamana snapshot invalidation built `LIKE` patterns by interpolating the caller namespace directly (`{namespace}::vamana::%`) with no `ESCAPE` clause. Since namespace names may legally contain `_` (a SQLite `LIKE` single-character wildcard), invalidating namespace `a_b` also matched and deleted namespace `aXb`'s Vamana snapshot rows.
- Escapes `%`, `_`, and `\` in the namespace before building the pattern and adds `ESCAPE '\'` at both call sites: `invalidate_vamana_snapshots` in `crates/kkernel/src/reindex.rs` and `invalidate_snapshot` in `crates/khive-pack-knowledge/src/knowledge/vamana.rs`. Mirrors the existing `escape_like` helper already used for the same bug class in `crates/khive-pack-git/src/ingest.rs`.
- Issue #818 is related but out of scope for this branch.

## Test plan
- [x] Added regression tests in both crates seeding an underscore-bearing namespace (`a_b`) alongside a namespace differing only by that wildcard-position character (`aXb`); asserts the unrelated namespace's snapshot survives invalidation.
- [x] Verified both new tests fail against the pre-fix `LIKE` pattern (both rows deleted) and pass after the fix.
- [x] `cargo test -p kkernel -p khive-pack-knowledge` — all pass (from `crates/`, `CARGO_TARGET_DIR=./target`)
- [x] `cargo clippy -p kkernel -p khive-pack-knowledge --all-targets -- -D warnings` — clean

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/vamana.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/vamana.rs
@@ -818,8 +818,23 @@ async fn live_content_hash(
 /// Called after any vector-corpus mutation to guarantee `ensure_ann_for_model` cannot
 /// load a snapshot that no longer matches the live corpus.  Best-effort: if
 /// the `retrieval_snapshots` table doesn't exist yet, the call is a no-op.
+/// Escape SQLite `LIKE` wildcard characters (`%`, `_`) and the escape
+/// character itself (`\`) so a caller-supplied namespace is matched literally
+/// under `LIKE ... ESCAPE '\'` rather than as a pattern (#819: an
+/// underscore-bearing namespace like `a_b` must not also match `aXb`).
+fn escape_like(input: &str) -> String {
+    let mut out = String::with_capacity(input.len());
+    for c in input.chars() {
+        if matches!(c, '\\' | '%' | '_') {
+            out.push('\\');
+        }
+        out.push(c);
+    }
+    out
+}
+
 pub(crate) async fn invalidate_snapshot(rt: &KhiveRuntime, namespace: &str) {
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Diff read: escape helper covers backslash/percent/underscore, the DELETE now carries ESCAPE, and the regression test pins the exact a_b-vs-aXb crossover. Approving. One hygiene note (no new CI round warranted): the new helper was inserted between invalidate_snapshot's doc comment and its fn, so the 'Called after any vector-corpus mutation' paragraph now attaches to escape_like in rustdoc — fold the reorder into the comment-hygiene sweep.

PR #825, opened 2026-07-10
fix(gtd): push status/assignee/priority filters into SQL for next/tasks

### LINKED ISSUE #772: bug(gtd): next/tasks pre-fetch a fixed 500-row window before filtering, silently hiding older actionable or matching tasks
**Mechanism:** `handle_next` fetches only the newest 500 `task` notes via `list_notes(token, Some("task"), 500, 0)` and applies status/assignee filtering afterward, in memory. `handle_tasks` fetches `offset + limit + 500` notes the same way and applies status/assignee/priority filtering afterward. Neither pushes these predicates into the SQL layer (`crates/khive-pack-gtd/src/handlers.rs`).

**Failure scenario:** If 500 or more task notes have been created or updated more recently than an older actionable task, that task never appears in `gtd.next`'s candidate set regardless of priority — a p0 can be silently hidden by newer inbox/done churn. Similarly, `gtd.tasks(status="done")` can return an empty result even when completed tasks exist, if enough newer non-done tasks fill the fixed window first.

**Anchors:** `crates/khive-pack-gtd/src/handlers.rs` (`handle_next`, `handle_tasks`).

**Fix direction:** Push status/assignee/priority predicates into the `NoteFilter`/SQL query so filtering happens before the row-count cap, and paginate over the filtered result set (keyset pagination or an explicit continuation token) instead of a fixed unfiltered pre-fetch window.

ohdearquant:
## Summary

Fixes #772. `gtd.next` and `gtd.tasks` pre-fetched a fixed unfiltered recency
window of task notes (500 rows, or `offset+limit+500`) and applied
status/assignee/priority filtering afterward in Rust. Any actionable or
matching task older than that window was silently invisible regardless of
priority; `gtd.tasks` additionally always fetched from offset 0, so deep
pages could never reach rows beyond the window.

- Added `FilterOp::In` and `FilterOp::NotInOrMissing` to `khive-storage`
  (compiled in `khive-db`'s `build_note_filter_where`), additive only -
  existing `khive-pack-comm`/`schedule`/`session` call sites are unaffected.
- `handle_next` now pushes `status IN (next, active)` + optional assignee
  into SQL and pages through every matching row, keeping the existing
  dependency-blocker enrichment and priority sort unchanged.
- `handle_tasks` now pushes status/assignee/priority into SQL and uses real
  `PageRequest{limit, offset}` pagination instead of Rust-side skip/take over
  an always-offset-0 fetch.

## Test plan

- [x] New regression tests confirmed to FAIL on pre-fix source, PASS after:
  - `next_finds_actionable_task_older_than_fixed_window`
  - `tasks_finds_done_task_older_than_fixed_window`
  - `tasks_pagination_returns_disjoint_pages` (non-regression correctness check)
- [x] `cargo test -p khive-storage -p khive-db -p khive-pack-gtd` - all pass
- [x] `cargo clippy -p khive-storage -p khive-db -p khive-pack-gtd --all-targets -- -D warnings` - clean
- [x] `cargo fmt --all -- --check` - clean
- [x] `cargo check --workspace` - clean

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/crates/khive-db/src/stores/note.rs
+++ b/crates/khive-db/src/stores/note.rs
@@ -871,6 +871,62 @@ impl NoteStore for SqlNoteStore {
         .await
     }
 
+    async fn query_notes_filtered_bounded(
+        &self,
+        namespace: &str,
+        filter: &NoteFilter,
+        max_rows: u32,
+    ) -> Result<Vec<Note>, StorageError> {
+        for pf in &filter.property_filters {
+            validate_json_path(&pf.json_path)?;
+        }
+        if let Some((path, _)) = &filter.order_by {
+            validate_json_path(path)?;
+        }
+
+        let namespace = namespace.to_string();
+        let filter = filter.clone();
+        let limit_i64 = i64::from(max_rows) + 1;
+
+        self.with_reader("query_notes_filtered_bounded", move |conn| {
+            let (where_sql, mut data_params) = build_note_filter_where(&namespace, &filter)?;
+            data_params.push(Box::new(limit_i64));
+            let limit_idx = data_params.len();
+
+            // Tie-break on `id` in addition to the primary sort key so the
+            // snapshot ordering is fully deterministic even when many rows
+            // share the same `created_at` (or the same custom sort value).
+            let order_clause = match &filter.order_by {
+                Some((path, dir)) => {
+                    let dir_str = match dir {
+                        SortDir::Asc => "ASC",
+                        SortDir::Desc => "DESC",
+                    };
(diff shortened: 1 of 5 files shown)


oceanwaves630 (approved):
Content review: In/NotInOrMissing SQL semantics are right (NULL handling, empty-set edges), the bounded single-statement snapshot for next avoids the COUNT+page race, and the EqOrMissing default-value handling for inbox/p2 preserves legacy rows. Regression coverage is thorough, including the exact-bound boundary and disjoint-pages cases. ADR-019 amendment matches the implementation. Approved.

PR #826, opened 2026-07-10
fix(runtime): guard link()/link_many() against concurrent hard-delete (#769)

### LINKED ISSUE #768: bug(runtime): hard delete purges edges/indexes before the entity row in a separate transaction, crash window loses edges permanently
**Mechanism:** `KhiveRuntime::delete_entity` (and the analogous `delete_note`) performs cascade cleanup (`graph.purge_incident_edges`, `remove_from_indexes`) and the base-row delete as separate, independently-committing async calls, not inside one transaction (`crates/khive-runtime/src/operations.rs`, cascade-then-delete split around the `delete_entity` handler). `KhiveRuntime::update_entity` and `KhiveRuntime::merge_entity` have the same shape (`crates/khive-runtime/src/curation.rs`): the row write commits, then reindexing and the audit-event append run as later, independently-fallible steps. Merge's SQL core does run in one `BEGIN IMMEDIATE` transaction, but its reindex and audit append execute after that transaction commits.

**Failure scenario:** A process crash, disk-full error, or SQLite write failure between `purge_incident_edges`/`remove_from_indexes` and `delete_entity` leaves the entity row live with all its incident edges permanently gone (the reverse of the intended cascade guarantee: no dangling edges, but also no edges at all). For update/merge, a crash after the row commits but before reindex/event-append leaves a durably-changed row with a stale FTS/vector index and a missing audit trail, while the caller receives an error for a mutation that in fact persisted.

**Anchors:** `crates/khive-runtime/src/operations.rs` (`delete_entity`), `crates/khive-runtime/src/curation.rs` (`update_entity`, `merge_entity`).

**Fix direction:** Move cascade edge-purge, SQLite-backed index cleanup, base-row mutation, and the audit-event append into one `BEGIN IMMEDIATE` transaction on the single-writer path. Where non-transactional work is unavoidable (embedding computation), treat it as a durable post-commit job with retry/reconciliation rather than a step whose failure can leave the primary mutation half-applied or misreported.

### LINKED ISSUE #769: bug(runtime): canonical link()/link_many() validate endpoints then insert separately, concurrent hard-delete creates dangling edges
**Mechanism:** `KhiveRuntime::link` calls `validate_edge_relation_endpoints` (an async existence check), then constructs and inserts the edge via `graph.upsert_edge`, which issues `edge_upsert_statement` — an unconditional `INSERT ... ON CONFLICT` with no re-check of endpoint existence. `link_many`/`build_edge` follow the same shape. `graph_edges` carries no foreign key constraints (`crates/khive-db/sql/graph-ddl.sql`), so nothing at the storage layer rejects an edge to a nonexistent record.

A commit-time guarded variant already exists: `edge_insert_guarded_by_endpoints_statement` in `crates/khive-db/src/stores/graph.rs` builds exactly `INSERT ... SELECT ... WHERE EXISTS(...) AND EXISTS(...)` against both `entities` and `notes` — but it is wired only into the atomic batch-apply path. Its doc comment reasons the canonical path does not need it because no other op's write interleaves between its own validation and write; that covers intra-batch ordering, not concurrent requests from different callers.

**Failure scenario:** Concurrent MCP calls: request A validates endpoint X (exists), yields; request B hard-deletes X (cascades and commits); request A resumes and inserts the edge. The edge is now durably dangling — no FK catches it and nothing re-validates at insert time. Reachable through concurrent MCP calls or separate WAL-connected processes.

**Anchors:** `crates/khive-runtime/src/operations.rs` (`link`, `link_many`/`build_edge`), `crates/khive-db/src/stores/graph.rs` (`edge_upsert_statement` vs `edge_insert_guarded_by_endpoints_statement`), `crates/khive-db/sql/graph-ddl.sql`.

**Fix direction:** Route canonical `link`/`link_many` through `edge_insert_guarded_by_endpoints_statement` (or an equivalent commit-time `WHERE EXISTS` guard) inside the writer transaction, same as the atomic path. Zero affected rows should surface as `NotFound` rather than a silent no-op success.

ohdearquant:
## Summary

Fixes #769: canonical `link()`/`link_many()` validated edge endpoints with a
separate async read, then wrote the edge with an unconditional
`INSERT ... ON CONFLICT`. A concurrent hard-delete landing between the
validation read and the write could leave a durably dangling edge, since
`graph_edges` carries no foreign key.

Root cause: `KhiveRuntime::link`/`link_many` called
`validate_edge_relation_endpoints` (an async read), then
`graph.upsert_edge`/`upsert_edges` (a separate, unconditional write). The
atomic `link` path already had a commit-time `WHERE EXISTS` guard
(`edge_insert_guarded_by_endpoints_statement`), but it was wired only into
the atomic-batch path, not the canonical one.

## Fix

- Added `GraphStore::upsert_edge_guarded`/`upsert_edges_guarded`, reusing the
  existing guarded `INSERT ... SELECT ... WHERE EXISTS(...)` builder for the
  single-edge path, and a pre-check-then-write batch loop (inside one
  `BEGIN IMMEDIATE`) for the batch path so a missing endpoint aborts the
  whole batch with zero rows written.
- Wired `KhiveRuntime::link`/`link_many` to the guarded calls. Zero affected
  rows now surfaces as `RuntimeError::NotFound` instead of a silent no-op
  success.
- Broadened the guard's existence check from entities/notes only to
  entities, notes, events, and edges, matching
  `substrate_exists_by_id`'s actual definition of "exists" (the `annotates`
  relation may target any substrate). The narrower check was initially
  written and caught real regressions in existing `annotates`-to-event and
  annotates-to-edge tests before being widened; those tests pass unchanged
  now.
- `link_with_target_backend` (the cross-backend coordinator path) has the
  same structural shape but is out of scope here pending a
  `SubstrateCoordinator`-owner review noted in the prep brief; flagged as a
  follow-up, not silently left unaddressed.
- Round-2 codex REJECT fixes (commit `e7459661`): hard-delete's incident-edge
  purge and the endpoint row delete now run in ONE transaction (was two
  independently-committing storage calls), the `GraphStore` trait addition
  got `Unsupported` defaults so external implementors keep compiling, and
  the guarded-write failure type now names the exact missing endpoint(s)
  (and batch index) instead of a generic message.
- Round-2 codex Medium fixes (this update): the guarded-write failure
  outcome is now produced entirely inside the storage transaction itself
  (`GuardedWriteOutcome`/`GuardedBatchOutcome`) instead of being
  reconstructed by the runtime with a post-hoc read after the write was
  already refused; and the concurrency regression suite now includes
  file-backed tests with explicit `.await`-ordering for both
  `write_queue_enabled` states, not just in-memory `tokio::join!` races.

Refs #768, cascade subcase only. `e7459661`'s commit message says "Fixes
#768" — that is not accurate as a full close. That commit implements only
the hard-delete's edge-purge cascade running atomically with the row
delete. #768 also covers SQLite-backed index cleanup, the audit-event
append, and `update_entity`/`merge_entity` consistency (both currently
commit their row write, then reindex/append as separate, independently
fallible steps) — none of which this PR touches. #768 should stay open
until those remain. See the tracking comment on #768 for the itemized
breakdown.

## Test plan

- [x] `cargo test -p khive-db` (new guarded-upsert unit tests, single and
      batch, endpoint-missing and endpoint-live cases)
- [x] `cargo test -p khive-runtime` (new deterministic write-time-guard
      regression tests for `link` and `link_many`; file-backed
      write-queue-on/off ordering tests; all pre-existing `link_*`/
      `annotates_*`/hard-delete tests pass unchanged)
- [x] `cargo clippy -p khive-storage -p khive-db -p khive-runtime --all-targets -- -D warnings`

### DIFF
--- a/crates/khive-db/src/stores/graph.rs
+++ b/crates/khive-db/src/stores/graph.rs
@@ -52,6 +52,25 @@ const EDGE_NATURAL_KEY_CONFLICT_SET: &str = "weight = excluded.weight, \
      metadata = excluded.metadata, \
      target_backend = excluded.target_backend";
 
+/// A `WHERE`-clause fragment asserting the id bound to `id_param` (an SQL
+/// placeholder like `?3`) resolves to a live edge endpoint — an
+/// undeleted entity or note, an event (append-only, no `deleted_at`), or an
+/// undeleted edge (the `annotates` relation's target may be any substrate,
+/// including another edge; ADR-002/ADR-055). Shared by
+/// [`edge_insert_guarded_by_endpoints_statement`] and
+/// [`edge_endpoints_exist`] (#769) so the two "does this endpoint still
+/// exist" probes — the guarded single-row insert and the guarded batch
+/// pre-check — can never drift on which substrates count as valid
+/// endpoints.
+fn endpoint_exists_clause(id_param: &str) -> String {
+    format!(
+        "EXISTS (SELECT 1 FROM entities WHERE id = {id_param} AND deleted_at IS NULL) \
+         OR EXISTS (SELECT 1 FROM notes WHERE id = {id_param} AND deleted_at IS NULL) \
+         OR EXISTS (SELECT 1 FROM events WHERE id = {id_param}) \
+         OR EXISTS (SELECT 1 FROM graph_edges WHERE id = {id_param} AND deleted_at IS NULL)"
+    )
+}
+
 /// The exact natural-key-upserting `INSERT ... ON CONFLICT` this store's
 /// `upsert_edge` issues. Canonicalizes symmetric-relation endpoints first,
 /// matching `upsert_edge`'s own call to `canonical_edge_endpoints`.
@@ -132,16 +151,15 @@ pub fn edge_insert_guarded_by_endpoints_statement(
     now: i64,
     metadata: Option<&str>,
 ) -> SqlStatement {
+    let src_exists = endpoint_exists_clause("?3");
+    let tgt_exists = endpoint_exists_clause("?4");
     SqlStatement {
         sql: format!(
             "INSERT INTO graph_edges \
               (namespace, id, source_id, target_id, relation, weight, \
                created_at, updated_at, metadata) \
               SELECT ?1, ?2, ?3, ?4, ?5, ?6, ?7, ?7, ?8 \
-              WHERE (EXISTS (SELECT 1 FROM entities WHERE id = ?3 AND deleted_at IS NULL) \
-                     OR EXISTS (SELECT 1 FROM notes WHERE id = ?3 AND deleted_at IS NULL)) \
-                AND (EXISTS (SELECT 1 FROM entities WHERE id = ?4 AND deleted_at IS NULL) \
-                     OR EXISTS (SELECT 1 FROM notes WHERE id = ?4 AND deleted_at IS NULL)) \
+              WHERE ({src_exists}) AND ({tgt_exists}) \
               ON CONFLICT(namespace, source_id, target_id, relation) DO UPDATE SET \
                   {EDGE_NATURAL_KEY_CONFLICT_SET}"
         ),
@@ -892,13 +910,11 @@ fn edge_endpoints_exist(
     source_id: Uuid,
     target_id: Uuid,
 ) -> Result<bool, rusqlite::Error> {
+    let src_exists = endpoint_exists_clause("?1");
+    let tgt_exists = endpoint_exists_clause("?2");
+    let sql = format!("SELECT ({src_exists}) AND ({tgt_exists})");
     conn.query_row(
-        "SELECT \
-            (EXISTS (SELECT 1 FROM entities WHERE id = ?1 AND deleted_at IS NULL) \
-             OR EXISTS (SELECT 1 FROM notes WHERE id = ?1 AND deleted_at IS NULL)) \
-         AND \
-            (EXISTS (SELECT 1 FROM entities WHERE id = ?2 AND deleted_at IS NULL) \
-             OR EXISTS (SELECT 1 FROM notes WHERE id = ?2 AND deleted_at IS NULL))",
+        &sql,
         rusqlite::params![source_id.to_string(), target_id.to_string()],
         |row| row.get::<_, bool>(0),
     )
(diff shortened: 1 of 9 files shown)


oceanwaves630 (approved):
Content-level review of the full diff at 1b4c0e6e.

Verified:
- `endpoint_exists_clause` is the single source for both the guarded INSERT's WHERE and the standalone probe, so the two existence checks cannot drift on which substrates count. Widening the clause to events and edges is correct, not a loosening: `annotates` legitimately targets any substrate, and relation typing is enforced at prepare-time validation — this guard is existence-only.
- The refused-insert-to-probe seam runs on one connection with no await between, and the flag-off singleton path now wraps both in one BEGIN IMMEDIATE with rollback on every error path. The seam-barrier test is the right construction: a cfg(test) rendezvous keyed on the exact (source, target) pair, forcing the racer to land inside the seam window instead of sleep-guessing — deterministic on both the pre-fix (racer lands, probe lies) and fixed (racer blocks behind the open transaction) shapes.
- Batch guard pre-checks every entry before any INSERT inside the same write-locked transaction, so refusal leaves nothing to roll back and the all-or-nothing contract matches create_many's. Refusal provenance (entry index + which endpoint) comes from the in-transaction pre-check, never a post-hoc re-read.
- `atomic_hard_delete_with_edge_purge` closes the delete-vs-guarded-write window from the other side: row delete + incident-edge purge in one atomic unit, with AffectedRowGuard::exactly(1) mapping a lost delete race to Ok(false), preserving the old bool contract. The interleaving argument in its doc comment is sound given every write funnels through the single writer.
- Trait defaults return Unsupported rather than silently falling back to the unguarded upsert — correct fail-closed choice.
- The four file-backed tests force both orderings on both write-queue configs with explicit await sequencing, closing the coverage gap the in-memory join-race tests couldn't (scheduler may serialize them).

One minor note, non-blocking: after a refused batch, `link_many`'s `affected != len` fallback branch is unreachable in practice (affected is only ever len or the refusal path returns first) — fine as defense, just noting it's not a live path.

macOS CI failure at this head is `recovery_tests::persistent_corruption_is_bounded_and_never_reports_false_success` in khive-pack-git — a crate this PR does not touch; same macOS flake family as the refetch_clone instances. Rerunning for green; approving on content.

Issue #780, opened 2026-07-09
comm.probe cursor can permanently skip messages: cursor is MAX(created_at) but created_at is stamped outside the writer critical section


ohdearquant:
## Behavior

`comm.probe` advances its cursor on `COALESCE(MAX(created_at), 0)` (crates/khive-pack-comm/src/handlers.rs:1375). `created_at` is assigned by application code via `Utc::now().timestamp_micros()` while the note is being built — BEFORE the write acquires the global writer mutex (crates/khive-db/src/pool.rs:352).

## The hazard

Writer A stamps T1, loses the writer-lock race; writer B stamps T2 > T1 and commits first. A prober that polls now sees B and advances its cursor to T2. A then commits its row carrying created_at = T1 < T2. Every subsequent `created_at > cursor` query excludes A's row forever — the message is not delayed, it is permanently invisible to the probe.

The single global writer mutex serializes COMMITS, but positions are stamped outside the critical section, so commit serialization does not imply position ordering. This is the exact anomaly the transactional-outbox pattern prevents, and why Kafka assigns offsets at the broker on append rather than trusting producer timestamps.

## Severity, honestly bounded

No observed message loss to date: fleet write concurrency is low, and the canonical monitor script independently re-alerts on anything unread past a staleness threshold, so a cursor-dropped message surfaces late rather than never. But that mitigation is a property of one consumer's belt-and-braces, not of the substrate — any programmatic consumer trusting `cursor_us` alone (which is what the verb advertises) loses messages under concurrency. That is a product-surface correctness bug.

## Same shape elsewhere (audit before generalizing)

- `gtd_lifecycle_audit` (khive-pack-gtd/src/vocab.rs:77) has the identical position hazard: `write_audit_record` (handlers.rs:106) stamps `Utc::now()` before acquiring the writer — AND is additionally a non-transactional dual write ("audit write failed (non-fatal)" — a transition can commit while its audit row is lost).
- gtd `updated_at` at handlers.rs:630/:697 is stamped in the decision phase, before the write.

## Requested fix

The probe cursor must be a DB-assigned monotonic value produced inside the same critical section as the row it orders: rowid, or an outbox-style sequence column populated by the writer. Never `created_at`/`updated_at`/any app clock read before the writer lock. Cursor semantics change from "microsecond wall clock" to "opaque monotonic token" — the probe contract should present it as opaque so the representation can change.

[...]

PR #827, opened 2026-07-10
fix(pack-comm): make comm.probe cursor commit-order safe (#780)

ohdearquant:
## Summary

- `comm.probe` advanced its cursor on `COALESCE(MAX(created_at), 0)`, but `created_at` is an application-clock read taken before a note's write acquires the writer critical section. Two concurrent writers can commit out of stamp order, so the cursor could advance past a row that commits *after* it but was stamped with an earlier `created_at` — that row becomes permanently invisible to every future probe.
- Fixed by keying `PROBE_SQL`'s `cursor_us` and the `since_us` filter on SQLite `rowid` instead of `created_at`. `rowid` is assigned by SQLite exactly once, inside the one active writer transaction, so it is monotonic with commit order by construction. Wire field names (`cursor_us`, `since_us`) are unchanged; the value is now documented as an opaque monotonic token, not a microsecond timestamp. Display fields (`created_at_us`, ordering) are untouched.
- **Paired fix (same consistency story, mandatory per the issue):** `comm.read` mutated the read flag via get-note + `upsert_note` (`INSERT OR REPLACE`), which on a primary-key conflict is a SQLite DELETE+INSERT that silently reassigns the row's `rowid`. With a `rowid`-keyed cursor, that churn could bump an already-probed message's `rowid` past the current cursor and resurrect it as "new" on a later poll. Added `NoteStore::update_note_properties`, a narrow `properties`/`updated_at` `UPDATE` that never touches `rowid`, and switched `handle_read` to use it.

## Root cause

`created_at` (khive-storage/src/note.rs `Note::new`) is stamped by the caller before any lock is acquired; the writer mutex (khive-db/src/pool.rs) only serializes *commit* order, not *stamp* order. A probe keyed on `created_at` conflates the two, so the newest-stamped-but-not-yet-committed row can "poison" the cursor against an older-stamped row that commits later.

## Test plan

- [x] Added `probe_survives_out_of_order_commit_vs_created_at` (khive-pack-comm/tests/probe.rs): plants a later-`created_at` message that commits first, then an earlier-`created_at` message that commits second; asserts the second is still visible via the first probe's cursor. Verified this fails against the old `created_at`-keyed SQL.
- [x] Added `probe_read_does_not_resurrect_message_via_rowid_churn`: probes past a message, marks it read via `comm.read`, probes again with the same cursor, asserts it does not reappear. Verified this fails when the rowid-cursor fix lands *without* the `handle_read` fix (confirming both halves are required together).
- [x] Updated `probe_cursor_advances_and_filters_since_us` and `probe_ignores_messages_addressed_to_other_actors`, which asserted `cursor_us` against literal `created_at` values that no longer hold once the cursor is `rowid`-based.
- [x] `cargo test -p khive-pack-comm -p khive-db -p khive-storage -p khive-runtime` — all green (275+14+50+130+16+774 passed, 0 failed)
- [x] `cargo clippy -p khive-pack-comm -p khive-db -p khive-storage -p khive-runtime --all-targets -- -D warnings` — clean
- [x] `cargo check --workspace` — clean

### DIFF
--- a/crates/khive-db/sql/007-notes-seq.sql
+++ b/crates/khive-db/sql/007-notes-seq.sql
@@ -25,3 +25,16 @@ CREATE TABLE IF NOT EXISTS notes_seq (
 );
 
 CREATE INDEX IF NOT EXISTS idx_notes_seq_note_id ON notes_seq(note_id);
+
+-- Backfill: every note that already existed on a pre-V7 (V6 and earlier)
+-- database has no `notes_seq` row yet. `comm.probe`'s query INNER JOINs
+-- `notes` to `notes_seq`, so without this backfill every pre-existing
+-- inbound message would silently vanish from `new_messages`, `cursor_us`,
+-- and `stale_unread_count` the moment a populated database is upgraded to
+-- V7 -- the exact bug this migration exists to fix, reintroduced at the
+-- boundary. Ordered by `(created_at, id)` so the backfilled sequence
+-- preserves the same visible ordering `comm.probe` already presents to
+-- callers. This migration runs exactly once per database (tracked by
+-- `_schema_migrations`), so the full-table scan here is a one-time cost.
+INSERT OR IGNORE INTO notes_seq (note_id)
+SELECT id FROM notes ORDER BY created_at ASC, id ASC;
--- a/crates/khive-db/sql/007-notes-seq.sql
+++ b/crates/khive-db/sql/007-notes-seq.sql
@@ -1,0 +1,27 @@
+-- V7: durable, non-reusing sequence for `notes` (khive #827).
+--
+-- `notes` has a TEXT PRIMARY KEY (`id`), so SQLite assigns it an *implicit*
+-- rowid. Two things make an implicit rowid unsafe as a durable cursor key:
+--   1. VACUUM may renumber implicit rowids (khive exposes `memory.vacuum`).
+--   2. SQLite reuses the highest rowid after that row is deleted (khive
+--      exposes a public hard delete of notes), so a later insert can land on
+--      a rowid a caller's cursor has already passed, permanently excluding
+--      that message from `comm.probe`.
+--
+-- `notes_seq` fixes both: its primary key is an explicit `INTEGER PRIMARY
+-- KEY AUTOINCREMENT` column. An explicit integer primary key is never
+-- renumbered by VACUUM (only implicit-rowid tables are), and AUTOINCREMENT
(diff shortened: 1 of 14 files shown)


oceanwaves630 (approved):
Content-level review of the full diff at f5805a8e.

Verified:
- The durable cursor closes both implicit-rowid hazards: VACUUM renumbering and highest-rowid reuse after hard delete, each pinned by its own test. `notes_seq.seq` is assigned by `assign_note_seq` inside the same transaction as the note insert on every insert path — `upsert_note` and `try_insert_note` via the new `with_writer_tx`, `batch_upsert_notes` inside the batch transaction — so a note can never commit without its sequence row.
- The upgrade-lineage coverage is complete: V7 backfill for V6 databases, V8 anti-join repair for the partial-ledger state left by the original backfill-less V7 (the round-3 finding — the test rebuilds that exact ledger byte-for-byte including a stale sqlite_sequence high-water mark), and the lazy once-per-backend repair for databases that never route through run_migrations. The once-per-backend gate is race-free as written (load-run-store all under the held writer guard) and observable via the run-count accessor rather than timing.
- The high-water-mark reset for pre-upgrade timestamp cursors is sound in both directions: issued cursors are always <= the HWM (MAX(seq) <= HWM, and the regression clamp floors at effective_since which is itself <= HWM), so no legitimate cursor can ever trip the reset; and the round-trip test pins that a legitimately high sequence value survives, which a fixed cutoff would have broken.
- The cursor-never-regresses clamp handles the hard-delete-of-highest-seq case without complicating the indexed query.
- `with_writer_tx` also fixes a real pre-existing bug beyond this feature: the old flag-off `upsert_notes` only rolled back on a failed COMMIT, so a mid-batch error escaped with BEGIN IMMEDIATE still open on the shared connection — the new test proves the batch rolls back fully AND the connection stays usable.
- `update_note_properties` (real UPDATE, not INSERT OR REPLACE) removes the read-path row churn; correctly documented as defensive now that the cursor no longer depends on rowid stability.

One non-blocking note: `handle_read` ignores `update_note_properties`' bool, so a message hard-deleted between the get and the update still reports read=true to the caller. Harmless (the message is gone), just asymmetric with the earlier not-found error.

macOS CI failure at this head is `recovery_tests::public_verb_partial_side_effects_survive_commit_snapshot_recovery` in khive-pack-git — untouched by this PR, same macOS flake family as the refetch_clone instances. Rerunning for green; approving on content.

PR #828, opened 2026-07-11
fix(tests): eliminate three racy tests behind main CI flakes

ohdearquant:
Fixes three distinct test races that have been flipping main CI red intermittently.

1. khive-runtime daemon shutdown_cleanup_skips_when_socket_was_rebound_by_a_replacement: inode recycling on some filesystems could make the original and replacement socket identities collide. The test now pins a synthetic mismatching SocketIdentity, which is all the negative assertion needs.

2. khive-db checkpoint busy_writer_skips_both_passive_and_truncate: six sibling tests mutate the process-wide checkpoint gauges without the serial(checkpoint_skip_metrics) tag the file requires, so they could clobber the gauges mid-assertion. All six are now tagged.

3. khive-pack-git recovery_tests public_verb_partial_side_effects_survive_commit_snapshot_recovery: the namespace-keyed one-shot fault injection could be consumed by an unrelated concurrent test. The thread-local arm_vector_fail_after primitive is now wired into the single-model note path and the test uses it.

Evidence: each fix passed 50/50 consecutive standalone runs; full per-crate suites green (khive-runtime 774, khive-db 247, khive-pack-git 70); fmt and clippy clean.

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -1562,7 +1562,7 @@ mod tests {
     /// at/above `truncate_high_water_pages` and no prior attempt has run, the
     /// escalation fires and stamps `last_attempt`.
     #[test]
-    #[serial(tx_registry)]
+    #[serial(tx_registry, checkpoint_skip_metrics)]
     fn truncate_attempts_when_high_water_crossed_with_no_prior_attempt() {
         let dir = tempfile::tempdir().unwrap();
         let path = dir.path().join("truncate_trigger.db");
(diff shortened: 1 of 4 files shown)


oceanwaves630 (approved):
All three fixes address root cause rather than papering over timing: gauge tests serialized into the existing group, fault injection moved from process-global namespace matching to a thread-local counter that concurrent tests cannot win, and the socket-identity test restructured so inode recycling cannot produce false equality (with a loud assert if socket_identity degenerates). Test-only plus cfg-gated injection plumbing. Approved.

PR #830, opened 2026-07-11
perf: n-sized tolerance factor for advisory floor + bench-1m calibration suite

### LINKED ISSUE #829: perf: bench_calibrate advisory floor must size k from n (noncentral-t), not hardcode mean-3*std
The calibration harness (scripts/perf/bench_calibrate.py, PR #821) prints an advisory floor computed as mean - 3*std. The signed benchmarking program spec retires that exact formula: the tolerance factor k must be sized from the run count n via the one-sided normal tolerance interval (noncentral-t), giving k ~= 3.98 at n=10 and larger at the harness default n=5. A hardcoded k=3 fires on pure noise at roughly 22 to 39 percent rates depending on distribution.

Required before any printed floor circulates as a promotion candidate:
- compute k from n (noncentral-t one-sided tolerance factor at the spec's coverage/confidence), or
- relabel the field as spread-only so nobody copies it into a gate.

Scope: small change in _advisory_floor plus a unit check that k(n=10) matches the spec value. Fold into wave-1 before the first calibration profiles are shared.

ohdearquant:
## Summary

Closes #829.

- **Task 1 (#829)**: `_advisory_floor` in `scripts/perf/bench_calibrate.py` hardcoded `mean - 3*std` regardless of `--runs`. Replaced the fixed `3.0` multiplier with `k(n)`, the one-sided normal tolerance factor (Wald-Wolfowitz approximation), sized from the run count `n` via a stdlib-only inverse-normal (Acklam's rational approximation). Pinned at coverage=0.99 / confidence=0.95 to match the calibration spec's worked example (`k(10) ~= 3.98`). Floors are now labeled `advisory-calibration-only` and `k`/`n` are printed alongside the floor in stdout, the markdown table, and the JSON payload. Below `n=3` the approximation is out of its valid domain (denominator `a <= 0`); the harness now reports std/CV without a floor instead of crashing, so `--runs 2` still works.
- Added a `ToleranceFactorSelfCheck` unittest (`python -m unittest scripts.perf.bench_calibrate`) asserting `k(10)` is within `0.05` of `3.98` and that `k` is monotonically decreasing in `n`.
- **Task 2**: added a `bench-1m` suite entry wrapping the exact hermetic per-PR gate command from `.github/workflows/bench-1m.yml`'s `ann-ci-gate` job (`bash scripts/bench_1m.sh --ci-synthetic`). Extracts assertion checks, per-N row metrics, and growth-exponent fits from the schema-versioned result JSON with fail-closed cardinality checks (`SchemaError` on drift), matching the style of the existing `pipeline`/`load` extractors. The manual/self-hosted-only `bench-1m-sift` job (needs a ~500MB SIFT-1M dataset) is intentionally not wrapped since it cannot run standalone.

Referenced spec: `.khive/workspaces/20260710/bench-program/SPEC-draft.md` (Principles and Calibration protocol sections) — not present in this worktree; parameters were reverse-derived from the pinned `k(10) ~= 3.98` acceptance target plus the given Wald-Wolfowitz formula, which resolves to coverage=99%/confidence=95% (coverage=99.9% under the same formula gives `k(10) ~= 5.16`, inconsistent with the pinned target).

## Test plan

- [x] `python3 -m py_compile scripts/perf/bench_calibrate.py`
- [x] `python3 -m unittest scripts.perf.bench_calibrate -v` — 3/3 pass
- [ ] `python3 scripts/perf/bench_calibrate.py --suite pipeline --runs 2 --allow-dirty` smoke run (kkernel-bench build permitting)
- [ ] `python3 scripts/perf/bench_calibrate.py --suite bench-1m --runs 2 --allow-dirty` smoke run (vec_bench release build permitting)

### DIFF
--- a/scripts/perf/bench_calibrate.py
+++ b/scripts/perf/bench_calibrate.py
@@ -61,11 +61,13 @@ import signal
 import statistics
 import subprocess
 import sys
+import tempfile
 import time
 import unittest
 
 REPO_ROOT = pathlib.Path(__file__).parent.parent.parent
 PERF_SCRIPTS_DIR = pathlib.Path(__file__).parent
+_BENCH1M_FIXTURE_PATH = PERF_SCRIPTS_DIR / "testdata" / "bench1m_result_fixture.json"
 PIPELINE_SCRIPT = PERF_SCRIPTS_DIR / "bench_pipeline_daemon.py"
 LOAD_SCRIPT = PERF_SCRIPTS_DIR / "bench_load_harness.py"
 
@@ -313,8 +315,9 @@ def _load_extract(run_dir: pathlib.Path, proc: subprocess.CompletedProcess) -> d
 # JSON can never be read by, or leak into, another run.
 
 _BENCH1M_DATASET = "SIFT-CI-synthetic"
+_BENCH1M_EXPECTED_SCHEMA_VERSION = "1.0"  # crates/khive-vamana/examples/vec_bench.rs schema_version
 _BENCH1M_EXPECTED_ROWS = 2  # --ci-synthetic fixes NS=10000,50000 (bench_1m.sh)
-_BENCH1M_EXPECTED_CHECKS = 3  # khive-vamana/ci-synthetic/clustered-128 in perf/targets.toml
+_BENCH1M_EXPECTED_ROW_NS = frozenset({10000, 50000})
 _BENCH1M_ROW_NUMERIC_FIELDS = (
     "build_ms",
     "iso_recall_beam",
@@ -331,6 +334,24 @@ _BENCH1M_FITS_NUMERIC_FIELDS = (
     "build_wallclock_exponent",
     "iso_recall_query_exponent_warm",
     "bruteforce_exponent",
+)
+# khive-vamana/ci-synthetic/clustered-128 target in perf/targets.toml. `scope`
+# governs the shape of a check's "measured" field in the result JSON
+# (evaluate_check in vec_bench.rs): "all_rows" serializes a Vec<f64> (one
+# value per row), "fits"/"max_n" serialize a single f64. A probe result with
+# an unexpected metric name or scope here is schema drift, not a metric to
+# silently accept.
+_BENCH1M_EXPECTED_CHECK_SCOPES = {
+    "recall_at_10": "all_rows",
+    "beam_growth_exponent": "fits",
+    "speedup_vs_brute_force": "max_n",
+}
+# gate_pass (1) + 2 metrics/check (measured, pass) + row fields + fits fields.
+_BENCH1M_EXPECTED_METRIC_COUNT = (
+    1
+    + len(_BENCH1M_EXPECTED_CHECK_SCOPES) * 2
+    + _BENCH1M_EXPECTED_ROWS * len(_BENCH1M_ROW_NUMERIC_FIELDS)
+    + len(_BENCH1M_FITS_NUMERIC_FIELDS)
 )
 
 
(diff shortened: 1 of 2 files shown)


ohdearquant:
## Round-1 findings addressed (a74bce22)

**1. Coverage discrepancy (99% vs 99.9%) — resolved explicitly in favor of 99%.**
Checked the governing calibration spec (Calibration protocol, item 4): it specifies "99%
coverage / 95% confidence" and pins k(10)~=3.98, which is what the code already implements
(`_TOLERANCE_COVERAGE = 0.99`). Recomputing at 99.9%/95% gives k(5)=7.5314 / k(10)=5.1556 —
inconsistent with the #829 pin. 99% is the correct reading. Added an explicit rationale
comment in `bench_calibrate.py` documenting both sides of the discrepancy, plus a
`test_coverage_is_99_not_99_9_percent` self-check that locks in k(5)=5.7504/k(10)=3.9400 at
99% and asserts the 99.9% reading does NOT match the pin, so this can't silently drift.
Flagging honestly per the task: if a maintainer knows of a spec revision that deliberately
widens this to 99.9%, that's a spec change requiring a fresh k(10) pin, not a code bug.

**2. HOME isolation broke rustup.** `_run_once` isolates `HOME` per run for state isolation,
but bench-1m's `cargo run --release` and the pipeline/load suites' `~/.cargo/bin/kkernel`
fallback resolve toolchain/binary paths relative to `HOME`. Now derives `CARGO_HOME`/
`RUSTUP_HOME` from the *original* `HOME` before swapping to the isolated scratch dir (only
when the caller hasn't already pinned them). Verified live: `--suite bench-1m --runs 2`
completes clean, no "rustup could not choose a version of cargo".

**3. Extractor crashed on real result JSON.** `recall_at_10` (scope=`all_rows`) serializes a
`Vec<f64>` in `evaluate_check` (vec_bench.rs), not a scalar — `float(check["measured"])`
raised `TypeError`. `_bench1m_extract` now branches on `scope`, reduces `all_rows` arrays to
the operator-binding value (min for `>=`/`>`, max for `<=`/`<`), and raises `SchemaError` on
any malformed `measured` shape. Added `scripts/perf/testdata/bench1m_result_fixture.json` —
a real captured `bash scripts/bench_1m.sh --ci-synthetic` result — plus a
`Bench1mExtractSelfCheck` test class exercising the real shape and rejection paths.

**4. Schema/cardinality guard didn't detect drift.** Now requires `schema_version == "1.0"`,
exact `{recall_at_10, beam_growth_exponent, speedup_vs_brute_force}` check identities with
their expected scopes (`all_rows`/`fits`/`max_n`), exact row-N set `{10000, 50000}`, and a
final metric-count guard (29). Each violation raises `SchemaError`. Verified with unit tests
that a `schema_version=999.0` probe and an unexpected assertion name/scope are now rejected.

## Verification
- `python3 -m py_compile scripts/perf/bench_calibrate.py` — OK
- `python3 -m unittest scripts.perf.bench_calibrate -v` — 10/10 pass (4 new: coverage
  discrepancy lock-in + 3 fixture-based bench-1m extractor tests, on top of the existing 3 +
  new `Bench1mExtractSelfCheck` cases)
- `cargo build --release --workspace` (from `crates/`) — clean
- `python3 scripts/perf/bench_calibrate.py --suite bench-1m --runs 2` — live, completes clean,
  extracts metrics, no crash
- `python3 scripts/perf/bench_calibrate.py --suite pipeline --runs 2` — live, completes clean,
  extracts metrics

oceanwaves630 (approved):
Reviewed at 17420195. The n-sized tolerance factor is implemented exactly to spec: I re-derived k(10) by hand from the Wald-Wolfowitz formula (z_p=2.3263, z_conf=1.6449, a=0.8497, b=5.1413) and got 3.9399, matching the pinned 3.9400 within rounding. The 99%-vs-99.9% resolution is locked on both sides by the self-check (99.9% gives k(10)=5.1556, correctly asserted NOT to match the pin), the domain guard degrades to floor-omitted rather than crashing at n=2, and the bench-1m extractor is properly fail-closed on every schema-drift axis including the per-check all_rows cardinality registry. The all_rows reduction to the operator-direction worst case is the right semantics. The HOME-swap CARGO_HOME/RUSTUP_HOME preservation is a good catch.

One non-blocking follow-up: --runs has no argparse minimum, and n=1 raises plain ValueError from _tolerance_factor_k, which the ToleranceDomainError-only handlers in _advisory_floor and main don't catch — so --runs 1 crashes with a traceback after the suite has already executed. A meaningless invocation for calibration, but a one-line floor on the flag (or catching ValueError alongside) would fail it fast before spending bench time. Fine to fold into the next perf touch. Approved.

PR #831, opened 2026-07-11
fix(db): decode FeedbackExplicit signal observation from target_id

### LINKED ISSUE #811: event decode: decode_signal_observation reads about_id that FeedbackExplicit never sets
Latent-bug FYI from the Tile-0 recon (atlas, 2026-07-10).

decode_signal_observation (khive-db event.rs:572-582) looks up payload about_id, but the emitted FeedbackExplicit payload never sets about_id — the emitter uses event.target_id via .with_target. Either dead code or a silently-empty projection for feedback observations.

Verify which and fix accordingly: if the projection is meant to surface feedback observations, decode from target_id; if dead, delete the lookup.

ohdearquant:
## Root cause

`decode_signal_observation` (`crates/khive-db/src/stores/event.rs`) projects `event_observations` rows for `FeedbackExplicit` events by reading `payload.about_id`. No emitter ever writes that field.

The only emitter, `brain.feedback` (`crates/khive-pack-brain/src/handlers.rs:1451` and `:1493`), builds the event via:

```rust
Event::new(namespace, "brain.feedback", EventKind::FeedbackExplicit, SubstrateKind::Event, "brain")
    .with_target(target)
    .with_payload(data)
```

`.with_target(target)` sets `event.target_id`, not `payload["about_id"]`. Every real `FeedbackExplicit` event therefore projects zero `Signal` observation rows — `decode_signal_observation` always hits its `payload_uuid(event, "about_id")? == None` early return and returns `Ok(Vec::new())`. This is a silently-empty projection, not a hard failure, so nothing surfaced it until #811's recon.

## ADR contract

[ADR-041](docs/adr/ADR-041-event-provenance-projection.md) §3/§Implementation is the governing contract and specified the (nonexistent) `payload.about_id` field. Checked against the emitter, `event.target_id` is the side that's actually shipped and consistent with every other target-carrying event kind (`EntityUpdated`, `NoteUpdated`, `TaskTransitioned`, `LinkCreated`'s source/target) — all of which the decoder already reads from `target_id`/well-known payload fields, never a role-specific alias. Verdict: **the decoder was wrong, not the emitter.** Adding a duplicate `about_id` payload field to the emitter would be a redundant field with no consumer.

Fixed the decoder side and added **Amendment A1** to ADR-041 documenting the corrected contract (`event.target_id`, not `payload.about_id`) and why the emitter side was judged correct.

## Fix

```rust
fn decode_signal_observation(event: &Event) -> Result<Vec<EventObservation>, rusqlite::Error> {
    let Some(entity_id) = event.target_id else {
        return Ok(Vec::new());
    };
    ...
```

Mirrors `decode_target_observation`'s existing `event.target_id` read.

## Test evidence

Added `feedback_explicit_projects_signal_observation_from_target_id` in `crates/khive-db/src/stores/event_tests.rs`: builds a `FeedbackExplicit` event the same way the real emitter does (`Event::new(...).with_target(target)`, no `about_id` in payload), appends it through `SqlEventStore::append_event`, and asserts one `event_observations` row with `role = 'signal'` and `entity_id = target` exists. Before the fix this test fails (0 rows); after the fix it passes.

```
$ cargo test -p khive-db
test result: ok. 249 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out  (contract tests)

$ cargo clippy -p khive-db --all-targets -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) — clean

$ cargo fmt --all -- --check
(clean)
```

Fixes #811

### DIFF
--- a/crates/khive-db/src/stores/event.rs
+++ b/crates/khive-db/src/stores/event.rs
@@ -453,8 +453,21 @@ fn decode_event_observations(event: &Event) -> Result<Vec<EventObservation>, rus
 }
 
 fn payload_uuid_array(event: &Event, field: &'static str) -> Result<Vec<Uuid>, rusqlite::Error> {
+    Ok(payload_uuid_array_opt(event, field)?.unwrap_or_default())
+}
+
+/// Like [`payload_uuid_array`], but distinguishes an absent `field` (`Ok(None)`)
+/// from a present-but-malformed one (`Err`). Callers that fall back across a
+/// chain of alternative field names (e.g. `decode_rank_observations`'s
+/// `selected`/`reranked`/`final_scores`) need that distinction: collapsing
+/// "missing" into `Ok(vec![])` makes every later field in the chain
+/// unreachable, since `Result::or_else` only fires on `Err`.
+fn payload_uuid_array_opt(
+    event: &Event,
+    field: &'static str,
+) -> Result<Option<Vec<Uuid>>, rusqlite::Error> {
     let Some(values) = event.payload.get(field) else {
-        return Ok(Vec::new());
+        return Ok(None);
     };
     let Some(array) = values.as_array() else {
         return Err(invalid_payload(event.kind, field, "expected array"));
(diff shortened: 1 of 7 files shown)


oceanwaves630 (approved):
Content-level review of the full diff at ac3b33ca.

Verified:
- The missing-vs-malformed split (`payload_uuid_array_opt` returning Ok(None) for an absent field) genuinely fixes the unreachable fallback chain: the old code returned Ok(empty) for a missing `selected`, so `or_else` — which fires only on Err — never reached `reranked`/`final_scores`, projecting zero selected rows for every legacy rerank event.
- The decoder split matches the typed contracts: RecallExecuted/SearchExecuted consult only `selected` (their ADR-041 projection field); RerankExecuted consults `final_scores` (ordered output, authoritative) with `reranked` as legacy fallback only when `final_scores` is absent, and a stray `selected` key on a rerank payload is never consulted — each pinned by its own test, including the both-present order-disagreement case.
- Tuple-shape decoding through the exact typed shapes rejects one-element, extra-element, and non-numeric-score tuples with the event insert rolled back, not silently accepted.
- The `target_id` fix is the right direction: the emitter always wrote `Event::with_target`, never a payload `about_id`, so the Signal projection had been silently empty for the ADR's entire implementation lifetime — reading the field the emitter actually writes, consistent with `decode_target_observation`. Substrate threading from the resolved target (entity/note) plus the historical-placeholder fallback to Entity preserves pre-fix data behavior, and the `observed_as_signal` lowering widens to match, with the end-to-end note-target test resolving through the canonical GQL query.
- ADR-041 Amendments A1/A2 document both changes accurately against the code.

One non-blocking note: `decode_rerank_observations` evaluates the `reranked` decode eagerly via `.or(...)?`, so a payload with valid `final_scores` but malformed `reranked` rejects the whole event. Fail-closed and consistent with the tuple-shape strictness — just noting it's stricter than lazy fallback would be.

Approving.

PR #832, opened 2026-07-11
fix(query): accept integer literals in GQL inline property maps

### LINKED ISSUE #755: query: GQL inline property maps reject integer literals; quoted numbers silently match nothing
## Summary

GQL inline node-property patterns reject integer literals, and the obvious workaround silently returns nothing. First-time users hit this immediately on the git-provenance corpus ("show me PR #54"):

| Form | Result |
|------|--------|
| `MATCH (n:pull_request {number:54}) RETURN n` | parse error: `expected string literal` |
| `MATCH (n:pull_request {number:'54'}) RETURN n` | silently empty (property is stored as a JSON number; string never matches) |
| `MATCH (n:pull_request) WHERE n.number = 54 RETURN n` | works |

The only working form is the non-obvious one. The inline-map parser accepts only string literals, while the WHERE clause compiler handles numeric comparison correctly.

## Expected

Inline property maps should accept the same literal types WHERE accepts — at minimum integer (and ideally float/boolean) literals — and compile to the same JSON-number comparison the WHERE path uses.

Alternatively (or additionally), if `{number:'54'}` can never match a stored JSON number, a type-mismatch warning would be better than a silent empty result.

## Where

`crates/khive-query` GQL parser — inline property-map literal grammar vs the WHERE expression grammar.

## Provenance

Found by first-consumer usage evidence on the ADR-088 git-provenance corpus (query-DX track, 2026-07-09). Ranked a top-5 friction item: every first-time user hits it, and the intermediate failure mode (quoted number → silent empty) actively misleads.

ohdearquant:
## Root cause

GQL inline node-property patterns (`{key: value}`) only accepted string literals — `parse_props()` in `crates/khive-query/src/parsers/gql.rs` called `parse_string_literal()` unconditionally, so `MATCH (n:pull_request {number:54}) RETURN n` failed to parse (`expected string literal`).

The documented workaround, `{number:'54'}`, silently matched nothing. Entity/note properties are stored as a JSON blob; the SQL compiler always bound property-map values as `QueryValue::Text` and compared with `json_extract(properties, '$.number') = ?N COLLATE NOCASE`. SQLite's `json_extract` returns a JSON number as an INTEGER/REAL storage class, and SQLite never considers TEXT equal to INTEGER/REAL regardless of collation — so a quoted numeric literal can never match a JSON-number property. The WHERE-clause path (`b.number = 54`) already worked because `ConditionValue::Number` binds a numeric `QueryValue::Float`.

`NodePattern.properties` was typed `HashMap<String, String>`, which structurally could not carry a numeric literal through to the compiler — the parser-level restriction and the compiler-level TEXT binding were two faces of the same root cause.

## Fix

- `NodePattern.properties` is now `HashMap<String, ConditionValue>` (`crates/khive-query/src/ast.rs`), reusing the `String`/`Number`/`Bool` literal type already used by WHERE clause conditions.
- GQL's `parse_props()` (`crates/khive-query/src/parsers/gql.rs`) now calls the existing `parse_value()` parser (already exercised by the WHERE-clause grammar), so inline maps accept string, integer, negative integer, float, and bool literals — the same literal grammar WHERE already accepts, per the issue's "Expected" section.
- SPARQL's triple-to-property lowering (`crates/khive-query/src/parsers/sparql.rs`) wraps string/kind objects in `ConditionValue::String` to match the new property-map type. Numeric-literal triples in SPARQL were already routed through the WHERE-condition path (`triples_to_ast`) rather than inline properties, so SPARQL numeric literals were unaffected by this bug — only the value-type plumbing changed.
- Added `compile_property_equality()` in the SQL compiler (`crates/khive-query/src/compilers/sql.rs`): string values bind as `TEXT` with `COLLATE NOCASE` (unchanged case-insensitive behavior), number values bind as `REAL` with no `COLLATE`, and bool values bind as `INTEGER` — matching how `json_extract` returns JSON numbers/booleans as numeric SQLite storage classes. Applied at all four inline-property-map compile sites: the fixed-length node path, the observation-target node path, and the variable-length (recursive CTE) start/end node paths.
- `entity_type` (lifted out of the property map into its own dedicated column) must still resolve to a string literal; a non-string value (`{entity_type: 54}`) now returns a clear parse error instead of silently producing a nonsensical comparison.

## Decided behavior for quoted numeric strings

A quoted numeric literal, `{number: '54'}`, is a deliberate string literal and **keeps matching JSON strings only** — it is not coerced to a number. This is intentional: `khive` properties are typed JSON, and silently reinterpreting a quoted string as a number would be its own source of surprising behavior (e.g. a property that is genuinely the string `"054"` or `"1e3"`). The silent-mismatch trap from the issue is closed by making the unquoted form (`{number: 54}`) work correctly, not by having the quoted form start matching numbers it was never meant to represent.

## Test evidence

`khive-query` unit + integration tests (`crates/khive-query/src/parsers/gql.rs`, `crates/khive-query/tests/compile_integration.rs`):
- Parser accepts integer, negative integer, float, and bool inline literals.
- `entity_type` with a non-string literal is rejected with a parse error.
- Compiled SQL binds a numeric `QueryValue` (no `COLLATE NOCASE`) for numeric inline properties, and still binds `QueryValue::Text` + `COLLATE NOCASE` for quoted strings — exercised across both the fixed-length and variable-length (recursive CTE) compile paths.

`khive-runtime` end-to-end regression against a real SQLite runtime (`crates/khive-runtime/tests/integration.rs`):
- `query_via_gql_inline_property_map_integer_literal_matches_json_number`: an entity created with JSON property `{"number": 54}` is matched by `MATCH (n:artifact {number: 54}) RETURN n`.
- `query_via_gql_inline_property_map_quoted_number_does_not_match_json_number`: the same entity returns zero rows for `MATCH (n:artifact {number: '54'}) RETURN n`, documenting the decided quoted-string behavior.

```
cargo test -p khive-query -p khive-runtime   # 65 + 142 + 46 + 8 unit/integration tests pass
cargo clippy -p khive-query -p khive-runtime --all-targets -- -D warnings   # clean
cargo fmt --all -- --check   # clean
```

Fixes #755

### DIFF
--- a/crates/khive-query/docs/design.md
+++ b/crates/khive-query/docs/design.md
@@ -92,6 +92,73 @@ edges that join against `event_observations`, not `graph_edges`.
 - Event nodes do not have `entity_type` or arbitrary `properties` -- these are
   rejected at compile time with an actionable error.
 
+### Inline Property-Map Literal Grammar (issues #755, #832)
+
+Node patterns support an inline property map: `(n:kind {key: value, ...})`. The
+grammar for `value` (shared with WHERE-clause condition values, via the same
+`parse_value` parser) is:
+
+```text
+value      = string | integer | float | bool
+string     = "'" ... "'" | '"' ... '"'
+integer    = ["-"] digit+                     -- no "." in the lexeme
+float      = ["-"] digit+ "." digit+          -- "." required
+bool       = "true" | "false"                 -- case-insensitive
+```
+
+**Type binding:**
+
+- `string` binds as SQL `TEXT` with `COLLATE NOCASE` (case-insensitive
+  equality against the property's stored JSON string).
+- `integer` binds as `QueryValue::Integer(i64)` -- the full `i64` range
+  (`i64::MIN..=i64::MAX`) is supported, including magnitudes beyond `f64`'s
+  exact-integer limit of 2^53. An integer lexeme outside `i64` range is a
+  **parse-time error** (`QueryError::Parse`), not a silent truncation or a
+  fallback to a lossy float.
+- `float` binds as `QueryValue::Float(f64)`. A float lexeme whose magnitude
+  overflows `f64` to `Infinity` (or otherwise fails to parse as finite) is a
+  **parse-time error** (`QueryError::Parse`); non-finite values never reach
+  the compiler. The compiler additionally re-checks `is_finite()` on every
+  numeric parameter it binds (in `compile_property_equality` and in WHERE
+  condition compilation) as defense-in-depth, returning `InvalidInput` if a
+  non-finite value is ever constructed by a caller outside the parser (e.g. a
+  hand-built AST).
+- `bool` binds as `QueryValue::Integer(0)` / `QueryValue::Integer(1)`, no
+  `COLLATE`.
+
+**`entity_type` is string-only.** `entity_type` is lifted out of the property
+map into `NodePattern.entity_type` and compiled to the dedicated
+`entity_type` column (never `json_extract`). A non-string `entity_type`
+value (e.g. `{entity_type: 54}`) is rejected at parse time with
+`QueryError::Parse`.
+
+**Why integer and float are distinct literal kinds, not one numeric type:**
+entity/note properties are stored as JSON, and SQLite's `json_extract`
+returns a JSON number as either its INTEGER or REAL storage class depending
+on whether the JSON literal had a decimal point. Prior to #832, every numeric
+literal parsed to `f64` and compiled to `QueryValue::Float` (`REAL`)
+regardless of source form; large integers (`2^53+1` and beyond, including
+`i64::MAX`/`i64::MIN`) silently rounded to the nearest representable `f64`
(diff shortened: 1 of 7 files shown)


oceanwaves630 (approved):
Reviewed at a3b1fe46. Splitting the numeric literal grammar at the lexeme (integer = no dot, float = digits on both sides) is the right fix for the 2^53 precision bug: integer literals now bind QueryValue::Integer(i64) exactly through all four compile paths, unified in compile_property_equality with COLLATE NOCASE correctly restricted to strings. i64 overflow and non-finite floats are parse-time errors rather than silent truncation, entity_type is enforced string-only in both GQL and SPARQL front-ends, and the tightened 1./-. rejection matches the documented grammar with regression guards on well-formed floats. Coverage is genuinely branch-complete: the variable-length tests route through the recursive-CTE WHERE and end-node inline-map compilers, and the runtime integration tests prove 2^53+1 and both i64 bounds match against real SQLite storage classes. The quoted-'54'-stays-TEXT decision is tested as deliberate behavior. Approved.

PR #833, opened 2026-07-11
perf: CI-tracked bench trends for e2e and component suites

ohdearquant:
## Summary

Adds CI-tracked, advisory-only trend tracking for both e2e (pipeline daemon + bench-1m synthetic) and component (Criterion) benchmarks, per the Leo-signed bench-program spec (`.khive/workspaces/20260710/bench-program/SPEC-draft.md`). This is Wave 2 territory (verb-latency / bench trend ledger), scoped down to what a single efficient workflow can carry without duplicating Wave 1's calibration harness.

- **`scripts/perf/bench_track.py`** (stdlib-only): appends one JSONL record `{schema_version, suite, sha, branch, timestamp, metrics, host}` to `bench-data/<suite>.jsonl` per run, and renders a compact markdown trend (last N runs, per-metric direction arrows, **no thresholds**). Reuses `bench_calibrate.py`'s `SUITES` registry and its `pipeline`/`load` extractors by import — does not re-parse their stdout/JSON. Adds two more sources of its own: arbitrary bench JSON (`--source json`, e.g. `bench_1m.sh --ci-synthetic`'s output) and a Criterion `estimates.json` tree walker (`--source criterion`).
- **`scripts/perf/publish_ledger.sh`**: publishes `bench-data/*.jsonl` to a dedicated orphan `perf-data` branch (never `main`), with a fetch+reset+retry loop (no force-push) so the two jobs below (and a nightly/push overlap) can't race each other's push.
- **`.github/workflows/bench-track.yml`**: two jobs, `components` (the 23 Criterion targets across 20 crates, compile-checked then run with `--quick`, wall-clock-bounded) and `e2e` (the pipeline daemon suite + the hermetic `bench-1m --ci-synthetic` gate, each a single informational run).
- **`scripts/perf/README.md`**: ledger schema, the `perf-data` branch, how to read trends.
- **`scripts/perf/test_bench_track.py`**: 13 unittest cases (record shape, ledger round-trip, trend rendering, all three metric sources).

## Design (spec references)

Per the spec's [Blocking-promotion ladder](.khive/workspaces/20260710/bench-program/SPEC-draft.md#blocking-promotion-ladder), every metric this workflow records is **Advisory** — measured, appended to the ledger, rendered as a trend — and never asserts pass/fail. Promotion to a blocking gate is out of scope here: it requires a `K>=10` same-SHA calibration run (`bench_calibrate.py`) plus a `>=2`-week, `>=60`-observation zero-alarm window (Gate 1 + Gate 2), which is Wave 1/2's calibration work, not this ledger-writer. This PR only builds the "measured, reported, trended" half of Principle 2 ("Advisory before blocking").

The [CI integration table](.khive/workspaces/20260710/bench-program/SPEC-draft.md#ci-integration) lists a `bench-verb-ledger.yml`-shaped surface (push-to-main + weekly cron, advisory) for Wave 2's Criterion suite, and this PR's `components` job follows that trigger/blocking shape. The `e2e` job additionally tracks the pipeline daemon suite and the bench-1m synthetic gate as informational data points, on the same cadence — both already exist as *blocking* PR gates (`bench-pipeline.yml`, `bench-1m.yml`'s `ann-ci-gate`); this workflow does not touch those, it only adds a second, non-blocking, trend-only observation of the same suites at push-to-main/nightly cadence.

## Efficiency measures

- **Never runs on `pull_request`** — push-to-main (path-filtered to `crates/**`, `scripts/perf/**`, the workflow file itself), nightly cron, and `workflow_dispatch` only.
- `Swatinem/rust-cache` on both jobs.
- `components` job: `cargo bench --workspace --all-targets --no-run` compile-check first (cheap, catches bench rot immediately), then `cargo bench --workspace -- --quick` under a `timeout`, bounded to fit the ~15 minute budget.
- `concurrency`: nightly-cron runs share one group with `cancel-in-progress: true` (an overlapping/stuck nightly never queues behind another); push/dispatch runs get their own per-ref group and are never cancelled (each carries a distinct commit's data point).
- Single repository variable `BENCH_TRACK_DISABLED` skips both jobs entirely (`if: vars.BENCH_TRACK_DISABLED != 'true'`).
- Raw output uploaded as a 90-day-retention artifact (per the spec's CI-integration disposition rule: "a metric that cannot be recovered from a past CI run without re-running the suite is not usable for the two-week advisory-observation requirement").

## What is informational vs future-blocking

**Informational (this PR):** every metric in `bench-data/components.jsonl`, `bench-data/pipeline.jsonl`, `bench-data/bench-1m.jsonl`. No thresholds, no pass/fail assertions anywhere in `bench_track.py` or `bench-track.yml`.

**Already blocking, untouched by this PR:** `bench-1m.yml`'s `ann-ci-gate` (`recall@10 >= 0.90`) and `bench-pipeline.yml`'s `pipeline-gate` (`P@K >= 0.70`) — both remain the real PR gates; this workflow's `e2e` job re-runs the same suites at push/nightly cadence purely to grow the trend ledger, never to gate.

**Future (out of scope here):** promotion of any component-bench or e2e metric to blocking requires the calibration protocol in the spec (K>=10 same-SHA run via `bench_calibrate.py`, then the 2-week/60-observation window) — tracked as Wave 1/2 follow-on work, not this PR.

## Test plan

- [x] `python3 -m py_compile scripts/perf/bench_track.py scripts/perf/bench_calibrate.py`
- [x] `actionlint .github/workflows/bench-track.yml` (clean)
- [x] `python3 -m unittest test_bench_track -v` (13/13 pass)
- [x] Local dry-run of `bench_track.py record --source criterion` against a synthetic `estimates.json` fixture — verified JSONL append + rendered trend markdown
- [x] Local dry-run of `publish_ledger.sh` against a scratch bare remote — verified orphan-branch creation, idempotent no-op rerun, append+republish, and multi-file publish (the `e2e` job's shape)
- [ ] First real run on push to `main` (creates the `perf-data` branch) — cannot be exercised from a draft PR; will verify once merged

### DIFF
--- a/.github/workflows/bench-track.yml
+++ b/.github/workflows/bench-track.yml
@@ -138,18 +138,32 @@ jobs:
           timeout 600 cargo bench "${pkg_args[@]}" -- --quick || true
 
       - name: Record trend ledger entry
+        id: record
+        # Captures bench_track.py's own exit code into $GITHUB_OUTPUT instead
+        # of letting a nonzero exit fail this step outright (round-3 finding:
+        # the step used to fail here, before the publish step ran, so an
+        # error record it had already written to local disk never reached
+        # perf-data - `if: success()` on publish silently skipped it). `set
+        # +e` plus an explicit `exit 0` keep this step green so publish/
+        # summary/upload below always run; the captured code is enforced by
+        # the final "Fail job" step after publish has had its chance.
         run: |
-          set -euo pipefail
+          set +e
           python3 scripts/perf/bench_track.py record \
             --suite components \
             --source criterion \
             --criterion-dir crates/target/criterion \
             --sha "$GITHUB_SHA" \
             --branch "$GITHUB_REF_NAME" \
+            --run-id "$GITHUB_RUN_ID" \
+            --run-attempt "$GITHUB_RUN_ATTEMPT" \
             --data-dir bench-data \
             --summary-out /tmp/components-trend.md
+          echo "exit_code=$?" >> "$GITHUB_OUTPUT"
+          exit 0
 
       - name: Publish ledger to perf-data branch
+        if: ${{ !cancelled() }}
         run: bash scripts/perf/publish_ledger.sh bench-data/components.jsonl
 
       - name: Write step summary
@@ -164,6 +178,14 @@ jobs:
           path: crates/target/criterion
           retention-days: 90
           if-no-files-found: ignore
+
+      - name: Fail job if trend ledger record errored
+        if: ${{ !cancelled() && steps.record.outputs.exit_code != '0' }}
+        env:
+          RECORD_EXIT_CODE: ${{ steps.record.outputs.exit_code }}
+        run: |
+          echo "bench_track.py record exited ${RECORD_EXIT_CODE} for suite=components shard=${{ matrix.shard }}; its status=error record was still published above (see 'Publish ledger to perf-data branch')." >&2
+          exit "${RECORD_EXIT_CODE}"
 
   # ─────────────────────────────────────────────────────────────────────────
   # e2e benches: the pipeline daemon suite (bench_calibrate's `pipeline`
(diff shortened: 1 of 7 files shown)


oceanwaves630 (approved):
Content-level review of the full diff at 5e3021d3 (workflow, bench_track.py, publish_ledger.sh, both test files).

Verified:
- The round-3 fixes are real in the code: aggregation keys on (sha, run_id, run_attempt) so a pass-then-fail rerun of one sha never blends metrics with the other run's gate/host provenance (and the exact scenario is pinned in `test_rerun_same_sha_different_run_id_does_not_mix_gate_status`); `subprocess.TimeoutExpired` is in `_cmd_record`'s except tuple with a test forcing a 1s-timeout suite; record steps capture exit codes with `set +e` so publish always runs before the final fail step re-raises the captured code.
- Legacy (pre-schema-v2) records without run_id/run_attempt aggregate via `.get()` defaults — mixed v1/v2 ledgers render without KeyError.
- `publish_ledger.sh` merge-on-copy: worktree history unioned with local lines via exact-line dedup, which is idempotent because records serialize with sort_keys=True; retry loop re-fetches and re-copies from the same local file, never force-pushes, and the orphan-creation race resolves on retry. The test file drives the real script against a local bare remote and pins the overwrite regression.
- Error records are written before the failure surfaces (`build_error_record` then return 1), and `_aggregate_shards` never masks an error shard with a later ok shard.
- Missing ledger file fails publication loudly (set -euo pipefail on the cat/cp), while a missing summary markdown is non-fatal by design.

One non-blocking observation: all push-to-main runs share one concurrency group (`bench-track-refs/heads/main`) with cancel-in-progress false — GitHub keeps at most one queued run per group, so a rapid burst of pushes can supersede a queued run's data point. Acceptable for an advisory tracker with the nightly cron backstop; worth a comment if it ever surprises someone reading a gap in the ledger.

Advisory-only posture is consistent throughout: no thresholds, no pass/fail assertions, gate exit codes carried as data. Approving.

PR #834, opened 2026-07-11
fix(db): escape LIKE wildcards in entity name-prefix resolution

### LINKED ISSUE #818: git ingest: unescaped LIKE wildcards in entity-name resolution page can hide exact matches

ohdearquant:
## Summary

- Fixes #818: entity name resolution (`resolve_uuid_async` → `resolve_name_async` in `khive-pack-kg/src/handlers/common.rs`) builds an `EntityFilter.name_prefix` from raw user input that `khive-db`'s `build_entity_where` bound unescaped as `name LIKE '<prefix>%'`. Names containing `%` or `_` (SQLite `LIKE` wildcards) broadened the match far beyond a literal prefix. Combined with the query's `LIMIT 100` page (ordered `created_at DESC`), enough wildcard-matched decoy rows created after the real entity could push the true exact match past the page boundary — the caller's post-fetch exact-match filter would then see zero candidates and report `NotFound`, even though the entity existed.
- Escapes `%`, `_`, and `\` in the prefix before binding and adds `ESCAPE '\'` to the `LIKE` clause in `crates/khive-db/src/stores/entity.rs`. This mirrors the `escape_like` helper pattern already used for the same bug class in `khive-pack-git/src/ingest.rs`, `kkernel/src/reindex.rs`, and `khive-pack-knowledge/src/knowledge/vamana.rs` (#819, PR #824). Those three crates sit downstream of `khive-db` in the dependency graph (`types → score → storage → db → query → runtime → pack-kg / pack-gtd → mcp`), so their existing helpers aren't reachable from `khive-db`; this adds a local `escape_like` copy in `khive-db` following the same established per-crate pattern rather than inventing a new cross-crate abstraction for a single call site.
- As defense in depth against page truncation even with the escape fix (a literal-prefix match could in principle still share a crowded page with 100+ other literally-prefixed rows), `query_entities` now ranks an exact case-insensitive name match first via `ORDER BY CASE WHEN LOWER(name) = ... THEN 0 ELSE 1 END, created_at DESC` whenever `name_prefix` is set. This mirrors the exact-before-pattern ordering already used in `find_document_for_path` (`khive-pack-git/src/ingest.rs`) — the issue's stated fix direction was both "escape wildcards" and "prefer exact-match candidates before pattern candidates so an exact hit can never lose to page truncation."

## Root cause

`crates/khive-db/src/stores/entity.rs:450` (pre-fix) bound `format!("{}%", prefix)` directly as a `LIKE` pattern with no escaping and no `ESCAPE` clause. `crates/khive-pack-kg/src/handlers/common.rs:199-244` (`resolve_name_async`) is the sole caller — it sets `name_prefix` to the raw, unresolved identifier string whenever a caller passes an entity name instead of a UUID or hex prefix (e.g. `link(source_id="50%off", ...)`), fetches page 1 (limit 100, `created_at DESC`), and then filters that page in Rust for a case-insensitive exact match. With enough decoy rows matching the *broadened, unescaped* pattern created after the real entity, the real entity fell outside the 100-row page and resolution failed.

## Test plan

- [x] Added `test_query_by_name_prefix_escapes_underscore_wildcard` and `test_query_by_name_prefix_escapes_percent_wildcard` in `crates/khive-db/src/stores/entity_tests.rs` — each seeds the real entity (`a_b` / `50%off`) plus 150 decoy entities matching only the *unescaped* wildcard form, created afterward so an unescaped/unordered page would truncate the real match. Verified these fail against the pre-fix code (exact match absent from the page) and pass after the fix.
- [x] Added `link_by_name_with_underscore_resolves_exactly` and `link_by_name_with_percent_resolves_exactly` in `crates/khive-pack-kg/tests/integration.rs` — exercise the actual `link` verb end-to-end with wildcard-bearing entity names plus a small decoy, proving the real-world resolution path still works.
- [x] `cargo test -p khive-db -p khive-pack-kg` (from `crates/`, `CARGO_TARGET_DIR=./target`) — 278 + 14 + 131 + 6 + 3 + 223 + 11 = all passing, including the 4 new tests.
- [x] `cargo clippy -p khive-db -p khive-pack-kg --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all -- --check` — clean.

### DIFF
--- a/crates/khive-db/src/stores/entity.rs
+++ b/crates/khive-db/src/stores/entity.rs
@@ -380,6 +380,22 @@ fn parse_uuid(s: &str) -> Result<Uuid, rusqlite::Error> {
     })
 }
 
+/// Escape SQLite `LIKE` wildcard characters (`%`, `_`) and the escape
+/// character itself (`\`) so a caller-supplied name is matched literally
+/// under `LIKE ... ESCAPE '\'` rather than as a pattern (#818: an
+/// entity named e.g. `a_b` must not also match `aXb`, and a name containing
+/// `%` must not silently widen into a broad substring scan).
+fn escape_like(input: &str) -> String {
+    let mut out = String::with_capacity(input.len());
+    for c in input.chars() {
+        if matches!(c, '\\' | '%' | '_') {
+            out.push('\\');
+        }
+        out.push(c);
+    }
+    out
+}
+
 fn build_entity_where(
     namespace: &str,
     filter: &EntityFilter,
@@ -448,8 +464,8 @@ fn build_entity_where(
     }
 
     if let Some(ref prefix) = filter.name_prefix {
-        params.push(Box::new(format!("{}%", prefix)));
-        conditions.push(format!("name LIKE ?{}", params.len()));
+        params.push(Box::new(format!("{}%", escape_like(prefix))));
+        conditions.push(format!("name LIKE ?{} ESCAPE '\\'", params.len()));
     }
 
     if !filter.tags_any.is_empty() {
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Both defects are real and independently fixed: LIKE wildcard escaping with the ESCAPE clause (consistent with the vamana fix), and exact-match-first ordering so page truncation can never strand a name-resolved record behind newer prefix-sharing rows. The ordering regression test is properly isolated — wildcard-free decoys, so it fails on the ORDER BY CASE removal alone. Approved.

PR #835, opened 2026-07-11
fix(pack-git): mask credentials in issue titles without dropping notes

### LINKED ISSUE #801: git pack: ingest_issues passes raw issue title into note name/properties — credential-shaped titles silently drop the issue note
Same mechanism as #763, sibling site. PR #785 fixed the commit-ingest path, but ingest_issues still passes the raw issue.title into properties.title and the note name. A credential-shaped issue title trips the secret gate and the issue note is silently dropped during ingest.

Evidence: reviewer leg on PR #785 proved the mechanism empirically during the gate (approval comment on #785 carries detail).

Fix shape: same as #785 — sanitize/neutralize the title before it reaches the gated write path, keep the raw title recoverable, and add a regression test with a credential-shaped issue title.

ohdearquant:
## Root cause

`ingest_issues` in `crates/khive-pack-git/src/ingest.rs` passed the raw GitHub `issue.title` straight into `properties.title` and the note `name` before dispatching `create`. The runtime secret gate (`khive-runtime::secret_gate`) recursively scans `properties` and the `name` on every `create` call. A credential-shaped issue title (e.g. `ghp_...`) trips that gate, `create` returns `Err`, and the pre-existing per-record failure handling in `ingest_issues` records a warning and retries the record on the next pass — the issue note never lands until the title stops looking like a credential.

This is the sibling site of #763/PR #785, which fixed the identical bug in `ingest_prs` (PR body/title). Issue ingestion never got the equivalent fix.

## Fix

Mirror the `ingest_prs` fix: mask the title once via `secret_gate::mask_secrets` before it is used anywhere gated, and use the masked `safe_title` for both `properties.title` and the note `name`. The original title is never stored in the raw form used to trip the gate; the masked value with `***MASKED***` markers is what gets persisted, so the record is created successfully instead of being rejected.

The gate itself is untouched — only the pack-level ingest path changes what it feeds into `create`.

## Loud failure on residual drops

The existing per-record path in `ingest_issues` already treats any `create` failure as loud, not silent: a failed record pushes `report.warnings.push(format!("create issue #{}: {e}", ...))`, does not increment `issues_ingested`, and leaves the cursor frozen so the record is retried on the next pass (same mechanism `ingest_prs`/`ingest_commits` use, and already covered by the `issue_hook_rejects_ungoverned_state_reason`-style tests in `acceptance.rs`). Masking the title removes the common trigger for that failure path; if a masked record still cannot be stored for some other reason, it is still counted (not silently dropped) via that same warnings/cursor-stall mechanism.

## Test evidence

Added two regression tests to `crates/khive-pack-git/tests/acceptance.rs`:

- `ingest_masks_credential_shaped_issue_title_without_dropping_note` — an issue titled with an obviously-fake `ghp_AAAA...` token shape ingests successfully (`issues_ingested == 1`), produces no drop warning, and both the stored `name` and `properties.title` carry `***MASKED***` with the raw token absent.
- `ingest_leaves_clean_issue_title_unmasked` — a clean, non-credential title passes through byte-for-byte unchanged, guarding against future over-aggressive masking.

```
cargo test -p khive-pack-git       # 70 unit + 34 acceptance passed, 0 failed
cargo clippy -p khive-pack-git --all-targets -- -D warnings   # clean
cargo fmt --all -- --check          # clean
```

Fixes #801

### DIFF
--- a/crates/khive-pack-git/src/hook.rs
+++ b/crates/khive-pack-git/src/hook.rs
@@ -95,8 +95,12 @@ impl KindHook for CommitHook {
     }
 }
 
-/// The governed `state_reason` value set for `issue` (ADR-088 §3).
-const ISSUE_STATE_REASONS: &[&str] = &["completed", "not_planned", "reopened", "duplicate"];
+/// The governed `state_reason` value set for `issue` (ADR-088 §3). `pub(crate)`
+/// so `ingest::MaskedIssueFields` can classify against the same set at the
+/// masking boundary, before an ungoverned (possibly credential-shaped) raw
+/// value can reach this hook's error path.
+pub(crate) const ISSUE_STATE_REASONS: &[&str] =
+    &["completed", "not_planned", "reopened", "duplicate"];
 
 /// `KindHook` shared by `issue` and `pull_request` — both require
 /// `properties.number` and `properties.project_id`, and, when present,
@@ -149,9 +153,12 @@ impl KindHook for IssueLikeHook {
         }
 
         if let Some(reason) = props.get("state_reason").and_then(Value::as_str) {
+            // The raw value is never interpolated into this error: it is
+            // caller-controlled (for `issue`, sourced from GitHub) and may be
+            // credential-shaped. Only the static governed set is echoed.
             if self.kind == "issue" && !ISSUE_STATE_REASONS.contains(&reason) {
                 return Err(RuntimeError::InvalidInput(format!(
-                    "issue properties.state_reason {reason:?} invalid — valid: {}",
+                    "issue properties.state_reason is not one of the governed values — valid: {}",
                     ISSUE_STATE_REASONS.join(", ")
                 )));
             }
--- a/crates/khive-pack-git/src/ingest.rs
+++ b/crates/khive-pack-git/src/ingest.rs
@@ -1505,6 +1505,7 @@ async fn ingest_issues(
 
             let raw_body = issue.body.unwrap_or_default();
             let content = secret_gate::mask_secrets(&raw_body).into_owned();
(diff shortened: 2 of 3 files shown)


oceanwaves630 (approved):
Content-level review of the full diff at db5488a1.

Verified:
- `MaskedIssueFields::new` consumes `GhIssue` by value with an exhaustive destructure (no `..`), so no call site can read a raw field back out and a future `GhIssue` field is a compile error at the boundary rather than a silent bypass. Title, body, author login, and labels all funnel through the secret gate; timestamps parse to canonical RFC3339 or are dropped with a warning that names only the field.
- The round-3 paging-floor fix is real and complete: the ENTIRE page is masked/canonicalized before the sort and before `last_updated_at` derivation, so a credential-shaped raw `updatedAt` (which sorts last under raw ASCII comparison) becomes None (which sorts first) and can never become the next `gh --search updated:>=` argv token. The full-page test forces exactly this branch at PAGE_LIMIT size and asserts the raw value absent from argv log, persisted cursor, and every stored record, while continuation still fires.
- Verified at source (decide_page_outcome, unchanged by this PR) that an all-None floor on a full page hits the `StopFloorStalled` arm — it can never fabricate a Continue floor, matching the conservative-stop claim.
- `StateReasonField::Rejected` never carries the raw string: classification happens at the boundary against the same governed set the hook uses (now pub(crate), one source of truth), the ingest warning names only the field, and the hook's own error path no longer interpolates the raw value — so a credential-shaped stateReason can no longer reach report.warnings through the create-failure path. The record is warn-and-skipped fail-closed per ADR-088 §3, pinned end-to-end.
- Clean-title passthrough test guards against over-masking regression; label and author-login masking each pinned against the recursive properties scan silently dropping the note.

Approving.

Issue #758, opened 2026-07-09
daemon: recovery lock released before spawned daemon is ready — double-spawn / kill-mid-boot window


ohdearquant:
## Summary

`kill_and_respawn` (crates/khive-mcp/src/daemon.rs:893) releases the recovery lock as soon as the replacement daemon process has been *spawned*, not when it is *ready*. `spawn_daemon` (daemon.rs:584) is fire-and-forget — `cmd.spawn()` returns immediately, before the child has acquired the daemon boot guard or bound the socket:

```rust
async fn kill_and_respawn(config_id: &str, namespace: &str) -> std::io::Result<RecoveryOutcome> {
    let _lock = acquire_recovery_lock();          // held for fn scope only
    match probe_daemon_identity(config_id, namespace, 500).await {
        ProbeOutcome::Alive | ProbeOutcome::Timeout => return Ok(RecoveryOutcome::Skipped),
        ProbeOutcome::Dead => {}
    }
    kill_stale_daemon_inner();
    spawn_daemon()?;                              // fire-and-forget cmd.spawn()
    Ok(RecoveryOutcome::Spawned)                  // _lock drops HERE
}
```

## Failure window

Between the lock drop and the child binding the socket, a second recoverer can:

1. acquire the (now free) recovery lock,
2. probe → the freshly spawned child has not bound yet → `ProbeOutcome::Dead`,
3. `kill_stale_daemon_inner()` — potentially killing the healthy, mid-boot child,
4. spawn another daemon.

Consequences range from a wasted double-spawn (if daemon-side boot-guard/bind exclusivity makes the loser exit cleanly) to a respawn storm under concurrent bridge processes each racing recovery — the exact multi-seat cold-start scenario the recovery lock exists to serialize.

## Fix directions

- Hold the recovery lock until the spawned child is confirmed ready: re-probe with bounded retries (or block on the daemon boot guard the way `wait_for_boot_quiescence_then_reprobe` at daemon.rs:1288 already does) before returning/dropping the lock.
- Alternatively (weaker): after acquiring the recovery lock, distinguish "dead" from "booting" before killing — e.g. treat a live boot-guard holder as `Timeout`/skip rather than `Dead`.

The first direction is preferred: it makes `RecoveryOutcome::Spawned` mean "spawned and reachable," which is what every call site actually wants.

## Provenance

[...]

Issue #667, opened 2026-07-06
Unguarded race between front-end local-dispatch and auto-spawned daemon on first-ever cold boot can corrupt FTS5 virtual-table state


ohdearquant:
## Description

On a brand-new database (no existing daemon/socket), the MCP front-end's request handler forwards to a warm daemon when reachable and falls back to fully local, in-process dispatch otherwise. When no daemon exists yet, the front-end spawns one and polls its socket for up to 5 seconds before giving up and falling back locally.

That first-ever spawn happens with **no lock held**, and the newly spawned daemon process independently runs the same startup migrations and pack schema initialization against the identical database file that the front-end process already ran for itself moments earlier. There is no cross-process coordination beyond ordinary SQLite file locking during this window.

This is asymmetric with the existing **stale-daemon recovery** path, which *does* serialize its kill-and-respawn under a dedicated recovery lock. The first-boot spawn path has no equivalent guard.

## Observed symptom

A `memory.recall` bootstrap call failed with:

```
storage: backend driver error in Text during fts_search: vtable constructor failed: fts_notes
```

when running a concurrency load/perf harness against a fresh scratch database with the full production pack set loaded, while the identical harness with one fewer pack loaded (dropping only the newest, unrelated pack) passed reliably. The pack whose presence flips the outcome does not itself touch the FTS table — its addition appears to add just enough incremental per-process startup work to push an already timing-sensitive condition over an observable threshold, rather than being a direct cause.

## Suspected root cause

The first-boot "spawn a daemon, poll its socket, fall back to local dispatch on timeout" path (`forward_or_spawn`, the `NoSocket` branch) holds no lock, unlike the stale-daemon recovery path (`kill_and_respawn`, which serializes under `acquire_recovery_lock`). This lets a freshly spawned daemon process and the front-end's own local fallback both independently perform full startup schema/migration work — including FTS5 virtual-table creation and `fts_search` — against the same database file with no coordination.

The exact SQLite/FTS5-level condition that turns this overlap into a virtual-table *construction* failure (as opposed to an ordinary `SQLITE_BUSY`-with-retry) has not been conclusively identified from static code reading. FTS5 registration spans several physical shadow tables that are not all visible/consistent in one atomic step from a second connection's perspective, which is a plausible fit, but confirming it needs a dynamic reproduction with tracing across both processes.

## Production relevance

This is **not** specific to the harness's single-file hermetic config:

[...]

Issue #645, opened 2026-07-05
daemon: stale-daemon recovery can unlink a fresh replacement daemon's socket/PID (availability race)


ohdearquant:
## Summary

Stale-daemon recovery can unlink the socket/PID files of a *fresh* replacement daemon. When
recovery spawns a replacement while an old daemon is still draining, the old daemon's
shutdown cleanup can unconditionally remove the shared socket/PID paths *after* the
replacement has already written them — leaving the healthy replacement unreachable.

## Mechanism

- `crates/khive-mcp/src/daemon.rs:337` and `daemon.rs:357` perform socket/PID removal on
  shutdown; `crates/khive-runtime/src/daemon.rs:600` is the runtime-side shutdown cleanup.
- The removal is unconditional: there is no PID-ownership check (verify the socket/PID file
  still belongs to *this* daemon before unlinking) and no recovery lock serializing
  respawn-vs-old-shutdown.
- Sequence: recovery detects a stale daemon → spawns replacement → replacement binds and
  writes socket+PID → old daemon finishes draining → old daemon's cleanup unlinks the shared
  paths that now belong to the replacement.

Affected paths: all daemon-forwarded request paths after a stale-daemon recovery event.

## Impact

Availability, not latency. After the race fires, the replacement daemon is running but has no
socket/PID advertised, so the next client falls back to cold local runtime (reloads
embedder/ANN state) or triggers another spawn — repeated cold recovery until the paths settle.

## Falsification / repro

Inspect the immediate socket/PID removal after SIGTERM and the runtime shutdown cleanup for
the absence of a PID-ownership check and a recovery lock. Reproduce by forcing a stale-daemon
recovery under a slow-drain old daemon and asserting the socket path is missing after the old
daemon exits despite a live replacement.

## Proposed direction

Guard unlink with ownership: only remove the socket/PID if it still references this daemon's
PID (compare-and-unlink), or hold a recovery lock across respawn so the old daemon's cleanup
cannot run after the replacement has claimed the paths.

ohdearquant:
Closing via backlog triage 2026-07-09 (verified against current main at 3d03c8f0): Fixed by same PR #698; `kill_stale_daemon_inner` (daemon.rs:699) rechecks PID ownership before delete (696-743), dedicated regression test at daemon.rs:4345.

PR #838, opened 2026-07-11
fix(khive-mcp): confirm daemon dead before killing to close recovery race

ohdearquant:
## Root cause

`kill_and_respawn` (`crates/khive-mcp/src/daemon.rs`) released the recovery lock as soon as the replacement daemon process had been *spawned*, not when it was *ready*. `spawn_daemon()` is fire-and-forget: `cmd.spawn()` returns as soon as the OS has forked+exec'd, well before the child ever reaches its own `acquire_daemon_boot_guard()` call (the same lock, acquired as literally the first thing daemon-mode does in `kkernel::main` — see `crates/kkernel/src/main.rs:322-325` — before it even builds the server).

That gap opened a race: a second recovering client could acquire the now-free lock, send its own bounded identity probe, see `NoSocket` (the child hasn't bound yet) — classified `Dead` — and proceed to `kill_stale_daemon_inner()` + `spawn_daemon()` again. Given the existing #645 defenses (pid-file/socket re-verification right before unlink, `write_pid_file_exclusive`'s daemon-side bind exclusivity), the concrete failure mode is a wasted duplicate spawn / respawn storm under concurrent bridge processes, not literally killing a bound, healthy child — but it's still exactly the double-spawn window the issue describes.

**Why the lock can't simply be held across the whole function**: the daemon acquires this exact lock as its first boot action, holding it across cold-boot schema init through bind + pid-write (`run_daemon_with_boot_guard`, `crates/khive-runtime/src/daemon.rs`). If the client held its own handle to the same lock while waiting for the child to become ready, the child could never acquire it to boot — every single recovery attempt would deadlock against the very daemon it's waiting for. This is documented explicitly in the existing "Deadlock note" comment in `run_daemon_with_boot_guard`.

## Fix

Added `confirm_genuinely_dead`, which reuses the exact "block on the shared boot guard, then re-probe" technique the codebase already uses reactively in `wait_for_boot_quiescence_then_reprobe` (`#667`) — but applies it *proactively*, as a gate before any kill decision, instead of only after a forward timeout:

- `quiesce_then_probe_identity` blocks (via `spawn_blocking`) on reacquiring the same boot/recovery lock a peer's in-flight `kill_and_respawn` or the daemon's own cold boot would be holding, then immediately drops it and re-probes identity. Successfully reacquiring-then-dropping proves nobody was mid-critical-section at that instant — deadlock-free, since it never runs while the caller's own lock handle is still open.
- `confirm_genuinely_dead` retries that up to `DEAD_CONFIRM_ROUNDS` (4, paced by `DEAD_CONFIRM_POLL_MS` = 75ms) times, returning early the moment a peer's boot is observed completing (`Alive`) or going slow (`Timeout` — NEVER-KILL-SLOW). Only returns `Dead` once every round agrees.
- `kill_and_respawn` now runs this confirmation between the initial under-lock probe and `kill_stale_daemon_inner()`/`spawn_daemon()`, so a `Dead` verdict is only trusted after a bounded, actively-verified wait rather than a single point-in-time read.

No sleeps are used as the synchronization mechanism — the real synchronization is the blocking `flock` on the boot/recovery lock; the small pacing sleep between rounds only avoids busy-spinning while polling for a peer that hasn't yet reached its own lock call.

## Test evidence

`confirm_genuinely_dead_waits_out_a_peers_held_boot_guard` (`crates/khive-mcp/src/daemon.rs`) drives the real boot/recovery lock from a background OS thread to deterministically simulate a peer's in-flight boot (same technique the existing `#667` test uses, synchronized via an `mpsc` channel signaled right after the thread acquires the lock — not guessed timing): a background thread holds the real boot guard for 300ms while a fake, already-bound, identity-matching listener answers instantly. Asserts `confirm_genuinely_dead` blocks for the full hold duration and returns `Alive`, not an early `Dead`.

(An earlier draft also added an end-to-end variant through `kill_and_respawn` itself, but `kill_and_respawn`'s own initial `acquire_recovery_lock()` is a synchronous blocking flock with no hook to signal from, so there was no deterministic way to sequence "initial probe completes" before "peer boot guard becomes held" without the same kind of guessed-timing race that made the first draft of this test flaky under full-suite parallel load. Dropped it rather than ship something non-deterministic; the remaining test directly covers the new mechanism the fix relies on.)

```
$ cargo test -p khive-mcp --lib
test result: ok. 189 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo clippy -p khive-mcp --all-targets -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) — clean

$ cargo fmt --all -- --check
(clean, no diff)
```

Refs #758 — this closes the double-spawn/respawn-storm consequence the issue describes via a bounded, actively-verified wait rather than an unbounded lock hold (impossible without deadlocking the daemon's own boot, as explained above), so it is not marked `Fixes` outright.

### DIFF
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -986,18 +986,32 @@ async fn confirm_genuinely_dead(config_id: &str, namespace: &str) -> ProbeOutcom
     // rounds; only give up (return `LockContended`) once every round agreed
     // nobody could confirm, exactly mirroring how `Dead` only becomes trusted
     // once every round agreed the daemon was absent.
-    let mut last_unresolved = ProbeOutcome::Dead;
+    //
+    // #838 round-2 Finding 1: `LockContended` is STICKY across rounds — once
+    // any round can't confirm quiescence, the aggregate must never collapse
+    // back to `Dead` just because a LATER round happened to observe it. The
+    // old code tracked only the last round's outcome, so a
+    // LockContended-then-Dead sequence overwrote the earlier contention and
+    // returned `Dead`, permitting kill+spawn on a call that never actually
+    // established quiescence across every round. `Dead` is only trustworthy
+    // when EVERY round agrees; a single contended round makes the whole call
+    // `LockContended` regardless of what any other round returned.
+    let mut saw_contention = false;
     for round in 0..DEAD_CONFIRM_ROUNDS {
         match quiesce_then_probe_identity(config_id, namespace, BOOT_FENCE_PROBE_TIMEOUT_MS).await {
-            ProbeOutcome::Dead => last_unresolved = ProbeOutcome::Dead,
-            ProbeOutcome::LockContended => last_unresolved = ProbeOutcome::LockContended,
+            ProbeOutcome::Dead => {}
+            ProbeOutcome::LockContended => saw_contention = true,
             other => return other,
         }
         if round + 1 < DEAD_CONFIRM_ROUNDS {
             tokio::time::sleep(std::time::Duration::from_millis(DEAD_CONFIRM_POLL_MS)).await;
         }
     }
-    last_unresolved
+    if saw_contention {
+        ProbeOutcome::LockContended
+    } else {
+        ProbeOutcome::Dead
+    }
 }
 
 /// Deadline for acquiring the recoverer-only lock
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Read the full diff at ae718d58e. The two-lock architecture is the right resolution of the round-1 deadlock tension: the recoverer lock is a separate file the daemon never acquires, so a recoverer can hold it across the entire confirm-through-spawn critical section without deadlocking against the very child it spawned — while the boot lock keeps its existing narrow role (per-round quiescence detection, bounded now, plus the kill+spawn step). Sticky LockContended aggregation is correct: Dead is only trustworthy under unanimity, so a single contended round must poison the whole call regardless of later rounds — and Alive/Timeout short-circuiting stays sound because those are positive observations that make killing unnecessary. The bounded flock helper correctly separates EWOULDBLOCK from hard errors, and Uncertain fails safe at every surface (behaves as Skipped, reported distinctly so logs never conflate confirmed-alive with gave-up). Both test barriers are cfg(test)-only with reset between tests, and the spawn-commit barrier's 80ms bound means a recoverer legitimately blocked at the real lock cannot deadlock the rendezvous. The round-2 test-oracle fix is the important one: the barriers make the recoverer lock the deciding factor rather than scheduler luck, and the independently repeated sabotage (lock removed → 4/4 double-spawn; restored → 5/5 both regressions) proves it. Approved.

Issue #844, opened 2026-07-11
flaky: pack-memory fr2_kg_merge_invalidates_warm_ann_without_subsequent_remember sanity panic on macOS CI (3 hits in one night)


ohdearquant:
High-frequency new flake blocking the merge ladder. `pack::note_mutation_hook_tests::fr2_kg_merge_invalidates_warm_ann_without_subsequent_remember` panics at the sanity assertion (khive-pack-memory/src/pack.rs:985: "warm-up recall must leave the ANN cache current before the merge under test") on macos-latest CI. Three independent hits on 2026-07-11 (runs 29136947650 on PR #826, 29137255585 on PR #833, 29137667058 on PR #835) — all on PRs that do not touch khive-pack-memory.

The failing assertion is the test's own precondition: the warm-up recall is expected to leave the ANN cache current, and on slow/contended macOS runners the async ANN warm apparently has not completed by the time the sanity check runs. Likely related to the ANN warm-race class previously fixed in the #750 lane — the warm task tracking may not cover the path this test's warm-up uses, or the test needs a deterministic warm-complete barrier instead of relying on recall having synchronous warm effects.

Distinct from the #842 pack-git-on-macOS family (different crate, different mechanism) and from #843 (daemon self-heal shared state). Root-cause candidates: deterministic await on the warm task handle in the test's warm-up, or a test-only seam that blocks until ANN generation matches.

ohdearquant:
Fixed by #846 (merged): the fire-and-forget ann::ensure_ann_background task got a test-only wait_until_warm_idle Notify seam (warming guard recorded synchronously, no started-vs-finished hole). Closure was missed when #846 landed — caught by a cursor status pass; #843 (same PR) was closed at the time.

Issue #843, opened 2026-07-11
flaky: daemon::tests::fire_pending_self_heal_is_a_no_op_when_nothing_is_armed intermittent under multi-threaded test runs


ohdearquant:
Observed during PR #838 round-3 work (branch fix/758-daemon-recovery-lock): `daemon::tests::fire_pending_self_heal_is_a_no_op_when_nothing_is_armed` (khive-mcp, re-exec self-heal bridge) failed intermittently (~2/6 runs) under default multi-threaded `cargo test -p khive-mcp --lib`, passes cleanly under `--test-threads=1` (774/774) and does not reproduce on the finding-1-only commit (3/3 clean). The #838 changes touch daemon-recovery locking, a different subsystem; the new two-recoverer barrier test likely shifts scheduling timing enough to expose a latent cross-test isolation race in the self-heal bridge.

This is NOT the khive-pack-git-on-macOS flake family tracked in #842 — different crate, different class (shared/global state between daemon tests rather than pack-git fixture state). Filing separately so the #842 sweep scope stays coherent. Candidate root-cause approaches per the #828 discipline: serial group or thread-local injection for whatever global the self-heal bridge tests share (SPAWN_COUNT-style statics / armed-state globals).

ohdearquant:
Root cause identified during PR #838's round-3 codex review (independent trace): `reset_self_heal_counters()` resets only the atomics, not `PENDING_SELF_HEAL` (crates/khive-mcp/src/daemon.rs:1273, :1489). The pre-existing protocol-mismatch test can leave that slot armed, and under multi-threaded ordering the "nothing armed" test consumes it (`REEXEC_INVOKED_COUNT == 1`). Fix shape: clear `PENDING_SELF_HEAL` in the reset helper (or serialize the two tests). Isolated the test passes 5/5; confirmed pre-existing and untouched by #838's diff.

ohdearquant:
Fix landed in #846 (merged): reset_self_heal_counters() now clears PENDING_SELF_HEAL. Sabotage evidence during the fix leg reproduced the exact predicted failure twice with the fix reverted; 10/10 multi-threaded green with it restored, and the #846 codex review independently re-verified every arming test arms after reset. Nothing remains open.

PR #846, opened 2026-07-11
fix(tests): deterministic ANN warm barrier (#844) + self-heal reset completeness (#843)

ohdearquant:
## Summary

Two independent flaky-test fixes, each landed as its own commit.

### #844 — `pack::note_mutation_hook_tests::fr2_kg_merge_invalidates_warm_ann_without_subsequent_remember`

Root cause: ANN warming is asynchronous. Both `memory.remember`'s write path
(`handlers/remember.rs`) and `memory.recall`'s stale-but-present branch
(`handlers/common.rs`) fire `ann::ensure_ann_background` as a fire-and-forget
background task. When the test's warm-up recall landed on the "cache present
but stale" fast path (because an earlier seeding `memory.remember`'s
background build had already raced ahead and installed something), the
recall returned immediately without waiting on the in-flight rebuild — so
the test's own `is_current` sanity assertion could observe a still-stale
cache.

`khive_runtime::track_background_task` is deliberately fire-and-forget in
production and hands back no `JoinHandle`, so there was nothing to await
directly. Added a minimal test-only seam in `ann.rs`: a `Notify`
(`AnnState::warming_idle`, `#[cfg(test)]`) fired from `WarmingGuard::drop`
whenever a key's single-flight `warming` guard releases, plus
`wait_until_warm_idle(ann, key)` which awaits it in a check-then-notify loop
(no missed-wakeup window, no polling sleep). The test now awaits this before
its sanity assertion instead of assuming synchronous completion. The
sanity assertion itself is unchanged.

Verified: `cargo test -p khive-pack-memory --lib` for this test — 20x
multi-threaded (default) + 5x single-threaded (`--test-threads=1`), **25/25
green**. Full crate suite: 237 passed, 0 failed, 1 ignored.

### #843 — `daemon::tests::fire_pending_self_heal_is_a_no_op_when_nothing_is_armed`

Root cause: `reset_self_heal_counters()` only zeroed the two invocation-count
atomics, leaving the `PENDING_SELF_HEAL` mutex slot untouched.
`forward_or_spawn_rejects_old_daemon_and_returns_protocol_mismatch_error`
exercises `forward_or_spawn`'s protocol-mismatch path end to end, which arms
`PENDING_SELF_HEAL` via `trigger_bridge_self_heal` but never takes the action
back out (it only asserts on the returned error). Under multi-threaded test
ordering, if that test runs before
`fire_pending_self_heal_is_a_no_op_when_nothing_is_armed` in the same binary,
the latter's `reset_self_heal_counters()` call left the armed slot in place,
and `fire_pending_self_heal()` then fired it, failing the "no-op" assertion.

Fix: `reset_self_heal_counters()` now also clears `PENDING_SELF_HEAL` via a
new `clear_pending_self_heal()` helper (both unix and non-unix cfg variants).
Every test that intentionally arms the slot does so AFTER calling
`reset_self_heal_counters`, never before, so this changes no test's intended
semantics.

Sabotage verification: reverted the fix and ran `cargo test -p khive-mcp
--lib` repeatedly to reproduce the original ordering-dependent flake. It
reproduced twice (of the runs completed before a 5-minute batch timeout),
both times failing at exactly the predicted assertion
(`daemon.rs:5242`, `left: 1, right: 0`) — confirming the root cause rather
than just asserting it. Restored the fix and ran the full `khive-mcp` lib
suite **10x multi-threaded, 10/10 green**.

## Test plan

- [x] `cargo test -p khive-pack-memory --lib` — fr2 test 20x multi-threaded + 5x single-threaded, 25/25 green; full crate suite 237 passed
- [x] `cargo test -p khive-mcp --lib` — 10x multi-threaded, 10/10 green (plus 2 reproduced failures pre-fix during sabotage verification)
- [x] `cargo clippy -p khive-pack-memory --all-targets -- -D warnings` clean
- [x] `cargo clippy -p khive-mcp --all-targets -- -D warnings` clean
- [x] `cargo fmt --all -- --check` clean

Refs #844, #843

### DIFF
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -1489,11 +1489,33 @@ pub(crate) static DRAIN_EXIT_INVOKED_COUNT: std::sync::atomic::AtomicUsize =
 pub(crate) fn reset_self_heal_counters() {
     REEXEC_INVOKED_COUNT.store(0, std::sync::atomic::Ordering::SeqCst);
     DRAIN_EXIT_INVOKED_COUNT.store(0, std::sync::atomic::Ordering::SeqCst);
+    clear_pending_self_heal();
 }
 
 #[cfg(all(test, not(unix)))]
 pub(crate) fn reset_self_heal_counters() {
     DRAIN_EXIT_INVOKED_COUNT.store(0, std::sync::atomic::Ordering::SeqCst);
+    clear_pending_self_heal();
+}
+
+/// Clear whatever `PENDING_SELF_HEAL` currently holds (#843). A test that
+/// exercises `forward_or_spawn`'s protocol-mismatch path end to end (e.g.
+/// `forward_or_spawn_rejects_old_daemon_and_returns_protocol_mismatch_error`)
+/// arms it via `trigger_bridge_self_heal` but never takes the action back
+/// out — that is `fire_pending_self_heal`'s job, and asserting the forwarding
+/// behavior alone has no reason to call it. Left armed, that leftover slot
+/// is invisible to `reset_self_heal_counters` resetting only the two
+/// invocation counters, so a later test in the same binary (e.g.
+/// `fire_pending_self_heal_is_a_no_op_when_nothing_is_armed`) can inherit it
+/// under multi-threaded test ordering and observe a spurious fire. Every
+/// existing test that intentionally arms the slot does so AFTER calling
+/// `reset_self_heal_counters`, never before, so clearing it here changes no
+/// test's semantics.
+#[cfg(test)]
+fn clear_pending_self_heal() {
+    *PENDING_SELF_HEAL
+        .lock()
+        .unwrap_or_else(std::sync::PoisonError::into_inner) = None;
 }
 
 /// Test double for [`reexec_in_place`]: a real `exec()` would replace the test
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Reviewed the full diff at 37836f7a1ad61be75ce8cefd30f297061be7630e.

**#844 (fr2 ANN warm flake)**: the `wait_until_warm_idle` barrier is the right shape — the `Notified` future is created before the predicate check (closing the lost-wakeup window per tokio's documented `Notify` guarantee), the loop covers chained re-enqueues, and the predicate is the `warming` single-flight set, which the guard enters synchronously inside `ensure_ann_background` before it returns — so the barrier cannot report idle while a rebuild is pending but unspawned. The fr2 sanity assertion itself is untouched, and the merge-invalidation subject of the test is preserved (verified by the round-1 sabotage: disabling the merge hook still fails the test at its final invalidation assertion).

**#843 (PENDING_SELF_HEAL leak)**: `clear_pending_self_heal()` in both cfg variants of `reset_self_heal_counters` closes the cross-test leak, and the doc comment's safety argument holds — every test that intentionally arms the slot does so after the reset call, so clearing at reset changes no test's semantics. Poison-tolerant lock access is correct for a test-teardown path.

All new state and helpers are `cfg(test)`-gated; no production-path changes. fr2 25/25 (20x multithreaded + 5x single-threaded) and khive-mcp changed surface 5/5 x201 per the PR's test evidence. Approve.

Issue #842, opened 2026-07-11
Flaky tests (macOS CI): khive-pack-git cache refetch_clone + recovery persistent_corruption


ohdearquant:
Two khive-pack-git tests failed together on macOS CI for PR #834 (run 29135105188), whose diff touches only khive-db/khive-pack-kg entity resolution — unrelated to pack-git:

- `cache::tests::refetch_clone_updates_an_existing_slot_to_the_remote_tip` (panicked cache.rs:758)
- `recovery_tests::persistent_corruption_is_bounded_and_never_reports_false_success` (panicked recovery_tests.rs:525)

Ubuntu passed the same suite. Likely macOS-specific timing or filesystem behavior (git clone/fetch against local fixtures), or shared temp-dir state under parallel execution. Same flake-family discipline as #837/#839/#840: root-cause the shared state or platform assumption, don't just serialize. If rerun goes green this is confirmed flaky rather than broken.

ohdearquant:
Another instance of the khive-pack-git shared-state flake class, on PR #823 macOS CI (run 29135672404): recovery_tests::corrupt_promisor_cache_self_heals_via_refetch_on_first_call panicked at recovery_tests.rs:393. Same fixture surface as the two originally filed here. Adding to the root-cause sweep scope.

ohdearquant:
Fourth instance: PR #830 macOS CI (run 29136627263) failed on cache::refetch_clone_updates_an_existing_slot_to_the_remote_tip — unrelated Python-only diff, confirming the shared-state class. Leo re-confirmed this belongs to the e8dca5f7 sweep.

ohdearquant:
Family-scope expansion (per Leo's ruling on the #826/#827 approval pass): the flake class is a khive-pack-git-on-macOS family, not one test. Two new instances observed on unrelated PR CI runs:

- `persistent_corruption_is_bounded_and_never_reports_false_success` (crates/khive-pack-git/tests/recovery_tests.rs:525) — failed on PR #826's macOS job
- `public_verb_partial_side_effects_survive_commit_snapshot_recovery` (crates/khive-pack-git/tests/recovery_tests.rs:737) — failed on PR #827's macOS job

Both folded into the sweep scope alongside the prior instances (#837, #839, #840, refetch_clone from #830's run, corrupt_promisor_cache from #823's run). Sweep must root-cause the shared-state class (serial groups / thread-local injection per the #828 discipline) and survey the pack-git fixture surface, not patch tests one at a time.

[...]

Issue #840, opened 2026-07-11
Flaky test: create_note_with_embedding_content_vector_failure_rolls_back_note_row_and_fts (khive-runtime, parallel execution)


ohdearquant:
Flakes under full-suite parallel execution, passes in isolation and on repeat runs. Confirmed pre-existing by reproducing on pre-fix HEAD via git stash on PR #832's branch. Likely shared-state coupling in the embedding/rollback path. Same flake family as #837/#839; root-cause the shared state, do not just #[serial] it.

ohdearquant:
Closed by #847 (merged). Root-cause trail for the pack-git macOS flake family:

- Historical failures were `CacheError::Io(NotFound)` inside `dir_size`/`evict_lru` walks immediately after a clone/fetch in the same test (verified against the original failed attempts of runs 29135105188, 29135672404, 29136627263, 29137613892).
- Load-bearing fix: `dir_size` tolerates a descendant vanishing mid-walk beneath an existing root (NotFound only; the walk root vanishing is a hard error, so eviction cannot succeed over a missing keep slot). Pinned by three regression tests including a Barrier-raced deletion.
- Defensive hardening: `-c maintenance.auto=false` on cache git spawns (git 2.47+ runs post-fetch auto-maintenance as a detached background child; spawn trace2-verified, suppression verified).
- Secondary: five previously unguarded real-git spawns now hold the crate ENV_MUTEX; `CacheError::Io` carries operation+path context.

Full analysis in the PR body (five review rounds, sabotage-verified).

ohdearquant:
Correction to the closure note above: this issue is not part of the pack-git cache mechanism. It was folded into the flaky-test sweep family (gtd lane, one-night ladder surfacing); it has not recurred across the many whole-workspace CI runs of the 9-PR ladder plus the #846/#847 verification runs since it was filed. Closing with the sweep; if the rollback test flakes again, reopen and it gets its own root-cause pass under the #828 determinism discipline.

Issue #839, opened 2026-07-11
Flaky test: fire_pending_self_heal_is_a_no_op_when_nothing_is_armed (khive-mcp daemon global state under parallel execution)


ohdearquant:
Fails under parallel full-suite execution (expected 0, observed 1), passes in isolation. Reproduced twice by an independent review leg on PR #823's branch at 9193335d; the test is untouched by that diff. Same process-global-state flake class as #837/#828. Root-cause the shared daemon global rather than serializing, per the #828 discipline.

ohdearquant:
Closed by #847 (merged). Root-cause trail for the pack-git macOS flake family:

- Historical failures were `CacheError::Io(NotFound)` inside `dir_size`/`evict_lru` walks immediately after a clone/fetch in the same test (verified against the original failed attempts of runs 29135105188, 29135672404, 29136627263, 29137613892).
- Load-bearing fix: `dir_size` tolerates a descendant vanishing mid-walk beneath an existing root (NotFound only; the walk root vanishing is a hard error, so eviction cannot succeed over a missing keep slot). Pinned by three regression tests including a Barrier-raced deletion.
- Defensive hardening: `-c maintenance.auto=false` on cache git spawns (git 2.47+ runs post-fetch auto-maintenance as a detached background child; spawn trace2-verified, suppression verified).
- Secondary: five previously unguarded real-git spawns now hold the crate ENV_MUTEX; `CacheError::Io` carries operation+path context.

Full analysis in the PR body (five review rounds, sabotage-verified).

ohdearquant:
Correction to the closure note above: this issue is not part of the pack-git cache mechanism. It belongs to the daemon self-heal shared-state class fixed in #846 — `reset_self_heal_counters()` cleared the atomics but not `PENDING_SELF_HEAL`, so a pending self-heal armed by one test leaked into the no-op assertion of another under parallel execution. Closed as part of the same flaky-test sweep (with #842/#843/#844).

Issue #837, opened 2026-07-11
Flaky test: fr2_kg_merge_invalidates_warm_ann_without_subsequent_remember (shared warm-ANN process-global state)


ohdearquant:
## Symptom

\`fr2_kg_merge_invalidates_warm_ann_without_subsequent_remember\` (crates/khive-pack-memory/src/pack.rs:985) failed on PR #828's CI run even though that test is not in #828's diff, then is expected to pass on rerun. Fourth racy test in the same class as the three fixed by #828.

## Suspected class

Same shared-warm-ANN / process-global state family as the #828 sweep: tests sharing warm-ANN generation state (or another process-global) race when run in the same test binary alongside concurrent memory-pack tests. Note the warm-ANN generation counter is already known to be process-local (#752), which may be the underlying coupling.

## Ask

Root-cause with the same discipline as #828 (identify the exact shared state, fix the isolation, do not just serialize with #[serial] unless the shared state is inherent). Check whether #752 resolves it structurally.

## Evidence

- PR #828 CI run: fr2 failure on a diff that does not touch khive-pack-memory.
- Reported by Leo during #828 review, 2026-07-11.

ohdearquant:
Closed by #847 (merged). Root-cause trail for the pack-git macOS flake family:

- Historical failures were `CacheError::Io(NotFound)` inside `dir_size`/`evict_lru` walks immediately after a clone/fetch in the same test (verified against the original failed attempts of runs 29135105188, 29135672404, 29136627263, 29137613892).
- Load-bearing fix: `dir_size` tolerates a descendant vanishing mid-walk beneath an existing root (NotFound only; the walk root vanishing is a hard error, so eviction cannot succeed over a missing keep slot). Pinned by three regression tests including a Barrier-raced deletion.
- Defensive hardening: `-c maintenance.auto=false` on cache git spawns (git 2.47+ runs post-fetch auto-maintenance as a detached background child; spawn trace2-verified, suppression verified).
- Secondary: five previously unguarded real-git spawns now hold the crate ENV_MUTEX; `CacheError::Io` carries operation+path context.

Full analysis in the PR body (five review rounds, sabotage-verified).

ohdearquant:
[...]

PR #847, opened 2026-07-11
fix(pack-git): close scratch-cache ENOENT race behind macOS flake family (#842)

ohdearquant:
## Rounds 3-5 (current) — codex round-2 REQUEST CHANGES: two Majors fixed in round 3; rounds 4-5 scope the evidence claims to exactly what was observed

Codex round-2 reviewed 98e3dd45 with git trace2 instrumentation and found the
round-2 fix incomplete on two points. Both are fixed in this round; see
"Verification" for the trace2 captures and sabotage test.

### Major 1 — `gc.auto=0` does not suppress the detached maintenance child

Round-2's root-cause claim above ("git's default `gc.autoDetach=true` ...
`git gc --auto` ... forks and detaches ... continuing to repack/prune loose
objects") **overclaimed the mechanism** and codex correctly called it
unproven: this crate's cache-slot test fixtures are tiny and sit well below
git's default gc thresholds, so "auto-gc pruned loose objects" was never
actually demonstrated as the historical mutator.

What **is** trace2-proven (scoped precisely in rounds 4-5, per codex): `git
fetch` runs auto-maintenance when `maintenance.auto` (default `true`) is
set, and since **git 2.47** that maintenance runs as a detached background
child (`git maintenance run --auto --no-quiet --detach`) that can outlive
the foreground command (on 2.46 and earlier it ran synchronously). The
trace evidence covers the **`fetch --refetch` path only** — the same trace
showed `clone` spawning no maintenance child, so the flag on the clone
builder is harmless defensive configuration with no trace evidence claimed.
I traced a real `git fetch --refetch` against a local origin with
`GIT_TRACE2_EVENT` (git 2.49):

- **With only `-c gc.auto=0`** (round-2's fix): the detached child still
  spawns —
  `"argv":["git","maintenance","run","--auto","--no-quiet","--detach"]`
  appears in the trace2 `child_start` events.
- **With `-c gc.auto=0 -c maintenance.auto=false`** (this round's fix): no
  `maintenance` child appears anywhere in the trace2 log for the same fetch —
  only the expected `git-upload-pack`/`pack-objects`/`index-pack` children the
  fetch itself needs.

**Scope of what the trace2 evidence proves (round-4 correction, per codex
round-3):** the traces prove the detached child *spawns* with default config
and is suppressed by `maintenance.auto=false`. They do not prove the child
executed a maintenance task that mutated `.git` in the historical CI
failures — in small repos the child typically finds no task to run and
exits quickly. So `maintenance.auto=false` is **defensive hardening**: when
one of the child's auto-maintenance tasks does fire it mutates the slot's
`.git` tree (commit-graph writes, pack maintenance, lock files) concurrently
with our walks, and this flag removes the one background mutator git itself
forks into our cache slots. The **load-bearing fix** for the observed ENOENT
flake family is the `dir_size` descendant-vanish tolerance (Major 2 section
below), which handles any concurrent mutator of the walked tree — git's
child, or our own racing `evict_lru`/`ensure_clone` repair. Added to every
git invocation builder in `cache.rs` (`clone`, `fetch`, `fetch_refetch`).
`gc.auto=0` is kept alongside it — a separate, harmless knob (disables
`git gc --auto`'s own opportunistic-gc check) — but it does not suppress the
child; the doc comment on `clone` in `cache.rs` carries this same honest
framing.

**Does disabling auto maintenance mean cache repos never gc?** No — growth is
bounded by wholesale eviction, not in-place compaction, so there's nothing
useful for maintenance/gc to do in a cache slot regardless:
`ensure_clone`/`refetch_clone` (`cache.rs`) delete a slot outright via
`remove_owned_entry` the moment `dir_size` measures it over
`digest_cache_clone_max_bytes` right after a fetch — a slot can be fetched
into repeatedly, but it can never accumulate objects past its own size cap
without being deleted and re-cloned fresh. Independently, `evict_lru`
(`cache.rs`) deletes whole least-recently-used slot directories once the
cache-wide `digest_cache_max_repos`/`digest_cache_max_bytes` caps are
exceeded. Both are whole-directory removals, never in-place object pruning.

### Major 2 — `dir_size` laundered a vanished walk **root** into `Ok(0)`

`dir_size` treated `NotFound` on the exact path it was asked to size
identically to `NotFound` on a descendant found while walking — both were
silently tolerated and contributed `0` bytes. That's correct for a
mid-walk descendant (the intended fix from round 2) but wrong for the walk
root itself: a caller sizing `keep` (`evict_lru`) or a slot it just
fetched/cloned into (`ensure_clone`, `refetch_clone`) got a false `Ok(0)`
instead of an error when that root had actually vanished — `evict_lru` could
then report success over a missing `keep` slot, or record a
concurrently-evicted candidate as a phantom `0`-byte entry instead of simply
having nothing to weigh.

Fixed: `dir_size` now distinguishes the two cases by comparing the popped
path to the function's own `path` argument. `NotFound` on that exact root
path is a hard error (`CacheError::Io`); `NotFound` on anything discovered
while walking beneath an existing root is still tolerated, unchanged from
round 2. Audited every call site for which semantic it needs:

- `ensure_clone`, `refetch_clone` (`cache.rs`): already propagate via `?` —
  correct, since the root just finished being fetched/cloned into, so a
  vanish is a real problem to surface, not a maybe-absent slot.
- `install_fresh_clone`'s staging-dir sizing: already propagates via `?` with
  its existing cleanup `inspect_err` — a private, freshly-created staging
  directory vanishing is a genuine bug.
- `evict_lru`'s `keep` argument: propagates via `?` (unchanged code) — every
  caller touches or freshly installs `keep` immediately before calling
  `evict_lru`, so a vanish is a real problem, not an expected race.
- `evict_lru`'s per-candidate `dir_size(&p)` call: this is the one site that
  legitimately needs tolerance — another `evict_lru`/`ensure_clone` repairing
  the same root can delete a listed candidate between the `read_dir` listing
  and this call. Now explicitly matches `CacheError::Io(NotFound)` and skips
  the entry instead of aborting the whole eviction pass.

New regression tests pin both halves of the contract:
`dir_size_errors_when_the_root_itself_is_missing` (deterministic — root never
existed), `dir_size_errors_when_the_root_is_removed_mid_walk` (a genuine
Barrier-raced two-thread deletion, mirroring the existing tolerant-descendant
test but against an *empty* root so a single `rmdir` can race the walker's
first two syscalls — a populated root's own directory entry is removed last
by `remove_dir_all`, well after the walker), and
`evict_lru_errors_when_keep_itself_is_missing`.

## Round 2 root cause (superseded above) / fix

Every historical macOS failure in this flake family is a **scratch-cache I/O
error, `ENOENT`** — e.g.
`cache::tests::refetch_clone_updates_an_existing_slot_to_the_remote_tip`
(`cache.rs:742`) failing inside `refetch_clone` with
`CacheError::Io(No such file or directory)`. That test already holds
`cache::ENV_MUTEX`, sets its **own** unique `tempfile::tempdir()` scratch
root, and installs no `PATH` shim — the round-1 fix below does not touch it
at all, and could not have caused or cured this failure.

I audited every place this crate touches the scratch cache filesystem,
specifically checking the fleet lead's cross-process-interference hypothesis:

- **Every** test that overrides `KHIVE_GIT_DIGEST_SCRATCH_ROOT` (in
  `cache.rs` and `recovery_tests.rs`, both in the lib test binary) sets it to
  its own freshly created `tempfile::tempdir()` and holds `ENV_MUTEX` for the
  whole test. No two tests share a root, and no test leaks the override past
  its own scope (paired `set_var`/`remove_var` inside every guarded test).
- `tests/acceptance.rs` (a **separate test binary/process**) never
  constructs a `DigestSource::Remote` — every `git.digest` call there passes
  a local filesystem path as `source`. `GitPack::handle_digest` only calls
  into `cache::{ensure_clone,refetch_clone,reclone}` for a `Remote` source
  (`handlers.rs:144-160`); a local source never touches `cache.rs` at all.
  So the acceptance binary cannot be racing the lib binary's scratch cache —
  there's no shared filesystem location between them to race on, because one
  side never visits it.

That rules out cross-test and cross-binary path sharing as the mechanism.
What's proven now (see "Major 1" above for the precise evidence scope): git
2.47+ *spawns* a detached maintenance child after `fetch` — a background
process that, when one of its tasks fires, mutates the same `.git` tree our
`dir_size`/`evict_lru` walks traverse. Whether such a task actually fired in
the historical failures is not proven (spawn-only evidence); it is **not**,
as originally (and wrongly) claimed, loose-object pruning by `gc --auto`,
which the tiny test fixtures here never grow large enough to trigger. The
proven, load-bearing fix for the observed ENOENT class is the `dir_size`
descendant-vanish tolerance, which covers any concurrent mutator of the
walked tree.

### Fix applied in round 2 (still correct, extended in round 3 above)

- `dir_size`, and `evict_lru`'s directory listing, tolerate an entry
  vanishing **mid-walk beneath an existing root**: a `NotFound` there shrinks
  the running total or skips the listing entry instead of aborting the whole
  call — unchanged by round 3, which only tightens what counts as "mid-walk"
  versus "the root itself" (see Major 2 above).
- `CacheError::Io` now carries the operation and path it happened on at the
  three sites this was asked for (`touch`, `dir_size`, `evict_lru`) instead
  of an opaque `io::Error` — a future `ENOENT` here will name what vanished
  and which function was walking it.

## Round 1 (kept — internally correct, but not the cause of the ENOENT family)

The original PR round closed a **different, real** gap: two tests in
`ingest.rs`'s `recovery_classifier_tests` and three tests in
`tests/acceptance.rs` spawned real `git`/dispatched `git.digest` without
holding `cache::ENV_MUTEX`/the acceptance binary's own `ENV_MUTEX`, so they
could transiently run a concurrently-installed `PATH` shim or race a sibling
test's `KHIVE_TEST_GIT_FAIL_*` env vars. That fix (guarding all five) is
correct and stays — every test that spawns `git`/`gh` now holds the
appropriate mutex before doing so.

## Verification (round 3)

- **trace2, both directions (Major 1)**: real `git fetch --refetch` against a
  local origin, captured via `GIT_TRACE2_EVENT`.
  - `-c gc.auto=0` alone: `child_start` events include
    `["git","maintenance","run","--auto","--no-quiet","--detach"]`.
  - `-c gc.auto=0 -c maintenance.auto=false`: no `maintenance` child anywhere
    in the log for the same fetch.
- **Sabotage, Major 2**: reverted `dir_size`'s root/descendant distinction to
  round-2's uniform tolerance, ran the three new pinned tests —
  `dir_size_errors_when_the_root_itself_is_missing`,
  `dir_size_errors_when_the_root_is_removed_mid_walk`, and
  `evict_lru_errors_when_keep_itself_is_missing` all failed as expected
  (the first two asserting `0` was returned instead of an error; the barrier
  test additionally reported it never observed the error across its loop).
  Restored the fix; `git diff` confirmed clean before committing.
- `cargo test -p khive-pack-git`: **15/15** whole-crate green runs,
  multi-threaded, foreground batches of 5 — exact counts every time: 74 lib
  tests (71 prior + 3 new: the two `dir_size` root-vanish tests and
  `evict_lru_errors_when_keep_itself_is_missing`) + 41 acceptance tests, 0
  failures.
- `cargo clippy -p khive-pack-git --all-targets -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- Two `recovery_tests.rs` assertions hardcoded the exact `git` argv string
  logged by the `PATH` shim (`"-c core.hooksPath=/dev/null -c gc.auto=0 clone
  "`) to count clone/reclone invocations; this round's new `-c
  maintenance.auto=false` flag shifts that string, so both were updated to
  match the new argv shape and re-verified green.

## Verification (round 2, retained)

- **New regression test**
  (`cache::tests::dir_size_tolerates_a_subdirectory_removed_mid_walk`):
  races a real concurrent `std::fs::remove_dir_all` against `dir_size` via a
  `std::sync::Barrier` (not a sleep), with a wide fan of sibling directories
  to widen the window, looped 200x per run so the race is virtually certain
  to land at least once.
- Honest framing on reproducibility: this is a genuine two-thread race, not
  a single-shot deterministic repro — the exact interleaving isn't
  guaranteed identical every run. What is deterministic and repeatable:
  `dir_size` and `evict_lru` walking a tree that changes under them mid-walk
  hard-fails before the fix and tolerates it after, on every run.

Refs #842 #837 #839 #840

### DIFF
--- a/crates/khive-pack-git/src/cache.rs
+++ b/crates/khive-pack-git/src/cache.rs
@@ -345,23 +345,28 @@ fn remove_dir_all_retrying(path: &Path) -> std::io::Result<()> {
 }
 
 /// `-c maintenance.auto=false` on every clone/fetch into a cache slot, as
-/// defensive hardening: since git 2.30, `fetch`/`clone` unconditionally
-/// check `maintenance.auto` (default true) after they finish and, if set,
-/// launch `git maintenance run --auto --detach` as a detached background
-/// child. The child spawn is trace2-proven in both directions
-/// (`GIT_TRACE2_EVENT` on a live `git fetch --refetch`: with default config
-/// the child forks; with `maintenance.auto=false` it does not). When one of
-/// its auto-maintenance tasks fires, that child mutates the slot's `.git`
-/// tree (commit-graph writes, pack maintenance, lock files) concurrently
-/// with any `dir_size`/`evict_lru` walk of the same slot. Whether such a
-/// task actually fired in issue #842's historical macOS ENOENT failures is
-/// not proven -- in small repos the child typically finds no task to run
-/// and exits quickly -- so the load-bearing fix for that flake family is
-/// the descendant-vanish tolerance in `dir_size`; this flag removes the one
-/// background mutator git itself forks into our cache slots. `gc.auto=0`
-/// alone does **not** suppress the child (trace2-verified); it is kept
-/// alongside because it disables `git gc --auto`'s separate
-/// opportunistic-gc check, harmless to also turn off here.
+/// defensive hardening. `git fetch` runs auto-maintenance after it
+/// finishes when `maintenance.auto` (default true) is set, and since git
+/// 2.47 that maintenance runs as a *detached background child*
+/// (`git maintenance run --auto --detach`) that can outlive the foreground
+/// command; on 2.46 and earlier it ran synchronously. The spawn is
+/// trace2-proven in both directions on the `fetch --refetch` path
+/// (`GIT_TRACE2_EVENT`, git 2.49: with default config the child forks; with
+/// `maintenance.auto=false` it does not). The same trace showed `clone`
+/// spawning no maintenance child; the flag is applied to the clone builder
+/// too purely as harmless defensive configuration, with no trace evidence
+/// claimed for that path. When one of the detached child's tasks fires it
+/// mutates the slot's `.git` tree (commit-graph writes, pack maintenance,
+/// lock files) concurrently with any `dir_size`/`evict_lru` walk of the
+/// same slot. Whether such a task actually fired in issue #842's historical
+/// macOS ENOENT failures is not proven -- in small repos the child
+/// typically finds no task to run and exits quickly -- so the load-bearing
+/// fix for that flake family is the descendant-vanish tolerance in
+/// `dir_size`; this flag removes the one background mutator git itself can
+/// fork into our cache slots. `gc.auto=0` alone does **not** suppress the
+/// child (trace2-verified); it is kept alongside because it disables
+/// `git gc --auto`'s separate opportunistic-gc check, harmless to also turn
+/// off here.
 ///
 /// This does not mean a cache slot is naturally garbage-collected some other
 /// way instead: no cache-slot repo is ever gc'd or maintenance'd by us.
--- a/crates/khive-pack-git/src/cache.rs
+++ b/crates/khive-pack-git/src/cache.rs
@@ -336,10 +336,20 @@ fn remove_dir_all_retrying(path: &Path) -> std::io::Result<()> {
     Err(last_err.expect("loop always sets last_err before exiting"))
 }
 
+/// `gc.auto=0` on every clone/fetch into a cache slot: git's default
+/// `gc.autoDetach=true` lets an opportunistic `gc --auto` fork and keep
+/// running in the background *after* this command already returned success
+/// -- exactly the kind of concurrent mutator that can delete a loose object
+/// `dir_size`/`evict_lru` are mid-walk over in the same slot (issue #842's
+/// macOS-only ENOENT flake family). No cache-slot repo is ever gc'd by us;
+/// eviction deletes the whole directory instead, so there is nothing for
+/// git's own gc to usefully do here.
 fn clone(url: &str, dest: &Path) -> Result<(), CacheError> {
     let status = Command::new("git")
         .arg("-c")
         .arg("core.hooksPath=/dev/null")
+        .arg("-c")
+        .arg("gc.auto=0")
         .arg("clone")
         .arg("--filter=blob:none")
         .arg(url)
@@ -359,6 +369,8 @@ fn fetch(repo: &Path) -> Result<(), CacheError> {
     let status = Command::new("git")
         .arg("-c")
         .arg("core.hooksPath=/dev/null")
+        .arg("-c")
+        .arg("gc.auto=0")
         .arg("-C")
         .arg(repo)
         .arg("fetch")
@@ -383,6 +395,8 @@ fn fetch_refetch(repo: &Path) -> Result<(), CacheError> {
     let status = Command::new("git")
         .arg("-c")
         .arg("core.hooksPath=/dev/null")
+        .arg("-c")
+        .arg("gc.auto=0")
         .arg("-C")
         .arg(repo)
         .arg("fetch")
@@ -400,8 +414,20 @@ fn fetch_refetch(repo: &Path) -> Result<(), CacheError> {
     Ok(())
 }
 
+/// Wrap an I/O error with the operation and path it happened on -- a bare
+/// `CacheError::Io(e)` at these call sites used to surface as an opaque
+/// "No such file or directory" with no way to tell which of the many paths
+/// `dir_size`/`touch`/`evict_lru` touch actually disappeared.
+fn io_err(op: &str, path: &Path, e: std::io::Error) -> CacheError {
+    CacheError::Io(std::io::Error::new(
+        e.kind(),
+        format!("{op} {}: {e}", path.display()),
+    ))
+}
+
 fn touch(repo_dir: &Path) -> Result<(), CacheError> {
-    std::fs::write(repo_dir.join(MARKER_FILE), b"")?;
+    let marker = repo_dir.join(MARKER_FILE);
+    std::fs::write(&marker, b"").map_err(|e| io_err("touch: write marker", &marker, e))?;
     Ok(())
 }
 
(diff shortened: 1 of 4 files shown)


oceanwaves630 (approved):
Reviewed the full diff at b215b263a8ad57d5bcb40ce691e1445d93ddc610.

The dir_size contract is the load-bearing fix and it is precisely scoped: the walk root vanishing surfaces as CacheError::Io(NotFound) while only descendants vanishing beneath a still-existing root are tolerated, and the tolerance is NotFound-only at all three walk sites — permission and I/O errors still propagate, now with operation+path context via io_err. The caller audit is visible in the diff itself: ensure_clone/refetch_clone propagate (the root was just fetched into), and evict_lru distinguishes keep (propagates — every caller just touched or installed it, so a vanish is a real problem) from a listed candidate deleted by a concurrent repair on the same root (skipped — nothing left to weigh). The pre-fix harm paths (eviction reporting success over a missing keep slot, phantom zero-byte candidates) are each pinned by a dedicated regression test, and both race tests document their harness honestly, including why the root-vanish variant needs an empty root to make the race reachable.

The maintenance rationale comment claims exactly what the evidence shows: detached auto-maintenance since git 2.47 with 2.46-synchronous noted, trace2 evidence scoped to the fetch --refetch path on 2.49, the clone flag stated as defensive configuration with no trace evidence claimed, and the historical task-fire explicitly left unproven with the dir_size tolerance named as the load-bearing fix. The growth-bounds paragraph correctly grounds why cache slots never need git's own maintenance: over-cap slots are deleted whole and re-cloned, never compacted in place.

The five previously unguarded real-git spawn sites (two recovery_classifier tests, three acceptance digest_verb tests) now hold the crate-wide ENV_MUTEX, closing the secondary shim race, and the recovery_tests argv assertions match the new builder flags exactly. Approve.

PR #848, opened 2026-07-11
feat(code-pack): kkernel code-ingest admin path + default load (ADR-085 Amendment 3)

ohdearquant:
## What

Wires up the code pack's existing `ingest_findings_json` mapper (previously only reachable by
linking `khive-pack-code` into a caller's own binary) so a `findings.json` audit sweep can
actually reach the graph, and turns on the pack by default.

1. **`kkernel code-ingest <findings.json>`** (`crates/kkernel/src/code_ingest.rs`): a new admin
   CLI subcommand, following the same shape as `kkernel git-ingest`. The whole document is
   parsed and validated fail-closed BEFORE any runtime or database construction, so an invalid
   document (or a `--dry-run`) has zero filesystem effect: no database file is created, no
   migrations run, and an existing database is left byte-identical. Every entity/note field,
   including nested property values, is preflighted through the runtime secret gate before the
   first write (explicit content-derived UUIDv5 ids are required for idempotent re-ingest and
   cannot flow through the shared `create` dispatch path, which accepts no caller-supplied id).
   Records are keyed by content-derived id, so re-running the same sweep reports them as skipped
   rather than duplicating or overwriting a finding's curated lifecycle state. `--dry-run`
   validates and reports would-create counts; against an existing database it opens a read-only
   connection (no migrations, `query_only` pragma). `--source-run`, `--db`, and `--namespace`
   mirror the existing admin ingesters. The command fails loud if the configured pack set does
   not include `code`.
2. **Producer status normalization** (`crates/khive-pack-code/src/ingest.rs`): producer
   `fixed` maps to `kind_status=resolved` and `false_positive` to `invalid` per the ADR-085
   governed lifecycle; the raw producer value is preserved in `audit_status`.
3. **`code` joins the default pack set** (`crates/khive-runtime/src/config.rs`), and
   `khive-mcp` now links `khive-pack-code` (dependency + force-link for inventory
   registration) so the pack actually resolves in the server binary; without that link the MCP
   server fails to boot under the new default config. The `finding` note kind and the pack's 22
   additive edge rules are now live wherever khive-mcp or `kkernel` run without an explicit
   `--pack`/`KHIVE_PACKS` override. The pack currently contributes zero MCP verbs, so the total
   verb count is unchanged at 78.
4. **Doc updates** across the default-pack surfaces (`CLAUDE.md`, `AGENTS.md`, READMEs, guides,
   config example, perf scripts, CLI help), and an ADR-085 amendment recording this as a
   deliberate surface change, with the no-verb statements scoped to the findings surface.

## Why

The pack's `finding` note kind and fail-closed `ingest_findings_json` validator already existed
in source (ADR-085), but had no way to reach a real database short of writing a custom binary,
and the pack wasn't loaded by default. A durable audit pipeline that produces `findings.json`
sweeps needs those findings to land as queryable graph records rather than staying flat local
files. This closes that gap with an admin-only path: no MCP verb is added, so agents never hold
a bulk-ingest capability; only an operator or a runner invoking the CLI writes findings into the
graph.

## Surface change

Turning on the `code` pack by default makes the `finding` note kind and its edge rules live on
the production surface. This is called out explicitly in the ADR-085 amendment as a deliberate
change, not an incidental one. The pack stays verb-less on the findings surface (no new MCP
surface); ingestion stays an admin/runner-side path. The ADR's accepted-but-unimplemented
source-ingest verb design is marked deferred and is unaffected.

## Verification

- `cargo test --workspace`: all green, zero failures. This run is what surfaced the missing
  `khive-pack-code` link in the `khive-mcp` binary (27 server tests failed with
  `unknown pack name "code"` before the link landed): a boot failure that per-crate test runs
  cannot catch.
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all -- --check`: clean.
- Black-box CLI tests (`crates/kkernel/tests/code_ingest_cli.rs`) against the compiled binary:
  fresh `--dry-run` leaves a nonexistent `--db` path nonexistent (exit 0); an invalid document
  against a nonexistent path creates nothing (exit 1); an invalid document against an existing
  database leaves it byte-identical; a normal ingest creates the expected entity/note/edge plus
  FTS and vector rows (vector metadata labeled with the canonical `entity.body` field); and
  `--db`/`--namespace`/`--source-run` all reach storage.
- Functional proof against a scratch database (never the production db): valid ingest, idempotent
  re-run (all skipped), dry-run against fresh and existing databases (no mutation either way),
  invalid document rejected pre-write, a finding whose evidence embeds a fake AWS key rejected by
  the secret gate before any write, and the code-pack-absent guard failing loud.

## Test plan

- [x] `cargo test --workspace` (zero failures)
- [x] Black-box CLI tests against the compiled `kkernel` binary
- [x] Manual functional proof against a scratch database (create, idempotent re-run, dry-run
      no-mutation, invalid input, secret-gate rejection, pack-absent guard)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,9 +17,10 @@ khive gives your agent:
 
 All 10 packs load by default. **78 public verbs** across the packs: the `git` pack
 contributes the `git.digest` verb plus the commit/issue/pull_request provenance note kinds
-and a batch ingester; the `code` pack contributes zero verbs by design, since its `finding`
+and a batch ingester; the `code` pack currently contributes zero verbs: its `finding`
 note kind is written only through the `kkernel code-ingest` admin CLI path, never a verb
-an agent can call (ADR-085 D1, Amendment 3). Regenerate via `request(ops="verbs()")`
+an agent can call (ADR-085 D1, Amendment 3; Amendment 2's `code.ingest` verb is accepted
+but unimplemented). Regenerate via `request(ops="verbs()")`
 before editing this line.
 
 If you're working on khive itself (writing code in this repo), see `CLAUDE.md` instead.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,9 +15,12 @@ khive gives your agent:
 9. **Brain** — Bayesian profile tuning from feedback signals
 10. **Session** — persist and resume agent-session records
 
-All 9 packs load by default. **78 public verbs** across the packs — the `git` pack
+All 10 packs load by default. **78 public verbs** across the packs: the `git` pack
 contributes the `git.digest` verb plus the commit/issue/pull_request provenance note kinds
-and a batch ingester (regenerate via `request(ops="verbs()")` before editing this line).
+and a batch ingester; the `code` pack contributes zero verbs by design, since its `finding`
+note kind is written only through the `kkernel code-ingest` admin CLI path, never a verb
+an agent can call (ADR-085 D1, Amendment 3). Regenerate via `request(ops="verbs()")`
+before editing this line.
 
 If you're working on khive itself (writing code in this repo), see `CLAUDE.md` instead.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -167,9 +167,12 @@ request(ops="[{\"tool\":\"v1\",\"args\":{...}}, ...]")
 ```
 
 Verbs come from whichever packs are loaded via `KHIVE_PACKS` (env) or `--pack` (CLI). Default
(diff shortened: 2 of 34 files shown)


oceanwaves630 (approved):
Gate read at 67210e24: snapshot-copy dry-run path verified (db and -wal copied together, TempDir lifetime held by named binding, target path never opened); both WAL regressions assert byte-identity and presence over the full db/-wal/-shm set at the target path, with the binary-boundary variant additionally proving no new files via directory-listing equality; status-mapping plumbing test and doc-surface updates accurate against the live pack list. Full review chain converged cleanly.

PR #851, opened 2026-07-11
fix(bench): stop libtest harness from rejecting criterion --quick in bench-track shards

ohdearquant:
## Problem

Bench Trend Tracking has failed on every main push since it shipped (all 4 component shards, first run 29141259783 through 29157051310). Every shard reports:

```
[bench_track] error: no criterion estimates.json found under crates/target/criterion - did `cargo bench` run?
```

## Root cause

The workflow runs `cargo bench -p <crates> -- --quick`. Cargo forwards trailing bench args to every benchmark-capable target, and each crate's library still has the default libtest bench harness enabled, so the very first target invoked is `unittests src/lib.rs`, which rejects the flag and aborts the whole invocation before any Criterion bench runs:

```
Running unittests src/lib.rs (target/release/deps/khive_bm25-...)
error: Unrecognized option: 'quick'
error: bench failed, to rerun pass `-p khive-bm25 --lib`
```

The `|| true` in the workflow swallows the abort, so the failure only surfaces later as the missing-estimates ledger error.

## Fix

Set `[lib] bench = false` on the 20 crates in the bench-track shard matrix (the standard Criterion workspace pattern), so `cargo bench` invokes only the Criterion `[[bench]]` targets, which understand `--quick`.

## Verification

Before (main, unpatched):

```
$ cargo bench -p khive-bm25 -- --quick
     Running unittests src/lib.rs (target/release/deps/khive_bm25-040223458872e244)
error: Unrecognized option: 'quick'
error: bench failed, to rerun pass `-p khive-bm25 --lib`
```

After (this branch):

```
$ cargo bench -p khive-bm25 -- --quick
Benchmarking memory_usage/docs/1000: Analyzing
memory_usage/docs/1000  time:   [49.041 µs 49.176 µs 49.717 µs]
remove_document/1k_corpus
                        time:   [2.1738 ms 2.1859 ms 2.2345 ms]
```

and `target/criterion/` now contains per-bench estimate directories, which is exactly what `bench_track.py record` consumes.

Metadata-only change: 20 files, 60 insertions, no code or dependency changes.

### DIFF
--- a/crates/khive-bm25/Cargo.toml
+++ b/crates/khive-bm25/Cargo.toml
@@ -10,6 +10,9 @@ keywords.workspace = true
 categories.workspace = true
 description = "BM25 (Okapi BM25) keyword index with deterministic scoring"
 
+[lib]
+bench = false
+
 [dependencies]
 khive-score = { version = "0.3.0", path = "../khive-score" }
 serde = { workspace = true }
--- a/crates/khive-brain-core/Cargo.toml
+++ b/crates/khive-brain-core/Cargo.toml
@@ -10,6 +10,9 @@ keywords.workspace = true
 categories.workspace = true
 description = "Brain primitives — Beta posteriors, section types, profile state, weight derivation"
 
+[lib]
+bench = false
+
 [dependencies]
 khive-runtime = { version = "0.3.0", path = "../khive-runtime" }
 serde = { workspace = true }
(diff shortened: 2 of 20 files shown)


oceanwaves630 (approved):
Verified: pure manifest change, [lib] bench = false across 19 crates so criterion flags reach the intended bench targets instead of the default libtest harness. No code paths touched.

Issue #849, opened 2026-07-11
kg: no working exact-name entity lookup (GQL name equality returns empty; resolve() misses existing names)


ohdearquant:
## Problem

There is no reliable primitive for "give me the entity whose name is exactly X". Reported by a heavy agent consumer as a dedup-check gap: before minting an entity, an agent wants to check whether one with that exact name already exists. Today every path fails or is unreliable:

1. **GQL equality on name returns empty.** `MATCH (e:entity) WHERE e.name="LoRA" RETURN e.id, e.name` returns `rows: []` even though entities with that exact name exist. `CONTAINS` and `IN` forms are parse errors.
2. **`resolve()` misses existing exact names.** `resolve(refs=["RoLoRA"], kind="entity")` returns `not_found`, while `search(kind="entity", query="RoLoRA")` returns the RoLoRA entity in the same session. If resolve's hybrid fallback is intended as the exact-ref primitive, it is not doing exact-name matching before falling through.
3. **`search()` is FTS-ranked**, so an exact-name target can be buried under fuzzy neighbors; callers cannot distinguish "exists with this exact name" from "similar things exist".

## Repro (live server, v0.3.0)

```
query(query="MATCH (e:entity) WHERE e.name=\"LoRA\" RETURN e.id, e.name")  -> {"rows":[]}
resolve(refs=["RoLoRA"], kind="entity")                                    -> not_found
search(kind="entity", query="RoLoRA", limit=3)                             -> RoLoRA present in hits
```

## Ask

Two candidate shapes (either resolves the friction):

- Fix GQL `WHERE e.name = "..."` equality so it compiles to an indexed exact match on `entities.name`, or
- Make `resolve()` perform a case-sensitive exact-name lookup step (after UUID/slug passthrough, before hybrid fallback) and return it as high-confidence `Resolved`.

The dedup-check use case wants a cheap, deterministic yes/no plus id. Search-then-filter client-side works but is the workaround, not the primitive.

## Impact

Duplicate-mint prevention for every agent writer; `search`-before-`create` discipline (KG protocol) currently cannot be made exact.

PR #852, opened 2026-07-11
fix(kg): exact-name entity lookup in resolve before hybrid fallback (#849)

ohdearquant:
## Summary

`resolve(refs=["RoLoRA"], kind="entity")` returned `not_found` even though the `RoLoRA` entity existed and `search(kind="entity", query="RoLoRA")` found it in the same session. Fixes Fork B of #849 (the GQL `WHERE e.name = "..."` equality gap from the same issue is out of scope here).

## Root cause

Two compounding bugs in the resolve pipeline:

1. **No exact-name check against storage.** `resolve`'s only "exact" matching happened against the actor's session-scoped reference ring (`crates/khive-runtime/src/reference_resolution.rs`, stage 2). Any entity that already existed but was never `create`d/`get`/`update`d/`delete`d/`merge`d/`link`ed by this actor in this session was invisible to that stage, and the pipeline fell straight to the hybrid-search fallback — which is FTS/RRF-ranked, not exact-match, and can legitimately return `NotFound` or `Ambiguous` for a name that unambiguously exists.
2. **`kind` was forwarded unnormalized.** `crates/khive-pack-kg/src/handlers/resolve.rs:46` (before this change) passed the raw `kind` string straight into the fallback's `entity_kind` filter, unlike `create`/`list`/`search`, which normalize `kind` through `resolve_kind_spec` first. Passing the substrate label `kind="entity"` (as in the reported repro) became an `EntityFilter.kinds = ["entity"]` filter at `crates/khive-runtime/src/retrieval.rs:459`. No real entity ever matches that — `entities.kind` only ever holds a granular value like `"concept"` — so the fallback's alive-set filter silently dropped every candidate before scoring ever ran.

## Fix

- Added a deterministic, case-sensitive exact-name storage lookup as a new stage in `resolve_reference`, placed after the id-string passthrough and ring stages and before the hybrid-search fallback:
  - Single exact match → `Resolved` at a fixed 0.98 confidence (above both ring bands, below the 1.0 certainty of an id-string passthrough).
  - Multiple exact matches → `Ambiguous`, listing the candidates (never a silent pick).
  - No exact match → falls through to hybrid search unchanged.
  - Scoped to the caller's namespace and `deleted_at IS NULL`, matching the rest of the by-name lookup surface (`crates/khive-pack-kg/src/handlers/common.rs::resolve_name_async`).
- Normalized `resolve`'s `kind` param through `resolve_kind_spec`, so the bare substrate label `"entity"` means "no kind filter" (consistent with `create`/`list`/`search`) and a non-entity kind (e.g. `"note"`) is rejected with a clear error instead of silently over-filtering to zero matches.

## Tests

Added to `crates/khive-pack-kg/src/dispatch.rs`:
- exact-name hit resolves at high confidence, even for an entity never referenced through this session's ring (also the literal #849 regression case: `kind="entity"`)
- two entities sharing the exact same name resolve as `Ambiguous`
- a non-exact ref still falls through to the hybrid-search fallback (existing behavior preserved)
- a soft-deleted entity is not matched by its old exact name
- a granular `kind` filter narrows two same-name entities down to one match
- a non-entity `kind` is rejected with an error

Updated one pre-existing test (`resolve_search_result_sets_never_populate_the_ring`) whose ref happened to be an entity's exact name: it now correctly resolves via the new exact-name stage instead of the (no-longer-reached) hybrid fallback; the ring-emptiness invariant it was written to check still holds.

## Test plan

- [x] `cargo test -p khive-pack-kg` — 786 passed, 0 failed
- [x] `cargo test -p khive-runtime` — passed (reference_resolution + integration suites)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean, 0 warnings

Root cause: `crates/khive-pack-kg/src/handlers/resolve.rs:46` (unnormalized `kind` forwarding) and `crates/khive-runtime/src/reference_resolution.rs` (no exact-name-against-storage stage between the ring and hybrid-search stages).

### DIFF
--- a/crates/khive-db/src/stores/entity.rs
+++ b/crates/khive-db/src/stores/entity.rs
@@ -468,6 +468,15 @@ fn build_entity_where(
         conditions.push(format!("name LIKE ?{} ESCAPE '\\'", params.len()));
     }
 
+    if let Some(ref exact) = filter.name_exact {
+        params.push(Box::new(exact.clone()));
+        // `entities.name` has no `COLLATE NOCASE` (see sql/schema.sql), so
+        // `=` is already SQLite's default case-sensitive BINARY comparison.
+        // `COLLATE BINARY` is spelled out here so this predicate stays
+        // correct even if the column's default collation ever changes.
+        conditions.push(format!("name = ?{} COLLATE BINARY", params.len()));
+    }
+
     if !filter.tags_any.is_empty() {
         let placeholders: Vec<String> = filter
             .tags_any
--- a/crates/khive-pack-kg/src/dispatch.rs
+++ b/crates/khive-pack-kg/src/dispatch.rs
@@ -1237,9 +1237,11 @@ mod tests {
     /// `create`/`get`/`update`/`delete`/`merge`/`link` on THIS registry) must
     /// not resolve via the ring's high-confidence exact-match stage. It is
     /// created directly on the runtime (bypassing dispatch, hence bypassing
-    /// admission entirely) so the only way `resolve` can find it at all is
-    /// the hybrid-search fallback (stage 3) — proven by a confidence below
-    /// the ring's fixed 0.95 exact-match / 0.7 substring-match bands.
+    /// admission entirely) so the only way `resolve` can find it via the ring
+    /// is stage 2 — proven by a confidence that never equals the ring's fixed
+    /// 0.95 exact-match / 0.7 substring-match bands (it instead comes from
+    /// the stage-3 exact-name storage lookup, #849, since the ref is this
+    /// entity's exact name).
     #[tokio::test]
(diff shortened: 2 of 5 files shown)


oceanwaves630 (approved):
APPROVE at 72f70359. Verified content-level: stage-3 exact-name lookup now issues a storage-level binary-equality predicate (name = ? COLLATE BINARY), so case-variant decoys never enter the page and created_at ordering can no longer evict the exact target; cardinality decision reads page.total (storage COUNT under the same predicate), not the LIMIT-capped page length; kind param resolves through the shared kind-spec path and rejects non-entity kinds loudly. Regression tests pin both failure modes (120 newer case-variant decoys; 11 identical names exceeding the 10-row page).

Issue #452, opened 2026-07-02
khive-gate-rego: Wrong-shaped policy output is echoed into client-visible deny reasons and audit logs [high]


ohdearquant:
Severity: **high** | Confidence: high | Priority: P1 | Categories: security, errors
Crate: `khive-gate-rego` | Finding id: `GATEREGO-AUD-001`

## Finding
Request secrets can be returned to the caller, written to structured tracing, and persisted in audit events on malformed-policy paths.

## Failure scenario
Install a malformed Rego policy that returns input.args as the decision object. Dispatch propose with changeset.entity.properties.api_key = "AKIAFAKEKEY000000000". RegoGate serializes that non-GateDecision object into decision_json, logs it, and puts it in GateDecision::Deny.reason. VerbRegistry emits/persists an AuditEvent from that decision and returns RuntimeError::PermissionDenied with the same reason before proposal.rs reaches secret_gate::check_json, exposing the secret in the client error and audit sinks.

## Evidence
- `crates/khive-gate-rego/src/gate.rs:243` — Wrong-shaped policy result is logged as raw decision_json
- `crates/khive-gate-rego/src/gate.rs:248` — Wrong-shaped policy result is embedded in generated Deny reason
- `crates/khive-runtime/src/pack.rs:911` — Runtime builds AuditEvent from the gate decision before dispatching to the pack
- `crates/khive-runtime/src/pack.rs:941` — Runtime persists the audit event when an EventStore is configured
- `crates/khive-runtime/src/pack.rs:970` — Runtime returns PermissionDenied with the gate-supplied reason
- `crates/khive-pack-kg/src/handlers/proposal.rs:95` — Proposal secret gate runs inside the handler, after the runtime gate consultation
- `crates/khive-pack-kg/tests/integration.rs:5888` — Existing tests demonstrate proposal changesets can contain secret-shaped api_key material that must be blocked

## Recommendation
Do not include raw policy output in RegoGate-generated deny reasons or warn fields. Return a generic non-GateDecision denial and, if needed, log only a redacted/truncated type summary that cannot contain caller args.

## Verification
Add a khive-gate-rego regression policy that returns input.args containing an api_key and assert the Deny reason/logs omit the value. Add a runtime dispatch regression asserting PermissionDenied and AuditEvent.deny_reason are redacted.

---
Source: automated static-analysis sweep.

ohdearquant:
[...]

PR #853, opened 2026-07-11
fix(gate-rego): mask malformed-policy output from deny reasons and audit logs (#452)

ohdearquant:
## Root cause

`RegoGate::check` embedded the raw wrong-shaped policy decision value (`decision_json`) directly into the generated `Deny` reason and into the `tracing::warn!` log. A malformed or misconfigured Rego policy that returns `input.args` (or any other caller-supplied value) as its decision object — instead of a proper `{"decision": "allow"|"deny", ...}` shape — would have that value forwarded verbatim.

The runtime dispatch path (`VerbRegistry::dispatch_with_identity` / `pack.rs`) builds `AuditEvent`s from the gate decision and returns `RuntimeError::PermissionDenied` using that same reason string before the pack handler's own `secret_gate::check_json` ever runs. So a request secret (API key, token) anywhere in the request would leak into the client-visible error and into persisted audit events.

## Fix

`crates/khive-gate-rego/src/gate.rs`: on a non-`GateDecision` policy result, the Deny reason and the tracing warn no longer include the raw policy output. They now carry only a type-shape summary (`object`/`array`/`string`/`number`/`boolean`/`null`/`unparsable`) computed from the JSON value's top-level type — never any field name or value.

## Tests

New regression test `malformed_policy_echoing_input_args_does_not_leak_secret`: a policy that returns `input.args` as its decision is evaluated against a `propose` request whose changeset contains a fake AKIA-style key (`AKIAFAKEKEY000000000`). Asserts the resulting `Deny` reason contains neither the key value nor the `api_key` field name.

```
cargo test -p khive-gate-rego   # 24 passed
cargo fmt --all -- --check      # clean
cargo clippy -p khive-gate-rego --all-targets -- -D warnings   # clean
```

## Test plan
- [x] Regression test plants a fake secret and asserts it never surfaces in the Deny reason
- [x] Existing gate-rego test suite (poisoned mutex, entrypoint validation, wrong-shape, undefined rule) still green


ohdearquant:
## Automated code review — round 4

**VERDICT: APPROVE**

Re-review at head `34edac1b`. Net diff contains only `crates/khive-gate-rego/Cargo.toml` and `crates/khive-gate-rego/src/gate.rs` — the accidentally committed build directory is gone, no summary artifacts, no em-dashes or absolute paths in added lines.

- Deserialization-error log path now emits only a fixed category and JSON shape; the caller-derived secret can no longer leak through the unknown-variant error text
- Regression coverage captures tracing output and asserts absence of both the secret string and field names
- Focused gates green at head: 25 tests plus doc tests

No findings.

oceanwaves630 (approved):
Lane A (small security fix, secret-redaction in gate logging + regression test) — codex APPROVE on-PR, net diff clean. Approving per MERGE FLOW v3.

Issue #466, opened 2026-07-02
khive-quant: Public train and encode APIs panic or create malformed vectors on invalid external input [high]


ohdearquant:
Severity: **high** | Confidence: high | Priority: P1 | Categories: errors, tests
Crate: `khive-quant` | Finding id: `QUANT-AUD-002`

## Finding
Invalid shape data can panic a library consumer or silently create encoded vectors that no longer match the codec dimensions. Those malformed vectors feed into distance computations and can produce wrong ANN acquisition scores.

## Failure scenario
Sq8Codec::train(&[vec![0.0], vec![1.0, 2.0]]) sets dims = 1 from the first row, then the second row reaches d = 1 and indexes min[1], panicking from public API input. In release builds, GsSq8Codec::encode(&[]) after training a one-dimensional codec skips the debug assertion and returns an empty code vector; is_in_distribution(&[]) returns true because the zipped iterator is empty, and l2_sq between malformed empty codes returns 0.0 instead of a dimension error.

## Evidence
- `crates/khive-quant/src/lib.rs:227` — Sq8Codec::train_flat is public and returns Self, not Result.
- `crates/khive-quant/src/lib.rs:228` — train_flat panics on dims == 0.
- `crates/khive-quant/src/lib.rs:230` — train_flat panics when flat input length is not divisible by dims.
- `crates/khive-quant/src/lib.rs:276` — Sq8Codec::train indexes min/max using every later row's enumerate index without row-length validation.
- `crates/khive-quant/src/lib.rs:303` — Sq8Codec::encode relies on debug_assert_eq for vector length.
- `crates/khive-quant/src/lib.rs:330` — Sq8Codec::encode_flat_par divides by caller-supplied dims without a zero/divisibility check.
- `crates/khive-quant/src/lib.rs:453` — GsSq8Codec::train_flat repeats the public infallible train_flat pattern.
- `crates/khive-quant/src/lib.rs:526` — GsSq8Codec::train indexes min/max using every later row's enumerate index without row-length validation.
- `crates/khive-quant/src/lib.rs:577` — GsSq8Codec::encode relies on debug_assert_eq for vector length.
- `crates/khive-quant/src/lib.rs:593` — GsSq8Codec::encode_flat_par divides by caller-supplied dims without a zero/divisibility check.

## Recommendation
[...]

PR #854, opened 2026-07-11
fix(quant): typed validation errors instead of panics on invalid train/encode shapes (#466)

ohdearquant:
## Root cause

`Sq8Codec` and `GsSq8Codec`'s public train/encode APIs panicked or produced malformed encoded vectors on invalid external input:

- `train_flat` panicked on `dims == 0` and on a flat length not divisible by `dims`.
- `train` indexed `min`/`max` by every row's enumerate index with no row-length check, so a ragged corpus (rows of different lengths) panicked on out-of-bounds indexing instead of erroring.
- `encode` relied on `debug_assert_eq` for vector-length matching, which compiles out in release builds; a mismatched-length vector silently produced a truncated or malformed code vector instead of erroring — e.g. `encode(&[])` on a trained codec returned an empty code vector that `is_in_distribution` vacuously accepted and `l2_sq` scored as `0.0`.
- `encode_flat_par` divided `vectors.len() / dims` with no zero or divisibility check.

Those malformed vectors feed into distance computations and can produce wrong ANN acquisition scores.

## Fix

Added a `QuantError` enum and `try_train_flat` / `try_train` / `try_encode` / `try_encode_flat_par` fallible variants on both codecs that validate `dims > 0`, a non-empty corpus, flat-length divisibility, rectangular (non-ragged) rows, and exact encode length before touching any index.

The existing panicking `train_flat` / `train` / `encode` / `encode_flat_par` are kept as thin wrappers over the `try_*` variants — unchanged signatures, so `khive-vamana`'s existing call sites are unaffected — but now panic with a typed, descriptive `QuantError` message instead of an out-of-bounds index panic or a debug-only assert.

## Tests

```
cargo test -p khive-quant   # 31 passed
```

Covers: empty corpus, zero dims, flat-length remainder, ragged rows (both codecs), shorter/longer encode input, `encode_flat_par` zero-dims/remainder/dims-mismatch, and a roundtrip test asserting the `try_*` and panicking APIs produce identical output on valid input.

```
cargo fmt --all -- --check
cargo clippy -p khive-quant --all-targets -- -D warnings
```
Both clean.

## Test plan
- [x] Every panic site named in the issue now has a corresponding regression test
- [x] `khive-vamana`'s existing `GsSq8Codec::train_flat` call sites compile unchanged (workspace `cargo check` green)


ohdearquant:
## Automated code review — round 4

**VERDICT: APPROVE**

Re-review at head `9c7be615`. Net diff contains only `crates/khive-quant/src/lib.rs` — the accidentally committed build directory is gone, no summary artifacts, no em-dashes or absolute paths.

- `try_encode_par` now validates every row length before parallel encoding and returns `Result<_, QuantError>`; the panicking `encode_par` remains only as the documented wrapper
- Short-row and long-row regression tests added
- Focused gates green at head

No findings.

oceanwaves630 (approved):
Lane A (small fallible-validation fix + regression tests) — codex APPROVE on-PR, net diff clean. Approving per MERGE FLOW v3.

PR #857, opened 2026-07-11
fix(query): make substrate node labels (entity/note) satisfiable in GQL/SPARQL (#849)

ohdearquant:
## Root cause

The SQL compiler emitted every node label directly as a `kind = <label>` equality (`crates/khive-query/src/compilers/sql.rs`, fixed-length path and the CTE variable-length twin). Stored `kind` values are always granular (`concept`, `task`, `observation`, ...), so any GQL or SPARQL query using a substrate label such as `MATCH (e:entity)` was unsatisfiable by construction, for every predicate.

## Fix

Substrate labels (`entity`/`note`/`event`/`edge`) now compile against the union source's `substrate_kind` column (or to no filter where the FROM source is already that substrate's table), mirroring the substrate/granular split `resolve_kind_spec` applies on the verb-dispatch surface. Granular labels keep the existing `kind = <label>` emission. Applies to both GQL and SPARQL, which share these compiler sites.

## Behavior change

Queries with substrate labels that previously returned empty results now return rows. This is the intended correction of #849.

## Tests

Compiler-level regression tests in `tests/compile_integration.rs`: substrate label satisfiability against fixture data, granular-label emission unchanged, note parity, SPARQL equivalent.

## Verification status

The implementation session was interrupted before its full test run completed; `cargo fmt` and clippy pass locally. CI on this PR is the authoritative gate, and an independent review pass follows before ready.

Co-Authored-By: Leo <noreply@khive.ai>

### DIFF
--- a/crates/khive-query/Cargo.toml
+++ b/crates/khive-query/Cargo.toml
@@ -16,6 +16,7 @@ thiserror = { workspace = true }
 
 [dev-dependencies]
 criterion = { workspace = true }
+rusqlite = { workspace = true, features = ["bundled"] }
 
 [[bench]]
 name = "parse_bench"
--- a/crates/khive-query/src/compilers/sql.rs
+++ b/crates/khive-query/src/compilers/sql.rs
@@ -216,6 +216,51 @@ fn namespace_filter(alias: &str, opts: &CompileOptions, params: &mut Vec<QueryVa
     }
 }
 
+/// Compile a node label's `kind` predicate against a primary-substrate union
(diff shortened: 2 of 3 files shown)


oceanwaves630 (approved):
Verified at head: the SPARQL substrate-label regression test now uses a parseable pattern with an edge triple and executes the compiled SQL against the edge fixture. The kind_filter_predicate routing (substrate labels to substrate_kind, granular kinds to kind, event-arm skip) was already verified sound at the prior head.

PR #859, opened 2026-07-11
fix(memory): bounded ANN wait in recall with lexical fallback (#836)

### LINKED ISSUE #836: memory.recall semantic path blocks 300s+ under fresh-write ANN rebuild (ADR-107 stale-serve gap)
## Report

Field report (dogfooding, 2026-07-11 ~00:36-00:44 UTC): \`memory.recall\` with a semantic query timed out at the 300s MCP ceiling twice in a row on a warm kkernel daemon. In the same window, on the same daemon:

- \`knowledge.search\` (atom ANN+FTS) returned in under 1s
- \`gtd.tasks\` fast
- ~20 create/link/update writes all succeeded fast

So the stall is isolated to the memory-pack semantic recall path, not the daemon or the shared stores.

## Context preceding the hang

The session had just written ~3 memories and ~20 kg notes. Hypothesis: the memory ANN index was rebuilding after the write burst and recall BLOCKED on the rebuild instead of stale-serving.

ADR-107 specifies stale-serve + off-thread rebuild exactly to prevent this. Either:

1. the stale-serve path does not cover the semantic-query case in memory recall, or
2. the rebuild holds a lock the read path waits on.

## Repro sketch

Burst of memory writes in a session, then a semantic \`memory.recall\` shortly after, before the index warm completes.

## Severity

P0-adjacent: recall is a core product surface; a 5-minute hang on \`memory.recall\` is a launch-quality defect for khive-cloud. The reporter routed around it via \`knowledge.search\`, but a real user cannot.

## Ask

Audit the memory recall semantic path against the ADR-107 stale-serve implementation: verify the read path never awaits an in-flight rebuild, and add a regression test that issues recall during a forced rebuild and asserts bounded latency (stale results acceptable).

ohdearquant:
## Summary

`memory.recall` timed out at the 300s MCP ceiling on three independent occasions in production on 2026-07-11. In each case the stall was isolated to the memory-pack semantic recall path; every other verb on the same warm daemon (`knowledge.search`, `gtd.tasks`, ~20 create/link/update writes) returned in under a second.

## Root cause

`collect_recall_vector_hits` (`crates/khive-pack-memory/src/handlers/common.rs`) calls `ensure_ann_for_model` on a cold cache miss for a model's vector index. That call acquires a per-model single-flight lock (`ann.rs::model_warm_lock`) before it will do anything. The daemon's boot-time background warm (`warm_existing_memory_indexes`, triggered from `crates/khive-runtime/src/daemon.rs`) holds that same lock for the full duration of a from-scratch corpus build. A recall landing while boot warm is still building therefore blocked on lock acquisition for the entire build, with no bound.

This is distinct from the write-triggered rebuild path ADR-107 already covers (a write bumping a model's generation counter and re-triggering a background rebuild does not block recall). The gap was specifically the synchronous cold-miss path, which had no timeout at all.

## Fix

`collect_recall_vector_hits` now wraps the `ensure_ann_for_model` await in a bounded `tokio::time::timeout` (default 8s). The bound is tunable per request via `RecallConfig.ann_ready_timeout_ms`, or process-wide via `KHIVE_MEMORY_ANN_READY_TIMEOUT_MS`.

On timeout, that model's vector leg contributes zero candidates for this recall and the hybrid pipeline degenerates to its lexical (FTS) arm instead of falling through to the sqlite-vec exact-search fallback (which is itself an O(corpus) scan, unsuited to a bounded-latency path). Each result in a degraded response is stamped `"degraded": "ann_unavailable"`, following the same per-result convention already used for `multilingual_routed`. A degraded recall with zero FTS hits still returns an empty array, not an error.

The ANN build itself, the boot-time warm task, and ADR-107's invalidation/generation semantics are unchanged. A later recall against the same model benefits from the build once it installs, exactly as before.

## Tests

Three new regression tests in `khive-pack-memory` (`handlers/recall.rs`):
- a recall issued while the model's warm lock is held (simulating boot warm mid-build) returns within the bounded wait, with FTS-only results marked `degraded: "ann_unavailable"`
- the normal, uncontended path is unchanged: no `degraded` marker
- a degraded recall whose FTS leg also has no matches returns an empty array, not an error

`cargo test -p khive-pack-memory`: 323 passed, 0 failed.
`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-pack-memory` all pass clean.

## Test plan

- [x] `cargo test -p khive-pack-memory` (323 passed)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-pack-memory`

Closes #836.

### DIFF
--- a/crates/khive-pack-memory/src/ann.rs
+++ b/crates/khive-pack-memory/src/ann.rs
@@ -403,6 +403,21 @@ async fn model_warm_lock(ann: &SharedAnn, key: &AnnKey) -> Arc<tokio::sync::Mute
         .clone()
 }
 
+/// #836 test seam: acquire and hold `key`'s per-model warm single-flight
+/// lock, the same one `ensure_ann_for_model` blocks on when a concurrent
+/// caller (e.g. the daemon's boot-time `warm_existing_memory_indexes`) is
+/// mid-build. Returning the owned guard lets a test simulate "boot warm is
+/// building this model from scratch" without depending on real build
+/// latency — the guard is simply held for as long as the test wants, then
+/// dropped to release it.
+#[cfg(test)]
+pub(crate) async fn hold_model_warm_lock_for_test(
+    ann: &SharedAnn,
+    key: &AnnKey,
+) -> tokio::sync::OwnedMutexGuard<()> {
+    model_warm_lock(ann, key).await.lock_owned().await
+}
+
 #[cfg(test)]
 impl AnnState {
     pub(crate) fn warm_route_count(&self) -> usize {
--- a/crates/khive-pack-memory/src/config.rs
+++ b/crates/khive-pack-memory/src/config.rs
@@ -94,6 +94,18 @@ pub struct RecallConfig {
     /// exhausted. When `None`, falls back to the `ANN_OVERFETCH_MAX_ROUNDS`
     /// env var (default 3). Pass `Some(1)` to disable widening entirely.
     pub ann_overfetch_max_rounds: Option<usize>,
+
+    /// Bounded wait (milliseconds) for a cold-miss `ensure_ann_for_model`
+    /// call on the recall path before degrading that model's contribution
+    /// to FTS-only (#836). Recall and the daemon's boot-time background
+    /// warm (`warm_existing_memory_indexes`) both funnel through
+    /// `ensure_ann_for_model`'s per-model single-flight lock; without a
+    /// bound, a recall landing while boot warm holds that lock for a
+    /// from-scratch corpus build waits out the full build (300s+ observed
+    /// in production). When `None`, falls back to the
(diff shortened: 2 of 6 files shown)


oceanwaves630 (approved):
Verified the detach fix at head: the build now runs on a tracked background task raced via oneshot, so a timeout abandons only the wait — the phase-event pair closes and later recalls find a warm index. The self-build regression test (degrade, detached install, then clean vector path) proves the exact failure mode that concerned me, and the sender-dropped arm handles a panicked build task without surfacing it as the recall's error. #836 stays open pending production hang-window correlation per the agreed closure discipline.

PR #860, opened 2026-07-11
feat(pack-git): git write surface - commit, branch, push (ADR-108)

ohdearquant:
## Summary

Adds a git write surface to the git pack: three thin verbs, `git.commit`, `git.branch`, and `git.push`. This amends ADR-088's "no write-back" stance per ADR-108.

## Scope

Exactly three verbs, matching the ADR-108 resolutions:

- `git.commit` — stage and commit (optionally scoped to specific paths)
- `git.branch` — create a branch from HEAD or an explicit start point
- `git.push` — push a branch to a remote

Explicitly out of scope: tag, merge, checkout, pull, fetch, a `publish_branch` verb, and force-push under any argument combination. An explicit `force` argument is always rejected rather than silently ignored, and any non-boolean `force` value is rejected as well.

`repo` and `branch` are ordinary verb arguments, consistent with the rest of the verb surface. Authorization is enforced at the Gate (ADR-018), not inside these handlers.

## Security posture

- **Argv-only execution.** Every git invocation goes through `std::process::Command::args`, never a shell string. There is no code path where a caller-supplied value is interpolated into a shell command line.
- **Fixed allowlist.** Only the specific git subcommands and flags needed for commit/branch/push are ever constructed; nothing else is reachable.
- **Restrictive character-set validation before argv.** Branch names, remote names, and paths are validated against a restrictive character set, checked for leading dashes (to prevent flag injection), and checked for path traversal, before any value is placed into an argument vector. This validation logic lives in its own module so it can be reviewed and hardened in isolation.
- **No force-push, ever.** Force flags are not in the allowlist, and an explicit `force` argument (of any type) is rejected with a clear error rather than silently dropped.
- **Audit trail.** Every write verb dispatch already produces a gate-check audit event through the standard dispatch path. Successful writes additionally record a supplementary audit event carrying the repository, branch, and (for commits) the resulting commit SHA.

## Test coverage

- Unit tests for the validation and argv-construction module, including injection-shaped inputs: leading dashes, `--upload-pack`-style flag injection, whitespace, semicolons, path traversal, and other malformed input.
- Handler-level integration tests exercising all three verbs against real, disposable git repositories created fresh in a temp directory for each test (never touching any production store), covering both the successful path and each rejection path (empty message, injection-shaped paths/names, non-repo paths, non-boolean or explicit `force`, nonexistent branch).

## Test plan

- [x] `cargo test -p khive-pack-git` — unit + integration tests pass
- [x] `cargo check --workspace` — full workspace compiles
- [x] `cargo clippy -p khive-pack-git --all-targets -- -D warnings` — clean
- [x] Verb catalog and documentation updated to reflect the new verb count


oceanwaves630 (approved):
Reviewed at head b3cd00d6. Delta since the prior approved head 59603ca0 verified independently via merge-tree recompute: manual conflict resolutions confined to 7 docs files plus the verb-count assertion in tests/smoke_test.py (79/81 -> 82, matching the live registry). CI green across the matrix. Approving.

PR #861, opened 2026-07-11
fix(ci): bench-track advisory lane warns instead of failing on ledger record errors

ohdearquant:
## Why

The Bench Trend Tracking workflow is an advisory data-collection lane, but its deliberate fail-on-record-error step painted main red on every run where a Criterion tree came back empty. A red X on main is reserved for correctness CI.

## What

Both "Fail job if trend ledger record errored" steps become "Warn ... (advisory, never fails)":

- the status=error ledger record still publishes to perf-data (unchanged, happens earlier),
- the warning lands in the job summary and as a workflow warning annotation,
- a tracking issue ("bench-track: trend ledger record errored (advisory)") is created or commented on,
- the step exits 0.

Both jobs gain `issues: write` for the tracking-issue upsert. This workflow never runs on pull requests (push-to-main, nightly cron, manual dispatch only), so the warn semantics apply to all of its runs. YAML validated.

Co-Authored-By: Leo <noreply@khive.ai>

### DIFF
--- a/.github/workflows/bench-track.yml
+++ b/.github/workflows/bench-track.yml
@@ -190,6 +190,7 @@ jobs:
           RECORD_EXIT_CODE: ${{ steps.record.outputs.exit_code }}
           GH_TOKEN: ${{ github.token }}
         run: |
+          set +e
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
APPROVE at 4425da3c. Verified content-level: both trend-ledger fail steps are now advisory warn steps — set +e opens each block, every command is guarded non-fatal, and the step ends in an unconditional exit 0, so a data-collection error can no longer paint main red; the status=error record still publishes, the warning lands in the job summary and workflow annotations, and standing issue #863 gets a guarded reopen+comment (no cross-shard create race). issues:write scope added minimally to the two jobs. This completes the main-red semantics order: red X on main is now reserved for correctness CI.

PR #878, opened 2026-07-12
docs(adr): ADR-104 Amendment 1 - prior-preserving evidence decay for per-entity posteriors

ohdearquant:
Amendment to the accepted ADR-104, adding Stage D: per-entity posterior state decomposes into evidence-only counts plus a read-time Beta(1,1) prior, with lazy exponential decay (configurable half-life, default 30 days). Decay-to-neutral doubles as the recovery mechanism for negatively-judged memories. Explicitly rejects pessimistic cold-start priors and a separate coverage/exploration route, with recorded reasoning and the revisit trigger. Snapshot format advances to version 2 with a defined v1 migration. Docs-only; Stage D implementation follows sign-off. Draft pending codex + Leo spec gate.


oceanwaves630 (changes requested):
SPEC GATE verdict (Leo): APPROVE-WITH-FIXES — design ratified, one quantitative defect must be corrected before Stage D implementation merges.

RATIFIED: evidence/prior decomposition with read-time Beta(1,1) (posterior mean form is correct), lazy exponential decay with prior-preserving updates, no-writes-on-read, the v1->v2 migration (prior subtraction + decay clock starting at migration), and all three deliberate rejections including the recorded revisit trigger for the exploration route.

FIX 1 (required, prose): the recovery claim 'returns to within 0.05 of neutral in roughly two half-lives' is magnitude-dependent and as stated only holds for a single negative event. With alpha_ev=0, beta_ev=B the decayed mean is 1/(2+gB); within 0.05 of 0.5 requires gB <= 0.222, so time-to-neutral is H*(log2(B)+2.17). B=1: ~2.2 half-lives (matches the text). B=4: ~4.2. B=10: ~5.5 (over five months at the 30-day default). State the formula or the magnitude condition; do not ship the unconditional two-half-life claim.

FIX 2 (required, gate): the Stage D behavior test must pin the evidence magnitude it exercises, at minimum one case at B=1 and one at B>=8, each with the correct expected bound derived from the formula above (and state the mean-to-multiplier mapping the 0.02 tolerance is measured through). As written the gate passes trivially with light evidence while the heavy-negative recovery story goes untested.

Both fixes are doc+test-spec edits; no design change needed. Re-request on push and Stage D unblocks.

oceanwaves630 (approved):
SPEC GATE re-check (Leo): APPROVED at head 28167ec9. Both fixes verified in the diff — recovery prose now carries the magnitude-dependent formula (mean = 1/(2+gB), time-to-neutral ~ H*(log2 B + 2)), and the Stage D gate pins both magnitudes (B=1 within 0.02 at two half-lives; B>=8 provably NOT at two half-lives, within by H*(log2 B + 2.2)) with the clamp mapping computed explicitly in assertions. Stage D implementation unblocked.

oceanwaves630 (approved):
SPEC GATE round-2 sign-off (Leo): APPROVED at head 59a5b54e, explicitly covering the Stage-B accrual amendment.

EXPOSURE-EXCLUSION RULING RATIFIED: RecallHit/NoteAccessed excluded from per-entity evidence. I desk-verified the self-reinforcement claim independently: steady state of a daily +1.0 under H=30 decay is alpha* = g/(1-g) with g = 2^(-1/30), about 42.8 (43.8 with the fresh hit), mean 0.978, multiplier 1.143 at w_ent 0.3 — exposure alone would durably order candidates, which inverts the term's purpose. Exclusion is the correct cut; global posteriors keep exposure signal as shipped; the revisit trigger now correctly names reintroduction of any automatic exposure signal as a decision that must be retaken.

Also verified in the diff: evidence-as-mass semantics (correction 2.0 / explicit 1.5 / implicit 0.1 under the cumulative implicit fold-gate clamp, B in mass units throughout including the gate cases), event-time write clock with dt = max(event_time - last_event_at, 0) and the byte-identical replay gate per ADR-032, and H boundary rejection (zero, negative, non-finite). Stage D gate now pins B=1, B=2.0 correction, B>=8, exposure no-op, backward-clock no-op, and replay determinism.

Merge rides the codex narrow re-review; Stage D implementation is GO against this head.

oceanwaves630 (approved):
Spec-gate re-approval at head 8cf6e5ab (round 3). All four deltas verified by desk read: (1) the three-way write branch resolves round 2's internal inconsistency where the prose said a regressed timestamp 'never grows evidence' while the clamped-dt formula still added weight at g=1; the explicit drop is deterministic and now gate-pinned byte-identical. (2) Read-clock clamp g=1 on regression is correct; g>1 would amplify evidence. (3) Persisted updated_at as the migration anchor makes v1-to-v2 conversion a pure function of stored bytes (ADR-032 replay determinism); fail-loud on missing timestamp is right. (4) NoteAccessed full no-op scoping is factually accurate to the shipped reducer. Stage D gate additions cover every delta. Approved for merge; Stage D implementation may proceed against this head.

PR #882, opened 2026-07-12
docs(adr): ADR-108 - WebAssembly support for khive-vamana

ohdearquant:
## Summary

Proposes ADR-108: making khive-vamana usable in the browser, per direction to bring khive's ANN search to wasm.

Two layers:

- **Layer A (khive-vamana + khive-quant)**: `parallel` and `mmap` features (default-on, zero native behavior change) with sequential/owned fallbacks, an unconditional `to_bytes`/`from_bytes` persistence API sharing the fingerprinted layout with the file path, an entropy seam (seeded or getrandom js), and a wasm32 `cargo check` CI gate.
- **Layer B (new crate khive-vamana-wasm)**: wasm-bindgen cdylib with build/search/insert/remove/serialize over typed arrays, SQ8 acquisition tier on by default (memory is the binding constraint in a tab), persistence delegated to the JS host (OPFS/IndexedDB/fetch), npm packaging.

Non-goals v0: browser embedding, wholesale khive-in-browser, explicit SIMD128 kernels (measure first), WebGPU, wasm threads.

Acceptance gates: wasm check in CI, sequential/parallel parity on seeded builds, bytes round-trip + native-file interop, fingerprint rejection, headless-browser smoke with SQ8, gzipped size budget.

Rollout: Layer A PR first (independently reviewable), Layer B second, demo consumer separate.

Spec gate: routed to review on open, before any dependent implementation.


oceanwaves630 (approved):
Spec-gate approval at head a1b478b1 (full read). Layer A/B split is right: A is independently useful, zero native behavior change (default-on gates reproduce today's dep set), and the acceptance gates are concrete and testable. Strongest element: the parity gate makes the sequential fallback a tested contract — note it also functions as an honest detector if the 'parallel build is already deterministic' claim (ADR-052) turns out weaker than believed; if parity fails there, fix the claim, not the gate. Bytes API with fingerprint validation on both paths + native-writer/wasm-reader interop gate is the correct compatibility surface. SQ8-default in the tab, host-owned persistence, no-threads/no-SIMD128-without-numbers non-goals all correct. Non-blocking: the Layer B size budget is 'agreed budget TBD' — pin the number in the Layer B PR, not after. Layer A implementation is UNBLOCKED on this approval; re-gate at new head if codex r1 forces material changes.

oceanwaves630 (approved):
Spec gate sign-off (Leo) at de703cdb42027be63a3979249c6b30dc542f4e76, codex APPROVE r6. Delta read r4-r6: SQ8-only v0 with no acquisition-mode selector in config or either persistence format and a fully-specified deferral contract; closed JS config object mapping field-for-field to VamanaConfig with pre-construction validation matching core invariants; JS numeric boundary contract (finite integral safe integers, no implicit truncation, k=0 rejected, k>live returns all live) with its own acceptance gate. Insert/remove atomicity and recycled-ordinal mapping pinned as gates. Size gate normative at 500,000 bytes gzip -9 with ADR as source of truth. Ready for Layer A.

PR #883, opened 2026-07-12
docs(adr): ADR-104 Amendment 2 - closing the Stage D contract

ohdearquant:
Closes the four gaps codex r4 found on #878 (post-merge), per the ratified plan: Stage D implementation is gated on this amendment.

- **Normative signal-to-mass table** covering the legacy `useful`/`not_useful`/`wrong` ladder (1.5/1.5/2.0, an intentional change from the shipped unweighted 1.0, effective at Stage D) alongside the semantic signals.
- **Determinism restated**: ranking is a pure function of (store, query, profile state, read time, decay policy); eval artifacts record the read clock, Stage D ranking tests pin it.
- **Decay policy persisted**: `(policy_version, half_life_days, mass_table_version)` stamped into v2 snapshots; fold always uses the state's own policy; config consulted only at creation and explicit migration; changing H is a recorded policy-version bump, never silent adoption.
- **Storage claim reconciled**: no new table/column/store, but versioned snapshot-payload fields; `brain_profile_snapshots.updated_at` named as the v0/v1 migration anchor applied before deserialization discards row metadata.

Gate additions cover each: per-signal mass/polarity assertions, pinned-clock ranking tests, config-differs-from-snapshot replay, double-migration byte-identity.

### DIFF
--- a/docs/adr/ADR-104-posterior-serving-recall.md
+++ b/docs/adr/ADR-104-posterior-serving-recall.md
@@ -405,9 +405,18 @@ versioned input stamped into persisted state**, superseding Amendment 1's
     policy_version: u32,
     half_life_days: f64,
     mass_table_version: u32,
-    mass_table: [{ signal: FeedbackSignal, polarity: positive | negative, evidence_mass: f64 }, ...]
+    mass_table: [{ signal: EvidenceFeedbackSignal, polarity: positive | negative, evidence_mass: f64 }, ...]
   }
   ```
+
+  `EvidenceFeedbackSignal` is a new closed persisted-row enum containing exactly the
+  eight wire names in the signal table above. Neither shipped type covers all eight:
+  `FeedbackSignal` holds only the three legacy names and `FeedbackEventKind` holds
+  the five semantic names (`crates/khive-brain-core/src/signal.rs`). Stage D defines
+  the new enum with pinned serde names equal to the wire names, a one-way `From`
+  mapping from each `FeedbackSignal` and `FeedbackEventKind` value, and a test
+  asserting the enum has exactly eight members matching the table. The existing
+  public feedback types and their consumers are unchanged.
 
   `mass_table` is the complete signal-to-polarity-and-mass table above, in the row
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Spec-gate approval at head 8a1622cb (same-day, full read). All four r4 gaps closed to my ruling: (1) mass table complete + normative + versioned; the legacy 1.0->1.5/2.0 remap is right per the canonical signal ladder (useful/not_useful ARE explicit judgments; wrong IS correction-class) and codifying the shipped 1.0 would freeze an accident — the per-row gate assertion makes it safe. (2) Determinism restated with read clock as explicit input, clocks pinned in tests. (3) Decay policy as persisted replay state is the correct resolution of my Amendment-1 miss: config consulted only at creation/explicit migration, immutable per lineage, version-bump-never-silent, and the config-differs-from-snapshot test pins that config cannot leak into the fold. Migration as pure fn of (bytes, configured policy) keeps wall clock a non-input. (4) Storage claim honestly scoped; load-before-deserialization ordering named. One non-blocking note: policy_version vs mass_table_version — keep the version->table resolution immutable in code (version-indexed tables, no in-place edits) exactly as the text mandates, since state persists the version number, not the rows. If codex r1 raises material findings, re-gate at the new head per the per-head rule; otherwise Stage D implementation is UNBLOCKED against this contract.

oceanwaves630 (approved):
Leo re-gate: APPROVE at be9611be (codex-stable head, APPROVE r4). The two r3 Medium fixes read correct on the delta: (1) EvidenceFeedbackSignal as a NEW closed persisted-row enum with pinned serde wire names, one-way From mappings from both shipped types, and an exact-eight-membership test — the right call vs widening either existing public type. (2) The migration-boundary field-for-field match between the inline mass_table and the immutable registry entry named by mass_table_version is correctly scoped to the admin boundary ONLY — replay still reads the inline table exclusively, so the Amendment-2 replayability principle is intact while mass_table_version stops being a free label. Both gained matching Stage D negative gates. Flip ready + auto-merge; Stage D impl leg is go.

PR #888, opened 2026-07-12
fix(runtime): token-boundary trigger matching in secret gate to stop path-slug false positives

### LINKED ISSUE #870: Secret scanner false-positive blocks document writes when a path slug contains 'auth' mid-string
## Summary
The write-time secret scanner blocks legitimate `create`/`update` of a document whose path/slug contains a sensitive keyword (e.g. 'auth') glued mid-string inside a hyphenated filename. A file path is not a secret. Repro'd on both create AND update. (Meta: a comm reply merely quoting the path was also blocked — the bug bit its own bug report.)

## Repro (v0-ADR migration, 2026-07-11)
Creating document `c87845b0` (ADR-051 archive) failed: the archive filename slug contains the word 'auth' mid-string, which the scanner read as a high-entropy token adjacent to a sensitive keyword.

Workaround used: split the path value into `file_dir` + `file_name` (both wrote). Cost: that doc now breaks path-field uniformity vs the other 84 v0-ADR docs (single path field).

## Related
Prior secret-gate path FP work (#632/#577) narrowed trigger-word FPs; this is a residual/adjacent case where the sensitive word is a substring of a longer alnum slug rather than a standalone token.

## Ask
Loosen the scanner so a filesystem-path-shaped value with an embedded keyword substring is not flagged (measure-first: confirm the exact token/entropy heuristic firing). Then the 051 doc can revert to a single path field for uniformity.

## Provenance
Dogfooding finding during the v0-ADR migration. Reported by lambda:atlas.

ohdearquant:
Closes #870.

The secret gate's trigger-word scan matched substrings anywhere in a token, so benign path slugs (e.g. `docs/api-key-rotation.md`) self-triggered the gate. This change moves trigger matching to token boundaries and adds an inline-credential check (trigger + assignment + high-entropy value) so true positives are preserved.

- `contains_trigger`: token-boundary matching instead of raw substring
- `has_inline_credential_trigger`: keeps blocking real `key=<entropy>` assignments
- Regression tests in both directions: benign slugs pass, inline credentials still block

### DIFF
--- a/crates/khive-runtime/src/secret_gate.rs
+++ b/crates/khive-runtime/src/secret_gate.rs
@@ -916,7 +916,9 @@ fn has_assignment_credential_trigger(low_text: &str) -> bool {
             .rsplit(|c: char| !c.is_ascii_alphanumeric() && c != '_')
             .next()
             .unwrap_or_default();
-        contains_compound_trigger(label)
+        COMPOUND_TRIGGER_WORDS
+            .iter()
+            .any(|needle| label.contains(needle))
             || TRIGGER_WORDS
                 .iter()
                 .any(|tw| contains_bounded_word(label, tw))
@@ -941,7 +943,9 @@ fn has_inline_credential_trigger(raw_token: &str) -> bool {
 
     !low.contains(['/', '-', '.'])
         && low.contains('_')
-        && (contains_compound_trigger(&low)
+        && (COMPOUND_TRIGGER_WORDS
+            .iter()
+            .any(|needle| low.contains(needle))
             || TRIGGER_WORDS
                 .iter()
                 .any(|tw| contains_bounded_word(&low, tw)))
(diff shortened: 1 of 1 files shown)


ohdearquant:
Live repro during review window: a comm.send whose content contained an absolute workspace file path near the word "keys" was blocked by the gate ("content matches secret pattern high-entropy-token"), requiring a manual reword. Exactly the false-positive class this PR addresses.

ohdearquant:
Codex review round 5 (static, security-focused) at head 0c1465aa: **APPROVE** — 0 findings.

- Assignment labels and glued candidates now use unbounded compound substring matching (supersets of the prior match); prose-context boundary logic retained where it belongs.
- The round-4 escaping input xapi_keyv2=Xk9m... is a regression asserting block; all six earlier bypass-shape regressions retained.
- Reviewer traced the trigger path end-to-end and confirmed no new bypass class introduced.

Five review rounds total, each earlier round surfacing a real bypass class — archived internally. Marking ready + auto-merge.

oceanwaves630 (approved):
Lane C read (security class): full 315-line diff read. Mechanism sound — candidate token excluded from its own surrounding trigger context (kills path-slug self-triggering) while inline/assignment credential shapes are checked on the token itself; compound triggers keep suffix protection and gain an identifier-boundary prefix rule; adversarial regressions (separator-split secret_access_key, prefix-glued xapi_keyv2, spaced/JSON assignments, masker redaction) all covered. Note: this fix does not obviously cover the false-positive class I reported today (bare filesystem path flagged with a genuine trigger word in distant window prose) — keep that tuning row open.

PR #892, opened 2026-07-12
feat(query): GQL WHERE operators CONTAINS, STARTS WITH, IN, IS NOT NULL

### LINKED ISSUE #804: GQL: WHERE grammar lacks CONTAINS / STARTS WITH / IS NOT NULL
Dogfood receipt from KG enrichment consumption testing: the GQL WHERE grammar is comparison-only. These all fail with "expected comparison operator":

- MATCH ... WHERE n.name CONTAINS 'x'
- MATCH ... WHERE n.name STARTS WITH 'x'
- MATCH ... WHERE d.prop IS NOT NULL

Consequence: property-presence queries (all docs with a given property set) and name-prefix scans require paging the whole kind and filtering client-side, which runs straight into the truncation pain tracked in #777.

Ask: add CONTAINS, STARTS WITH, and IS NOT NULL to the GQL WHERE grammar (khive-query parser + SQL compiler). Exact equality already works and remains the primary key path.

Note the existing constraint that WHERE pushdown covers top-level columns only; IS NOT NULL on properties may need the properties-blob filtering story decided first — if so, split that into the design note.

### LINKED ISSUE #864: GQL grammar lacks CONTAINS and IN operators in WHERE clauses
Found during the #849 investigation (defect S2, split out per the fix-lane plan).

WHERE-clause forms like:

    MATCH (e:entity) WHERE e.name CONTAINS "LoRA" RETURN e.id
    MATCH (e:entity) WHERE e.name IN ["LoRA", "QLoRA"] RETURN e.id

are parse errors: the GQL grammar in crates/khive-query only supports equality and comparison operators, so substring and set-membership filters cannot be expressed at all. Agents fall back to kg search or client-side filtering, which defeats the query layer for these question classes.

Scope: grammar + AST + SQL compiler support for CONTAINS (compiles to LIKE with escaped pattern) and IN (compiles to an IN list with bound parameters), plus parser and compiler regression tests. SPARQL parity should be considered in the same change (FILTER CONTAINS / IN equivalents) or explicitly deferred with rationale.

Related: #849 (S1 substrate-label compilation and S3 resolve exact-name were fixed in #857 / #852; this is the remaining grammar-level defect).

ohdearquant:
Closes #864. Closes #804.

Full vertical for four GQL WHERE operators: lexer tokens, AST nodes, validation, and SQL compilation with parameterized values and %/_ escaping for the LIKE-backed forms. Includes the small ADR-008 operator-table update that documents the contract (kept in this PR because the table is the normative surface for the same change).

### DIFF
--- a/crates/khive-query/docs/design.md
+++ b/crates/khive-query/docs/design.md
@@ -65,6 +65,10 @@ ADR-008. It is intentionally split into three stages:
   preserving SQL OR/AND connectives rather than flattening to AND-only.
 - GQL WHERE grammar: `where_expr = and_expr ('OR' and_expr)*` where
   `and_expr = condition ('AND' condition)*`. AND binds tighter than OR.
+- GQL WHERE conditions support `=`, `!=`, `>`, `<`, `>=`, `<=`, `LIKE`,
+  `CONTAINS`, `STARTS WITH`, `IN` with a list literal, and `IS NOT NULL`.
+  `CONTAINS` and `STARTS WITH` treat `%`, `_`, and `\` as literal characters.
+  All condition values are emitted as bound SQL parameters.
 - Node kind strings are pack-agnostic and pass through the query layer unchanged.
   Kind validation is a pack-handler concern, not a query-layer concern.
 - `namespace` is always injected via `CompileOptions.scopes`, never from query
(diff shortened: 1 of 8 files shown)


ohdearquant:
codex review round 2: **APPROVE** at head 125e0b85 (round 1 APPROVE-WITH-FIXES findings verified applied). Verdict archived internally.

oceanwaves630 (approved):
Lane B: codex APPROVE at head + CI green verified mechanically; v3 approve-on-sight. Sampling audit continues per merge flow.

PR #895, opened 2026-07-12
test(kg): exact-name lookup regressions for CJK and spaced names (#849)

### LINKED ISSUE #849: kg: no working exact-name entity lookup (GQL name equality returns empty; resolve() misses existing names)

ohdearquant:
Closes #849.

Investigation outcome: the reported name-equality failures are not reproducible on main — the resolve exact-name tier and GQL name comparison handle CJK and spaced names correctly. This PR pins that behavior with regression tests (resolve exact-name tier: CJK + spaces + case-variant hybrid fallback; GQL compile integration for name equality) and clarifies the documented semantics in the API reference.

### DIFF
--- a/crates/khive-pack-kg/src/dispatch.rs
+++ b/crates/khive-pack-kg/src/dispatch.rs
@@ -1532,6 +1532,25 @@ mod tests {
                 Some(0.98)
             );
         }
+
+        let whitespace_variant = registry
+            .dispatch(
+                "resolve",
+                json!({"refs": ["Entity  Name With Spaces"], "kind": "entity"}),
+            )
+            .await
+            .expect("whitespace-variant resolve must succeed");
+        let whitespace_results = whitespace_variant
+            .get("results")
+            .and_then(|v| v.as_array())
+            .unwrap();
+        assert_ne!(
+            whitespace_results[0]
+                .get("confidence")
+                .and_then(|v| v.as_f64()),
+            Some(0.98),
+            "interior whitespace must be preserved by exact-name lookup: {whitespace_results:?}"
+        );
     }
 
     /// The storage exact-name tier is case-sensitive. A case-only variant
--- a/crates/khive-pack-kg/src/dispatch.rs
+++ b/crates/khive-pack-kg/src/dispatch.rs
@@ -1472,6 +1472,116 @@ mod tests {
         );
     }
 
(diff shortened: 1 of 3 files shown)


ohdearquant:
codex review round 2: **APPROVE** at head 0b097e74 (round 1 findings verified applied). Verdict archived internally.

oceanwaves630 (approved):
Lane B: codex APPROVE at head + CI green verified mechanically; v3 approve-on-sight. Sampling audit continues per merge flow.

PR #909, opened 2026-07-12
fix(pack-kg): include resource entity kind in create help schema

### LINKED ISSUE #899: create help=true schema advertises 8 entity kinds — resource (ADR-048) missing from the enum
The live `create` help schema still lists the 8-member entity_kind enum; `resource` (ADR-048) is absent even though the vocab accepts it. Integrators probing help= get a stale contract. Fix in the pack vocabulary/help generation so the enum is derived from the registered kinds, not a hand-written list.

ohdearquant:
Closes #899


oceanwaves630 (approved):
Lane-A approve: codex APPROVE on record, CI gate green; pack-registry scope re-verified during the Resource-kind check earlier today.

Issue #884, opened 2026-07-12
schedule.remind fires but delivers nothing observable to the creating actor


ohdearquant:
From live fleet repro 2026-07-12 (atlas, Leo-directed filing).

**Behavior**: a \`schedule.remind\` row transitions pending -> fired (\`fired_at\` stamps on the daemon tick) but no message is delivered to the creating actor's inbox and nothing else observable happens. \`schedule.schedule(action="comm.send(to=..., ...)")\` on the same daemon delivers trigger-to-inbox in ~1.5s.

**Why it is nasty**: the row looks healthy in \`schedule.agenda()\` (status pending, repeat set, fired_at advancing), so "remind me at 9am" is a silent no-op with no error anywhere. \`remind\` is the first verb new users try, so this shape matters commercially.

**Resolution options**:
- (a) \`remind\` delivers a message to the creating actor's inbox on fire (the semantics everyone assumes), or
- (b) rename/document \`remind\` as a display-only agenda entry and make \`schedule.schedule(action=comm.send)\` the documented reminder pattern.

The fleet has standardized on (b)'s pattern as a workaround. Preferred fix is (a): on fire, the executor delivers the reminder content to the creating actor's inbox, so the verb matches user intuition. Docs updated accordingly.

Repro fixtures exist in the live store (seven cancelled remind rows from 2026-07-12 ~17:15Z vs working schedule.schedule replacements); atlas offered to be test consumer on the fix.

PR #911, opened 2026-07-12
fix(pack-schedule): preserve ISO TZ offset in remind create-response rendering

### LINKED ISSUE #871: schedule.remind create-response humanize layer drops ISO TZ offset (renders future trigger as past)
## Summary

`schedule.remind`'s **create-response humanized rendering** drops the ISO timezone offset when computing relative time, so a trigger set for a moment in the future can render as if it were in the past (e.g. `"1h ago"` for a time 2.5h ahead). The stored data and the fire path are correct — this is a **presentation/humanize-layer** bug only.

## Repro

```
schedule.remind(content="...", at="2026-07-11T19:00:00-04:00")
```
At wall-clock ~20:25Z, the create response's humanized `trigger_at` rendered `"1h ago"`. `19:00-04:00` is `23:00Z` — 2.5h in the **future**. Rendering it `1h ago` is consistent with a naive read of `19:00` as `19:00Z` (=1h25m before 20:25Z), i.e. the `-04:00` offset was dropped when parsing for the relative-time calculation.

## Seam analysis (what is NOT broken)

The offset is handled correctly at all three data-path seams — verified by static read and confirmed against runtime evidence:

- **Write / validate** — `crates/khive-pack-schedule/src/handlers.rs:77` `at.parse::<DateTime<Utc>>()` converts `-04:00` → UTC correctly; `handle_remind` (`:794-802`) stores the caller's original offset string verbatim as `trigger_at` (round-trip faithful, confirmed: the agenda listing returns `-04:00` intact).
- **Candidate SQL** — `crates/khive-mcp/src/pending_events.rs:354` compares via SQLite `datetime(...)`, which normalizes offsets to UTC (the PR #782 r3 fix).
- **Fire check** — `crates/khive-mcp/src/pending_events.rs:482` `parse::<DateTime<Utc>>()` then `trigger_at > now` skip — offset-correct.

`schedule.cancel` also handled the offset fine.

## Locus

The bug is in the **humanize / relative-time rendering of `trigger_at` in the create response** (the presentation layer — see `crates/khive-runtime/src/presentation.rs`, which already carries a note at :436/:477 that `trigger_at` payload timestamps must not be compacted). Whatever computes the "Nh ago / in Nh" relative string is parsing `trigger_at` without honoring its offset — likely reading it as naive/UTC instead of `DateTime<FixedOffset>`.

## Fix direction

Parse `trigger_at` offset-aware (`DateTime<FixedOffset>` / `DateTime::parse_from_rfc3339`, then `.with_timezone(&Utc)`) before computing the relative-time delta in the humanize path, matching what the data-path seams already do. Add a regression test: a `remind` with a non-UTC offset must render a future-vs-past relative string consistent with the UTC instant, not the naive wall time.

## Environment

kkernel 0.3.0, binary built Jul 10 (post-#782). Surface: `schedule.remind` via production MCP.

ohdearquant:
Closes #871

## Root cause

`schedule.remind`/`schedule.schedule` return `trigger_at` as a top-level convenience
field in their create response (sibling to `id`/`full_id`), not nested under a
`"properties"` object.

Agent-mode presentation (`crates/khive-runtime/src/presentation.rs`, the shared
humanize layer used by every MCP response) only exempted caller-supplied payload
timestamps from compaction when they were nested under a literal `"properties"`
key (`inside_properties`). That guard correctly protects `trigger_at` returned by
`agenda`/`get` (nested), but did not cover the create response's top-level field,
which fell through to `compact_timestamp` and got rewritten into a relative
"Nh ago"/"in Nh" string — capable of rendering a real future trigger as if it were
in the past.

The lower-level offset arithmetic (`parse_iso8601_unix`/`compact_timestamp`) is
already correct — verified by pre-existing tests. The gap was that this field
should never be compacted at all, regardless of nesting.

## Fix

Added `PAYLOAD_TIMESTAMP_FIELDS` (currently `["trigger_at"]`), following the
existing `LIFECYCLE_NULL_PRESERVE`/`SCORE_FIELDS` field-name-set pattern, threaded
through `present` → `transform_agent` → `transform_field_agent`. A field named
`trigger_at` is now exempt from Agent-mode timestamp compaction at any nesting
depth, not just inside `properties`.

## Tests

- `khive-runtime/src/presentation.rs`: 5 new unit tests — top-level `trigger_at`
  with a non-UTC offset, UTC (`Z`), and offset-less input all round-trip
  unchanged; a sibling `created_at` field is still compacted (scoped fix, no
  blanket disable); the pre-existing nested-under-`properties` protection still
  holds.
- `khive-mcp/tests/integration.rs`: 2 new end-to-end tests driving
  `schedule.remind` through the real MCP `request` surface in default Agent mode,
  asserting the create response's top-level `trigger_at` is preserved verbatim
  (offset-bearing and UTC cases).

Full `khive-pack-schedule` suite re-run clean (105 tests), including the existing
H5 offset-preservation tests at the write/validate seam — confirms this is purely
a presentation-layer fix with no change to how `trigger_at` is stored or compared.

Firing/delivery logic (`khive-mcp/src/pending_events.rs`) is untouched — out of
scope per #884.

### DIFF
--- a/crates/khive-mcp/tests/integration.rs
+++ b/crates/khive-mcp/tests/integration.rs
@@ -3977,6 +3977,72 @@ async fn schedule_agenda_agent_preserves_properties_trigger_at_verbatim() -> any
     Ok(())
 }
 
+// ── #871: schedule.remind create-response preserves top-level trigger_at ─────
+
+/// `schedule.remind`'s create response returns `trigger_at` as a top-level
+/// convenience field (sibling to `id`/`full_id`), not nested under
+/// `"properties"`. In default Agent mode this must round-trip verbatim,
+/// offset intact — the humanize layer previously misread the wall-clock
+/// digits of a non-UTC-offset `at` as UTC and could render a future trigger
+/// as if it were in the past (#871).
+#[tokio::test]
+async fn schedule_remind_agent_preserves_top_level_trigger_at_with_offset() -> anyhow::Result<()> {
+    let client = connect_schedule_only().await?;
+    // Far enough in the future (relative to "now") that a naive/offset-blind
+    // parse landing inside the 24h relative-render window would have
+    // produced a bogus "Nh ago"/"in Nh" string instead of the exact ISO
+    // string being preserved.
+    let trigger_at = "2099-06-15T19:00:00-04:00";
+
+    let result = call(
+        &client,
+        "request",
+        json!({"ops": format!(
+            r#"schedule.remind(content="agent create-response trigger_at fidelity", at="{trigger_at}")"#
+        )}),
+    )
+    .await?;
+    let body: Value = serde_json::from_str(&first_text(&result))?;
+    let first = &body["results"][0];
+    assert_eq!(
+        first["ok"],
+        json!(true),
+        "schedule.remind must succeed: {first}"
+    );
+
+    let actual = first["result"]["trigger_at"].as_str().unwrap_or("");
+    assert_eq!(
+        actual, trigger_at,
+        "top-level trigger_at in the create response must be preserved verbatim, offset intact, in Agent mode"
+    );
+    Ok(())
+}
+
+/// UTC (`Z`) and offset-less `at` inputs must likewise round-trip unchanged
+/// through the create-response humanize layer (#871 regression coverage).
+#[tokio::test]
+async fn schedule_remind_agent_preserves_top_level_trigger_at_utc_and_offset_less(
+) -> anyhow::Result<()> {
+    let client = connect_schedule_only().await?;
+
+    let utc = "2099-06-15T23:00:00Z";
+    let result = call(
+        &client,
+        "request",
+        json!({"ops": format!(
+            r#"schedule.remind(content="agent create-response trigger_at utc", at="{utc}")"#
+        )}),
+    )
+    .await?;
+    let body: Value = serde_json::from_str(&first_text(&result))?;
+    let actual = body["results"][0]["result"]["trigger_at"]
+        .as_str()
+        .unwrap_or("");
+    assert_eq!(actual, utc, "UTC trigger_at must round-trip unchanged");
+
+    Ok(())
+}
+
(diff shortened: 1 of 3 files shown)


ohdearquant:
Codex review r1: APPROVE-WITH-FIXES at d89d9bd1 (archived: .khive/codex_reviews/codex_review_pr911_round1.md, local-only).

- Central fix verified real: top-level trigger_at in schedule create responses was compacted by the Agent-mode humanize layer (only properties-nested timestamps were exempt); PAYLOAD_TIMESTAMP_FIELDS follows the existing LIFECYCLE_NULL_PRESERVE / SCORE_FIELDS pattern and a workspace sweep found no other pack emitting a trigger_at field, so the any-depth exemption is not overbroad.
- Both Low findings (regression-narrative overclaim + mislabeled offset-less test) fixed in 633eca13 — wording/rename only, verified by full delta read; no logic change.
- Medium sibling finding (gtd.assign top-level due has the same fidelity class) deliberately NOT expanded into this PR — filed separately as #912.

Reviewed head == 633eca1337a42dedf768d69072d5bc005022706e.

oceanwaves630 (approved):
Lane-B approve at 633eca13: codex APPROVE-WITH-FIXES at d89d9bd1 with both Low findings fixed and the owning maintainer's delta read on record (comments/rename only). Sibling Medium correctly split to #912.

ohdearquant:
Codex round 2 (narrow, delta-only): **APPROVE-WITH-FIXES** at 26f611b4 — 1 Low (module docstring overstated Agent-mode timestamp rendering; <24h timestamps render relatively, older ones minute-granularity). Fix applied verbatim as prescribed at d442efae (comment-only delta). Round-1 APPROVE covered the code change at 633eca13. CI red at 633eca13 was the stale smoke expectation (offset_at[:16] minute-truncation) that this PR's #871 semantics correctly obsolete — smoke aligned in 26f611b4. Verdicts archived: .khive/codex_reviews/codex_review_pr911_round{1,2}.md

PR #913, opened 2026-07-12
fix(pack-schedule): isolate pack tests from the live store

### LINKED ISSUE #779: schedule pack tests write fixtures to the live store — 8 test events parked at trigger_at=2099 visible in agenda()

ohdearquant:
Closes #779

## Root cause

Every `khive-pack-schedule` Rust test already constructs its own `KhiveRuntime::memory()` (`db_path: None` -> a true in-memory SQLite connection, bypassing ambient CLI/env/config discovery entirely), and `tests/smoke_schedule.py` already spawns its binary with an explicit `--db :memory:` on every invocation. Neither path resolves an ambient or live store as currently written on `main`.

What was missing was an *enforced* invariant: nothing stopped a future edit to any of the five schedule test files from silently swapping in a file-backed or ambiently-resolved runtime, and nothing in the smoke test asserted its own isolation flag actually landed in the spawned argv. Given that review, the most plausible source of the observed live-store rows (8 -> 14 -> 42, accelerating) is out-of-band: fixture content copied into a live MCP session, or a stale/globally-installed `kkernel` binary invoked outside this repo's own build, while manually exercising the schedule pack -- not a `cargo test` or `python3 tests/smoke_schedule.py` run of the code as it exists on `main`.

## Changes

- New `crates/khive-pack-schedule/tests/support/mod.rs`: a single `memory_runtime()` choke point every one of the five schedule test files (`agenda`, `cancel`, `create_validation`, `integration`, `pack_registration`) now routes through. It constructs the in-memory runtime and asserts the resulting backend is not file-backed, so a future regression fails the test immediately -- in-process, no I/O against any real path -- instead of silently writing fixtures into whatever store the environment resolves.
- `tests/smoke_schedule.py`: `spawn_proc()` now builds argv from a named `ISOLATION_DB` constant and asserts `--db` landed on it before spawning, so a future edit that drops or changes the flag fails loud instead of quietly running against a default/live database.

## Out of scope

The one-time curation of the fixture rows already parked in the live database needs direct access to that store, which this change deliberately never touches (per safety constraints on this task). Flagging as a follow-up for whoever holds that access.

## Test plan

- [x] `cargo test -p khive-pack-schedule` -- 105 tests, 0 failed
- [x] `cargo clippy -p khive-pack-schedule --all-targets -- -D warnings` -- clean
- [x] `cargo fmt --all` -- no diff
- [x] `KKERNEL_BINARY=<local build> python3 tests/smoke_schedule.py` -- all 20 scenarios pass against a real built binary, confirming the new `ISOLATION_DB` guard doesn't regress behavior

### DIFF
--- a/crates/khive-pack-schedule/tests/agenda.rs
+++ b/crates/khive-pack-schedule/tests/agenda.rs
@@ -3,8 +3,10 @@
 use khive_pack_schedule::SchedulePack;
 use khive_runtime::{KhiveRuntime, VerbRegistry, VerbRegistryBuilder};
 
+mod support;
+
 fn build_registry() -> (VerbRegistry, KhiveRuntime) {
-    let runtime = KhiveRuntime::memory().expect("in-memory runtime");
+    let runtime = support::memory_runtime();
     let mut builder = VerbRegistryBuilder::new();
     builder.register(khive_pack_kg::KgPack::new(runtime.clone()));
     builder.register(SchedulePack::new(runtime.clone()));
@@ -191,7 +193,7 @@ async fn h2_agenda_finds_valid_event_past_corrupt_legacy_rows() {
     use khive_storage::Note;
     use serde_json::json;
 
-    let runtime = KhiveRuntime::memory().expect("in-memory runtime");
+    let runtime = support::memory_runtime();
     let mut builder = VerbRegistryBuilder::new();
     builder.register(khive_pack_kg::KgPack::new(runtime.clone()));
     builder.register(SchedulePack::new(runtime.clone()));
--- a/crates/khive-pack-schedule/tests/cancel.rs
+++ b/crates/khive-pack-schedule/tests/cancel.rs
@@ -3,8 +3,10 @@
 use khive_pack_schedule::SchedulePack;
 use khive_runtime::{KhiveRuntime, VerbRegistry, VerbRegistryBuilder};
 
+mod support;
+
 fn build_registry() -> (VerbRegistry, KhiveRuntime) {
-    let runtime = KhiveRuntime::memory().expect("in-memory runtime");
+    let runtime = support::memory_runtime();
     let mut builder = VerbRegistryBuilder::new();
     builder.register(khive_pack_kg::KgPack::new(runtime.clone()));
     builder.register(SchedulePack::new(runtime.clone()));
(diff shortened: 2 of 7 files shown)


ohdearquant:
Codex review r1: APPROVE, 0 findings (archived: .khive/codex_reviews/codex_review_pr913_round1.md, local-only).

- Load-bearing negative claim independently confirmed from origin/main: all 14 baseline schedule-test runtime constructors are explicit KhiveRuntime::memory() (db_path None → StorageBackend::memory, is_file_backed=false); no ambient file-backed resolution path exists, and --db :memory: cannot select the live store on either MCP boot path (resolve_db_anchor returns None; multi-backend configs are forced to BackendKind::Memory).
- Choke point coverage complete: support/mod.rs memory_runtime() is the only direct constructor left in the pack tests; assertion is non-vacuous (guards the live backend flag that a file-backed path would set).
- Smoke argv guard verified to fail before subprocess.Popen.
- Tests-only diff, Lane A.

Reviewed head == 2cae2432c539a1c21bc1e5bc96cf5dc22ee5a90f.

oceanwaves630 (approved):
Lane-A approve: tests-only, codex APPROVE zero findings with the negative claim independently verified from origin/main.

PR #915, opened 2026-07-12
feat(pack-knowledge): slow-request and timeout logging for compose

### LINKED ISSUE #887: knowledge.compose has no daemon-side slow-request or timeout logging
During a heavy-load window (system load avg 70-90), two consecutive knowledge.compose calls hit the 300s client timeout while memory.recall and kg search on the same daemon returned normally. Scanning the daemon log for that window shows zero compose-path entries: no start/finish record, no duration, no timeout error. The compose rerank leg is embedder-CPU-bound, so load starvation is the working hypothesis, but the daemon offers no evidence either way.

Ask: emit a structured log line for compose requests that exceed a slow threshold (e.g. 10s), carrying phase timings (suggest/fetch/rerank/trim) and atom counts, plus an explicit log record when a request is abandoned because the client disconnected. Without this, client-side timeouts are unattributable.

Not a behavior change to compose itself; observability only.

ohdearquant:
Closes #887

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/compose.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/compose.rs
@@ -335,9 +335,21 @@ fn tokenize(text: &str) -> Vec<String> {
 // *before* its (possibly fallible, possibly long-running) work starts; both
 // `finish` and `Drop` flush whatever phase is still open — `last..now` — into
 // that phase's bucket before emitting, so an in-flight phase is never lost.
-// The same flush also picks up the tail of the final phase (response-JSON
-// construction after the last measured stage), which a pre-`finish` mark
-// would otherwise miss.
+//
+// Round-2 codex review (#915) found two follow-ups, both addressed here and
+// in `search.rs`: (1) the round-1 regression tests called the private
+// `flush_active` manually before `finish`/`drop`, so they exercised the
+// flush logic but not the actual call sites that must invoke it — `finish`
+// now returns its post-flush breakdown so a test can assert on the real call
+// site's output directly, and the `Drop` path is covered by a `tracing`
+// capture subscriber (see the test module). (2) `compose()` was calling
+// `finish` before constructing its response `Value` on three routine return
+// paths, so the response-JSON assembly and the trailing bit of whichever
+// phase was active went unattributed; every routine return now builds its
+// response into a local binding, calls `finish`, then returns it, so `finish`
+// is genuinely the last thing that happens on every routine path — the flush
+// picks up the tail of the final phase (including the response-JSON
+// construction), not just the last explicitly `begin`-marked stage.
 
 /// Compose requests whose total handler time reaches this are logged at WARN
 /// with a per-phase breakdown. 10s matches the example threshold named in
@@ -396,13 +408,17 @@ impl Phase {
 /// needs (round-1 codex review, Medium finding).
 ///
 /// `finish` must be the last thing called on every return path that
-/// completes the request (success or a business-logic error): it flushes the
-/// active phase, flags the timing as complete, and, if the total reaches
-/// [`COMPOSE_SLOW_THRESHOLD_MS`], emits the slow-request WARN. If `finish` is
-/// never reached — because the enclosing future was dropped mid-poll (client
-/// disconnect, cancellation, or daemon shutdown drain) — `Drop` performs the
-/// same flush and emits a distinct "abandoned" WARN, so a request that never
-/// produces a response is not silently invisible.
+/// completes the request (success or a business-logic error) — in
+/// particular, after the response `Value` is fully constructed, not before
+/// (round-2 codex review, Low finding: response-JSON assembly is real work
+/// and belongs inside whichever phase is still active, typically `Trim`).
+/// `finish` flushes the active phase, flags the timing as complete, and, if
+/// the total reaches [`COMPOSE_SLOW_THRESHOLD_MS`], emits the slow-request
+/// WARN. If `finish` is never reached — because the enclosing future was
+/// dropped mid-poll (client disconnect, cancellation, or daemon shutdown
+/// drain) — `Drop` performs the same flush and emits a distinct "abandoned"
+/// WARN, so a request that never produces a response is not silently
+/// invisible.
 ///
 /// `query_bytes` records the query's UTF-8 *byte* length, not a char count —
 /// `str::len()` reads a value the string already carries (O(1)), unlike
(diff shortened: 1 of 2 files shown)


ohdearquant:
## Codex review round 2 (head 0299fd85)

APPROVE-WITH-FIXES

PR #915 round-2 re-review — `feat/887-compose-slow-logging` at `0299fd85a9244aa6f3c57bc46d72ce5a42d2c986`

Findings: 0 Blocker, 0 High, 1 Medium, 1 Low

### [Medium] The new finish/Drop regression tests do not exercise finish or Drop flushing

Evidence: `crates/khive-pack-knowledge/src/knowledge/compose.rs:746-756` calls the private `flush_active()` at `:750` and asserts its result at `:751-755` before calling `finish()` at `:756`. The Drop test repeats that pattern at `:765-777`: it flushes at `:771`, observes the total at `:772-776`, and only then calls `drop(t)` at `:777`. Thus deleting `self.flush_active()` from `finish()` (`:478-480`) or from `Drop::drop()` (`:496-500`) leaves both named regression tests green.

Why this matters: the refactor correctly fixes the round-1 defect today, but these tests do not protect the two call sites whose behavior is the regression contract. A future edit can reintroduce a missing in-flight phase on ordinary error completion or cancellation without failing the suite.

Suggested fix: capture the emitted `tracing` event in each test (or expose a test-only emission record) after calling `finish()` without a prior manual flush, and after dropping without a prior manual flush. Assert that the relevant phase field is nonzero; keep the current helper-level check only as a separate unit test if useful.

### [Low] `finish()` still precedes part of response JSON construction

Evidence: the successful path calls `timing.finish(count)` at `crates/khive-pack-knowledge/src/knowledge/search.rs:1596`, then constructs the outer response `json!` at `:1597-1600`. The auto-suggest-error and no-atoms early exits likewise finish before constructing their response values at `:1327-1338` and `:1405-1415`. This conflicts with the tracker contract that `finish` is the last operation on each routine return path (`crates/khive-pack-knowledge/src/knowledge/compose.rs:398-401`).

Why this matters: this is a small tail, but it is not attributed to `trim` (or to the active early-exit phase), contrary to the stated full handler breakdown. It is most visible as a documentation/measurement-boundary mismatch rather than a compose-result correctness risk.

Suggested fix: construct each response in a local `Value`, call `timing.finish(...)`, then `return Ok(response)`. That preserves the exact response/error contract while placing all response construction inside the active phase.

## Looks Right

- The checkout exactly matches the requested round-2 head. The full delta since the round-1 head (`254905ce...HEAD`) is limited to `compose.rs` and `search.rs`; both `git diff --check origin/main...HEAD` and `git diff --check 254905ce...HEAD` are clean.
- The round-1 in-flight-loss defect is fixed in the implementation. `begin` accumulates the previously active `last..now` interval and opens the next phase (`compose.rs:445-451`); `finish` flushes the active phase before setting `completed` (`:478-480`); an unfinished future takes the same flush through `Drop` (`:496-500`). All four phase slots are rendered even when a phase has no work (`:466-472`).
- Every post-tracker routine error/early exit closes the active phase: suggest failure at `search.rs:1325-1338`; no suggested domains at `:1348-1360`; loader/parser and atom/rerank/section-load errors through `try_or_finish!` at `:1296-1305`, `:1370-1388`, `:1429-1440`; no atoms at `:1404-1415`; and success at `:1596`. The only pre-tracker returns are validation/deserialization paths at `:1253-1282`, before any await or timing state exists. Future cancellation instead drops the active tracker.
- Await attribution is correctly placed: suggest is opened before `Self::suggest(...).await` (`:1295`, `:1310-1316`); fetch before domain/atom loads (`:1363`, `:1370-1388`) and section loading (`:1438-1440`); rerank before atom reranking (`:1429-1430`) and the section query embed (`:1443-1469`); trim begins only after those awaits (`:1487`).
- Repeated phases accumulate rather than overwrite: `begin` adds into the enum-indexed slot (`compose.rs:447-449`), and the two `Fetch` spans are exercised at `:712` and `:716` with an accumulated assertion at `:720-725`. `finish` then prevents a double flush/log in Drop via `completed` (`:479-480`, `:498-500`).
- The round-1 hot-path finding is resolved. The tracker stores `[Duration; Phase::COUNT]` rather than a `Vec` (`compose.rs:417-424`) and records `query.len()` (`:428-438`), an O(1) UTF-8 byte length. Eager capture is sound because `raw_query` is moved into response JSON before the later `finish()` call (`search.rs:1581-1596`); the byte-length regression test correctly distinguishes the 13-byte query from its character length (`compose.rs:788-796`).
- The `origin/main...HEAD` code delta is observability-only: the compose ranking, result construction, and score inputs are unchanged; formerly direct `?` propagation now returns the same `RuntimeError` after closing timing via `try_or_finish!` (`search.rs:1296-1305`). This preserves ADR-051's hybrid scoring and response behavior.

## Commands Run

- `git status --short --branch`, `git rev-parse HEAD`, and `gh pr view 915 --json ...`: the local branch is clean and matches PR #915 head `0299fd85a9244aa6f3c57bc46d72ce5a42d2c986` (base `main`).
- `git diff --find-renames --find-copies 254905ce...HEAD -- crates/khive-pack-knowledge/`, complete changed-hunk/source and exit-path review, and `git diff --color=never --unified=8 origin/main...HEAD -- .../search.rs`: completed.
- `git diff --check origin/main...HEAD` and `git diff --check 254905ce...HEAD`: passed with no whitespace errors.
- `lsof /private/tmp/khive-shared-target/debug/.cargo-build-lock`: the shared build lock was held by Cargo PID 90810; per instruction, `cargo test -p khive-pack-knowledge` was not run.

## What I Did Not Check

- Focused Cargo tests were skipped because the required shared Cargo build lock was busy.
- Runtime tracing capture was not run; the test-coverage finding is established from the test ordering and the tracker call sites.

## Re-Review Guidance

Narrow re-review after adding integration-level assertions for the actual `finish` and Drop emissions, and moving `finish` after response construction on the ordinary response paths.

Domain utility: SKIPPED — no lore connector was available; the accepted ADR and local crate contracts supplied the relevant review standard.

ohdearquant:
## Codex review round 3 (head 2ceb87d7)

APPROVE

PR #915 round-3 narrow re-review — `feat/887-compose-slow-logging` at `2ceb87d7e27643057c8f412c44fd350e1aad4a28`

## Verified Resolutions

- **Round-2 Medium — `finish()` regression coverage is real.** `crates/khive-pack-knowledge/src/knowledge/compose.rs:794-801` opens `Suggest`, sleeps, and directly calls `finish()` with no intervening `begin` or manual `flush_active()`. The assertion at `:802-807` reads the returned post-flush breakdown and requires `suggest_ms >= 5`. Removing `self.flush_active()` from `finish()` at `:506` leaves the active phase total at zero and fails this test.
- **Round-2 Medium — Drop regression coverage is real.** `compose.rs:873-894` opens `Rerank`, sleeps, then drops the tracker under the local capture subscriber without manually flushing. `Drop::drop()` flushes before emitting its event at `:525-535`; the test identifies that event and rejects a `("rerank", 0)` breakdown at `:896-915`. Removing `self.flush_active()` from `Drop` at `:528` therefore fails the assertion. The former direct-helper check remains expressly unit-only at `:763-782`.
- **Round-2 Low — response construction precedes completion.** The auto-suggest-error path constructs `response` at `crates/khive-pack-knowledge/src/knowledge/search.rs:1325-1337`, then finishes and returns it at `:1338-1339`. The no-atoms exit does the same at `:1406-1418`, and the success path builds its full outer `Value` at `:1582-1602` before `finish` at `:1603-1604`. The additional no-matching-domains exit also follows that order at `:1349-1362`. No routine `Ok` return after tracker construction calls `finish` before creating its response; error-only `try_or_finish!` exits have no handler response value to construct (`:1296-1305`).
- **Private signature change has no production response-contract effect.** `ComposeTiming::finish` remains `pub(super)` and returns only its post-flush fixed breakdown (`compose.rs:505-521`). Every production call site discards that return (`search.rs:1301`, `:1338`, `:1361`, `:1417`, `:1603`); the return is consumed only by the targeted unit test at `compose.rs:801-807`.
- **Scope and hygiene.** `git diff --name-status 0299fd85..HEAD` reports only `crates/khive-pack-knowledge/src/knowledge/compose.rs` and `search.rs`; `git diff --check 0299fd85..HEAD` is clean. The checkout remains clean.

## Commands Run

- `git rev-parse HEAD`, `git diff --name-status 0299fd85..HEAD`, full changed-source/call-site review, and `git diff --check 0299fd85..HEAD`: head matches `2ceb87d7...`; exactly the two allowed files changed; no whitespace errors.
- Non-blocking advisory-seat probe on `/private/tmp/khive-seat-build.lock`: busy (exit 75). Per instruction, `cargo test -p khive-pack-knowledge` was not started; no `.cargo-build-lock` file was created or locked.

## What I Did Not Check

- Focused Cargo tests were skipped because the shared build seat was held. The requested regression claims were verified statically from the tests and tracker call sites.

## Re-Review Guidance

No further re-review is needed for the two round-2 findings.

Domain utility: SKIPPED — this narrow Rust code review was governed by the local tracker and response-construction contracts; no external domain corpus was needed.

oceanwaves630 (approved):
Lane B: codex round-3 APPROVE (0 findings) at 2ceb87d7 on-PR, required checks green (cancelled entry = non-required bench profile). Approving; arm your squash.

Issue #889, opened 2026-07-12
memory.recall can hit 300s client timeouts while search/get/neighbors on the same daemon stay healthy


ohdearquant:
Two independent reports on the same day, same daemon:

- Agent session A: memory.recall timed out at 300s three times across two separate sessions (~13:30-14:00 local), while kg search, get, neighbors, and memory.remember all returned normally in the same sessions. Repro shape: `memory.recall(query=...)` via kkernel exec.
- Agent session B (different caller, ~13:30-13:50 local): knowledge.compose timed out at 300s twice while memory.recall and kg search returned fine.

System context: host load average was 70-90 during the window (heavy parallel builds), so CPU starvation is a contributing factor — but the failures were path-specific per caller rather than uniform, and the daemon log shows no trace for either path (see #887).

Suspects worth ruling in/out:
- sqlite-vec KNN IN-subquery non-monotonicity path in the vector store (affects memory.recall's vector leg)
- rerank leg in compose (embedder-CPU-bound)
- lock contention on the shared connection when a long vector query pins a reader

Ask: (1) daemon-side per-request duration logging for recall/compose (blocked on #887), (2) a bounded internal deadline on the vector leg so recall degrades to FTS-only with a warning instead of hanging past client timeouts, or at minimum an explicit slow-path log.

ohdearquant:
Retry from the atlas seat 2026-07-12 ~14:23 EDT: memory.recall(query=..., limit=3) via kkernel exec hung past a 5-minute kill — third reproduction. Caveat: NOT a quiet window (load averages 69-88 at probe time, build fan-out ongoing), so this confirms the symptom under load but does not yet isolate load vs service. The quiet-window datapoint is still owed — will run the identical probe when host load drops below ~10 and note the result here.

ohdearquant:
Incident timeline (2026-07-12, all EDT), correcting the recovery attribution:

- ~13:30 first embedding-path hang (compose); recall/remember hangs follow across 3 independent callers. SQL-path verbs healthy throughout. Writes were never lost — every timed-out remember committed; the hang was post-commit.
- ~14:20 wedged daemon inspected: 158% CPU with zero embedding-path throughput, while system load had already dropped from ~90 to ~40 — inconsistent with pure load contention. SIGTERM ignored; SIGKILL at 14:22. Unsymbolicated 3s process sample captured (available on request).
[...]

PR #919, opened 2026-07-13
fix(pack-memory): bound memory.recall with a fail-soft deadline

ohdearquant:
## Summary

`memory.recall` has no end-to-end deadline. Under sustained load the full pipeline
(profile resolution, FTS + vector candidate gather with its ANN-overfetch retry loop,
hydration, fusion/scoring, MMR, and supersedes suppression) can run long enough that
MCP clients hit their own upstream ceiling (300s observed in production, #889) while
the call is still in flight server-side.

- Wraps `memory.recall` in a bounded `tokio::time::timeout` via a new
  `handle_recall_with_deadline` dispatch wrapper in `khive-pack-memory`.
- On timeout, the call now returns a typed `RuntimeError::DeadlineExceeded` error
  immediately instead of hanging toward the client-side ceiling.
- Budget defaults to 30s (`KHIVE_MEMORY_RECALL_DEADLINE_MS` env, or a
  `params.config.recall_deadline_ms` per-request override) — well above every
  per-stage latency observed in the #889 load-harness repro (single-digit-to-
  low-hundreds of milliseconds per stage at 20-way concurrency) and well short
  of the 300s ceiling it exists to preempt.
- This is layered on top of, and distinct from, `#836`'s `ann_ready_timeout_ms`,
  which bounds only a single cold-miss ANN-build wait inside the vector leg and
  degrades in-band to an FTS-only result. This deadline bounds the *entire*
  pipeline end to end.

## Root-cause investigation (honest state)

Repro work in `.khive/workspaces/20260712/889-recall-stall/` (three load-harness
runs against scratch databases, `search`/`get`/`neighbors` used as the healthy
comparator per the reported symptom) found:

- The previously-suspected sqlite-vec KNN IN-subquery bug is **not** the current
  mechanism — ruled out by reading `VectorStore::search` at current HEAD.
- `memory.recall` is structurally and measurably slower than `kg search(kind=entity)`
  under concurrent load: a consistent ~1.7-2.2x p50 gap at every tested concurrency
  level (N=1 through N=20), confirmed at two different corpus configurations
  (15x size-asymmetric and exactly-equal), ruling out corpus size as the sole cause.
- The ANN-overfetch retry loop and the embed/vector leg both stayed cheap throughout
  and were ruled out as the dominant contributor at repro scale.
- The leading structural explanation: `khive-pack-brain`'s `BrainPack` holds a single
  process-wide `std::sync::Mutex<BrainState>` covering all profiles/bindings/namespaces.
  `memory.recall` makes two `registry.dispatch()` hops into it on every call
  (`brain.resolve` then `brain.profile`) that `kg search` never makes at all — consistent
  with the `profile_resolve` stage showing the most erratic p50-to-tail spread of any
  stage under concurrency in both full sweeps.
- A targeted cold-vs-warm-connection test (checking whether the symptom is tied to the
  first call of a fresh session/connection rather than raw concurrency) did **not**
  reproduce a cold-call penalty at repro scale — an honest negative result.
- The worst latency reproduced (~1.5s p99 at N=20) is roughly two orders of magnitude
  short of the reported 300s incidents. The magnitude gap is most plausibly explained by
  production-scale corpus size and/or daemon-lifecycle events not triggered by this
  repro (a WAL checkpoint spike to 2.5x+ the TRUNCATE threshold was found independently
  in daemon logs from an incident window; the `config_id`-mismatch/respawn path in
  `daemon.rs`'s `forward_or_spawn`/`kill_and_respawn` is verb-agnostic and untested by
  this repro's setup, and is the strongest remaining unconfirmed lead).

Given that gap, this PR ships the fail-soft mitigation only (bounds the symptom
regardless of mechanism, matches the pattern `#836` already established for the
narrower ANN-wait case). The `Mutex` contention finding is real but not confirmed as
the sole or dominant cause of the acute 300s cases, so a structural fix (sharding or
replacing that lock) is intentionally left as a follow-up pending stronger evidence
rather than attempted in this PR.

## Tests

Three new regression tests in `khive-pack-memory/src/handlers/recall.rs`:

- `recall_889_deadline_exceeded_returns_typed_error_not_hang` — 1ms override, proves
  the deadline actually races the real pipeline and returns promptly (elapsed < 5s)
  with a typed `DeadlineExceeded { operation: "memory.recall", budget_ms: 1, .. }`,
  not just that the code compiles.
- `recall_889_normal_path_succeeds_within_default_deadline` — no-regression check on
  the happy path under the default 30s budget.
- `recall_889_generous_override_succeeds` — a 120s override succeeds identically to
  the default, proving the override is symmetric (not only-ever-tightens).

## Gates (scoped to the two touched crates, run from `crates/` with a shared
`CARGO_TARGET_DIR` given lane contention)

- `cargo check -p khive-pack-memory -p khive-runtime` — clean
- `cargo test -p khive-pack-memory recall_889` — 3 passed, 0 failed
- `cargo clippy -p khive-pack-memory -p khive-runtime --all-targets -- -D warnings` — clean
- `cargo fmt -p khive-pack-memory -p khive-runtime -- --check` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-pack-memory -p khive-runtime` — clean

Full workspace `cargo test` was not run given shared-lane build contention at the time
of this PR; the scoped gates above cover every file this PR touches.

Ref #889

### DIFF
--- a/crates/khive-pack-memory/src/handlers/recall.rs
+++ b/crates/khive-pack-memory/src/handlers/recall.rs
@@ -929,7 +929,9 @@ mod tests {
 
     use async_trait::async_trait;
     use khive_pack_kg::KgPack;
-    use khive_runtime::{EmbedderProvider, KhiveRuntime, Namespace, VerbRegistryBuilder};
+    use khive_runtime::{
+        EmbedderProvider, KhiveRuntime, Namespace, RuntimeError, VerbRegistryBuilder,
+    };
     use khive_storage::Entity;
     use lattice_embed::{EmbedError, EmbeddingModel, EmbeddingService};
     use serde_json::Value;
@@ -4803,5 +4805,160 @@ mod tests {
     fn entity_lookup_candidates_empty_query_returns_empty() {
         assert!(crate::scoring::entity_lookup_candidates("").is_empty());
         assert!(crate::scoring::entity_lookup_candidates("   ").is_empty());
+    }
+
+    /// #889: an absurdly small deadline override (1ms) must abort
+    /// `handle_recall` with a typed `DeadlineExceeded` error rather than let
+    /// the caller wait for the pipeline to finish. 1ms is comfortably below
+    /// this uncontended in-memory pipeline's minimum real wall time on any
+    /// machine, so the timeout branch fires deterministically without needing
+    /// to hold a lock or otherwise force contention (contrast with the #836
+    /// tests above, which simulate a specific held lock to force their
+    /// narrower ANN-wait timeout). Proves the wrap actually races the
+    /// deadline against the real pipeline and returns promptly on the
+    /// timeout branch, not just that the code compiles.
+    #[tokio::test]
+    async fn recall_889_deadline_exceeded_returns_typed_error_not_hang() {
+        let rt = KhiveRuntime::memory().expect("in-memory runtime");
+        let ns = Namespace::parse("local").expect("local namespace");
+        let token = rt.authorize(ns).expect("authorize local");
+
+        rt.create_note(
+            &token,
+            "memory",
+            None,
+            "issue 889 deadline exceeded test note",
+            Some(0.7),
+            None,
+            vec![],
+        )
+        .await
+        .expect("create note");
+
+        let mut builder = VerbRegistryBuilder::new();
+        builder.register(KgPack::new(rt.clone()));
+        builder.register(MemoryPack::new(rt.clone()));
+        let registry = builder.build().expect("registry");
+
+        let start = std::time::Instant::now();
+        let result = registry
+            .dispatch(
+                "memory.recall",
+                serde_json::json!({
+                    "query": "889 deadline exceeded test",
+                    "limit": 10,
+                    "config": { "recall_deadline_ms": 1 }
+                }),
+            )
+            .await;
+        let elapsed = start.elapsed();
+
+        assert!(
+            elapsed < std::time::Duration::from_secs(5),
+            "#889 a timed-out recall must return promptly, never hang toward \
+             an upstream client-side ceiling (300s observed in production); \
+             took {elapsed:?}"
+        );
+
+        match result {
+            Err(RuntimeError::DeadlineExceeded {
+                operation,
+                budget_ms,
+                ..
+            }) => {
+                assert_eq!(operation, "memory.recall");
+                assert_eq!(budget_ms, 1);
+            }
+            other => panic!("#889 expected DeadlineExceeded, got: {other:?}"),
+        }
+    }
+
+    /// #889: the default (generous, 30s) deadline must not affect a normal,
+    /// fast, uncontended recall — no new error, no behavior change on the
+    /// happy path relative to pre-#889 behavior.
+    #[tokio::test]
+    async fn recall_889_normal_path_succeeds_within_default_deadline() {
+        let rt = KhiveRuntime::memory().expect("in-memory runtime");
+        let ns = Namespace::parse("local").expect("local namespace");
+        let token = rt.authorize(ns).expect("authorize local");
+
+        rt.create_note(
+            &token,
+            "memory",
+            None,
+            "issue 889 normal path recall note",
+            Some(0.7),
+            None,
+            vec![],
+        )
+        .await
+        .expect("create note");
+
+        let mut builder = VerbRegistryBuilder::new();
+        builder.register(KgPack::new(rt.clone()));
+        builder.register(MemoryPack::new(rt.clone()));
+        let registry = builder.build().expect("registry");
+
+        let result = registry
+            .dispatch(
+                "memory.recall",
+                serde_json::json!({
+                    "query": "889 normal path recall",
+                    "limit": 10
+                }),
+            )
+            .await
+            .expect("recall must succeed within the default deadline");
+
+        let results = result.as_array().expect("recall result must be an array");
+        assert!(
+            !results.is_empty(),
+            "normal recall must surface the seeded note"
+        );
+    }
+
+    /// #889: a generous per-request override (well above real wall time) must
+    /// behave identically to the default — the override path is symmetric,
+    /// not "only ever tightens".
+    #[tokio::test]
+    async fn recall_889_generous_override_succeeds() {
+        let rt = KhiveRuntime::memory().expect("in-memory runtime");
+        let ns = Namespace::parse("local").expect("local namespace");
+        let token = rt.authorize(ns).expect("authorize local");
+
+        rt.create_note(
+            &token,
+            "memory",
+            None,
+            "issue 889 generous override recall note",
+            Some(0.7),
+            None,
+            vec![],
+        )
+        .await
+        .expect("create note");
+
+        let mut builder = VerbRegistryBuilder::new();
+        builder.register(KgPack::new(rt.clone()));
+        builder.register(MemoryPack::new(rt.clone()));
+        let registry = builder.build().expect("registry");
+
+        let result = registry
+            .dispatch(
+                "memory.recall",
+                serde_json::json!({
+                    "query": "889 generous override recall",
+                    "limit": 10,
+                    "config": { "recall_deadline_ms": 120_000 }
+                }),
+            )
+            .await
+            .expect("recall must succeed under a generous override");
+
+        let results = result.as_array().expect("recall result must be an array");
+        assert!(
+            !results.is_empty(),
+            "normal recall must surface the seeded note under a generous override"
+        );
     }
 }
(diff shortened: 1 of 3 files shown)


ohdearquant:
## Codex review round 1 (head 5769f6fc)

APPROVE-WITH-FIXES

Reviewed PR #919 at `5769f6fcd5fbc853b3c18ddc2c0efb170fefba1f`.

### [Medium] Validate positive deadline values instead of silently coercing them

Evidence: `crates/khive-pack-memory/src/pack.rs:509` accepts only `as_u64()` and silently falls through for negative or non-numeric request values at `:512-513`; `:516` then accepts zero as a zero-millisecond timeout. The environment parsing at `:543-546` likewise accepts `0`. `RecallParams` parses the nested config through `RecallConfig` (`crates/khive-pack-memory/src/handlers/common.rs:260-268`), but `RecallConfig` has no deadline field and uses permissive `#[serde(default)]` (`crates/khive-pack-memory/src/config.rs:40-42`), so those malformed request values do not later fail validation.

Why this matters: `recall_deadline_ms: 0` makes normal recall immediately time out, while negative/non-numeric request values are silently converted to the environment/default budget. This conflicts with the requested invalid-value behavior and ADR-033's requirement that invalid recall config return a clear per-op error rather than silently coerce.

Suggested fix: parse the deadline with an explicit positive-integer validator before starting the timeout. Reject invalid per-request values as `InvalidInput`; treat invalid/zero environment values as a logged/config-ledgered fallback to 30,000 ms. Add coverage for `0`, negative, and non-numeric request/env values, plus the intended precedence.

### [Medium] Make the deadline regression test control a genuinely slow pipeline stage

Evidence: `crates/khive-pack-memory/src/handlers/recall.rs:4810-4819` calls the 1 ms case deterministic, but the test at `:4821-4873` only seeds the normal in-memory implementation and supplies `recall_deadline_ms: 1` (`:4843-4853`). It does not hold the contention point or inject a delayed embedder/storage path. The normal and generous tests at `:4880-4917` and `:4924-4962` also do not cover invalid env/request inputs or the MCP batch envelope.

Why this matters: the test does fail if this wrapper is deleted because it expects `DeadlineExceeded`, but whether the real in-memory path takes more than 1 ms is host/load dependent. It is therefore not a deterministic regression proof for the deadline behavior under the contention that caused #889.

Suggested fix: register a test-only embedding provider (or controlled storage/brain stage) that waits on a `Notify`/paused Tokio clock, then assert a comfortably sized deadline returns `DeadlineExceeded` promptly. Exercise it through an MCP parallel batch with a succeeding sibling so the required `{ok:false, error}` isolation and deadline text are also covered.

## Looks Right

- The timeout is installed at the public `memory.recall` dispatch arm (`crates/khive-pack-memory/src/pack.rs:461-464`) and wraps the complete `handle_recall` future (`:515-519`), rather than one stage. That function includes profile resolution (`crates/khive-pack-memory/src/handlers/recall.rs:154-200`), the FTS/vector candidate pipeline, fusion (`:361`), scoring/ranking, and response construction.
- Precedence for valid values is coherent: a valid per-request unsigned value wins (`crates/khive-pack-memory/src/pack.rs:509-513`); otherwise the process-wide environment value is used, falling back to 30,000 ms (`:539-552`).
- `RuntimeError::DeadlineExceeded` names the operation and budget in its display text (`crates/khive-runtime/src/error.rs:270-287`). The MCP parallel dispatcher maps a non-`Khive` runtime error to one `{ok:false, tool, error}` entry (`crates/khive-mcp/src/server.rs:915-932`) and joins independent operations without aborting siblings (`:935-945`).
- The normal success path is an exact passthrough (`crates/khive-pack-memory/src/pack.rs:521`), so it preserves the former response shape and ordering. Deadline expiry drops the handler before its post-result serve-ledger enqueue and recall-feedback mutation (`crates/khive-pack-memory/src/handlers/recall.rs:810-864`); background ANN work is separately tracked by the existing pipeline. I found no half-committed serve-ledger or feedback write on the timeout path.
- The added direct-runtime test is not a pre-cancelled-future assertion and explicitly checks the typed error, operation, and budget (`crates/khive-pack-memory/src/handlers/recall.rs:4863-4873`).

## Commands Run

- `git status --short --branch`, `git log -1`, `gh pr view 919 --json ...`: checkout is clean and matches the requested PR head `5769f6fcd5fbc853b3c18ddc2c0efb170fefba1f`.
- `git diff 737b2cc9abbde470f5a533307e24cef51c4eda84..HEAD`, changed-file and adjacent runtime/MCP/ADR inspection: reviewed the three changed files plus the full recall path, request dispatcher, ADR-021, ADR-033, and ADR-091.
- `git diff --check 737b2cc9abbde470f5a533307e24cef51c4eda84..HEAD`: passed.
- `lsof /private/tmp/khive-shared-target/debug/.cargo-build-lock`: lock held by Cargo PIDs 4642, 5108, and 5792; therefore `cargo test -p khive-pack-memory recall_889` was intentionally not run, as required.

## What I Did Not Check

- No Cargo test, check, clippy, fmt, or doc gate was started because the required shared build lock was busy.
- No live reproduction of WAL starvation or native sqlite-vec cancellation behavior; this is a static review.

## Re-Review Guidance

Narrow re-review after positive-value validation and a deterministic delayed-stage/batch-envelope test are added.

Domain utility: MEDIUM — ADR-033 and ADR-091 supplied the relevant recall-validation and WAL/cancellation contracts; no separate domain composition was available.

ohdearquant:
## Codex review round 2 (head aea1838a)

APPROVE

Narrow round-2 review of PR #919 at `aea1838ae5add1576dc747d4f0ea51369cb138a0`, limited to the two round-1 Medium fixes and their delta from `5769f6fc`.

## Resolved Round-1 Findings

- **Deadline override validation is now strict at the dispatch entry.** `crates/khive-pack-memory/src/pack.rs:512-515` invokes `parse_recall_deadline_override` before starting the timeout. `:544-560` accepts only a positive JSON unsigned integer; absent and `null` fall through, while `0`, negative values, and strings return `RuntimeError::InvalidInput`. The full dispatch test verifies the per-op error path for zero at `crates/khive-pack-memory/src/handlers/recall.rs:4964-4983`; the pure parser coverage exercises zero, negative, non-numeric, and array inputs at `:4899-4911`.
- **Invalid environment input cannot fail recall dispatch.** `crates/khive-pack-memory/src/pack.rs:575-600` is total over `Option<&str>`: absent input, zero, negative input, non-numeric input, and empty input all produce the 30,000 ms fallback (with warnings for supplied invalid values). `recall_deadline_ms` consumes that pure `u64` result without an error path at `:611-621`; focused coverage is at `crates/khive-pack-memory/src/handlers/recall.rs:4920-4934`.
- **The timeout regression is deterministic.** `SlowEmbedService::embed` blocks on test-controlled `Notify` at `crates/khive-pack-memory/src/handlers/recall.rs:4823-4836`. The test consumes exactly one setup permit, holds the subsequent query embedding, uses a 50 ms deadline, and releases only after dispatch returns at `:4995-5068`. Removing the timeout wrapper would leave that dispatch waiting on the held notification rather than allowing a host-speed-dependent pass.
- **Sibling isolation is adequately tested at the correct lower boundary.** The new `tokio::join!` test holds one recall past its deadline and requires an independent `stats` dispatch to succeed at `crates/khive-pack-memory/src/handlers/recall.rs:5115-5155`. `khive-pack-memory` correctly has no upward `khive-mcp` dependency (`crates/khive-pack-memory/Cargo.toml:15-37`), while MCP depends on the pack (`crates/khive-mcp/Cargo.toml:16-22`). The unchanged MCP batch executor independently maps registry errors to `{ok:false, tool, error}` and `join_all`s every operation at `crates/khive-mcp/src/server.rs:915-945`; therefore the registry-boundary test covers the reachable concurrent behavior and leaves no material wire-level gap.

## Scope Check

- The `5769f6fc..aea1838a` delta changes only `crates/khive-pack-memory/src/pack.rs` and `crates/khive-pack-memory/src/handlers/recall.rs`.
- Its only production behavior changes are the intended strict request override validation and fail-open environment parsing. The remaining changes replace the flaky test with the controlled embed fixture and add focused coverage. `git diff --check` passed.

## Commands Run

- `git status --short --branch`, `git rev-parse HEAD`, `gh pr view 919 --json ...`: clean review checkout at the requested head.
- `git diff --check 5769f6fc..HEAD`, targeted changed-file/ADR-033/runtime/MCP dependency and dispatch inspection: passed/static-reviewed.
- Ruby non-blocking flock check and `lsof /private/tmp/khive-shared-target/debug/.cargo-build-lock`: lock held by Cargo PID 91661. Per instruction, `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-pack-memory recall_889` was not run.

## What I Did Not Check

- No Cargo test was started because the required shared build lock was busy.
- This was intentionally a narrow re-review; no unrelated PR behavior was re-reviewed.

Domain utility: SKIPPED — the accepted ADR-033 contract and repository code supplied the needed review context; no separate domain-composition source was available.

oceanwaves630 (approved):
Codex r2 clean at this head; typed deadline + deterministic timeout regression close the #889 symptom lane. Approving.

Issue #580, opened 2026-07-04
WAL checkpoint starvation under multi-session MCP concurrency: 15.5GB WAL on 3.7GB db, writer timeouts


ohdearquant:
## Observed (live incident, 2026-07-04)

- `~/.khive/khive.db` = 3.7GB; `khive.db-wal` = **15.5GB** (15,512,941,272 bytes); `-shm` = 30MB.
- Symptom: repeated `sqlite: invalid data: timed out after 5s waiting for sqlite writer connection` on `comm.send` (and presumably any write) while several concurrent implementer agents were running.
- `PRAGMA wal_checkpoint(PASSIVE)` → `0|3768965|44`: not busy, 3,768,965 frames in the log, **only 44 checkpointable**. A long-lived read snapshot pins the WAL tail, so checkpoints cannot advance and the WAL grows without bound.
- Process census at incident time: 1 `kkernel mcp --daemon` + 7 `kkernel mcp [--db ...]` stdio sessions, all with live parents, some sessions >24h old.

## Root-cause class

An MCP stdio session (or the daemon) holds a read transaction open for the lifetime of some long operation — or leaks one — and under multi-seat concurrency there is effectively always at least one old snapshot alive. SQLite WAL checkpointing can never pass the oldest reader's mark, so the WAL only grows. The 5s writer-timeout failures are the downstream symptom (writer + wal-index operations degrade as the WAL and shm grow).

## Why this matters (khivedb P0)

This is exactly the deployment shape we sell: many agents, one substrate, long-lived sessions. Unbounded WAL growth + write timeouts under normal load is a production-killer defect class, not an ops quirk.

## Suggested investigation

1. Audit session/daemon code paths for read transactions held across awaits or across the whole session (connection-pool reader checkout that is never returned; `BEGIN` without commit on read paths).
2. Consider a periodic `wal_checkpoint(TRUNCATE)` maintenance hook in the daemon plus a WAL-size watchdog metric (log a warning past N MB).
3. Consider `PRAGMA journal_size_limit` so a successful checkpoint shrinks the file.
4. Repro: open one connection, `BEGIN; SELECT 1;` and hold it; run sustained writes from a second connection; watch `wal_checkpoint(PASSIVE)` frame counts freeze while WAL grows.

## Prior related lessons

Historic WAL bloat from embedding UPDATEs was mitigated by FTS trigger narrowing; this incident is a distinct mechanism (reader-pinned checkpoint starvation), observed with that fix already in place.

[...]

PR #921, opened 2026-07-13
fix(db): ADR-091 Plank 1 background age sweep over tx_registry

ohdearquant:
## Summary

Closes the implementation gap between ADR-091 and current `khive-db` for issue #580. Live incident tonight: the daemon logged

```
WARN khive_db::checkpoint: WAL high-water mark exceeded; sustained WAL pressure — a long-lived
reader may be pinning an old snapshot that PASSIVE cannot reclaim wal_pages=52054 high_water=6000
```

Detection fired correctly (8.7x the 6000-page ceiling), but nothing escalated further — Plank 1 (bounded transaction lifetime) had never actually been built.

## Gap table (ADR-091 vs. `crates/khive-db/src/checkpoint.rs` @ `85d30db9`)

| ADR-091 mandate | Status before this PR | Status after |
|---|---|---|
| **Plank 0** — open-transaction registry (`tx_registry`), per-tick oldest-entry debug trace, WARN on `warn_pages`/`high_water_pages` crossing naming the pinning caller | **Shipped.** `khive-storage/src/tx_registry.rs` registers at all 7 stores + `sql_bridge.rs` + `pool.rs` + `writer_task.rs`; `checkpoint.rs`'s `log_tx_registry_oldest_warn`/`log_tx_registry_snapshot_warn` fire on crossing; also exposed on the daemon metrics-read surface (`daemon.rs` `build_metrics_snapshot`). | Unchanged. |
| **Severity ladder** (INFO → WARN → ALARM, khive#617) | **Shipped.** `CheckpointSeverityState` implements the full 3-cycle-sustained-pressure ladder. | Unchanged. |
| **Plank 2** — TRUNCATE escalation past `truncate_high_water_pages`, rate-limited, backoff on writer-busy skip | **Shipped.** `maybe_truncate`, `note_truncate_outcome`, the whole `KHIVE_WAL_TRUNCATE_*` knob set. | Unchanged. |
| **Plank 1** — bounded caller-controllable transaction lifetime: `KHIVE_TX_WARN_SECS`/`KHIVE_TX_MAX_AGE_SECS`, soft-cap WARN, "cooperative stale-op guard" (reject further statements / roll back a late `commit()`), background registry sweep logging any stale entry every tick | **Not built.** `tx_registry.rs`'s own module doc: *"This is observe-only: no enforcement reads the registry in this plank."* Zero occurrences of either env var in the tree. This is the gap tonight's incident exposed: after the one-shot high-water WARN, sustained pressure went silent. | **Partially built, adapted** (see below). |

### Why "adapted," not a literal transplant

Plank 1's per-statement guard ("reject further `execute`/`query_row`/`query_all` calls; roll back a `commit()` past the cap") was designed against `SqliteTransaction`/`begin_tx`. **That API no longer exists in this codebase** — ADR-067 (`write-owner-daemon`, landed after ADR-091) introduced `SqlAccess::atomic_unit`, a closure-scoped write API where the caller's closure runs inside the writer task's own transaction and must complete on its first poll (enforced at runtime on the write-queue path). This *is*, in substance, Plank 1's own named follow-up ("closure-scoped transaction API... eliminating the 'held across an await' class of risk entirely... tracked here as a candidate future ADR") — already delivered for every production write path by a later ADR. There is no live call site left holding a caller-controlled handle across multiple statements for a per-statement reject to intercept.

**What ships in this PR**: the part of Plank 1 that still applies verbatim — "applied uniformly to every registered transaction regardless of which mechanism created it." `TxAgeSweepState` (`checkpoint.rs`) checks `tx_registry::oldest()`'s age against `KHIVE_TX_WARN_SECS` (default 30s) / `KHIVE_TX_MAX_AGE_SECS` (default 120s) on every observed checkpoint tick, independent of WAL page pressure, and escalates to `warn!`/`error!` on each below→above crossing — same debounce idiom as the existing WAL-pressure severity ladder, so a sustained stale span logs once per rung rather than spamming every 500ms tick. This is visibility, not reclamation, matching the ADR's own accepted gap ("a transaction... held idle across an await with no further calls... does not force-close the connection").

### What's still open (named, not invented)

`crates/khive-db/src/stores/graph.rs`'s `traverse` opens a deferred read transaction and holds it across a `roots.chunks(400)` loop — its own comment already calls it "the most WAL-pin-relevant span in the store," and it's registered in `tx_registry`, so this PR's sweep does surface it. But *bounding or aborting* that traversal past an age cap is a genuine new design decision (which cap, whether a partial-result error is acceptable to callers) that ADR-091's own Inventory item (3) explicitly left as an open question, not a specified mechanism — so it is **not** implemented here; it's recorded as a follow-up in the ADR amendment. Also un-ruled-out: the pin may be outside this process entirely (a separate `kkernel mcp` stdio session), which `tx_registry` — a process-local singleton — structurally cannot see; that's the ADR's own already-deferred "route reads through the daemon" alternative.

## Changes

- `crates/khive-db/src/checkpoint.rs`: `CheckpointConfig::{tx_warn_secs, tx_max_age_secs}` (+ env parsing), `TxAgeRung`/`TxAgeEmission`/`TxAgeSweepState` (pure state machine, mirrors `CheckpointSeverityState`), `log_tx_age_emission`, wired into `run_checkpoint_task`'s tick loop. Module-level doc updated to explain the Plank-1-vs-`atomic_unit` divergence.
- `docs/adr/ADR-091-wal-snapshot-lifetime.md`: Config-summary status updated for the two knobs; new "2026-07-12 amendment" section recording the `atomic_unit` divergence and the open `graph.rs traverse` / cross-process follow-ups.

## Tests (8 new, `-p khive-db`)

- `tx_age_sweep_empty_registry_emits_nothing`, `tx_age_sweep_fresh_entry_emits_nothing` — no false positives.
- `tx_age_sweep_warn_fires_once_on_crossing`, `tx_age_sweep_stale_fires_once_on_crossing` — edge-triggered, no per-tick spam.
- `tx_age_sweep_already_stale_entry_emits_both_rungs_same_tick`, `tx_age_sweep_rearms_after_entry_clears` — boundary + re-arm.
- `tx_age_sweep_uses_configured_thresholds_not_hardcoded_defaults` — closes the env-var-to-emitted-rung loop.
- `log_tx_age_emission_carries_label_for_both_rungs` — `CaptureSubscriber`-based, asserts actual message text + label.
- `tx_age_sweep_names_long_lived_reader_pinning_wal_past_high_water` — **the incident regression**: opens a real `BEGIN DEFERRED` reader pinning a WAL snapshot (same isomorphism as the existing `checkpoint_high_water_does_not_block_behind_reader` test), registers it in `tx_registry`, drives writes past `high_water_pages`, and asserts the sweep escalates to `Stale` naming that exact reader.

## Gate table

| Gate | Result |
|---|---|
| `cargo test -p khive-db --lib checkpoint::` | 37 passed |
| `cargo test -p khive-db` (full) | 289 unit + 10 contract + 0 doctest, all passed |
| `cargo clippy -p khive-db --all-targets -- -D warnings` | clean |
| `cargo fmt --all -- --check` (via pre-commit) | clean |
| `deno fmt` (ADR doc) | clean |

All run with `CARGO_TARGET_DIR=/private/tmp/khive-shared-target`, `cargo` invoked from `crates/`.

## Test plan

- [x] Unit tests for `TxAgeSweepState` cover empty/fresh/warn-crossing/stale-crossing/both-same-tick/re-arm.
- [x] Integration test reproduces the incident shape (real pinned reader + writes past high-water) and asserts the new mitigation fires.
- [x] Full `khive-db` suite green, no regressions.
- [ ] Deploy + watch the daemon's own logs for the new `ADR-091 Plank 1` warn/error lines under real production WAL pressure (orchestrator follow-up, not verifiable from a PR).

Refs #580, ADR-091.

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -34,7 +34,9 @@
 //! high-water snapshot enumeration) ride the SAME crossing gates — they are
 //! not independently rate-limited, so they never repeat on consecutive ticks
 //! that remain above a threshold. Only the per-tick `debug!` trace of the
-//! oldest open entry fires unconditionally.
+//! oldest open entry, and the ADR-091 Plank 1 age sweep below, run
+//! unconditionally on every tick — including a Skipped one; the sweep's own
+//! emissions stay edge-triggered per rung via `TxAgeSweepState`.
 //!
 //! ADR-091 Plank 2: rare TRUNCATE escalation. The periodic tick above stays
 //! PASSIVE-only and non-blocking; on top of it, `checkpoint_once` also
@@ -67,14 +69,19 @@
 //! is the ADR's own named follow-up ("closure-scoped transaction API"),
 //! already delivered for writes by a later ADR. What ships here instead is
 //! the part of Plank 1 that still applies to every registered span
-//! regardless of which mechanism created it: on each observed tick,
-//! `TxAgeSweepState` checks `khive_storage::tx_registry::oldest()`'s age
-//! against `tx_warn_secs`/`tx_max_age_secs` and escalates to `warn!`/`error!`
-//! on each below→above crossing (same debounce idiom as the WAL-pressure
-//! ladder — a sustained stale span logs once per rung, not once per tick).
-//! This is visibility, not reclamation: nothing here force-closes a stale
-//! span, matching the ADR's own accepted gap for a transaction "held idle
-//! across an await with no further calls."
+//! regardless of which mechanism created it: on EVERY tick — Skipped as well
+//! as Observed, since a registered `WriterGuard::transaction` span holds the
+//! writer mutex for its whole lifetime and would otherwise make the busiest,
+//! most relevant tick invisible to this sweep — `TxAgeSweepState` checks
+//! `khive_storage::tx_registry::oldest()`'s age against
+//! `tx_warn_secs`/`tx_max_age_secs` and escalates to `warn!`/`error!` on each
+//! below→above crossing (same debounce idiom as the WAL-pressure ladder — a
+//! sustained stale span logs once per rung, not once per tick), also
+//! re-arming both rungs if the oldest entry's identity changes between ticks
+//! so a departed span's latched state cannot suppress its replacement. This
+//! is visibility, not reclamation: nothing here force-closes a stale span,
+//! matching the ADR's own accepted gap for a transaction "held idle across
+//! an await with no further calls."
 
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
@@ -285,9 +292,10 @@ pub struct CheckpointConfig {
     /// shared open-transaction registry (`khive_storage::tx_registry`,
     /// covering every registered span regardless of which call site created
     /// it) is surfaced at `tracing::warn!` on the background sweep this
-    /// module runs every observed tick, independent of WAL page pressure —
-    /// a registered span can go stale while `wal_pages` sits well under
-    /// `warn_pages`.
+    /// module runs on EVERY tick — Skipped as well as Observed — independent
+    /// of WAL page pressure: a registered span can go stale while
+    /// `wal_pages` sits well under `warn_pages`, or while the writer is busy
+    /// and no WAL page count is sampled at all this tick.
     ///
     /// Overridable via `KHIVE_TX_WARN_SECS`.
     /// Default: 30 seconds.
@@ -404,6 +412,29 @@ impl CheckpointConfig {
                     cfg.tx_max_age_secs = Duration::from_secs(n);
                 }
             }
+        }
+
+        // The severity ladder assumes tx_warn_secs < tx_max_age_secs (Warn
+        // fires before Stale as an entry ages). A reversed or equal pair —
+        // whether from one misconfigured var or the interaction of both —
+        // would invert or collapse that ordering (e.g. WARN_SECS=120,
+        // MAX_AGE_SECS=30 emits Stale at 30s and never reaches the Warn
+        // crossing until 120s), so both are rejected together rather than
+        // silently honored. Resetting both to their defaults (rather than
+        // just clamping one) avoids guessing which of the two the operator
+        // actually meant to change.
+        if cfg.tx_warn_secs >= cfg.tx_max_age_secs {
+            let default = Self::default();
+            tracing::warn!(
+                configured_tx_warn_secs = cfg.tx_warn_secs.as_secs_f64(),
+                configured_tx_max_age_secs = cfg.tx_max_age_secs.as_secs_f64(),
+                fallback_tx_warn_secs = default.tx_warn_secs.as_secs_f64(),
+                fallback_tx_max_age_secs = default.tx_max_age_secs.as_secs_f64(),
+                "KHIVE_TX_WARN_SECS must be strictly less than KHIVE_TX_MAX_AGE_SECS; \
+                 both transaction-age thresholds were rejected and reset to their defaults"
+            );
+            cfg.tx_warn_secs = default.tx_warn_secs;
+            cfg.tx_max_age_secs = default.tx_max_age_secs;
         }
 
         cfg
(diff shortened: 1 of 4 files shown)


ohdearquant:
## Codex review round 1 (head 1992c344)

APPROVE-WITH-FIXES

PR #921 review — `fix/580-adr091-wal-mitigation` at `1992c34412e5919348c20ce72c6c1122918b4f74`.

Findings: 1 High, 4 Medium, 0 Low

### [High] The age sweep is skipped whenever the checkpoint cannot acquire the writer

Evidence: `crates/khive-db/src/checkpoint.rs:730` runs `checkpoint_once`, but `:733-736` immediately `continue`s for `CheckpointTick::Skipped`; the only `TxAgeSweepState::observe(tx_registry::oldest())` call is later at `:754`. A registered `WriterGuard::transaction` holds the writer while registered (`crates/khive-db/src/pool.rs:221-222`), so the sweep cannot report precisely that long-running registered transaction.

Why this matters: The stated Plank 1 policy is to sweep on every checkpoint tick for every registered span. A writer-busy episode is a normal checkpoint tick (and is separately counted as such), but it silently bypasses the registry lookup. This leaves the sweep dependent on writer availability rather than solely on age, contrary to the requested WAL-independent diagnostic.

Suggested fix: Advance `tx_age_state` and emit its logs before the `CheckpointTick::Skipped` early-continue, then retain the existing WAL-threshold state behavior for observed ticks only. Add a task-level regression that holds the pool writer and a stale registry entry across several intervals and asserts the age alert still occurs.

### [Medium] Reversed transaction-age thresholds are accepted and emit ERROR before WARN

Evidence: `CheckpointConfig::from_env` accepts the two positive values independently at `crates/khive-db/src/checkpoint.rs:393-407`. The sweep evaluates both independently at `:610-625`. Thus `KHIVE_TX_WARN_SECS=120` and `KHIVE_TX_MAX_AGE_SECS=30` produces `Stale` at 30 seconds and only later produces `Warn` at 120 seconds. The invalid/zero tests at `:1567-1600` and `:1692-1745` do not cover this ordering.

Why this matters: A deployment typo reverses the intended severity ladder and violates the required `WARN < MAX` invariant rather than failing safe.

Suggested fix: After environment parsing, reject or normalize `tx_warn_secs >= tx_max_age_secs` (for example, restore both transaction thresholds to their defaults) and add a serialized environment test for the reversed pair.

### [Medium] A stale replacement can be suppressed, leaving the new caller unnamed

Evidence: `TxAgeSweepState` remembers only two above-threshold booleans (`crates/khive-db/src/checkpoint.rs:574-580`) and emits only when either changes from false to true (`:613-625`). If stale entry A is observed, then A closes while already-stale entry B becomes the oldest before the next observed tick, both booleans remain true and B generates no emission. The comment at `:590-596` incorrectly assumes a replacement will naturally be fresh. The re-arm test at `:2438-2466` inserts an observed empty/fresh state, so it misses this sequence.

Why this matters: The alert can remain latched to a departed caller while a different long-lived transaction is now pinning the database—the exact caller identification this PR is meant to provide is lost.

Suggested fix: Have the registry expose a stable entry identity (or opened timestamp) with `oldest`, reset/re-emit the relevant rungs when that identity changes, and test stale-A → stale-B without an intervening below-threshold observation.

### [Medium] The amended ADR still asserts enforcement that this PR deliberately does not implement

Evidence: The amendment correctly says the shipped feature is logging only and that no reject/rollback exists at `docs/adr/ADR-091-wal-snapshot-lifetime.md:600-611`. But the still-live Plank 1 text requires statement rejection and rollback at `:303-338`; the config table describes `KHIVE_TX_MAX_AGE_SECS` as a rollback guard at `:458-463`; and Failure modes/Consequences repeat that claim at `:505-515` and `:545-553`. In contrast, `crates/khive-db/src/checkpoint.rs:638-657` only logs WARN/ERROR, with no reclamation.

Why this matters: ADR-091 is an accepted contract. Its internal contradiction would lead a future implementer or operator to believe a stale transaction is blocked or rolled back when it is only observed.

Suggested fix: Mark the superseded Plank 1 enforcement paragraphs as historical (or rewrite them) and update the config summary, Failure modes, and Consequences to describe the implemented background age sweep and its two explicitly deferred enforcement decisions.

### [Medium] The 37 checkpoint tests do not exercise the task-level sweep wiring

Evidence: The only production wiring is `crates/khive-db/src/checkpoint.rs:754-756`. Most new tests invoke the state machine directly, including the no-transactions case at `:2301-2306`. The incident-style test runs `checkpoint_once` at `:2554-2558` and then directly calls `tx_age_state.observe(...)` at `:2571-2572`; it never starts `run_checkpoint_task`. Therefore deleting `:754-756` leaves the state-machine and integration assertions intact.

Why this matters: The tests validate threshold calculations and log formatting, but not the required behavior that a real checkpoint task sweeps an aged registry entry while WAL pressure remains below its thresholds (or while the task is skipped for a busy writer).

Suggested fix: Add async task-level tests that capture the emitted age alert for (1) a stale labeled entry with a healthy WAL, (2) an empty registry, and (3) a writer-busy tick. Those tests should fail if the call at `:754-756` is removed.

## Looks Right

- For an observed checkpoint, the new sweep at `crates/khive-db/src/checkpoint.rs:748-756` is outside `warn_pages`, `high_water_pages`, and TRUNCATE branches; WAL pressure does not gate it.
- Invalid and zero transaction environment values fall back to defaults (`checkpoint.rs:393-407`, `:1567-1600`, `:1692-1745`).
- All 21 non-test `tx_registry::register` calls found under `crates/` pass an explicit `Some("…")` label, including the seven store families; the logger also renders an unlabeled fallback at `checkpoint.rs:639-655`.
- The changed implementation is observe-and-escalate only: `log_tx_age_emission` uses `tracing::warn!`/`error!` (`checkpoint.rs:638-657`) and adds no kill, rollback, or force-close behavior.
- The amendment candidly defers a graph traversal lifetime bound and cross-process pins at `docs/adr/ADR-091-wal-snapshot-lifetime.md:613-627`; those deferrals match the code’s process-local registry.

## Commands Run

- `git status --short`, `git rev-parse HEAD`, and `gh pr view 921 …`: checkout was clean and matched PR head `1992c34412e5919348c20ce72c6c1122918b4f74`.
- `cargo fmt --check` from `crates/`: passed.
- Static count: 32 `#[test]` plus 5 `#[tokio::test]` = 37 checkpoint tests.
- Shared-build gate: a non-blocking exclusive-lock probe on `/private/tmp/khive-shared-target/debug/.cargo-build-lock` returned busy. `cargo test -p khive-db --lib checkpoint::` was therefore not run, as requested.

## What I Did Not Check

- Focused Cargo tests were skipped because the required shared build lock was not available.
- No multi-process runtime experiment was run; cross-process pins remain an explicitly documented deferral.

## Re-Review Guidance

Narrow re-review after the five fixes above, with the focused checkpoint test run once the shared build lock is free.

Domain utility: SKIPPED — the accepted ADR corpus and adjacent runtime code supplied the required review contract.

ohdearquant:
## Codex review round 2 (head 11406c74)

APPROVE

# PR #921 narrow round-2 re-review

Reviewed `fix/580-adr091-wal-mitigation` at `11406c746600f48207c8098a19f4a4735527481a` against delta `1992c344..HEAD`.

Findings: 0 High, 0 Medium, 0 Low

## Round-1 fixes verified

- The transaction-age sweep is now executed at `crates/khive-db/src/checkpoint.rs:798` before the only `CheckpointTick::Skipped` early-continue at `:805`. `checkpoint_once` converts both writer contention (`:1010-1015`) and PASSIVE-checkpoint errors (`:1020-1027`) into a tick result rather than returning from `run_checkpoint_task`; consequently the sweep runs on skipped and checkpoint-error paths.
- `CheckpointConfig::from_env` rejects equal or reversed age thresholds and restores both defaults at `crates/khive-db/src/checkpoint.rs:417-438`. The new equal/reversed tests passed.
- `TxAgeSweepState` stores the oldest `TxId` and resets both latches before evaluating a changed identity at `crates/khive-db/src/checkpoint.rs:612-680`. A stale A -> stale B replacement therefore emits B immediately; repeated changes, including synthetic A -> B -> A, likewise re-evaluate and cannot suppress a legitimate crossing. The replacement regression is at `:2638-2681`.
- `TxId` is used only as an identity value returned by `oldest` (`crates/khive-storage/src/tx_registry.rs:74-95`). Although its tuple field is public to support identity-only state-machine tests, no public registry operation accepts a caller-provided `TxId`, and `TxHandle.id` remains private (`:40-42`); constructing a value cannot insert, remove, or otherwise violate registry invariants.
- ADR-091 now clearly records the old reject/rollback design as superseded at `docs/adr/ADR-091-wal-snapshot-lifetime.md:303-320` and `:521-528`; the live config table, failure mode, consequences, and amendment consistently describe logging-only observe-and-escalate behavior at `:474-485`, `:548-559`, `:581-613`, and `:643-663`. I found no live promise of rejection, rollback, or forced reclamation.
- The three task-level tests spawn `run_checkpoint_task`, capture tracing, and assert the production wiring: healthy stale entry `crates/khive-db/src/checkpoint.rs:3051-3107`, empty registry `:3112-3156`, and writer-busy-every-tick `:3165-3241`. Removing the sole production sweep call at `:798-800` would make the healthy and writer-busy tests fail; the latter also proves the skipped path was actually exercised (`:3215-3219`).

## Commands run

- `git rev-parse HEAD`: matched expected head `11406c746600f48207c8098a19f4a4735527481a`.
- `git diff --check 1992c344..HEAD`: passed.
- Shared-build lock probe on `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: free.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-db --lib`: the three new task-level tests passed; 294/295 tests passed. The sole failure was the pre-existing `tx_age_sweep_names_long_lived_reader_pinning_wal_past_high_water` test, which selected a concurrently registered `graph_upsert_edges` entry as genuinely oldest. Its source is unchanged in this delta and it passes in isolation, so this is an existing test-isolation race rather than a delta regression.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-storage --lib`: passed, 49/49.

## What I did not check

- This was the requested static, round-2 narrow review; I did not run workspace-wide CI or multi-process WAL experiments.

Domain utility: SKIPPED — the accepted ADR and directly adjacent code/tests supplied the needed contract.

oceanwaves630 (approved):
Lane C read at 11406c746600f48207c8098a19f4a4735527481a. Verified the sweep runs before the Skipped early-continue (and on checkpoint-error paths via checkpoint_once returning tick results), the identity-reset in TxAgeSweepState::observe fires both latch resets before the same-tick age check so a stale-successor replacement emits immediately, the reversed/equal env pair rejection restores both defaults, and TxId exposure is identity-only (TxHandle.id private, register() is the sole insertion path). ADR-091 now consistently observe-and-escalate with no live rollback/rejection promise. Pre-existing test-isolation race correctly filed as #926 rather than papered over. Auto-merge may arm.

PR #922, opened 2026-07-13
feat(storage): BlobStore content-addressed blob capability (ADR-111)

### LINKED ISSUE #292: feat(storage): BlobStore capability for large opaque payloads (content-addressed, object_store backend)
## Motivation

Entities increasingly need to carry a large, opaque raw payload **by reference** — a `document`
entity's source PDF, or rendered image tiles for a visual-retrieval path. SQLite is not a blob
backend: large BLOBs inflate the WAL and page cache and degrade the FTS/vector hot paths. Today
there is no first-class place for these bytes, so they either bloat the core DB or live ad-hoc
outside it with no GC story.

## Proposal

A new `BlobStore` capability trait in `khive-storage` (same trait family as ADR-005), with bytes
stored **outside** the core DB and addressed by content hash.

- **Content-addressed by blake3.** The hash is the object key; identical bytes dedupe for free.
- **Backend = the `object_store` crate** (one trait over local filesystem + S3 + GCS + Azure).
  **Filesystem-first**; **S3-standard** for cloud (R2 / MinIO are the same API, so the cloud swap
  is a config change, not a code change; per-tenant isolation is a bucket/prefix).
- **Reference column.** A dedicated **indexed `content_ref` column** on entities (one migration),
  not a key buried in the `properties` JSON — so GC can join on it cheaply.
- **GC.** An orphaned-blob sweep mirrors the existing `orphan_sweep` (ADR-044 §5): a blob with no
  referencing row is collectable.

## Scope

Ships **standalone** — no dependency on any embedding or multimodal work. A `document` entity
carrying its source PDF by pointer is reason enough on its own.

## Deliverables

- [ ] `BlobStore` trait in `khive-storage` (put / get / delete / exists, content-addressed).
- [ ] Filesystem `object_store` impl; S3-standard reachable by config.
- [ ] Migration: indexed `content_ref` column on entities.
- [ ] Orphan-blob GC mirroring `orphan_sweep`.
- [ ] New storage-capability ADR (ADR-005 family) recording the trait, the content-addressing
      scheme, the backend choice, and the GC contract.
- [ ] Round-trip test: put/get/delete + content-address dedup (same bytes → same key).

Labels: enhancement, needs-design.

ohdearquant:
## Summary

Closes #292. Adds a `BlobStore` capability trait (ADR-005 family) for storing large opaque payloads outside the core SQLite DB, addressed by content hash, plus a filesystem-backed implementation and an `entities.content_ref` reference column.

- `BlobStore` trait in `khive-storage`: `put` / `get` / `exists` / `delete`, content-addressed by **blake3** (`ContentRef`, canonical lowercase 64-hex). New `StorageCapability::Blob` variant.
- `FsBlobStore` in `khive-db`: 2-level hex-sharded filesystem layout; atomic publish via `tempfile::NamedTempFile` (same shard dir) → write + fsync + length-verify → `.persist()`; hash-first dedup on `put` (identical bytes never re-write).
- **Fail-closed free-space floor**: default 100GB, configurable, checked via `fs4::available_space` before any write. Refuses with `StorageError::CapacityFloor { capability, volume, available_bytes, floor_bytes }` — no silent degrade, no auto-spill.
- **Blob root resolution**: `KHIVE_BLOB_ROOT` env > caller-supplied `config_root` > default `<db_dir>/blobs`. Errors when none apply (e.g. in-memory backend with no override).
- `entities.content_ref`: new indexed column via migration **V10** (`010-entities-content-ref.sql`, additive `ALTER TABLE` + partial index), mirrored in the non-versioned `entities-ddl.sql` path. `Entity` struct + every read/write call site updated (khive-db entity store, khive-runtime curation/portability, khive-vcs sync).
- **Orphan GC**: `orphan_sweep` (dry_run + real) mirrors the existing ADR-044 `orphan_sweep` pattern — admin/CLI-only, deliberately **no MCP verb**. This is the only deletion path besides an explicit `delete(content_ref)`.
- **ADR-111** (`docs/adr/ADR-111-blob-store.md`) records the full design and two deliberate deltas from the issue's proposal:
  1. Hand-rolled `std::fs` + `tempfile` + `blake3` filesystem backend instead of the `object_store` crate — `BlobStore` is already the backend-swap seam, and ADR-086 set a defer-until-a-real-consumer-needs-it precedent for exactly this kind of abstraction.
  2. A pre-existing gap in `khive-runtime::curation::merge_entity_sql`'s `INSERT OR REPLACE` (it already drops `entity_type` on entity merge) is flagged but not fixed here — `content_ref` is deliberately left out of that same statement for consistency; the in-memory merge result struct still carries it forward.

## Test plan

- [x] `cargo test -p khive-storage -p khive-db` — 413 tests, 0 failures (includes new `FsBlobStore` unit tests: put/get/exists/delete round-trip, blake3 digest verification, dedup, capacity-floor refusal + error contents, orphan_sweep dry_run/real, blob-root resolution precedence; plus `content_ref` round-trip/default/query/batch-upsert tests and two migration tests for V10).
- [x] `cargo clippy -p khive-storage -p khive-db --all-targets -- -D warnings` — clean
- [x] `cargo clippy -p khive-runtime -p khive-vcs --all-targets -- -D warnings` — clean (ripple-site scope from the `content_ref` field addition)
- [x] `scripts/lint-sql.sh` — clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-storage -p khive-db` — clean

### DIFF
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -98,6 +98,11 @@ lru = "0.12"
 sha2 = "0.10"
 hex = "0.4"
 blake3 = { version = "1", default-features = false }
+# Cross-platform filesystem free-space query for BlobStore's fail-closed
+# floor check (khive#292). Pure Rust (rustix on Unix, windows-sys on
+# Windows) — no libc statvfs FFI to hand-maintain across the release
+# matrix's win32-x64 target.
+fs4 = "1.1"
 regorus = "0.10"
 rmcp = "1.7"
 futures = "0.3"
--- a/crates/khive-db/Cargo.toml
+++ b/crates/khive-db/Cargo.toml
@@ -31,10 +31,12 @@ crossbeam-queue = { workspace = true }
 parking_lot = { workspace = true }
 sha2 = { workspace = true }
 sqlite-vec = { workspace = true, optional = true }
+blake3 = { workspace = true }
+fs4 = { workspace = true }
+tempfile = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["full", "test-util"] }
-tempfile = { workspace = true }
 rusqlite = { workspace = true, features = ["bundled", "column_decltype"] }
 khive-storage = { version = "0.4.0", path = "../khive-storage" }
 khive-types = { version = "0.4.0", path = "../khive-types", features = ["serde"] }
--- a/crates/khive-db/sql/010-entities-content-ref.sql
+++ b/crates/khive-db/sql/010-entities-content-ref.sql
@@ -1,0 +1,12 @@
+-- V10: BlobStore content_ref reference column on entities (khive#292).
+--
+-- Nullable, indexed pointer into a BlobStore (khive-storage::BlobStore) —
+-- a hex BLAKE3 digest, not a key buried in `properties`, so orphan-GC can
+-- join on it cheaply. Storage does not validate that the referenced blob
+-- exists; callers publish the blob before setting this column.
+
+ALTER TABLE entities ADD COLUMN content_ref TEXT;
+
+CREATE INDEX IF NOT EXISTS idx_entities_content_ref
+    ON entities(content_ref)
+    WHERE content_ref IS NOT NULL;
--- a/crates/khive-db/sql/entities-ddl.sql
+++ b/crates/khive-db/sql/entities-ddl.sql
@@ -14,7 +14,8 @@ CREATE TABLE IF NOT EXISTS entities (
     updated_at     INTEGER NOT NULL,
     deleted_at     INTEGER,
     merged_into    TEXT,
-    merge_event_id TEXT
+    merge_event_id TEXT,
+    content_ref    TEXT
 );
 
 CREATE INDEX IF NOT EXISTS idx_entities_namespace ON entities(namespace);
@@ -23,3 +24,8 @@ CREATE INDEX IF NOT EXISTS idx_entities_kind_entity_type ON entities(namespace, 
 CREATE INDEX IF NOT EXISTS idx_entities_name ON entities(namespace, name);
 CREATE INDEX IF NOT EXISTS idx_entities_created ON entities(created_at DESC);
 CREATE INDEX IF NOT EXISTS idx_entities_merged_into ON entities(namespace, merged_into);
+-- BlobStore content_ref reference (khive#292) — mirrors migration 010 for
+-- callers that create the table via ensure_entities_schema() without ever
+-- running the versioned migration chain (e.g. StorageBackend::memory() test
+-- setups that apply ENTITIES_DDL directly).
+CREATE INDEX IF NOT EXISTS idx_entities_content_ref ON entities(content_ref) WHERE content_ref IS NOT NULL;
--- a/crates/khive-db/src/backend.rs
+++ b/crates/khive-db/src/backend.rs
@@ -13,7 +13,7 @@ use rusqlite::OptionalExtension;
 use crate::error::SqliteError;
 use crate::pool::{ConnectionPool, PoolConfig};
 use crate::sql_bridge::SqlBridge;
-use crate::stores::{entity, event, graph, note, sparse, text, vectors};
+use crate::stores::{blob, entity, event, graph, note, sparse, text, vectors};
 
 /// Concrete storage backend providing capability traits.
 pub struct StorageBackend {
@@ -574,6 +574,23 @@ impl StorageBackend {
         )))
     }
 
+    /// Get a `BlobStore` rooted per khive#292's precedence chain:
+    /// `KHIVE_BLOB_ROOT` env var > `config_root` (a caller-resolved
+    /// `khive.toml` override — `khive-db` has no TOML parser of its own) >
+    /// beside this backend's database directory. `floor_bytes` overrides the
+    /// default 100 GB fail-closed free-space floor (`None` keeps the
+    /// default). Errors if none of the three roots apply — e.g. an in-memory
+    /// backend with no override and no env var has nowhere to default to.
+    pub fn blob_store(
+        &self,
+        config_root: Option<&Path>,
+        floor_bytes: Option<u64>,
+    ) -> Result<Arc<dyn khive_storage::BlobStore>, SqliteError> {
+        let root = blob::resolve_blob_root(self.data_dir().as_deref(), config_root)?;
+        let floor = floor_bytes.unwrap_or(blob::FsBlobStore::DEFAULT_FLOOR_BYTES);
+        Ok(Arc::new(blob::FsBlobStore::new(root, floor)?))
+    }
+
     /// Is this a file-backed backend?
     pub fn is_file_backed(&self) -> bool {
         self.is_file_backed
@@ -1014,6 +1031,51 @@ mod tests {
             result.is_err(),
             "upsert_document on a read-only backend must reject, not silently no-op"
         );
+    }
+
+    #[tokio::test]
+    async fn blob_store_roundtrip_via_public_api() {
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("blob_backend.db");
+        let backend = StorageBackend::sqlite(&path).unwrap();
+
+        // Explicit floor_bytes=0, not the default 100GB — the free space on
+        // whatever volume runs this test is not this test's concern (and a
+        // dev machine or CI runner legitimately may not clear 100GB free).
+        let store = backend.blob_store(None, Some(0)).unwrap();
+        let bytes = b"backend-level blob roundtrip".to_vec();
+        let content_ref = store.put(bytes.clone()).await.unwrap();
+        assert_eq!(store.get(&content_ref).await.unwrap(), bytes);
+    }
+
+    #[test]
+    fn blob_store_defaults_root_beside_db_file() {
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("blob_default.db");
+        let backend = StorageBackend::sqlite(&path).unwrap();
+
+        // `blob_store` creates the root directory eagerly (`FsBlobStore::new`),
+        // so its existence at the expected default path is directly
+        // observable without reaching into the trait object.
+        let _store = backend.blob_store(None, None).unwrap();
+        assert!(
+            dir.path().join("blobs").is_dir(),
+            "default root must be created beside the database file"
+        );
+    }
+
+    #[test]
+    fn blob_store_errors_for_in_memory_backend_with_no_override() {
+        let backend = StorageBackend::memory().unwrap();
+        assert!(backend.blob_store(None, None).is_err());
+    }
+
+    #[test]
+    fn blob_store_accepts_explicit_root_for_in_memory_backend() {
+        let dir = tempfile::tempdir().unwrap();
+        let backend = StorageBackend::memory().unwrap();
+        let store = backend.blob_store(Some(dir.path()), None);
+        assert!(store.is_ok());
     }
 
     #[test]
--- a/crates/khive-db/src/migrations.rs
+++ b/crates/khive-db/src/migrations.rs
@@ -118,6 +118,8 @@ const V8_UP: &str = include_str!("../sql/008-notes-seq-repair.sql");
 
 const V9_UP: &str = include_str!("../sql/009-entities-name-ci-index.sql");
 
+const V10_UP: &str = include_str!("../sql/010-entities-content-ref.sql");
+
 /// DDL for the `_embedding_models` registry table.
 ///
 /// Shared between the V1 schema and the belt-and-suspenders creation in
@@ -171,6 +173,11 @@ pub const MIGRATIONS: &[VersionedMigration] = &[
         version: 9,
         name: "entities_name_ci_index",
         up: V9_UP,
+    },
+    VersionedMigration {
+        version: 10,
+        name: "entities_content_ref",
+        up: V10_UP,
     },
 ];
 
--- a/crates/khive-db/src/migrations_tests.rs
+++ b/crates/khive-db/src/migrations_tests.rs
@@ -608,3 +608,56 @@ fn v6_implicit_mass_upsert_on_conflict() {
         .expect("count rows");
     assert_eq!(count, 1, "upsert must not create a second row");
 }
+
+// ── V10: entities.content_ref (khive#292) ───────────────────────────────────
+
+#[test]
+fn v10_adds_content_ref_column_and_partial_index() {
+    let mut conn = open_memory();
+    run_migrations(&mut conn).expect("migrations should succeed");
+    assert!(
+        column_exists(&conn, "entities", "content_ref"),
+        "V10 must add entities.content_ref"
+    );
+    assert!(
+        index_exists(&conn, "idx_entities_content_ref"),
+        "V10 must create idx_entities_content_ref"
+    );
+}
+
+#[test]
+fn v10_content_ref_defaults_null_and_accepts_a_value() {
+    let mut conn = open_memory();
+    run_migrations(&mut conn).expect("migrations should succeed");
+
+    conn.execute(
+        "INSERT INTO entities (id, namespace, kind, name, tags, created_at, updated_at) \
+         VALUES ('e1', 'local', 'concept', 'NullRef', '[]', 0, 0)",
+        [],
+    )
+    .expect("insert without content_ref");
+    let null_ref: Option<String> = conn
+        .query_row(
+            "SELECT content_ref FROM entities WHERE id = 'e1'",
+            [],
+            |row| row.get(0),
+        )
+        .expect("read content_ref");
+    assert_eq!(null_ref, None, "content_ref must default to NULL");
+
+    let digest = "a".repeat(64);
+    conn.execute(
+        "INSERT INTO entities (id, namespace, kind, name, tags, created_at, updated_at, content_ref) \
+         VALUES ('e2', 'local', 'concept', 'WithRef', '[]', 0, 0, ?1)",
+        rusqlite::params![digest],
+    )
+    .expect("insert with content_ref");
+    let stored_ref: Option<String> = conn
+        .query_row(
+            "SELECT content_ref FROM entities WHERE id = 'e2'",
+            [],
+            |row| row.get(0),
+        )
+        .expect("read content_ref");
+    assert_eq!(stored_ref, Some(digest));
+}
--- a/crates/khive-db/src/stores/blob.rs
+++ b/crates/khive-db/src/stores/blob.rs
@@ -271,9 +271,34 @@ impl BlobStore for FsBlobStore {
         let owned_guard = self.write_lock.clone().lock_owned().await;
         let root = self.root.clone();
         let floor_bytes = self.floor_bytes;
+        // Round-3 Medium (codex r3, PR #922): `sync_hook::take` is the
+        // test-only seam that lets regression tests observe/control
+        // exactly when this call is inside the guarded section, replacing
+        // fixed-sleep/fixed-duration-poll timing assumptions with
+        // deterministic, event-driven synchronization. `#[cfg(test)]`-
+        // gated end to end -- zero effect on non-test builds.
+        #[cfg(test)]
+        let hook = sync_hook::take(&root);
         tokio::task::spawn_blocking(move || {
-            let _owned_guard = owned_guard;
-            put_blocking(&root, floor_bytes, bytes)
+            // The guard lives in this inner block so it is dropped BEFORE
+            // the test hook's `done` signal fires below -- a test that
+            // waits on `done` and then immediately asserts the lock is
+            // free needs that ordering to hold exactly, not "usually".
+            #[cfg_attr(not(test), allow(clippy::let_and_return))]
+            let result = {
+                let _owned_guard = owned_guard;
+                #[cfg(test)]
+                if let Some(h) = &hook {
+                    let _ = h.reached.send(());
+                    let _ = h.release.recv();
+                }
+                put_blocking(&root, floor_bytes, bytes)
+            };
+            #[cfg(test)]
+            if let Some(h) = &hook {
+                let _ = h.done.send(());
+            }
+            result
         })
         .await
         .map_err(|e| StorageError::driver(StorageCapability::Blob, "put", e))?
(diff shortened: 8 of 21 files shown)


ohdearquant:
## Codex review round 1 (head fd8d635d)

REJECT

Findings: 0 Blocker, 3 High, 0 Medium, 0 Low

Reviewed PR #922 at `fd8d635d29753af0189b52bc0ac4bec63b7873e3` against `main` (`85d30db9d6eefc8d4346bfa2a2a8a2f1e1e57f9c`).

### [High] `ContentRef` deserialization bypasses the canonical-key invariant

Evidence: `crates/khive-storage/src/blob.rs:29` derives `Deserialize` for the transparent `ContentRef(String)`, while the only length/lowercase-hex validation is in `ContentRef::from_hex` at `crates/khive-storage/src/blob.rs:42`. `shard_path` then slices the unchecked string at `crates/khive-db/src/stores/blob.rs:29`.

Why this matters: `serde_json::from_str::<ContentRef>("\"x\"")` can construct an invalid public `ContentRef` without calling `from_hex`; passing it to `get`, `exists`, or `delete` reaches the unchecked `[0..2]`/`[2..4]` slices and panics. Longer malformed values also invalidate the documented opaque canonical-digest/path-component contract. This contradicts ADR-111's canonical 64-character lowercase-hex requirement.

Suggested fix: Implement `Deserialize` manually (or deserialize through a validated helper) so it calls `ContentRef::from_hex`; add serde rejection tests for short, uppercase, and non-hex values.

### [High] The capacity floor neither reserves nor accounts for the pending write

Evidence: `crates/khive-db/src/stores/blob.rs:77` checks only `available < floor_bytes`; the input length is not considered before the write at `crates/khive-db/src/stores/blob.rs:96`, and there is no lock or recheck before `persist` at `crates/khive-db/src/stores/blob.rs:118`. In contrast, `crates/khive-storage/src/error.rs:112` documents the error as accounting for the pending write.

Why this matters: with `available == floor_bytes + 1`, a new two-byte blob passes the check and can leave the volume below the configured floor. Concurrent puts can make the same race worse: each observes sufficient space and both write. This does not satisfy the requested fail-closed floor or the error's documented semantics.

Suggested fix: Reject when `available - required_write_bytes < floor_bytes` (using checked/saturating arithmetic), and serialize/check-and-publish writes for a root or otherwise reserve capacity so concurrent `FsBlobStore` puts cannot all pass a stale observation. Add boundary-size and concurrent-put regression tests.

### [High] Blob deletion is not safe against live or newly created entity references

Evidence: `BlobStore::delete` exposes an unconditional physical deletion at `crates/khive-storage/src/blob.rs:150`, implemented as `remove_file` at `crates/khive-db/src/stores/blob.rs:220`. `orphan_sweep` accepts a caller-built snapshot of `live_refs` at `crates/khive-storage/src/blob.rs:105`; `FsBlobStore` copies that snapshot and deletes files absent from it at `crates/khive-db/src/stores/blob.rs:235` and `crates/khive-db/src/stores/blob.rs:243`. It has no database anti-join, transaction, or coordination with an entity update. ADR-111 instead recommends a snapshot query and states that the design avoids removing a shared or briefly in-flight blob (`docs/adr/ADR-111-blob-store.md:181` and `docs/adr/ADR-111-blob-store.md:186`).

Why this matters: an entity can acquire a `content_ref` after the caller obtains the sweep snapshot but before `remove_file`; the sweep then deletes a blob now referenced by a live entity. Any caller of the public `delete` method can also immediately create that dangling reference. The ADR's safety claim is therefore not true for this implementation.

Suggested fix: Provide a database-coordinated admin sweep that selects and deletes under the entity writer/transactional boundary, and make reference-aware deletion the supported path (or make direct deletion explicitly unsafe and restrict it to verified offline maintenance). Amend ADR-111 to state the remaining concurrency guarantee precisely, and test the race.

## Looks Right

- V10 is appended after V9 in `crates/khive-db/src/migrations.rs:177`; V1's `sql/schema.sql` hash is unchanged. The migration is transactionally ledgered and rerunning migrations is tested. Both the versioned SQL and `entities-ddl.sql` add the nullable column and matching partial index; targeted tests assert the column/index.
- Normal blob publication creates the temp file in the target shard (`crates/khive-db/src/stores/blob.rs:87`-`95`), syncs and length-checks it, then calls `persist`; this is same-filesystem atomic publish. Hash-first dedup is present, and concurrent same-digest replacement cannot expose a torn file.
- The trait uses the repository's established `async_trait` pattern, remains object-safe behind `Arc<dyn BlobStore>`, and routes filesystem/space-query failures to `StorageError` rather than silently accepting them. `CapacityFloor` has the capability, volume, available bytes, and floor fields.
- Entity SQL upsert and all three read/query projections thread `content_ref`; the local storage-`Entity` literal sweep found the expected updated literals in `khive-db`, `khive-runtime`, and `khive-vcs`. The deliberate pre-existing merge omission remains exactly as ADR-111 documents (`crates/khive-runtime/src/curation.rs:1188`).
- No blob/orphan verb or handler registration was added to MCP, packs, or kkernel. ADR-111 records the intended object-store and configurable-floor deltas.

## Commands Run

- `git status --short --branch`, `gh pr view 922 ...`, and SHA checks: local `HEAD` exactly matches the requested PR head; worktree remained clean.
- Migration/DDL, public-type/literal, ADR, and verb-registration `rg`/source inspections: completed.
- After confirming `/private/tmp/khive-shared-target/debug/.cargo-build-lock` was not held: `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo test -p khive-storage -p khive-db` (from `crates/`): passed — 304 `khive-db` unit tests, 56 `khive-storage` unit tests, 53 integration/compliance tests, and doc tests.

## What I Did Not Check

- Full workspace tests, Clippy, formatting, and platform-specific filesystem behavior were not run; the requested targeted crate tests passed.
- Power-loss durability was not fault-injected. The atomic-publish confirmation is a source review of same-directory tempfile + rename, not a crash-fuzz result.

## Re-Review Guidance

Request a focused re-review after the three data-safety fixes. It should include deterministic deserialization validation, floor-boundary/concurrency tests, and an entity-reference race test for the chosen deletion design.

Domain utility: SKIPPED — the available repository memory was sufficient; no separate domain corpus was needed for this storage-contract review.

ohdearquant:
## Codex review round 2 (head 85dabc26)

REJECT

Findings: 0 Blocker, 1 High, 1 Medium, 0 Low

Focused round-2 re-review of PR #922 at `85dabc26a2d974e4c28f86025217cf6a1b101552` (the requested head, and a direct descendant of the round-1 head `fd8d635d29753af0189b52bc0ac4bec63b7873e3`).

### [High] H2's floor guard is not actually per-root or cancellation-safe

Evidence: `crates/khive-db/src/stores/blob.rs:179-208` gives every `FsBlobStore::new` call a fresh `tokio::sync::Mutex<()>`; `crates/khive-db/src/backend.rs:584-591` constructs a fresh `FsBlobStore` for every `StorageBackend::blob_store` call, even when those calls resolve to the same root. The claimed guarantee is instead a *per-root* mutex in `docs/adr/ADR-111-blob-store.md:116-123`. In addition, `crates/khive-db/src/stores/blob.rs:220-225` retains the borrowed guard only in the outer async `put` future, while the `spawn_blocking` closure owns only `root`, `floor_bytes`, and `bytes`.

Why this matters: two independently obtained stores for the same root have independent locks, so their `put`s can again observe the same free-space snapshot and both persist. Also, cancelling/dropping a `put` future drops `_write_guard`, while an already-started blocking write continues without the guard; a second put can pass its check before the first persist completes. Thus the write-size-aware `crosses_floor` arithmetic is correct, but the check-through-persist serialization does not close the round-1 TOCTOU for all public call paths.

Suggested fix: share one `Arc<tokio::sync::Mutex<()>>` per canonical root (for example, cache it in `StorageBackend` or inject it into `FsBlobStore`), and acquire an owned guard that is moved into the blocking closure so cancellation cannot release it before `persist`. Add regression tests using two separately constructed stores with one root and an aborted/in-flight put.

### [Medium] ADR-111's normative `ContentRef` declaration still specifies the rejected deserialization path

Evidence: `docs/adr/ADR-111-blob-store.md:49-51` still shows `#[derive(..., Serialize, Deserialize)]` under `#[serde(transparent)]`, whereas the shipped type derives only `Serialize` at `crates/khive-storage/src/blob.rs:35-37` and manually validates every deserialize at `crates/khive-storage/src/blob.rs:39-46`. The same ADR later says the opposite at `docs/adr/ADR-111-blob-store.md:292-294`.

Why this matters: ADR-111 is accepted and its primary type declaration would reintroduce the original unsafe construction route if followed. It is internally contradictory and does not exactly describe the implementation.

Suggested fix: change the Decision-section example to derive only `Serialize` and show or state the manual `Deserialize` implementation that calls `ContentRef::from_hex`.

## Looks Right

- H1 is fixed in executable code: `ContentRef`'s field is private, its only constructors are validating `from_hex` and digest-derived `from_digest_bytes` (`crates/khive-storage/src/blob.rs:49-80`), and source inspection found no `From`/`TryFrom` bypass. Transparent serialization emits the raw string accepted by the manual deserializer; valid and invalid serde inputs are covered at `crates/khive-storage/src/blob.rs:304-345`.
- `crosses_floor` uses `saturating_sub` (`crates/khive-db/src/stores/blob.rs:65-80`), so its `u64` calculation neither underflows nor overflows; the exact boundary, exact-floor, below-floor, and oversized-write cases are tested at `crates/khive-db/src/stores/blob.rs:401-430`. For a normal, awaited call on one store, the guard does span availability check through `persist` (`crates/khive-db/src/stores/blob.rs:220-225` and `82-139`).
- Reads are not serialized behind the write mutex: `get` directly dispatches only its filesystem read at `crates/khive-db/src/stores/blob.rs:228-245`. No lock cycle is visible in the normal awaited write path.
- H3 is now honestly bounded in the trait and ADR: `delete` and `orphan_sweep` state the offline/quiesced-writer requirement at `crates/khive-storage/src/blob.rs:177-216`, and ADR-111 §8 matches the snapshot-only implementation at `docs/adr/ADR-111-blob-store.md:206-235`. The regression test explicitly preserves the documented unsafe race at `crates/khive-db/src/stores/blob.rs:527-569`.
- Issue [#924](https://github.com/ohdearquant/khive/issues/924) is open and accurately tracks a DB-coordinated transactional sweep, including the storage-vs-runtime design boundary and exclusion of a one-off quick fix.
- The round-2 delta is confined to the three named files; `git diff --check` passes and no unrelated regression was found in that delta.

## Commands Run

- `git rev-parse HEAD`, branch/status, ancestry, and PR metadata checks: local checkout matches the requested head; the round-1 head is an ancestor; worktree stayed clean.
- `gh pr view 922 --repo ohdearquant/khive ...` and `gh issue view 924 --repo ohdearquant/khive ...`: PR and follow-up scope verified.
- `git diff --check fd8d635d29753af0189b52bc0ac4bec63b7873e3..HEAD`: passed.
- The specified shared lockfile was atomically acquired with a non-blocking Perl advisory lock (the host has no `flock` executable). The first targeted test attempt exposed stale shared-target artifacts; while holding that lock, `cargo clean -p khive-storage -p khive-db` cleared the incompatible cache entries (Cargo reported 23.7 GiB removed), then `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-storage -p khive-db` passed: 311 `khive-db` unit tests, 10 DB contract tests, 60 `khive-storage` unit tests, 33 compliance tests, 10 vector tests, and both doc-test suites.

## What I Did Not Check

- Full-workspace tests, Clippy, formatting, cross-process same-volume writers, and fault-injected power-loss behavior were out of this focused re-review scope.

## Re-Review Guidance

Request another narrow H2/H1-doc re-review after the root-shared, cancellation-safe floor reservation is tested and ADR-111's type declaration is corrected.

Domain utility: SKIPPED — repository ADRs and source were sufficient for this focused storage-contract review.

ohdearquant:
## Codex review round 3 (head 18e4b21a)

APPROVE-WITH-FIXES

### [Medium] The two new guard regressions are schedule-dependent, and one fails in the required package run

Evidence: `crates/khive-db/src/stores/blob.rs:595-606` starts both independently constructed stores but has no synchronization point between either old implementation's free-space observation and its write; if the first old per-instance-mutex put happens to finish before the second reaches `put_blocking`, the second observes the reduced space and returns `CapacityFloor`, satisfying the assertions at `crates/khive-db/src/stores/blob.rs:616-625`. Thus this test can pass on the round-2 implementation it is intended to reject. The cancellation test uses a 10 ms scheduling delay at `crates/khive-db/src/stores/blob.rs:661-663` and treats a fixed 5 s polling window at `crates/khive-db/src/stores/blob.rs:683-694` as proof that the detached write leaked its guard. The required command failed this assertion at line 691 after 312 other `khive-db` tests passed; an immediately following isolated run passed in 4.35 s. The runtime outcome is therefore timing-sensitive rather than a reliable regression signal. It also calls the newly introduced `write_lock_for_root` at line 672, so unmodified round-2 source fails to compile this test rather than demonstrating the expected old runtime behavior.

Why this matters: The high-severity floor guard is now correctly implemented, but the claimed regressions do not reliably prove it and currently leave the required scoped test gate non-clean. A future reintroduction of the per-instance guard can pass depending on scheduling and filesystem throughput.

Suggested fix: Add a test-only synchronization hook inside the blocking critical section. For the cross-instance case, coordinate both old-code availability checks deterministically before allowing either persist; for cancellation, signal only after the owned guard has entered the blocking closure, abort the caller, assert the lock is still unavailable, then release the hook and await an explicit completion signal. Do not use fixed sleeps or fixed-duration I/O polling as correctness evidence.

## Looks Right

- The prior High is resolved in implementation: `write_lock_for_root` canonicalizes after `FsBlobStore::new` creates the directory and propagates any canonicalization error with `?` (`crates/khive-db/src/stores/blob.rs:243-245`, `:204-212`); it neither panics nor holds the registry's `StdMutex` across an await. Poison recovery takes the valid inner map before a short, synchronous entry operation (`:206-212`). The static registry retains only one small `Arc<Mutex<()>>` per canonical root, not guards; that is bounded by the process's configured/constructed roots.
- The guard is an owned guard from the shared `Arc` (`crates/khive-db/src/stores/blob.rs:271`) and is moved into the `spawn_blocking` closure (`:274-277`). Dropping the outer future cannot drop that captured guard before `put_blocking` returns. `get` and `exists` do not acquire the write lock (`:282-306`).
- `StorageBackend::blob_store` still constructs a fresh `FsBlobStore` on every factory call (`crates/khive-db/src/backend.rs:584-592`), and the constructor's canonical-root registry is therefore sufficient to share the lock without a backend change.
- ADR-111's Decision example now derives only `Serialize` and supplies the validating manual `Deserialize` implementation (`docs/adr/ADR-111-blob-store.md:49-72`), matching shipped `ContentRef` (`crates/khive-storage/src/blob.rs:35-46`). Its §5 canonical-root and owned-guard description (`docs/adr/ADR-111-blob-store.md:136-160`) matches the implementation. The configurable-floor explanation at `docs/adr/ADR-111-blob-store.md:288-294` agrees with `StorageBackend::blob_store`'s optional `floor_bytes` at `crates/khive-db/src/backend.rs:584-591`.

## Commands Run

- Non-blocking Perl `flock` probe on `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: exit 0 (free) before the required run.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-storage -p khive-db` from `crates/`: failed. `khive-db` reported 312 passed and 1 failed: `stores::blob::tests::aborting_the_outer_put_future_does_not_release_the_guard_before_persist_completes` at `blob.rs:691`. No stale-target `E0432`/`E0405` diagnostic occurred, so no clean-and-retry was warranted.
- After a second free-lock probe, `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-db stores::blob::tests::aborting_the_outer_put_future_does_not_release_the_guard_before_persist_completes -- --exact`: passed (1 passed, 312 filtered, 4.35 s). Cargo did subsequently wait for the shared build lock, confirming the failure above is timing-sensitive under suite conditions.
- `git diff --check 85dabc26..HEAD`: passed; the delta contains exactly `crates/khive-db/src/stores/blob.rs` and `docs/adr/ADR-111-blob-store.md`.

## What I Did Not Check

- This is the requested narrow round-3 re-review only: no workspace-wide lint, full test suite, or cross-process lock semantics were run. ADR-111 explicitly limits this guard to one process.

## Re-Review Guidance

- Narrow re-review after deterministic test coordination is added. The implementation and ADR amendments need no further design change for the previously rejected High and Medium.

Domain utility: SKIPPED — this narrowly scoped storage-concurrency and ADR review was fully covered by repository contracts and source inspection.

ohdearquant:
## Codex review round 4 (head 1be35752)

APPROVE-WITH-FIXES

### [Medium] Cross-instance regression still accepts the old per-instance-lock implementation under delayed scheduling

Evidence: `crates/khive-db/src/stores/blob.rs:776-782` makes the sole assertion that B did not enter A's critical section a 200 ms `recv_timeout`. The test then releases A at `crates/khive-db/src/stores/blob.rs:784-788`, and treats B reaching the checkpoint afterward as success at `crates/khive-db/src/stores/blob.rs:791-800`. With the old per-instance mutexes, B can be delayed longer than that arbitrary window (or reach just after it expires but before A is released); it then reaches after the release and all assertions pass. Tokio provides no scheduling or blocking-pool-start guarantee within 200 ms. Thus the timeout is the correctness decision for detecting the old defect, rather than merely a bounded failure backstop.

Why this matters: The test will normally catch the old code when B happens to run promptly, but does not deterministically reject it. This leaves the round-3 Medium unresolved despite eliminating the former sleep and polling loop.

Suggested fix: Once A has signalled `reached` while holding the guard, directly assert `store_b.write_lock.try_lock().is_err()` before releasing A. The current canonical-root registry makes that deterministic; old per-instance mutexes make it succeed immediately. B's ordinary successful `put` can then run after A is released if that coverage is still wanted. This removes the arbitrary timeout from the pass/fail proof.

## Looks Right

- The cancellation regression is event-driven: `crates/khive-db/src/stores/blob.rs:290-295` sends `reached` only after the owned guard is captured in the blocking closure and holds it until `release`; the test aborts only after that signal (`:836-850`), observes the lock as unavailable (`:852-858`), then releases and waits for the explicit `done` event (`:860-872`). No sleep or polling loop remains in this test. The inner scope drops the guard before the test-only `done` send (`:283-301`).
- The release-build no-op claim holds on source inspection. `sync_hook` is defined only under `#[cfg(test)]` (`:398-452`), and every `put` reference to its value or channels is similarly test-gated (`:280-281`, `:290-294`, `:297-300`). In non-test builds the only surviving structural change is `let result = { let _owned_guard = owned_guard; put_blocking(...) }; result` (`:287-301`): as in 18e4b21a, the guard spans all of `put_blocking` and drops immediately after it returns, before the blocking closure returns. The non-test `cfg_attr` at `:287` affects only Clippy lint handling.
- The hook registry is isolated by canonical root and protected by a short synchronous mutex (`:411-451`); hooks are FIFO and consumed once. The two instrumented tests each hold a distinct `tempfile::TempDir` root (`:745-748`, `:824-827`), so parallel module tests cannot consume one another's hooks. The cross-instance test queues B only after A has consumed its hook (`:755-771`).
- `18e4b21a..HEAD` changes only `crates/khive-db/src/stores/blob.rs`; `git diff 18e4b21a..HEAD -- docs/` and the ADR-111-specific diff are empty. The implementation prior to the test seam, including `put_blocking` and the canonical-root lock registry, is unchanged.

## Commands Run

- Non-blocking Ruby `flock` probe on `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: free.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-db stores::blob` from `crates/`: passed, 23 passed / 0 failed.
- `git diff --check 18e4b21a..HEAD`: passed.
- `git diff --quiet 18e4b21a..HEAD -- docs/adr/ADR-111-blob-store.md` and `git diff --quiet 18e4b21a..HEAD -- docs/`: both confirmed no documentation change.

## What I Did Not Check

- I did not rerun the claimed plain non-test check, Clippy, or docs build; the release-build conclusion above is a line-by-line cfg and lifetime review.
- I did not independently reproduce the claimed five isolated and five module-suite repetitions. The one scoped module run passed, but cannot remove the static scheduling counterexample above.

## Re-Review Guidance

- Narrow re-review after replacing the cross-instance timeout-based negative observation with a deterministic lock-state assertion. No implementation or ADR design change is otherwise needed.

Domain utility: SKIPPED — this narrowly scoped Rust concurrency review was fully covered by the source and repository contracts.

ohdearquant:
## Codex review round 5 (head f78348f8)

APPROVE

# PR #922 — Round-5 narrow re-review

The sole Round-4 Medium is resolved. This review is deliberately limited to commit range `1be35752..f78348f8b3f51cc166ba32f97cd4b631fe3d1d37` and the requested regression-proof checks.

## Looks Right

- The cross-instance test installs only store A's hook, waits for its `reached` signal at `crates/khive-db/src/stores/blob.rs:751-759`, and immediately checks `store_b.write_lock.try_lock().is_err()` at `crates/khive-db/src/stores/blob.rs:761-769`. With the canonical-root registry this is the same `Arc<Mutex<()>>`, so it fails immediately; with the old per-instance locks, B's independent lock is unheld and the assertion fails immediately. There is no scheduler-dependent second task in that proof.
- The ordering is strict: `put` obtains the owned lock at `crates/khive-db/src/stores/blob.rs:271`, moves it into the closure at `:282-290`, and sends `reached` only afterward at `:291-294`. The test asserts lock sharing before sending A's release at `:774`; therefore the old per-instance implementation has no window to pass.
- The pass/fail proof of both instrumented tests has no timeout, sleep, polling, or fixed-duration wait. The removed `recv_timeout_blocking` has no remaining callers; the sole current `recv_timeout` text is the historical explanatory comment at `crates/khive-db/src/stores/blob.rs:724-740`.
- The liveness-only store-B put remains after A completes at `crates/khive-db/src/stores/blob.rs:771-781`. The cancellation test and `sync_hook` seam are outside the changed hunks and retain their event-driven checks at `:384-451` and `:785-839`.
- `git diff 1be35752..HEAD` contains only `crates/khive-db/src/stores/blob.rs` (+32/-51); `docs/` is empty. `git diff --check` is clean.

## Commands Run

- Non-blocking Perl flock probe of `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: succeeded; the requested scoped test was safe to start.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-db stores::blob` (from `crates/`): passed — 23 passed, 0 failed.
- `git grep -n -F 'recv_timeout_blocking' HEAD`: no matches.
- `git diff --quiet 1be35752..HEAD -- docs/`: exited 0 (no documentation changes).

## What I Did Not Check

- Full workspace CI and unrelated PR behavior; this was the requested final narrow re-review.

## Re-Review Guidance

- No further review is needed for the Round-4 Medium. Any later review should be limited to new changes.

Domain utility: SKIPPED — this narrow lock-regression review was fully determined by the local implementation, diff, and scoped test.

oceanwaves630 (approved):
Gating read at f78348f8: ADR-111 byte-untouched since 18e4b21a; the sole delta since the freeze is confined to crates/khive-db/src/stores/blob.rs and every inserted line in the impl block is #[cfg(test)]-gated (sync_hook seam + hook plumbing), with release builds compiling the identical guard-covered shape. Deterministic try_lock().is_err() cross-instance mutual-exclusion proof verified in tests. Transactional-sweep follow-up confirmed filed as #924.

PR #925, opened 2026-07-13
feat(runtime): compose pack-declared entity-type subtypes at boot (ADR-017)

ohdearquant:
## Summary

`EntityTypeRegistry::global()` only knew builtin `(EntityKind, entity_type)` subtypes — pack-declared subtypes were invisible at runtime, a dead letter for any pack wanting to extend the entity-type vocabulary the same way `EDGE_RULES` already lets packs extend the edge endpoint contract (ADR-017).

This mirrors the EDGE_RULES boot-composition shape exactly (no new mechanism invented):

- `Pack::ENTITY_TYPES` (khive-types) — new associated const, default empty, additive-only, alongside `Pack::EDGE_RULES`.
- `PackRuntime::entity_types()` (khive-runtime) — object-safe mirror, default empty.
- `VerbRegistry::all_entity_types()` (khive-runtime) — aggregates every loaded pack's `ENTITY_TYPES`, not deduplicated (same posture as `all_edge_rules()`).
- `call_register_entity_type_validators` threads the composed aggregate into every pack's `register_entity_type_validator` (signature widened to accept it), mirroring how `all_edge_rules()` is installed once for the whole registry.

Three production consultation sites moved off `EntityTypeRegistry::global()` (builtin-only) onto `EntityTypeRegistry::with_extra(registry.all_entity_types())` (builtin + composed pack extras):
- `khive-pack-kg::dispatch::KgPack::register_entity_type_validator` (runtime-layer validator, installed on the pack's own runtime per multi-backend deployment rules)
- `khive-pack-kg::handlers::common::validate_entity_type` (handler-layer check used by create/search/list)
- `khive-pack-schedule::handlers::validate_entity_type_for_replay` (schedule pre-validation — kept in bit-for-bit parity with the real create handler per its own doc contract; discovered mid-task, out of the originally-named crate scope but fixed to avoid a real parity regression)

`khive-pack-brain`'s `BrainPackRuntime` delegating wrapper updated to forward the widened `register_entity_type_validator` signature (the only other production override of that method).

**Dead-letter remediation**: `khive-pack-git` declares `"adr"` as a `Document` entity-type subtype (aliases: `architecture_decision_record`, `decision_record`). `git::ingest::find_document_for_path` is the only document-entity-aware code in this repo (matches pre-existing document entities by git-tracked `source_uri` path); ADR corpora are the most source-grounded missing Document subtype for that path. I do not have access to any live database and did not attempt to query one — this is a defensible inference from the OSS repo's own code and ADR-heavy culture, not a confirmed match to any specific production data gap. If the actual gap is a different Document subtype, this const is trivially extended.

No taxonomy change: the 9 closed `EntityKind` kinds (ADR-001) are untouched. `entity_type` remains the separate, finer-grained properties-level axis. No verb surface changes.

## Test plan

- [x] `cargo check -p khive-types -p khive-runtime -p khive-pack-kg -p khive-pack-git -p khive-pack-schedule -p khive-pack-brain` — clean
- [x] `cargo test -p khive-types -p khive-runtime -p khive-pack-kg` — 139+6+3+224 = all passed, incl. 3 new khive-runtime unit tests (`all_entity_types_empty_when_no_pack_declares_extras`, `pack_declared_entity_type_validates_through_composed_registry`, `overlapping_pack_declared_entity_types_compose_without_conflict`) and 1 new khive-types unit test (`pack_entity_types_default_empty`); khive-runtime full lib suite 798 passed/0 failed
- [x] `cargo clippy -p khive-types -p khive-runtime -p khive-pack-kg --all-targets -- -D warnings` — clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-types -p khive-runtime` — clean
- [x] Additional (beyond formal gate scope, touched by this diff): `cargo test -p khive-pack-git -p khive-pack-schedule -p khive-pack-brain` — clean, incl. 2 new git-pack acceptance tests (`git_pack_adr_entity_type_validates_through_create`, `adr_entity_type_rejected_without_git_pack_loaded`) proving genuine composition (rejected when GitPack isn't loaded, accepted with its alias when it is)

Note: the shared CI build lane (`CARGO_TARGET_DIR` shared across concurrent worktrees fleet-wide) produced two transient artifact-corruption failures during gating (a missing test binary, and a spurious `E0438` on `Pack::ENTITY_TYPES` twice, in two different downstream crates) — both confirmed environmental via scoped `cargo clean -p khive-types` + rebuild reproducing a clean pass each time, not defects in this diff.

Draft per standing review process — codex leg + Leo sign-off before ready/auto-merge, not self-flipped.


ohdearquant:
## Codex review round 1 (head 2fd2f8c3)

REJECT
Findings: 0 Blocker, 2 High, 1 Medium, 0 Low

Reviewed PR #925 (`feat/entity-types-pack-composition` → `main`) at
`2fd2f8c3ce654c8b663720fc2f23d2848617bee9`.

### [High] Conflicting entity-type declarations silently use load-order semantics

Evidence: `crates/khive-runtime/src/pack.rs:1549` explicitly keeps duplicate entity-type declarations; `crates/khive-types/src/entity_type.rs:445` and `crates/khive-types/src/entity_type.rs:451` overwrite kind-qualified canonical and alias keys. Accepted `docs/adr/ADR-001-entity-kind-taxonomy.md:128` instead requires a boot error for a duplicate `(base_kind, canonical_name)` unless it is explicitly shared, and `docs/adr/ADR-001-entity-kind-taxonomy.md:130` requires an alias collision to fail at boot.

Why this matters: two independently loaded packs can declare `Document` aliases such as `record` for different canonical types. The later pack in topological registration order replaces `document:record`, while the first bare alias remains due to `or_insert`; swapping otherwise independent pack registration changes which value is accepted and stored. This is not equivalent to duplicate edge rules, which are membership-only, and it contradicts the accepted collision contract. The new duplicate-tolerant test codifies the wrong behavior.

Suggested fix: validate all builtin and pack `EntityTypeDef`s during `VerbRegistryBuilder::build`, rejecting duplicate canonical pairs and every normalized, kind-qualified alias collision with both pack names in the error. If identical sharing is truly needed, add an explicit shared-ownership representation and test it; otherwise change `overlapping_pack_declared_entity_types_compose_without_conflict` to require a build error.

### [High] The public `PackRuntime` trait change breaks existing out-of-tree validator implementations

Evidence: `crates/khive-runtime/src/pack.rs:216` changes the existing `register_entity_type_validator` trait method from one runtime argument to two arguments. `crates/khive-pack-brain/src/pack.rs:201` has to change its delegating implementation solely to match that new signature.

Why this matters: the new `Pack::ENTITY_TYPES` and `PackRuntime::entity_types()` defaults do preserve compatibility for a pack that uses neither feature nor overrides validator setup. They do not preserve it for any external pack that already overrides `register_entity_type_validator`, even if that pack declares no entity types: Rust rejects its now-mismatched trait implementation. This is a source-breaking public-pack API change.

Suggested fix: retain the old method signature and add a distinct, defaulted composed-registration hook (for example, `register_entity_type_validator_with_types`). Its default should call the old hook, and `call_register_entity_type_validators` should invoke the new hook. Override only the new hook in `KgPack` and delegating wrappers.

### [Medium] The added acceptance coverage does not pin runtime boot wiring or schedule parity for pack types

Evidence: the Git fixture builds the registry at `crates/khive-pack-git/tests/acceptance.rs:116` and returns it at `crates/khive-pack-git/tests/acceptance.rs:120` without calling `call_register_entity_type_validators`; production boot does call it at `crates/khive-mcp/src/serve.rs:1363`. The schedule entity-type tests build only KG plus schedule at `crates/khive-pack-schedule/tests/create_validation.rs:9` and exercise only builtin `paper` at `crates/khive-pack-schedule/tests/create_validation.rs:946`.

Why this matters: the Git test proves handler-layer aggregation, but it would still pass if the `call_register_entity_type_validators` aggregate were absent or passed builtin-only data; a production handler call would then be rejected by the runtime validator. Likewise, reverting schedule to `EntityTypeRegistry::global()` would leave all current schedule tests green, so the stated replay-parity expansion is unpinned.

Suggested fix: make the Git fixture run the production validator-registration sequence and assert direct `KhiveRuntime::create_entity`/`create_many` accepts `Document + adr`. Add schedule singleton and bulk tests with a loaded test/Git type-declaring pack that assert `schedule.schedule` accepts the type and that the corresponding KG create succeeds; retain the no-declaring-pack rejection case.

## Looks Right

- The aggregate itself walks the loaded packs in topological order, and the KG create, list, and search paths now resolve known-kind `entity_type` filters against the composed data.
- Builtin-only behavior remains the same for the handler paths: `with_extra([])` includes the builtins, and `update` has no `entity_type` mutation surface.
- Schedule now consults the same composed `EntityTypeRegistry::with_extra(registry.all_entity_types())` data as KG, eliminating the prior builtin-vocabulary mismatch when a valid pack declaration is present.
- `adr` is a `Document` subtype rather than a new `EntityKind`; the diff does not alter the closed taxonomy or MCP verb surface. That substrate choice is consistent with ADR-001's document category.

## Commands Run

- `git status --short --branch`, `git rev-parse HEAD`, `gh pr view 925 ...`, and full `main...HEAD` diff/source/ADR tracing: PR head matched the requested `2fd2f8c3ce654c8b663720fc2f23d2848617bee9`; worktree was clean.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-types -p khive-runtime -p khive-pack-kg`: not started. The mandated `flock` utility is unavailable, so the shared build lock could not be positively verified as free.

## What I Did Not Check

- No Cargo tests were run for the reason above; Git and schedule acceptance execution is therefore unverified.
- This was a static review only; no MCP/daemon live smoke test was run.

## Re-Review Guidance

Request a narrow re-review after collision validation, the backward-compatible trait hook, and the two targeted test additions are in place.

Domain utility: SKIPPED — no lore-domain connector was available; the accepted ADR corpus provided the governing contract.

ohdearquant:
## Codex review round 2 (head f49e2ce8)

REJECT

Findings: 0 Blocker, 1 High, 1 Medium, 0 Low

### [High] Boot collision detection does not use ADR-001 normalization

Evidence: ADR-001 requires `trim -> lowercase -> snake_case -> alias resolution` at write time in `docs/adr/ADR-001-entity-kind-taxonomy.md:104-107` and requires an alias resolving to two canonical names to fail boot in `docs/adr/ADR-001-entity-kind-taxonomy.md:128-130`. `check_extra_collisions` builds its canonical and alias keys directly from `def.type_name` and `alias` in `crates/khive-types/src/entity_type.rs:496` and `crates/khive-types/src/entity_type.rs:508-519`; it never calls the normalizer defined at `crates/khive-types/src/entity_type.rs:22-40`.

Why this matters: two pack declarations whose aliases normalize to the same qualified wire key (for example `"Architecture Decision Record"` and `"architecture-decision-record"` for different Document types) are not rejected. This does not satisfy the round-1 H1 requirement for normalized kind-qualified alias collision detection, even though raw canonical collisions, raw alias collisions, built-in ownership, and two-owner error text are correctly handled.

Suggested fix: construct (or validate) every canonical and alias key through the same `to_snake_case` path in registry registration and in `check_extra_collisions`, then add tests for normalized alias-vs-alias and alias-vs-canonical collisions, including a built-in-vs-pack case.

### [Medium] The new direct-runtime test does not fail when aggregate validator wiring is absent

Evidence: the fixture calls `call_register_entity_type_validators` at `crates/khive-pack-git/tests/acceptance.rs:118-124`, but the new test at `crates/khive-pack-git/tests/acceptance.rs:3942-3964` only asserts that `create_many(... entity_type: "adr")` succeeds. With no validator installed, `KhiveRuntime::validate_entity_type_for_kind` deliberately returns the input unchanged at `crates/khive-runtime/src/runtime.rs:619-635`; therefore that positive-only test also passes if aggregate wiring is deleted altogether.

Why this matters: the positive `adr` assertion detects a builtin-only validator, but not an absent one. It consequently does not meet the requested M3 proof that deleting aggregate runtime wiring fails a test.

Suggested fix: retain the positive Git `adr` direct-`create_many` assertion and add a direct `create_many` case with an unregistered Document subtype that asserts `RuntimeError::InvalidInput`. Together, the two assertions distinguish composed, builtin-only, and absent validators.

## Looks Right

- H2 is resolved by source inspection: the public one-argument `PackRuntime::register_entity_type_validator` signature remains unchanged at `crates/khive-runtime/src/pack.rs:212-219`; the new default hook delegates to it at `crates/khive-runtime/src/pack.rs:227-245`; and the registry invokes the new hook at `crates/khive-runtime/src/pack.rs:1662-1666`. An out-of-tree one-argument override therefore still compiles and executes.
- The H1 canonical duplicate path is wired into `VerbRegistryBuilder::build` at `crates/khive-runtime/src/pack.rs:609-617` and its errors name both owners at `crates/khive-types/src/entity_type.rs:497-502`. Raw kind-qualified aliases and built-in-vs-pack collisions also go through the same owner-tracked map.
- M3 schedule coverage is effective: the Git-backed singleton and bulk schedule tests at `crates/khive-pack-schedule/tests/create_validation.rs:980-1029` would fail if schedule replay reverted to `EntityTypeRegistry::global()`.
- The default production pack list contains all 11 requested packs at `crates/khive-runtime/src/runtime.rs:1072-1092`. Static inspection found Git's `adr` as the sole pack-declared entity type (`crates/khive-pack-git/src/vocab.rs:97-101`), which neither collides with a builtin nor with another default pack declaration; the new collision validation therefore does not reject the legitimate default set.
- No unrelated tracked-file changes or whitespace errors were introduced in `2fd2f8c3..f49e2ce8`.

## Commands Run

- Non-blocking Ruby `flock` probe on `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: free.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-types -p khive-runtime -p khive-pack-git -p khive-pack-schedule` (from `crates/`): passed.
- `git diff --check 2fd2f8c3..f49e2ce8`: passed.

## What I Did Not Check

- I did not run a separate temporary downstream crate solely to compile an out-of-tree old-hook override; the unchanged public signature and default virtual dispatch establish that compatibility statically.
- This was the requested static, narrow re-review; I did not rerun the full workspace or an MCP server integration suite.

## Re-Review Guidance

- Narrow re-review after the normalized collision keys/tests and the direct-runtime negative assertion are added.

Domain utility: SKIPPED — this narrow Rust/ADR contract review was fully covered by repository sources and scoped tests.

ohdearquant:
## Codex review round 3 (head 48665fe3)

APPROVE

## Verified

- The round-2 High is resolved. `EntityTypeRegistry::new` normalizes canonical and alias keys before both qualified and bare lookup insertion at `crates/khive-types/src/entity_type.rs:451-460`; `register` uses the identical path at `crates/khive-types/src/entity_type.rs:548-556`; and `resolve` normalizes inbound wire values at `crates/khive-types/src/entity_type.rs:574-582`. `lookup` is private, and these are its only construction/mutation paths.
- `check_extra_collisions` now normalizes canonical names and aliases before constructing the qualified seen key at `crates/khive-types/src/entity_type.rs:511-535`. Therefore aliases `"Architecture Decision Record"` and `"architecture-decision-record"` for different `Document` types both become `document:architecture_decision_record`; the second declaration reaches the existing collision error rather than silently composing.
- The three regressions directly pin normalization: alias-vs-alias (`crates/khive-types/src/entity_type.rs:1012-1031`), alias-vs-canonical (`crates/khive-types/src/entity_type.rs:1033-1053`), and built-in-vs-pack (`crates/khive-types/src/entity_type.rs:1055-1070`). Before this delta each pair used distinct raw strings and `check_extra_collisions` would return `Ok(())`, so each test would fail without the normalizer.
- Registration normalization preserves already-working mixed-case wire aliases: `resolve("Proof Goal")` remains covered by `crates/khive-types/src/entity_type.rs:932-937`. Its source alias is already snake_case, so the new registration key is identical to the old key, and resolution still returns canonical `goal`.
- The round-2 Medium is resolved. The direct `create_many` negative assertion requires `RuntimeError::InvalidInput` at `crates/khive-pack-git/tests/acceptance.rs:3982-4006`. If aggregate wiring were deleted, the absent-validator path returns the unmodified subtype at `crates/khive-runtime/src/runtime.rs:619-634`, so `expect_err` would fail. Combined with the positive Git `adr` assertion at `crates/khive-pack-git/tests/acceptance.rs:3954-3973`, this distinguishes composed, built-in-only, and absent validation.
- The stated delta changes only `crates/khive-types/src/entity_type.rs` and `crates/khive-pack-git/tests/acceptance.rs`; none of the round-2 confirmed-good runtime/pack, schedule, or Git vocabulary files were touched. `git diff --check f49e2ce8..HEAD` passed.

## Commands Run

- Non-blocking Ruby probe of `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: free.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -q -p khive-types -p khive-runtime -p khive-pack-git -p khive-pack-schedule` (from `crates/`): passed (exit 0). No `E0432` or `E0438` occurred, so no clean/retry was needed.
- `git diff --check f49e2ce8..HEAD`: passed.

## What I Did Not Check

- Requested narrow static re-review only; no full-workspace or MCP-server integration suite.

## Re-Review Guidance

- No further review is needed for the two round-2 findings.

Domain utility: SKIPPED — the narrow Rust/ADR validation was fully covered by repository sources and scoped tests.

oceanwaves630 (approved):
Lane C read at 48665fe3bd3f5c1c9731183b6571b9be3ba49601. Verified: the new register_entity_type_validator_with_types hook is defaulted and delegates to the unchanged one-arg hook (out-of-tree packs keep compiling and behaving identically); boot collision rejection walks the same kind-qualified keyspace new/register populate, routed through the same to_snake_case normalization (ADR-001:104-107), so case/separator variants collide exactly as they would in the composed lookup; error names both contributing owners including builtin; the composed aggregate is built once and passed to every pack, mirroring the EDGE_RULES posture. r2's normalization High is closed with the three regression shapes tested. Auto-merge may proceed.

PR #918, opened 2026-07-13
docs(adr): ADR-103 Amendment 1 - batch-scaled cost_unit + warm-hook attribution

ohdearquant:
## Summary

Adds Amendment 1 to `docs/adr/ADR-103-resource-attribution-model.md`, covering two Stage 1 spec gaps: (1) `cost_unit` must scale with per-dispatch batch size and per-item embedding-model fan-out rather than being a flat per-verb weight, specifying the formula, the emission seam, the additive payload shape, and the absent-`cost_unit` compatibility rule; and (2) the two daemon-startup embedder warmup hooks (`KgPack::warm()`, `KnowledgePack::warm()`) run outside any actor-attributed dispatch and are currently invisible on the event plane, and must emit the existing `PhaseStarted`/`PhaseCompleted` pattern with `work_class="warm"`, attributed to the daemon principal via the same token-minting pattern the ANN warm background task already uses.

This formalizes the findings of the 2026-07-13 embedder-attribution spike into normative ADR text. Docs-only: exactly one file changed, no code.

Status on the amendment is `Proposed`, pending Leo's spec gate before any dependent implementation work picks it up.

## Test plan

- [x] Pre-commit hooks passed (deno fmt, lint ADR references, secret scan)
- [x] Diff scoped to exactly one file, additive only (no rewrite of accepted ADR-103 sections)
- [ ] Leo spec-gate review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #927, opened 2026-07-13
feat(runtime): emit resource.cost_unit per ADR-103 Amendment 1

ohdearquant:
## Summary

Implements ADR-103 Amendment 1 (`docs/adr/ADR-103-resource-attribution-model.md`,
merged to main via PR #918): emission of `resource.cost_unit` on every successful
verb dispatch's terminal audit event, plus phase-span attribution for the two
daemon-startup embedder-warmup hooks (`KgPack::warm`, `KnowledgePack::warm`) that
previously ran with zero observability.

This PR is emission only. `brain.event_counts` is intentionally NOT extended in
this PR (the amendment's Stage-2 read-path work is separate; the MCP-facing
description string for `event_counts` still correctly states `cost_unit` does not
exist on any payload yet at the time of writing, and needs a follow-up once this
merges).

## Part 1: `cost_unit` formula (`crates/khive-runtime/src/cost_unit.rs`, new file)

```
cost_unit = base_weight(verb) + per_item_weight(verb) x item_count x model_count
```

Computed with checked `i64` arithmetic: `checked_mul` at each product step,
`checked_add` for the final sum, clamping the WHOLE expression to `i64::MAX` on any
overflow rather than omitting the field, per Amendment 1 Part 1.

**Weight values (documented default, not invented):** the amendment commits only
to the formula's *shape* and to `per_item_weight(verb) = 0` for every verb outside
its closed embedding-bearing family. It leaves specific weight *magnitudes*
unspecified, describing them as "deterministic, hand-set constants ... fixed at
implementation time and not measured." This PR ships `base_weight = 1` uniformly
across every verb and `per_item_weight = 1` for every embedding-bearing verb, as
the documented default pending a dedicated per-verb weights table. No weight value
in this PR was invented beyond what the ADR explicitly left open; a follow-up issue
should track the weights table.

**Embedding-bearing verb classification (closed set, per the amendment):** `create`
(singleton only), `update`, `memory.remember`, `memory.recall`, `knowledge.search`,
`knowledge.compose`, `knowledge.index`. Everything else, including bulk
`create(items=[...])` (which intentionally skips embedding and backfills via a
separate `reindex` call, `operations.rs:4698-4709`) and `link` (no embedding-bearing
path, singleton or bulk), gets `base_weight` only, regardless of item/created count.

**`item_count`:**
- `create` singleton, `memory.remember`, `update`, `memory.recall`,
  `knowledge.search`, `knowledge.compose`: always `1`.
- `knowledge.index`: `result["total"]`, the full paged corpus count, never the
  internal `batch_size` chunking ceiling (`clamp(1, 1000)`, an embed-grouping page
  size, not the dispatch's total work).

**`model_count`:**
- `memory.remember` with an explicit `embedding_model` override: `1`, registry
  never consulted.
- `memory.remember` without an override, and singleton `create`: reads
  `PackRuntime::registered_embedding_model_names().len()` for the pack that owns
  the verb, evaluated lazily via `FnOnce` so every other dispatch never touches the
  embedder registry. Zero registered models is a valid result: the whole
  `per_item_weight x item_count x model_count` term vanishes and `cost_unit`
  reduces to `base_weight(verb)` alone.
- `update`, `memory.recall`, `knowledge.search`, `knowledge.compose`,
  `knowledge.index`: always `1` (a constant, single embedding call, never a
  fan-out), and never reads the registry.

New `PackRuntime::registered_embedding_model_names()` hook (default `Vec::new()`)
lets the dispatch loop query the pack that already owns the verb without needing a
separate runtime handle; overridden by `KgPack` and `MemoryPack`.

**Error/denied dispatches omit the `resource` object entirely** (not just
`cost_unit`), preserving the existing "absence has exactly two meanings"
convention the amendment states for `resource.cost_unit` specifically (a
pre-amendment event, or a dispatch that errored). Implementation decision: rather
than invent a third, unspecified payload shape (`resource.work_class` alone, no
`cost_unit`), this PR omits the whole object on any non-`Ok` dispatch outcome. This
mirrors `counts_by_work_class`'s existing absence-is-exclusion convention on the
read side.

## Part 2: warm-hook phase-span attribution

`KgPack::warm()` and `KnowledgePack::warm()` run at daemon construction / first
pack install, outside `dispatch()` entirely, so there is no caller-supplied token.
**Correction to my task briefing:** I was told this warm-hook attribution was
already shipped in the Stage-1 lineage (PR #732/#737) and not to re-implement it.
I verified this at source before writing any code (grepping both files for
`work_class`/`PhaseStarted`) and found neither hook had ANY phase-span emission
before this PR: only `khive-pack-memory/src/ann.rs`'s ANN-rebuild background task
had the pattern, which the ADR cites as the mechanism to *reuse*, not as evidence
the two `warm()` hooks already had it. This PR implements Part 2 fresh.

Both hooks now:
1. Mint a `NamespaceToken` via `runtime.authorize(Namespace::local())` (warn + skip
   telemetry, but still run the warmup, on mint failure).
2. Emit `PhaseStarted` (`work_class: "warm"`, `phase: "kg_embedder_warm"` /
   `"knowledge_embedder_warm"`, `corpus_size: None`).
3. Emit exactly one terminal event: `PhaseCancelled` if
   `is_benign_shutdown_cancellation` classifies the embed error as a
   `spawn_blocking` cancellation from shutdown, `PhaseCompleted` otherwise.

`KnowledgePack::warm()`'s span wraps *only* the configured-embedder branch's
spawned task body, not the unconditional `vamana::warm_known_snapshots` call above
it, so no phase event fires for a corpus with no embedder configured.

`is_benign_shutdown_cancellation` moved from `khive-pack-memory/src/ann.rs` (a
private fn) to a new `khive-runtime/src/phase_events.rs` (re-exported at the crate
root), so both new call sites reuse the exact same typed-downcast classification
`ann.rs` originated instead of duplicating it. `ann.rs`'s own 4 call sites
(including 3 existing unit tests) needed no changes beyond the import.

**Note on my task briefing's test list:** I was asked to pin "exactly-one-terminal-
event with cost_unit on warm hooks." Per the ADR, phase-span events (`PhaseStarted`
/ `PhaseCompleted` / `PhaseCancelled`) never carry `resource.cost_unit`; that field
is exclusive to the per-dispatch `Audit` event via `resource_payload`. Warm hooks
are background phase work, not a verb dispatch, so there is no `item_count`/
`model_count` to compute for them. The tests below pin exactly-one-terminal-event
as literally specified; they do not assert `cost_unit` on those events because the
ADR does not put it there.

## Tests

- `crates/khive-runtime/src/cost_unit.rs`: 17 unit tests covering formula
  arithmetic and overflow clamping, zero-model vanishing (both `create` and
  `memory.remember`), bulk-create base-only (small and 250-item batch),
  `knowledge.index` full-total vs. `batch_size` ceiling (including a missing-total
  default and an `i64::MAX` clamp case), `memory.remember` explicit-model override,
  and the constant-model-count verb family never touching the registry (asserted
  positively via a panicking `unreachable_model_count` closure).
- `crates/khive-runtime/src/pack.rs`: 8 new integration tests exercising the real
  dispatch path: resource present on non-embedding and embedding-scaling success,
  zero-registered-models base-only, resource omitted on dispatch error, on an
  unknown verb, on gate denial, and on both link-singleton success and failure.
- `crates/khive-pack-kg/src/dispatch.rs`:
  `kg_warm_emits_exactly_one_phase_started_and_one_terminal_event`, deterministic
  (in-memory runtime has no embedder configured, so `embed()` errors via the
  non-cancellation path and the terminal event is `PhaseCompleted`; the test
  asserts `completed + cancelled == 1` either way).
- `crates/khive-pack-knowledge/src/pack.rs` (new `mod tests`):
  `knowledge_warm_emits_no_phase_events_when_no_embedder_configured`, pinning the
  configured-only branch gating: zero phase events when no embedder is registered,
  proving `warm_known_snapshots` does not independently emit and the branch does
  not fire when unconfigured.

**Known coverage gap:** `KnowledgePack::warm()`'s *affirmative* path (embedder
configured, phase events actually emitted) is not covered by an automated test in
this PR. That branch fires inside a detached `tokio::spawn` with no
await-completion signal back to the caller, and building a reliable
non-flaky test for it (poll-with-timeout on the event store) was judged out of
scope for this PR's time budget given the deterministic negative test already pins
the gating logic that matters (the branch scoping bug this Part 2 work exists to
prevent). Filing a follow-up issue is recommended if affirmative-path coverage is
wanted.

## Gates (scoped to khive-runtime, khive-pack-kg, khive-pack-memory,
khive-pack-knowledge; run from `crates/` with
`CARGO_TARGET_DIR=/private/tmp/khive-shared-target`)

| Gate | Command | Result |
|---|---|---|
| check | `cargo check -p khive-runtime -p khive-pack-kg -p khive-pack-memory -p khive-pack-knowledge --all-targets` | exit 0 |
| test | `cargo test -p khive-runtime -p khive-pack-kg -p khive-pack-memory -p khive-pack-knowledge` | exit 0, 0 failed (all suites) |
| clippy | `cargo clippy -p khive-runtime -p khive-pack-kg -p khive-pack-memory -p khive-pack-knowledge --all-targets -- -D warnings` | exit 0 |
| fmt | `cargo fmt --all -- --check` | exit 0 |
| doc | `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-runtime -p khive-pack-kg -p khive-pack-memory -p khive-pack-knowledge` | exit 0 |
| pre-commit | cargo fmt / cargo clippy / secrets scan hooks | all passed on commit |

## Test plan

- [x] `cargo test` scoped to the four touched crates: 0 failures
- [x] New `cost_unit.rs` tests pin the formula, clamping, and zero-model behavior
- [x] New `pack.rs` tests pin resource presence/omission across success/error/
      denied/link paths through the real dispatch loop
- [x] New warm-hook tests pin exactly-one-terminal-event (KgPack) and
      zero-events-when-unconfigured (KnowledgePack)
- [ ] Reviewer: confirm the base_weight=1/per_item_weight=1 default and the
      omit-the-whole-object-on-error decision are acceptable pending a weights-
      table follow-up ADR/issue

This PR stays in draft until codex review approves.

### DIFF
--- a/crates/khive-pack-brain/src/tests.rs
+++ b/crates/khive-pack-brain/src/tests.rs
@@ -6356,6 +6356,73 @@ mod event_counts_tests {
         );
     }
 
+    /// ADR-103 Amendment 1 (PR #927): the dispatch-boundary audit-row emitter
+    /// now stamps `resource: {"work_class": "interactive"}` on EVERY audit
+    /// row outcome, denied and errored included -- only `resource.cost_unit`
+    /// is scoped to a successful dispatch (ADR-103 Decision (a) requires
+    /// `work_class` on every event; Amendment 1's omission rule covers
+    /// `cost_unit` alone). This pins that a failed and a denied dispatch's
+    /// audit rows are still counted in `counts_by_work_class` via the same
+    /// `payload.resource.work_class` fallback path other non-phase events
+    /// use: outcome plays no role in the split, only payload shape does.
+    #[tokio::test]
+    async fn failed_and_denied_audit_rows_count_via_resource_work_class_fallback() {
+        let (pack, rt) = make_pack();
+        let registry = empty_registry();
+        let token = rt.authorize(Namespace::local()).unwrap();
+
+        let mut errored = Event::new(
+            token.namespace().as_str(),
+            "probe",
+            EventKind::Audit,
+            SubstrateKind::Event,
+            "lambda:khive",
+        );
+        errored.created_at = 1_000_000;
+        errored.outcome = khive_types::EventOutcome::Error;
+        errored.payload = json!({"resource": {"work_class": "interactive"}});
+        rt.events(&token)
+            .expect("event store")
+            .append_event(errored)
+            .await
+            .expect("seed errored audit event");
+
+        let mut denied = Event::new(
+            token.namespace().as_str(),
+            "list",
+            EventKind::Audit,
+            SubstrateKind::Event,
+            "lambda:khive",
+        );
+        denied.created_at = 1_000_000;
+        denied.outcome = khive_types::EventOutcome::Denied;
+        denied.payload = json!({"resource": {"work_class": "interactive"}});
+        rt.events(&token)
+            .expect("event store")
+            .append_event(denied)
+            .await
+            .expect("seed denied audit event");
+
+        let result = pack
+            .dispatch(
+                "brain.event_counts",
+                json!({"since": micros_to_iso(0)}),
+                &registry,
+                &token,
+            )
+            .await
+            .expect("brain.event_counts must succeed");
+
+        assert_eq!(result["total"], json!(2), "got: {result}");
+        assert_eq!(
+            result["counts_by_work_class"]["interactive"],
+            json!(2),
+            "a failed dispatch and a denied dispatch must both still count under \
+             counts_by_work_class -- work_class is stamped on every event outcome \
+             (ADR-103 Decision (a)), unlike cost_unit which is success-only: {result}"
+        );
+    }
+
     /// `cost_unit` does not exist on any event payload in the codebase today (ADR-103
     /// Stage 0 defines it; no Stage 1 producer emits it yet). This verb must not invent
     /// a `cost_unit` key — asserts its literal absence from the response shape so a
--- a/crates/khive-pack-kg/src/dispatch.rs
+++ b/crates/khive-pack-kg/src/dispatch.rs
@@ -9,7 +9,8 @@ use khive_runtime::pack::PackRuntime;
 use khive_runtime::{
     EntityTypeValidatorFn, KhiveRuntime, NamespaceToken, RuntimeError, VerbRegistry,
 };
-use khive_types::{EdgeEndpointRule, HandlerDef};
+use khive_storage::{PhaseCancelledPayload, PhaseCompletedPayload, PhaseStartedPayload};
+use khive_types::{EdgeEndpointRule, EventKind, HandlerDef};
 
 use crate::handler_defs::{handle_verbs, KG_HANDLERS};
 use crate::pack::{KgPack, KG_EDGE_RULES};
@@ -53,7 +54,88 @@ impl PackRuntime for KgPack {
     }
 
     async fn warm(&self) {
-        let _ = self.runtime.embed("khive warmup").await;
+        // ADR-103 Amendment 1 Part 2: `warm()` runs at daemon construction /
+        // first pack install, outside `dispatch()` entirely -- there is no
+        // caller-supplied token. Mint one the same way
+        // `khive-pack-memory`'s ANN background-rebuild task does
+        // (`rt.authorize(Namespace::local())`), so this daemon-startup
+        // embedder warmup is attributed to the daemon principal instead of
+        // remaining invisible on the event plane. A mint failure only
+        // removes this pass's telemetry -- the warmup itself still runs.
+        let token = match self.runtime.authorize(khive_runtime::Namespace::local()) {
+            Ok(token) => Some(token),
+            Err(e) => {
+                tracing::warn!(
+                    error = %e,
+                    "kg_embedder_warm: failed to mint ADR-103 phase-span attribution token; \
+                     warmup proceeds without telemetry"
+                );
+                None
+            }
+        };
+        if let Some(token) = &token {
+            khive_runtime::emit_phase_event(
+                &self.runtime,
+                token,
+                "kg.embedder_warm",
+                EventKind::PhaseStarted,
+                PhaseStartedPayload {
+                    work_class: "warm".into(),
+                    phase: "kg_embedder_warm".into(),
+                    corpus_size: None,
+                },
+            )
+            .await;
+        }
+        let phase_start = std::time::Instant::now();
+        let cpu_start = khive_runtime::process_resource_usage();
+        let result = self.runtime.embed("khive warmup").await;
+        if let Some(token) = &token {
+            let wall_us = phase_start.elapsed().as_micros() as i64;
+            let cpu_us =
+                khive_runtime::cpu_delta_us(cpu_start, khive_runtime::process_resource_usage());
+            match &result {
+                Err(e) if khive_runtime::is_benign_shutdown_cancellation(e) => {
+                    khive_runtime::emit_phase_event(
+                        &self.runtime,
+                        token,
+                        "kg.embedder_warm",
+                        EventKind::PhaseCancelled,
+                        PhaseCancelledPayload {
+                            work_class: "warm".into(),
+                            phase: "kg_embedder_warm".into(),
+                            wall_us,
+                            cpu_us,
+                        },
+                    )
+                    .await;
+                }
+                _ => {
+                    khive_runtime::emit_phase_event(
+                        &self.runtime,
+                        token,
+                        "kg.embedder_warm",
+                        EventKind::PhaseCompleted,
+                        PhaseCompletedPayload {
+                            work_class: "warm".into(),
+                            phase: "kg_embedder_warm".into(),
+                            wall_us,
+                            cpu_us,
+                        },
+                    )
+                    .await;
+                }
+            }
+        }
+        // `PackRuntime::warm` is documented infallible: errors are logged
+        // internally, never propagated. The embed outcome is already
+        // captured in the phase-span payload above (Completed vs
+        // Cancelled); nothing further to do with it here.
+        let _ = result;
+    }
+
+    fn registered_embedding_model_names(&self) -> Vec<String> {
+        self.runtime.registered_embedding_model_names()
     }
 
     fn register_entity_type_validator(&self, _runtime: &KhiveRuntime) {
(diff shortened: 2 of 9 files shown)


ohdearquant:
## Codex review round 1 (head 21e62cbf)

APPROVE-WITH-FIXES

Reviewed PR #927 at `21e62cbfabc7192988aa2c49f5859996226f95aa` against ADR-103 Amendment 1 (`docs/adr/ADR-103-resource-attribution-model.md:468-696`).

### [Medium] Error, denied, and unknown audit rows lose their required work class

Evidence: `crates/khive-runtime/src/pack.rs:1108-1114`, `:1310-1324`, and `:1407-1414` pass `None` as the complete `resource` value for denied, handler-error, and unknown-verb audit rows; `build_audit_storage_event` then inserts no `resource` object at `:2070-2073`. This is asserted intentionally by tests at `:4380-4384`, `:4409-4410`, and `:4441-4443`. ADR-103 requires the closed `work_class` to be stamped on every event (`docs/adr/ADR-103-resource-attribution-model.md:115-129`), while Amendment 1 only requires the nested `resource.cost_unit` to be absent for an errored dispatch (`:598-612`). `brain.event_counts` extracts interactive audit classification specifically from `payload.resource.work_class` (`crates/khive-pack-brain/src/handlers.rs:800-804`, `:2292-2301`).

Why this matters: Omitting the whole object satisfies the narrower “no cost_unit” condition but invents a different wire shape that silently excludes failed, denied, and unknown interactive requests from `counts_by_work_class`. Those rows have a known work class; only their cost cannot be computed without a successful result.

Suggested fix: Keep `resource: {"work_class": "interactive"}` on denied, errored, and unknown-verb Audit events, omitting only `resource.cost_unit`. Update the three omission tests to assert this shape and add an event-count regression covering a failed or denied Audit event.

## Looks Right

- The formula is implemented as `base_weight + per_item_weight * item_count * model_count` with checked multiplications and addition, clamping to `i64::MAX` (`crates/khive-runtime/src/cost_unit.rs:120-156`), as Amendment 1 requires (`docs/adr/ADR-103-resource-attribution-model.md:505-523`).
- The amendment leaves weight magnitudes as hand-set constants fixed at implementation time; the documented uniform `1`/`1` default is therefore permitted (`cost_unit.rs:8-15`, ADR-103 `:516-523`).
- The embedding-family rules match the amendment: bulk `create(items=...)` is base-only, `knowledge.index` uses returned full `total`, zero registered models eliminates the item term, and only explicit `memory.remember.embedding_model` overrides model count (`cost_unit.rs:29-109`).
- All successful `PackRuntime::dispatch` returns flow through the generic success arm, except successful singleton `link`, whose v2 audit path separately adds the same resource payload (`pack.rs:1235-1325`). Batch and chain operations each call `dispatch_with_identity` per operation (`crates/khive-mcp/src/server.rs:729-735`, `:915-932`); they do not bypass this seam. `help=true` is intentionally an earlier no-side-effect introspection short-circuit (`pack.rs:1011-1013`).
- Phase spans correctly contain `work_class`, phase identity, timing, and CPU data but no `cost_unit`, matching the separate phase-event payload contract (`docs/adr/ADR-103-resource-attribution-model.md:614-677`; `crates/khive-pack-kg/src/dispatch.rs:76-128`; `crates/khive-pack-knowledge/src/pack.rs:125-179`). Each new hook emits one start and exactly one terminal event when it executes.
- `git show HEAD^`/`git log -G` found no prior KG or Knowledge embedder-warm phase emission. The only matching history entry is this PR, so the new hooks do not double-emit an already-shipped span.
- The ANN cancellation predicate moved to the runtime and is still imported at every memory-pack use (`crates/khive-runtime/src/phase_events.rs:30-37`; `crates/khive-pack-memory/src/ann.rs:9-11`, `:858`, `:1031`, `:1244`). No orphaned local definition remains.
- `brain.event_counts` was not extended for cost units, as intended. Its existing resource-work-class fallback means successful audit rows participate without event-plane/schema changes; the finding above is limited to the omitted-resource outcomes.

## Commands Run

- `git status --short --branch`, `git rev-parse HEAD`, and `gh pr view 927 ...`: checkout is clean and HEAD matches the requested PR commit.
- Full ADR-103 read, including Amendment 1; `git diff --check HEAD^..HEAD`: no whitespace errors.
- Parent/history sweeps for KG/Knowledge phase events: no pre-existing warm-hook emitter found.
- Non-blocking Ruby probe of `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: free.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-runtime -p khive-pack-kg -p khive-pack-memory -p khive-pack-knowledge`: did not compile tests. Cargo used a cached `khive-storage` interface in which `tx_registry::oldest()` returns `(TxId, Duration, Option<String>)`, while this checkout's unchanged source declares `(Duration, Option<String>)`; `khive-runtime/src/daemon.rs:700` consequently fails with E0308. The source line is unchanged by this PR and this is not the specified E0432/E0438 retry case, so no clean/retry was performed.

## What I Did Not Check

- The scoped test suites could not execute past the shared-target cache/base mismatch above; I did not create a per-PR target directory or alter the shared cache outside the requested retry rule.
- I did not run a full workspace or external CI suite.

## Re-Review Guidance

Narrow re-review after the resource shape and its regression test are corrected. Re-run the scoped command after the shared target is coherently rebuilt.

Domain utility: SKIPPED — the accepted ADR and repository contracts supplied the necessary review standard directly.

ohdearquant:
Addressed codex r1's Medium finding at c793ca2d:

- `resource: {"work_class": "interactive"}` is now stamped on every audit row outcome (denied, errored, unknown-verb), not just Success. Only `resource.cost_unit` remains scoped to a successful dispatch per Amendment 1's omission rule.
- New `khive-runtime::cost_unit::base_resource_payload()` used at all three non-Success call sites in `pack.rs`.
- Updated the four existing tests that asserted `resource` was entirely absent on error/denied/link-failure paths to assert the `{"work_class": "interactive"}` shape instead.
- New regression in `khive-pack-brain/src/tests.rs`: `failed_and_denied_audit_rows_count_via_resource_work_class_fallback` proves an Error-outcome and a Denied-outcome Audit event both count in `counts_by_work_class["interactive"]`.
- Hit the flagged shared-target stale-cache hazard exactly as described; `pgrep -x cargo/rustc` confirmed no concurrent build, `cargo clean -p khive-storage -p khive-runtime` cleared it, all five gates (check/test/clippy -D warnings/fmt --check/doc -D warnings) reran clean on khive-runtime, khive-pack-kg, khive-pack-memory, khive-pack-knowledge, khive-pack-brain.

New head SHA: c793ca2d822a077110204856ed2edc3415153427

ohdearquant:
## Codex review round 2 (head c793ca2d)

APPROVE

Round-2 narrow re-review of PR #927 at `c793ca2d822a077110204856ed2edc3415153427`, limited to the prior Medium audit-resource omission and its fix delta (`21e62cbf..HEAD`).

## Looks Right

- The prior Medium is resolved. `base_resource_payload()` is exactly `{"work_class":"interactive"}` with no `cost_unit` key at `crates/khive-runtime/src/cost_unit.rs:176-188`. The denied audit path passes it at `crates/khive-runtime/src/pack.rs:1113-1118`; generic handler errors at `:1320-1336`; and the allowed-but-unknown-verb error path at `:1420-1433`. `build_audit_storage_event` inserts that object into the persisted audit payload at `:2089-2106`.
- All audit persistence outcomes are accounted for. Successful singleton `link` writes the unchanged `resource_payload()` into its v2 audit event at `crates/khive-runtime/src/pack.rs:1253-1279`, with the v1 fallback retaining the same payload at `:1282-1296`; generic successful dispatches continue to use `resource_payload()` at `:1320-1336`. The four `build_audit_storage_event` call sites are the denied path, singleton-link success fallback, generic result path, and unknown-verb error path. The gate-infrastructure failure branch deliberately creates no audit row, so it is not an unstamped outcome.
- The updated runtime tests pin the exact omission shape rather than merely the presence of `resource`: equality with `{"work_class":"interactive"}` excludes `cost_unit` for handler error (`crates/khive-runtime/src/pack.rs:4402-4413`), unknown verb (`:4438-4442`), denied dispatch (`:4473-4478`), and singleton-link error (`:4557-4561`). Existing success tests continue to assert the two-key `{work_class, cost_unit}` shape (`:4311-4314`, `:4519-4525`).
- The brain regression genuinely takes the resource fallback: its two stored `EventKind::Audit` rows have `Error` and `Denied` outcomes plus only nested `resource.work_class` (`crates/khive-pack-brain/src/tests.rs:6374-6404`), and asserts both count as interactive (`:6416-6423`). `brain.event_counts` increments from `event_work_class` at `crates/khive-pack-brain/src/handlers.rs:800-804`; that helper checks `payload.resource.work_class` when no top-level work class exists at `:2292-2301`.
- Delta containment holds: `git diff --name-only 21e62cbf..HEAD` contains only `crates/khive-runtime/src/{cost_unit.rs,lib.rs,pack.rs}` and `crates/khive-pack-brain/src/tests.rs`. No phase-span emitter or hook file changed, and the successful dispatch payload code is unchanged apart from surrounding error-path comments.

## Commands Run

- `git status --short --branch`, `git rev-parse HEAD`, `git diff --name-only 21e62cbf..HEAD`, and `git diff --check 21e62cbf..HEAD`: clean checkout at the requested head; four described files only; no whitespace errors.
- Non-blocking Ruby `flock` probe of `/private/tmp/khive-shared-target/debug/.cargo-build-lock`: free.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target cargo test -p khive-runtime -p khive-pack-brain` (from `crates/`): passed. `khive-pack-brain`: 217 unit tests plus 13 integration tests; `khive-runtime`: 825 passed, 5 ignored, plus 71 integration tests. The new fallback regression and all four runtime omission-shape tests passed.

## What I Did Not Check

- This was the requested narrow static re-review; I did not rerun the full workspace suite or external CI.

## Re-Review Guidance

- No further re-review is needed for the round-1 Medium.

Domain utility: SKIPPED — ADR-103 and the repository contracts supplied the necessary review standard directly.

oceanwaves630 (approved):
Lane C read at c793ca2d (first implementation under ADR-103 Amendment 1): cost_unit.rs implements the formula shape exactly — checked i64 arithmetic clamped to i64::MAX, per_item_weight=0 outside the closed embedding-bearing family, bulk create carved out, knowledge.index item_count from result total rather than the internal page ceiling. Uniform 1/1 default weights are within the amendment's implementation-time-constants latitude; a per-verb weights table is the right follow-up before Stage 3. work_class stamped on every dispatch outcome with cost_unit success-only (all four outcomes test-pinned), and phase spans carry work_class "warm" (valid closed-enum member, ADR-103:178) with no cost_unit key. No ADR text modified.

PR #930, opened 2026-07-13
fix(pack-kg): expose effective list limits (#894)

### LINKED ISSUE #894: list(): silently caps pages at 500 rows regardless of requested limit — silent truncation caused a near 48% corpus skip
Repro 2026-07-12 (atlas provenance sweep): list(kind="entity", limit=1000, offset=N) returns at most 500 rows per page with no error, no truncation marker, and no indication the requested limit was clamped. A pagination loop striding by the REQUESTED limit (1000) therefore silently skips every other 500-row block — our sweep initially missed ~48% of a 4,194-entity corpus and only caught it because row counts didn't reconcile. Ask: either honor limits up to the documented max, or return the effective limit / a has_more marker so callers can detect clamping. Silent clamp + caller-side stride = silent wrong reads (same failure class as #869).

ohdearquant:
## Summary

- Preserve the existing bare-array response for `list` requests whose limit is within the kind-specific cap.
- Return `items`, `requested_limit`, `effective_limit`, and `limit_clamped` when an offset-based request exceeds its cap.
- Preserve the edge cursor response shape and add the same limit metadata when its requested limit is clamped.
- Apply the explicit cap contract to entity, note, edge, event, and proposal listings.
- Document the response contract in verb help, the API guide, and `AGENTS.md`.

## Why

A caller could request more rows than the server cap, receive only the capped page, and then advance its offset by the requested limit. That silently skipped records. Machine-readable requested and effective limits let callers advance safely while preserving compatibility for ordinary list calls.

## Testing

- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target RUSTC_WRAPPER= cargo clippy -p khive-pack-kg --all-targets -- -D warnings`
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target RUSTC_WRAPPER= cargo test -p khive-pack-kg`
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target RUSTC_WRAPPER= cargo fmt --all -- --check`

Closes #894


ohdearquant:
## Codex review round 1 (head 5889403d)

APPROVE

Findings: 0 Blocker, 0 High, 0 Medium, 0 Low.

## Scope and Verdict

Reviewed `origin/main...5889403dd4d3cba7062b0f891f7973932d00654c` against issue #894 and the PR description. The implementation retains each pre-existing page cap and, for an explicitly over-cap request, returns the requested and effective limits plus `limit_clamped: true`. That gives an offset-paginating caller the value required to advance safely, without changing the legacy bare-array shape for ordinary requests.

## Looks Right

- `crates/khive-pack-kg/src/handlers/list.rs:90-152` preserves the requested entity limit before applying the 500-row cap and returns the metadata envelope only when clamped. The tagged and untagged entity paths share that rendering point.
- `crates/khive-pack-kg/src/handlers/list.rs:176-208` applies the runtime's 1000-edge cap before both offset and cursor queries. Cursor responses retain `edges` and `next_after`, with the added limit metadata at `:195-200`.
- `crates/khive-pack-kg/src/handlers/list.rs:217-367` and `:371-443` apply the same contract to note/message and event listings, including the event outcome-filter path.
- `crates/khive-pack-kg/src/handlers/proposal.rs:455-563` applies and reports the proposal cap without changing its un-clamped array response.
- `crates/khive-pack-kg/tests/integration.rs:10642-10882` exercises the actual truncation case for entities, notes, and both edge modes, while `:10607-10640`, `:10910-10954` verifies that metadata is emitted even when fewer records than the cap match. This would catch the silent-clamp failure class from #894.
- `crates/khive-pack-kg/src/handler_defs.rs:152-160`, `docs/guide/api-reference.md:216-221`, and `AGENTS.md:440` document the same caps, envelope fields, and compatibility behavior implemented by the handlers.
- The changes comply with ADR-023's handler-return `serde_json::Value` contract. ADR-078 keeps JSON as the lossless default for machine parsing, so the added metadata remains available to offset-paginating clients.
- No added public text contains internal reviewer or provenance markers, and the added diff contains no em dash. `git diff --check` is clean.

## Commands Run

- `gh issue view 894 -R ohdearquant/khive --json number,title,body,state,url,labels,comments`: confirmed the required signal is an effective limit and/or clamp marker, not a higher cap.
- `gh pr view 930 -R ohdearquant/khive --json number,title,body,state,url,baseRefName,headRefName,commits,files`: confirmed the claimed scope and local head.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target RUSTC_WRAPPER= cargo test -p khive-pack-kg`: passed after Cargo waited for the shared build-directory lock.
- `CARGO_TARGET_DIR=/private/tmp/khive-shared-target RUSTC_WRAPPER= cargo test -p khive-pack-kg --quiet`: passed.

## What I Did Not Check

- I did not run the full workspace suite, Clippy, or formatting checks; the requested scoped package test passed.

## Re-Review Guidance

No follow-up review is required for this change.

## Domain Utility

Domain utility: SKIPPED - no domain-composition connector was available; the accepted ADR corpus supplied the applicable contract.

ohdearquant:
Post-approval delta (after codex round-1 APPROVE at 5889403d): commit f983f6f0 is a test-only fix. The workspace CI run surfaced that khive-pack-git's acceptance test `issue_full_page_never_leaks_raw_updated_at_into_paging_floor` dispatched `list` above the issue-kind cap and asserted on a bare array; with the #894 clamp metadata it now pages through all persisted issues (handling both response shapes) and asserts full coverage. No production code changed since the approved head.

oceanwaves630 (approved):
Salvage lane verified: independent review APPROVE at the recorded head, CI green at current head, delta beyond the reviewed head limited to conflict-resolution/test-only commits as documented on the PR.

PR #862, opened 2026-07-11
feat(perf): flagship coverage manifest and validator (benchmark program PR1)

ohdearquant:
## Summary

PR 1 of the approved benchmark-overhaul plan (`.khive/workspaces/20260711/bench-overhaul/DESIGN.md`, PR 1 slice). Contracts and coverage tooling only - no benchmark execution, no runner code, no CI changes.

- `scripts/perf/flagship_workloads.toml`: the declarative flagship workload manifest. 84 scenarios encoding every F1-F10 requirement from DESIGN.md's gap audit, each with `surface` (`mcp_daemon` | `admin_cli` | `raw_daemon_control`), `scale`, cold/warm `state` + `settle` method, `required_percentiles`, `attempts` (sample count), and a `fixture_hash` (`sha256:<64 hex>`, hashed from the fixture's content-addressed name pending PR 2+'s fixture builder). Scenario IDs follow the `f<N>.<verb>.<arm>.<embedder>` convention.
- `scripts/perf/schemas/flagship-result-v1.json`: checked JSON Schema for the Distribution contract and the Ledger record (`FlagshipRecord`) contract from DESIGN.md's "Typed interfaces" section.
- `scripts/perf/flagship_schema.py`: stdlib-only Python module implementing the same two contracts as plain-dict validators (`validate_distribution`, `validate_record`) - this is what `coverage_validator.py` actually imports, matching the no-`jsonschema`-dependency convention of `bench_track.py`/`bench_calibrate.py`.
- `scripts/perf/coverage_validator.py`: reads the manifest plus a `perf-data`-style records directory (`*.jsonl`, one JSON object per line, `suite == "flagship-e2e"`) and reports one of five statuses per scenario: `measured` / `stale` / `missing` / `wrong-surface` / `confounded`.
- `scripts/perf/test_flagship_coverage.py`: 28 unit tests (stdlib `unittest`, run via `python3 -m unittest test_flagship_coverage`) covering manifest completeness (every F1-F10 feature has >=1 scenario), duplicate scenario IDs, invalid surfaces, fixture-hash format, schema validation, and every coverage-status branch (measured/stale/missing/wrong-surface/confounded, including the two freshness windows).

### The 0%-measured property

[...]

PR #946, opened 2026-07-13
fix(perf): require measurement evidence before coverage counts a row measured (#945)

### LINKED ISSUE #945: perf: flagship coverage validator certifies rows without measurement evidence (empty distributions count as measured)
## Summary

The flagship coverage validator (landed in #862) can mark a scenario "measured" from a row that contains no actual measurement evidence. Coverage percentages computed from it are therefore non-evidentiary: a fabricated or degenerate row can close a scenario and contribute to a 100% coverage claim.

## Evidence (current main)

- `scripts/perf/flagship_schema.py` (record validation): `metrics` and `distributions` are only required to be objects. No requirement for a percentile field, successful-sample count, error/timeout accounting, steady-state duration, daemon-engagement proof, raw-artifact existence, or correlated server duration.
- `scripts/perf/test_flagship_coverage.py` `_base_record`: the validator's own positive fixture uses `distributions: {}` and a single `metrics` key, and the suite expects that scenario to count as measured.
- `scripts/perf/coverage_validator.py`: marks a row measured after schema, identity, cohort, and freshness checks — none of which inspect measurement evidence.

## Why it matters

The coverage manifest's stated definition (design doc, seven conjunctive conditions) includes a timed real-transport path, proven daemon engagement, error/timeout evidence, and distributions. The implemented gate checks none of the evidence conditions, so the tool's "measured" verdict is weaker than the definition it reports against. Any completion claim built on it inherits the gap.

## Fix shape

Make the coverage contract mechanically equivalent to the evidence conditions it claims to certify:

1. Require non-empty distributions with a minimum successful-sample count for timed scenarios.
2. Require error/timeout accounting fields and reject rows that omit them.
3. Require daemon-engagement evidence (fallback count present AND a positive engagement proof once the correlation mechanism exists; until then, at minimum reject `daemon_fallback_count > 0` rows from "measured").
4. Require the referenced raw artifact to exist and hash-match when validating locally.
5. Add rejection-case tests for each: an empty-distributions row, a missing-error-accounting row, and a fallback-tainted row must all fail.

Scope note: this tightens the read/validation side only. Scenario semantics and the wider coverage definition are being handled in the benchmark-program spec amendment; this issue is the minimal correctness fix so the landed tool does not certify false coverage in the interim.

ohdearquant:
## Defect

The flagship coverage validator (`scripts/perf/coverage_validator.py` + `flagship_schema.py`, landed in #862) could mark a manifest scenario "measured" from a record carrying no actual measurement evidence: `metrics`/`distributions` were only required to be *objects*, with no requirement for a non-empty distribution, a successful-sample floor, error/timeout accounting, or a daemon-fallback / raw-artifact evidence check. The validator's own positive test fixture (`_base_record` in `test_flagship_coverage.py`) used `distributions: {}` and still passed.

This is the read/validation-side minimal fix per the issue. Scenario semantics and the wider coverage definition are being handled separately in a benchmark-program spec amendment; this PR does not touch the manifest or runner.

## The five tightenings

1. **Non-empty distributions + minimum successful samples** — `flagship_schema.validate_record` now rejects a record whose `distributions` object is empty, and `validate_distribution` rejects any distribution whose `successes` is below `MIN_SUCCESSFUL_SAMPLES` (1). Zero successful samples carries no latency evidence.
2. **Error/timeout accounting fields** — once `distributions` is required non-empty, the existing closed-object check on each distribution (`errors_by_code`, `timed_out`, …) actually runs; a distribution missing those fields is now rejected instead of silently skipped (it was never being validated when the dict was empty).
3. **Fallback-tainted rows rejected** — `validate_record` now rejects any record with `runtime.daemon_fallback_count > 0`. Until a positive daemon-engagement correlation mechanism exists, a fallback-engaged run cannot count as measured.
4. **Raw artifact existence + hash check, with explicit "unverified" when not resolvable** — `coverage_validator` gains an optional `--artifacts-dir` and a new `verify_artifact()` check. Once a scenario clears every other "measured" check, its `artifact.name` is resolved against `--artifacts-dir`; it must exist and sha256-match or the scenario is downgraded to `confounded`. When `--artifacts-dir` is not supplied (the common CI shape — only the `*.jsonl` summaries are fetched, not the raw artifacts), the per-scenario report records `artifact_verification: "unverified"` rather than silently passing as verified. This follows the same explicit-field pattern the report already uses (`status`, `reason`) since no prior "optional check" pattern existed to reuse.
5. **Rejection-case tests + upgraded fixture** — new tests assert an empty-distributions row, a missing-error-accounting row, and a fallback-tainted row each fail `validate_record`/`scenario_status`. The existing `_base_record` positive fixture now carries a real distribution (non-empty, `successes=1000`, full error/timeout accounting) and still passes.

## Evidence the old fixture would now fail

`test_empty_distributions_row_no_longer_counts_as_measured`, `test_fallback_tainted_row_no_longer_counts_as_measured`, and `test_missing_error_accounting_row_no_longer_counts_as_measured` in `test_flagship_coverage.py` reconstruct the pre-fix `_base_record` shapes (empty `distributions`, `daemon_fallback_count > 0`, distribution missing `errors_by_code`) and assert `scenario_status(...)` now returns `"confounded"`, not `"measured"` — i.e. the exact defect the issue reports is closed.

## Test results

```
uv run python -m pytest scripts/perf/test_flagship_coverage.py scripts/perf/test_mcp_bench_client.py -q
105 passed in 3.77s
```

Also verified via the doc-comment's alternate invocation form:
```
cd scripts/perf && uv run python -m unittest test_flagship_coverage -v
Ran 69 tests in 0.039s — OK
```

Python-only change under `scripts/perf/` — no `cargo`/`crates/` touched.

Closes #945

### DIFF
--- a/scripts/perf/coverage_validator.py
+++ b/scripts/perf/coverage_validator.py
@@ -215,6 +215,13 @@ def load_manifest(path: pathlib.Path) -> tuple[dict, list[str]]:
         if not isinstance(sc["required_percentiles"], list) or not sc["required_percentiles"]:
             errors.append(f"{label}: required_percentiles must be a non-empty array")
 
+        if "min_successes" in sc and (
+            isinstance(sc["min_successes"], bool)
+            or not isinstance(sc["min_successes"], int)
+            or sc["min_successes"] <= 0
+        ):
+            errors.append(f"{label}: min_successes must be a positive integer, got {sc['min_successes']!r}")
+
     return data, errors
 
 
@@ -437,6 +444,17 @@ def scenario_status(
 
     declared_min_successes = scenario.get("min_successes")
     if declared_min_successes is not None:
+        # load_manifest rejects malformed values, but a programmatic caller can
+        # bypass it - a bad declaration must confound, never raise.
+        if (
+            isinstance(declared_min_successes, bool)
+            or not isinstance(declared_min_successes, int)
+            or declared_min_successes <= 0
+        ):
+            return (
+                "confounded",
+                f"scenario declares malformed min_successes {declared_min_successes!r} (must be a positive integer)",
+            )
         required_successes = max(flagship_schema.TIMED_MIN_SUCCESSES, declared_min_successes)
         for dist_name, dist in latest.get("distributions", {}).items():
             successes = dist.get("successes") if isinstance(dist, dict) else None
--- a/scripts/perf/coverage_validator.py
+++ b/scripts/perf/coverage_validator.py
@@ -8,14 +8,19 @@ the shape a checked-out `perf-data` branch's `bench-data/` directory would
 hold once PR 2+ start writing them) and reports, per manifest scenario,
 exactly one of five statuses:
 
-  measured        latest matching record is schema-valid (which now also
-                  requires non-empty distributions with at least one
-                  successful sample, error/timeout accounting fields, and
-                  `daemon_fallback_count == 0` - khive#945), `status: "ok"`,
-                  measured through the manifest's declared surface, its
-                  `fixture_hash` matches the manifest, it is within the
+  measured        latest matching record is schema-valid, `status: "ok"`
+                  (which now also requires - for `"ok"` records only,
+                  khive#945 M3 - non-empty distributions whose `successes`
+                  clear `max(flagship_schema.TIMED_MIN_SUCCESSES, scenario
+                  min_successes)` - raise-only, 100-success global floor,
+                  khive#945 amendment §2.1 - error/timeout accounting
+                  fields, and `daemon_fallback_count == 0`), measured
+                  through the manifest's declared surface, its
+                  `fixture_hash` matches the manifest, its `artifact.name`/
+                  `artifact.sha256` are well-formed, it is within the
                   scenario's freshness window, and (when `--artifacts-dir`
-                  is supplied) its raw artifact exists and sha256-matches.
+                  is supplied) its raw artifact resolves inside that
+                  directory and exists with a matching sha256.
   stale           a measured-quality record exists but its age exceeds the
                   freshness window for the scenario's `runner_class`
                   (OQ7 ruling: 14 days self-hosted, 7 days hosted).
@@ -24,11 +29,16 @@ exactly one of five statuses:
                   manifest's declared `surface` for this scenario.
   confounded      the latest record exists but cannot be trusted as a
                   measurement: its own `status` is `"confounded"`,
-                  `"error"`, or `"insufficient_samples"`, its
-                  `workload.fixture_hash` does not match the manifest's
-                  `fixture_hash`, it fails `flagship_schema.validate_record`,
-                  or (when `--artifacts-dir` is supplied) its raw artifact
-                  is missing or sha256-mismatched.
+                  `"error"`, or `"insufficient_samples"` (a legitimately
+                  schema-valid state - non-`"ok"` records are exempt from
+                  the evidence requirements above precisely so honest
+                  failure evidence stays schema-valid without ever
+                  counting as measured), its `workload.fixture_hash` does
+                  not match the manifest's `fixture_hash`, it fails
+                  `flagship_schema.validate_record`, its `successes` fall
+                  below a scenario-declared `min_successes` raise, or its
+                  raw artifact is absent/invalid, escapes `--artifacts-dir`,
+                  is missing, or sha256-mismatches.
 
 Each scenario's report entry also carries `artifact_verification`, one of
 `"verified"`, `"hash_mismatch"`, `"missing"`, `"unverified"`, or `None`
@@ -36,7 +46,11 @@ Each scenario's report entry also carries `artifact_verification`, one of
 `"unverified"` means the raw artifact path was not locally resolvable in
 this run (no `--artifacts-dir` given, e.g. a CI run that only fetched the
 `*.jsonl` summaries) - it is recorded explicitly rather than silently
-passing as verified (khive#945 item 4).
+passing as verified (khive#945 item 4). An absent/malformed `artifact.name`
+or `artifact.sha256`, or a `name` that resolves outside `--artifacts-dir`
+(`../` traversal or an absolute path), is reported `"missing"` instead -
+never `"unverified"` - and downgrades the scenario straight to confounded
+(khive#945 M1/M2).
 
 No manifest scenario is ever reported "measured" when the records
 directory is absent or empty - the coverage percentage is 0% in that case,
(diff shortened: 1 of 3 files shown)


ohdearquant:
APPROVE-WITH-FIXES

Findings: 0 Blocker, 0 High, 3 Medium, 0 Low

### [Medium] An absent artifact digest can still produce a measured result

Evidence: `scripts/perf/flagship_schema.py:368-373` only checks that the
`artifact` object has `name` and `sha256` keys; it does not validate either
value. `scripts/perf/coverage_validator.py:272-277` returns `"unverified"`
before inspecting those fields when `--artifacts-dir` is omitted, and
`scripts/perf/coverage_validator.py:421-425` then preserves the otherwise
measured status. A focused reproduction with `artifact={"name":
"report.json", "sha256": None}` yields no `validate_record()` errors and a
report entry of `status="measured", artifact_verification="unverified"`.
This also diverges from the checked external contract, which requires a
64-character hexadecimal string at `scripts/perf/schemas/flagship-result-v3.json:448-462`.

Why this matters: `unverified` should mean that a valid artifact reference
could not be resolved locally, not that the record has no digest to verify.
This permits a digest-free row to contribute to coverage in the common
no-artifacts-dir CI mode.

Suggested fix: validate a non-empty string name and a 64-lowercase-hex
`sha256` before returning `unverified` (and align `validate_record()` with
the checked JSON schema). Invalid or absent values must downgrade the row to
`confounded`. Add report-level tests for `sha256=None` and an absent `sha256`.

### [Medium] Artifact verification is not confined to `--artifacts-dir`

Evidence: `scripts/perf/coverage_validator.py:275-285` forms the candidate
as `artifacts_dir / name` and hashes it without resolving or checking that it
remains below `artifacts_dir`. A focused reproduction with
`artifact.name="../<sibling>/outside.json"` and the outside file's digest
returns `"verified"`.

Why this matters: the PR promises to verify the raw artifact resolved against
the supplied artifact directory. A ledger row can instead certify an
unrelated readable file outside that directory as its evidence.

Suggested fix: resolve both the directory and candidate, require
`candidate.relative_to(resolved_artifacts_dir)` to succeed before reading,
and return a confounding verification state for absolute or escaping paths.
Add tests for both `../` traversal and absolute names.

### [Medium] Measurement-evidence rules must not make honest error records schema-invalid

Evidence: `scripts/perf/flagship_schema.py:38` declares `"error"`,
`"confounded"`, and `"insufficient_samples"` as valid record statuses, but
`scripts/perf/flagship_schema.py:261-274` rejects empty distributions and
zero-success distributions for every status, while
`scripts/perf/flagship_schema.py:315-331` rejects every positive fallback
count. Thus an honest future `status="error"` row for a scenario crash cannot
be schema-valid. The coverage path already has the correct view-layer guard:
`scripts/perf/coverage_validator.py:351-356` validates first and then returns
`confounded` for every non-`ok` status.

Ruling: this is not acceptable as a schema-wide rule, even though coverage is
the only consumer today and no runner exists yet. `validate_record()` is the
Ledger record contract, and its status enum explicitly models unsuccessful
historical records. Make the non-empty-distribution, successful-sample,
error/timeout-accounting, and fallback-zero requirements conditional on
`status == "ok"` (or put them in a coverage-only measurement-evidence
validator). This does not weaken coverage: after that change,
`scenario_status()` will still reject those rows via its existing
`scenario_status` check at `scripts/perf/coverage_validator.py:355-356`.
Add a regression test proving a structurally valid `status="error"` record
with zero samples/fallback evidence passes schema validation but is reported
`confounded` by coverage.

## Looks Right

- The required rejection cases for empty distributions, missing accounting,
  and fallback-tainted rows are covered at
  `scripts/perf/test_flagship_coverage.py:703-732`.
- The requested `--artifacts-dir` happy path and digest-mismatch path are
  tested at `scripts/perf/test_flagship_coverage.py:755-781`; the mismatch
  is carried through to a `confounded` report at
  `scripts/perf/test_flagship_coverage.py:783-803`.
- `artifact_verification` is an explicit per-scenario JSON field at
  `scripts/perf/coverage_validator.py:430-439`; `"unverified"` is distinct
  from `"verified"`, and a repository search found no downstream report JSON
  consumer that conflates the states. The invalid-digest case above must still
  be corrected before relying on that distinction.
- `git diff --check a3206932..0e9cf4f8` passed.

## Commands Run

- `uv run python -m pytest scripts/perf/test_flagship_coverage.py scripts/perf/test_mcp_bench_client.py -q`: **105 passed in 3.54s** (pytest was available; unittest fallback was not used).
- Focused temporary-directory reproductions: confirmed `sha256=None` becomes
  `measured/unverified` without `--artifacts-dir`, and confirmed a `../`
  artifact name can return `verified` outside the supplied directory.

## What I Did Not Check

- No flagship runner exists yet, so no runner-produced ledger record was
  available for end-to-end execution.

## Re-Review Guidance

Narrow re-review after the three fixes and their regression tests.

Domain utility: SKIPPED — this Python-only review was governed by the PR,
the repository's schema contract, and the loaded review rubric; no separate
domain composition service was available.

ohdearquant:
APPROVE-WITH-FIXES

### [Medium] Malformed `min_successes` can crash coverage validation

Evidence: `scripts/perf/coverage_validator.py:169-218` accepts an optional
`min_successes` manifest value without validating its type or range, and
`scripts/perf/coverage_validator.py:438-440` passes that unchecked value to
`max(flagship_schema.TIMED_MIN_SUCCESSES, declared_min_successes)`. A focused
reproduction with `min_successes="500"` reaches this code and raises
`TypeError: '>' not supported between instances of 'str' and 'int'` rather
than returning a manifest-validation error or a confounded report.

Why this matters: `min_successes` is the new signed-amendment control for a
scenario's measurement floor. A malformed TOML value turns the coverage gate
into a traceback, so its input contract is neither reliable nor
explainable.

Suggested fix: In `load_manifest`, validate an optional `min_successes` as a
non-boolean positive integer (and reject other values with the existing
manifest-error path). Add regression cases for a string, boolean, and
non-positive value; retain the defensive handling at the use site if a
programmatic caller can bypass manifest loading.

## Looks Right

- M1 is resolved. `scripts/perf/flagship_schema.py:409-419` now requires a
  non-empty artifact name and a lowercase 64-hex digest, while
  `scripts/perf/coverage_validator.py:302-311` performs the same validation
  before the no-artifact-directory `unverified` fallback. `sha256=None`
  fails schema validation and is reported `confounded` in a focused run.
- M2 is resolved for escaping references. The candidate is resolved and
  required to be below the resolved artifacts directory at
  `scripts/perf/coverage_validator.py:313-321`; focused `../outside.json`
  and an absolute outside path both return `missing`, never `verified`.
- M3 is resolved. Evidence requirements are conditioned on `status == "ok"`
  at `scripts/perf/flagship_schema.py:286-311` and `:361-368`, while
  `scripts/perf/coverage_validator.py:392-397` still makes every non-`ok`
  record `confounded`. A `status="error"`, empty-distributions,
  fallback-count-3 record has no schema errors and reports `confounded`.
- The 100-success global floor is enforced for `ok` records at
  `scripts/perf/flagship_schema.py:55-60` and `:302-310`. The manifest can
  only raise it at `scripts/perf/coverage_validator.py:438-449`: 99 is
  rejected, 100 passes, `min_successes=1` cannot lower the floor, and a
  declared 500 rejects 400 but accepts 500.
- The regression coverage directly exercises the repaired paths at
  `scripts/perf/test_flagship_coverage.py:403-453` and `:782-954`.
- `git diff --check 0e9cf4f8..cd2ac6f4` passed.

## Commands Run

- `uv run python -m pytest scripts/perf/test_flagship_coverage.py scripts/perf/test_mcp_bench_client.py -q`: **123 passed in 3.33s**.
- Focused `uv run python` reproductions: confirmed the four requested
  round-1 cases and both floor directions; also confirmed the malformed
  `min_successes` traceback above.

## What I Did Not Check

- No runner emits flagship records yet, so no runner-produced record was
  available for a full end-to-end validation.
- Remote CI jobs were still in progress when reviewed; this verdict is based
  on the local requested suite and focused reproductions.

## Re-Review Guidance

Narrow re-review after the manifest validation and regression tests for
`min_successes` are added.

Domain utility: SKIPPED — this Python-only review was governed by the PR,
the checked schema contract, and the loaded review rubric; no separate domain
composition service was available.

ohdearquant:
APPROVE

Narrow round-3 re-review of `cd2ac6f4..5c16d839` only.

## Looks Right

- `scripts/perf/coverage_validator.py:218-223` rejects a supplied `min_successes` when it is a boolean, non-integer, or non-positive value, and includes the offending `repr` in the manifest error. The boolean test intentionally precedes the integer test, so Python's `bool`-is-an-`int` subclass behavior cannot admit `true`/`false`.
- Direct reproduction through `load_manifest` with TOML `min_successes="500"` returned `f1.recall.warm.real: min_successes must be a positive integer, got '500'`; it did not raise a traceback.
- `scripts/perf/coverage_validator.py:445-457` applies the same boolean-first positive-integer check before `max(...)`, returning `confounded` to programmatic callers that bypass manifest loading. Direct calls with `"500"`, `True`, `0`, and `-5` each returned `confounded` with a `min_successes` reason.
- `scripts/perf/test_flagship_coverage.py:129-150` emits Python booleans as TOML `true`/`false`, before the generic integer path. Its two regression cases cover the four malformed shapes at `:180-189` and `:840-848`.
- `git diff --check cd2ac6f4..5c16d839` passed.

## Commands Run

- `uv run python` focused reproduction: confirmed the manifest error for `"500"` and `confounded` outcomes for `"500"`, `True`, `0`, and `-5` through `scenario_status`.
- `uv run python -m pytest scripts/perf/test_flagship_coverage.py scripts/perf/test_mcp_bench_client.py -q`: **125 passed in 3.73s**.

## What I Did Not Check

- This was deliberately limited to the round-3 `min_successes` delta; previously resolved round-1/round-2 findings and unrelated PR changes were not re-reviewed.
- Remote CI was not inspected; the verdict is based on the requested local validation.

## Re-Review Guidance

No further re-review is needed for this narrow fix.

Domain utility: MEDIUM — the composed validation/testing guidance reinforced the boundary checks, while the verdict rests on direct code inspection and reproduction.

oceanwaves630 (approved):
Gate pass. Verified at head 5c16d839: signed Amendment 1 §2.1 floor mechanism present (TIMED_MIN_SUCCESSES=100, raise-only via max(floor, declared min_successes), schema-level >=100 enforcement for ok records with the nearest-rank p99 rationale stated); error/timeout rows are writable but never count toward measured coverage (regression test present); malformed min_successes rejected at load and confounded at use-site. Three-round arc clean: r1 3 Medium -> r2 1 Medium -> r3 narrow APPROVE zero findings at this exact head, 125 tests. On merge, PR1 #862 conditional ratification completes and the flagship freeze lifts per Amendment 1 §10.

Issue #644, opened 2026-07-05
daemon: ambiguous response loss can double-forward a mutating request (data duplication)


ohdearquant:
## Summary

When the persistent daemon commits a mutating request but the connection closes before the
client reads the response frame, the client cannot distinguish "commit succeeded, response
lost" from "request never processed." The current recovery path treats a read/parse failure
as a transport failure, respawns, and re-forwards the identical request — double-applying a
non-idempotent write.

## Mechanism

- `crates/khive-mcp/src/daemon.rs:139` writes the request frame and `:142` reads the response
  frame. The parse-failure / respawn branch (`daemon.rs:513`, `daemon.rs:545`) re-forwards on
  read failure without classifying the op as read-vs-write or carrying an idempotency token.
- The write frame is sent *before* the read; a commit can therefore land on the server side
  while the response never reaches the client, and the retry re-sends the same mutation.

Affected ops: all daemon-forwarded non-idempotent verbs — `create`, `update`, `delete`,
`link`, `merge`, proposal/review verbs, GTD writes, `remember`, feedback/prune, comm writes,
schedule writes, knowledge writes, session writes.

## Impact

Correctness, not latency. In the (rare) window where the daemon commits then drops the
connection before responding, the client re-issues the write → duplicate entities/edges/
memories plus duplicate downstream indexing and event work. This is a data-duplication bug,
so it should be prioritized above the latency findings from the same sweep.

## Falsification / repro

Inspect the write-frame-before-read-frame ordering and the parse-failure retry branch;
confirm there is no idempotency token, no read/write op classification, and no guard against
re-forwarding a possibly-committed write. A fault-injection test (close the socket after the
server commits, before it writes the response) should reproduce a duplicate write.

## Proposed direction

Classify ops as idempotent vs mutating at the forward boundary. For mutating ops, either
(a) attach a client-generated idempotency token that the daemon dedups on, or (b) do not
auto-retry on ambiguous response loss — surface the ambiguity to the caller instead of
silently re-applying.

ohdearquant:
Closing via backlog triage 2026-07-09 (verified against current main at 3d03c8f0): Fixed by PR #698 (commit 4865fc07); `ambiguous_forward_error()` (daemon.rs:909-920) implements never-retry/never-fallback fix, with end-to-end exactly-once test (daemon.rs:3922).

PR #949, opened 2026-07-13
fix(daemon): strict mode fails request on fallback (#947)

### LINKED ISSUE #947: daemon: KHIVE_DAEMON_STRICT=1 must fail the request on fallback, not log-and-complete locally
## Summary

`KHIVE_DAEMON_STRICT=1` is documented as making daemon fallback an explicit failure, but the current implementation logs the fallback, increments the fallback counters, and still completes the request through local dispatch. A strict-mode caller therefore gets a successful response that may have executed on the wrong path.

## Why now

Benchmark Program SPEC Amendment 1 (signed 2026-07-13) §3 makes daemon-engagement proof a coverage requirement for flagship measurements. The interim v0 proof is "strict mode active AND `daemon_fallback_count == 0`" — which is only sound if strict mode genuinely prevents a fallback-satisfied request from being reported as success. With the current behavior, a strict-mode benchmark run can silently measure local dispatch and attribute the numbers to the daemon path.

## Expected behavior

Under `KHIVE_DAEMON_STRICT=1`, a request that would fall back to local dispatch returns an error to the caller (naming the fallback reason) instead of completing locally. Non-strict behavior is unchanged. Counters and the existing fallback telemetry stay as they are.

## Evidence

`crates/khive-mcp/src/daemon.rs` — the strict-mode branch records/logs fallback state but the request proceeds through the local path; the fallback accessors never convert the condition into a request failure (verified against main in the bench-design TBV, 2026-07-13).

## Scope

Small behavior fix at the fallback decision point + a test that a strict-mode request errors when the daemon is unreachable. Graduated fail-loud precedent: the config_id topology parity fix (silent local-dispatch elimination) established the pattern.

ohdearquant:
## Summary

`KHIVE_DAEMON_STRICT=1` is documented as making daemon fallback an explicit failure, but it previously only logged the fallback (elevating `config_mismatch`/`namespace_mismatch` to `error!` + a dedicated violation counter, per the graduated fail-loud policy from bba4c4a7) and still let the request complete through local dispatch. A strict-mode caller could therefore get a successful response that actually executed on the local path, not the daemon.

This PR makes every fallback path in `crates/khive-mcp/src/daemon.rs` reject the request outright under `KHIVE_DAEMON_STRICT=1`, naming the fallback reason in the error, instead of logging and completing locally. Non-strict behavior is unchanged.

## Why

Benchmark Program SPEC Amendment 1 §3 (signed 2026-07-13) makes daemon-engagement proof a coverage requirement for flagship measurements. The interim v0 proof is "strict mode active AND `daemon_fallback_count == 0`" — sound only if strict mode genuinely prevents a fallback-satisfied request from being reported as success. Before this fix, a strict-mode benchmark run could silently measure local dispatch and attribute the numbers to the daemon path.

This is the same graduated fail-loud pattern established by the config_id topology parity fix (bba4c4a7, "eliminate silent local-dispatch fallback"): that commit added `KHIVE_DAEMON_STRICT` and elevated `Illegitimate`-severity fallbacks (`ConfigMismatch`/`NamespaceMismatch`) to an `error!` log + `FALLBACK_STRICT_VIOLATIONS` counter, but explicitly kept the fallback itself always-on ("the fallback is never disabled"). This PR is the next graduation step: strict mode now also rejects the request, for every `FallbackReason` (not just the `Illegitimate` tier — that tier still only governs the WARN vs ERROR log level, which is untouched).

## Behavior

| Fallback reason | Non-strict (default) | Strict (`KHIVE_DAEMON_STRICT=1`) |
|---|---|---|
| `no_socket` (daemon unreachable) | Completes locally (`None`) | `Some(Err(..))` naming `no_socket` |
| `config_mismatch` | Completes locally (`None`) | `Some(Err(..))` naming `config_mismatch` |
| `namespace_mismatch` | Completes locally (`None`) | `Some(Err(..))` naming `namespace_mismatch` |
| `KHIVE_NO_DAEMON=1` (explicit opt-out) | Completes locally (`None`) | Completes locally (`None`) — unaffected; it's a deliberate opt-out, not a fallback, and nothing is recorded for it |

Fallback counters and the graduated WARN/ERROR structured log (`record_fallback`, D2-R1/D2-R3) are untouched — they still increment/log exactly as before. The new `fallback_or_reject` helper wraps `record_fallback` and is the single point that decides `None` vs `Some(Err(..))` at every one of the five call sites (`map_response` ×3, `forward_or_spawn` ×2).

## Test evidence

Added to `crates/khive-mcp/src/daemon.rs`:
- 3 fast unit tests against `fallback_or_reject` directly (non-strict returns `None` and still counts; strict `no_socket`/`config_mismatch` errors and names the reason + `KHIVE_DAEMON_STRICT` in the message).
- 2 integration tests driving `forward_or_spawn` against a genuinely unreachable daemon (no `KHIVE_NO_DAEMON` opt-out): non-strict still completes locally, strict mode rejects with the reason named.
- All env-var-reading tests use `#[serial]` (existing crate pattern via `serial_test`), robust to parallel test execution.

```
cargo fmt --all -- --check         → clean
cargo clippy -p khive-mcp --all-targets -- -D warnings → clean
cargo test -p khive-mcp            → 211 lib tests + 113 integration tests passed, 0 failed
```

Closes #947

### DIFF
--- a/crates/khive-mcp/docs/design.md
+++ b/crates/khive-mcp/docs/design.md
@@ -78,7 +78,11 @@
 
 - The `daemon.rs` module provides the client side: `forward_or_spawn` connects
   to a warm daemon, auto-spawns it on first use, and maps responses to MCP
-  error types. Any failure falls back to `None` so the caller dispatches locally.
+  error types. Ordinary fallback paths return `None` so the caller dispatches
+  locally. `KHIVE_DAEMON_STRICT=1` (#947) turns a recordable fallback into a
+  caller-visible per-op error instead of completing it locally; `KHIVE_NO_DAEMON`
+  and the `request` tool's `save_to` bypass remain intentional local paths,
+  unaffected by strict mode.
 - The daemon is bound to `~/.khive/khived.sock`. Namespace mismatches trigger
   local-dispatch fallback.
 - `warm_all` is called in a background task after the daemon socket is bound so
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -5,7 +5,7 @@
 //! first use, and maps responses to MCP error types. Ordinary fallback paths
 //! return `None` so the caller can dispatch locally. `KHIVE_DAEMON_STRICT=1`
 //! (#947) is the exception: it turns a recordable fallback into a
-//! caller-visible per-op error instead, via [`fallback_or_reject`].
+//! caller-visible per-op error instead, via `fallback_or_reject`.
 //! `KHIVE_NO_DAEMON` and `crate::server`'s `save_to` bypass remain
 //! intentional, unconditional local paths — neither is affected by strict
 //! mode, since nothing is ever recorded or falls back for them.
@@ -215,7 +215,7 @@ enum FallbackSeverity {
 /// structured event plus a dedicated violation counter (D2-R1) — see
 /// [`record_fallback`]. It also, independent of severity, rejects the request
 /// outright instead of letting it complete through local dispatch (#947) —
-/// see [`fallback_or_reject`]. Together these make an illegitimate mismatch
+/// see `fallback_or_reject`. Together these make an illegitimate mismatch
 /// impossible to miss AND make "strict mode active" a sound proof that no
 /// request in the window was served off the local fallback path.
 ///
@@ -320,7 +320,7 @@ pub(crate) fn reset_fallback_counters() {
 /// (D2-R3 — strict mode keys on `FallbackReason`, never on "did a fallback
 /// happen at all"). This function only records; it never decides whether the
 /// caller proceeds locally — every call site pairs it with
-/// [`fallback_or_reject`] (#947), which is what converts a strict-mode
+/// `fallback_or_reject` (#947), which is what converts a strict-mode
 /// fallback into a hard error instead of a local dispatch.
 fn record_fallback(
     reason: FallbackReason,
@@ -1738,7 +1738,7 @@ async fn wait_for_boot_quiescence_then_reprobe(frame: &DaemonRequestFrame) -> Bo
 /// already decided at the daemon and the caller must not dispatch locally.
 ///
 /// #947: under `KHIVE_DAEMON_STRICT=1`, the `NoSocket` case is instead
-/// `Some(Err(..))` — see [`fallback_or_reject`] — so it no longer joins
+/// `Some(Err(..))` — see `fallback_or_reject` — so it no longer joins
 /// `KHIVE_NO_DAEMON` as a safe-to-dispatch-locally outcome. `KHIVE_NO_DAEMON`
 /// itself is unaffected: it is the caller's explicit, unconditional opt-out
 /// of the daemon (not a fallback — nothing is ever recorded or counted for
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -207,10 +207,12 @@ enum FallbackSeverity {
 
 /// `KHIVE_DAEMON_STRICT=1` elevates `Illegitimate`-severity fallbacks
 /// (`ConfigMismatch`, `NamespaceMismatch`) from a WARN to an error-level
-/// structured event plus a dedicated violation counter (D2-R1). It never
-/// disables the fallback itself — the request still completes locally so the
-/// caller is never dropped (SPEC_DRAFT §3 D2) — it only makes an illegitimate
-/// mismatch impossible to miss.
+/// structured event plus a dedicated violation counter (D2-R1) — see
+/// [`record_fallback`]. It also, independent of severity, rejects the request
+/// outright instead of letting it complete through local dispatch (#947) —
+/// see [`fallback_or_reject`]. Together these make an illegitimate mismatch
+/// impossible to miss AND make "strict mode active" a sound proof that no
+/// request in the window was served off the local fallback path.
 ///
 /// No hosted-vs-local auto-detection signal exists in this codebase (there is
 /// no build-time or runtime "is this the hosted fleet image" flag anywhere in
@@ -311,8 +313,10 @@ pub(crate) fn reset_fallback_counters() {
 /// every other case (non-strict mode, or a `RolloutTransient`/`NoDaemon`
 /// reason regardless of mode) logs at `warn!`, exactly as before this change
 /// (D2-R3 — strict mode keys on `FallbackReason`, never on "did a fallback
-/// happen at all"). The fallback itself is never disabled either way — the
-/// caller always proceeds to dispatch locally after this call returns.
+/// happen at all"). This function only records; it never decides whether the
+/// caller proceeds locally — every call site pairs it with
+/// [`fallback_or_reject`] (#947), which is what converts a strict-mode
+/// fallback into a hard error instead of a local dispatch.
 fn record_fallback(
     reason: FallbackReason,
     config_id_client: &str,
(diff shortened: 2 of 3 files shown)


ohdearquant:
APPROVE-WITH-FIXES

Findings: 0 Blocker, 0 High, 1 Medium, 1 Low

### [Medium] Strict fallback is a tool/RPC error, not the required per-op error

Evidence: `crates/khive-mcp/src/daemon.rs:378` builds `McpError::internal_error` for a strict fallback. `crates/khive-mcp/src/server.rs:1492`–`1493` returns that `Result` directly from the MCP `request` handler, whose signature is `Result<String, McpError>` (`crates/khive-mcp/src/server.rs:1479`). The existing caller-boundary regression deliberately asserts the same `Err(McpError)` shape at `crates/khive-mcp/src/server.rs:2389`–`2402`.

Why this matters: The request wire contract represents operation failures inside the returned request envelope as `{\"ok\": false, ...}` entries. A strict fallback therefore reaches the caller and is not swallowed, but it bypasses the per-op result/summary contract required for this change; a batch cannot report its fallback as the specified per-op error.

Suggested fix: Convert a strict fallback at the `request`/operation-result boundary into the normal failed-op envelope (including the correct summary and chain-abort behavior), then add an end-to-end `request()` test for an unreachable daemon that asserts the envelope, fallback reason, and absence of local dispatch.

### [Low] Fallback documentation still states the pre-#947 invariant

Evidence: `crates/khive-mcp/src/daemon.rs:5`–`6` says every failure falls back to `None`, while strict fallback now returns `Some(Err(...))` at `crates/khive-mcp/src/daemon.rs:370`–`387`. The user-facing design note has the same unconditional statement at `crates/khive-mcp/docs/design.md:79`–`83`.

Why this matters: This contradicts the new strict-mode behavior and can lead future edits to reintroduce a local fallback or to misunderstand the transport contract.

Suggested fix: State that ordinary fallback paths return `None`, but `KHIVE_DAEMON_STRICT=1` turns a recordable fallback into a caller-visible error; retain the separate `KHIVE_NO_DAEMON` and `save_to` intentional local paths as exceptions.

## Looks Right

- The five production recordable fallback paths are routed through `fallback_or_reject`: three response mismatches at `crates/khive-mcp/src/daemon.rs:554`–`579`, plus recovery failure and boot-quiescence fallback at `crates/khive-mcp/src/daemon.rs:1791`–`1800` and `crates/khive-mcp/src/daemon.rs:1833`–`1845`.
- `fallback_or_reject` records before rejecting, so strict rejection increments the existing per-reason telemetry (`crates/khive-mcp/src/daemon.rs:375`–`387`); the strict unreachable-daemon test checks the `no_socket` counter (`crates/khive-mcp/src/daemon.rs:2619`–`2651`).
- With strict mode unset, the helper returns `None` and preserves local fallback (`crates/khive-mcp/src/daemon.rs:2465`–`2476`); the real unreachable-daemon test covers that mode (`crates/khive-mcp/src/daemon.rs:2590`–`2614`).
- `KHIVE_DAEMON_STRICT` is read live via `std::env::var`, not cached (`crates/khive-runtime/src/daemon.rs:1318`–`1326`; `crates/khive-mcp/src/daemon.rs:224`–`226`). New env-mutating tests are `#[serial]`.
- The only local-dispatch paths outside the five recordable fallbacks are the explicit `KHIVE_NO_DAEMON` opt-out (`crates/khive-mcp/src/daemon.rs:1741`–`1744`) and the pre-existing `save_to` bypass (`crates/khive-mcp/src/server.rs:1484`–`1496`); neither attempts then falls back from daemon dispatch.
- The implementation follows the existing graduated fail-loud structure introduced by `bba4c4a7`: it preserves `record_fallback` as the telemetry owner and isolates strict disposition in one helper.

## Commands Run

- `cargo clippy -p khive-mcp --all-targets -- -D warnings`: passed.
- `cargo test -p khive-mcp`: passed, 211 tests.

## What I Did Not Check

- I did not run an external MCP client process; source inspection and the existing caller-boundary test establish that strict fallback currently returns an `McpError`, rather than a request-result envelope.

## Re-Review Guidance

Narrow re-review after the response-shaping fix: verify single, batch, and chain error envelopes through `request()` under an unreachable daemon, while retaining the no-local-dispatch and telemetry assertions.

Domain utility: SKIPPED — no domain corpus was available; the accepted daemon ADRs, implementation, and tests provided the required evidence.

ohdearquant:
APPROVE

Reviewed only `c2ae20ec..dcc08aa8` for PR #949.

## Looks Right

- Strict fallback is now converted at the `request()` boundary into the normal per-op response rather than returned as an RPC error: `crates/khive-mcp/src/server.rs:1490`–`1509`. The marker is crate-private and is created at the one strict fallback decision point with a Boolean tag plus a reason (`crates/khive-mcp/src/daemon.rs:379`–`403`). Daemon-originated errors cannot supply structured `McpError.data`: their wire field is `Option<String>` (`crates/khive-runtime/src/daemon.rs:348`–`351`) and `map_response` constructs untagged errors with `data: None` (`crates/khive-mcp/src/daemon.rs:552`–`608`). Therefore protocol, oversize-frame, and ambiguous-forward errors do not accidentally match the marker.
- The strict envelope exactly matches normal error aggregation: single/parallel produce one failed entry per input op (`crates/khive-mcp/src/server.rs:1547`–`1564`) and chains fail op 0 then mark each remainder `aborted: true` (`crates/khive-mcp/src/server.rs:1547`–`1559`). Its `failed`/`aborted` arithmetic (`crates/khive-mcp/src/server.rs:1567`–`1576`) agrees with the normal chain contract (`crates/khive-mcp/src/server.rs:1013`–`1025`).
- A malformed request remains a normal RPC `invalid_params` error: the strict shaper reuses `parse_request(...).map_err(dsl_err_to_mcp)` (`crates/khive-mcp/src/server.rs:1540`), matching the parser's empty/invalid-input rejection (`crates/khive-request/src/parser/dispatch.rs:16`–`25`). It never invents a zero-op envelope.
- The new end-to-end regression uses a genuinely unreachable daemon under strict mode and asserts the single, batch, and chain envelopes, their summary values, the fallback reason, and unchanged local state (`crates/khive-mcp/src/server.rs:2506`–`2669`).
- The #644 ambiguous post-write boundary remains untagged (`crates/khive-mcp/src/daemon.rs:1281`–`1286`) and the unchanged `request()` regression still asserts the raw RPC error plus no local double-dispatch (`crates/khive-mcp/src/server.rs:2399`–`2504`).
- The daemon module documentation and design note now distinguish ordinary local fallback from strict per-op failure (`crates/khive-mcp/src/daemon.rs:1`–`11`; `crates/khive-mcp/docs/design.md:79`–`85`).

## Commands Run

- `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo clippy -p khive-mcp --all-targets -- -D warnings`: passed.
- `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo test -p khive-mcp`: passed — 212 library tests and 113 integration tests.
- `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo test -p khive-mcp server::tests::request_strict_fallback_lands_as_failed_op_envelope_not_rpc_error -- --exact`: passed.
- `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo test -p khive-mcp server::tests::request_returns_ambiguous_forward_error_without_local_double_dispatch -- --exact`: passed.
- `cargo fmt --all -- --check` and `git diff --check c2ae20ec..dcc08aa8`: passed.

## What I Did Not Check

- No external MCP-client process; the exercised handler-level tests cover the real unreachable-socket and post-write daemon paths relevant to this delta.

## Re-Review Guidance

- No further review is required for this narrow fix.

Domain utility: SKIPPED — the applicable request and daemon contracts were fully specified by repository ADRs, source, and regression tests.

oceanwaves630 (approved):
Gate pass. r2 clean APPROVE was at dcc08aa8; I verified the delta dcc08aa8..e257b1ce myself: 1 commit, 1 file, 4 doc-comment lines converting private intra-doc links to plain code spans — no code change, so the r2 verdict carries. Strict-mode fail-loud verified per Amendment 1 §3: fallback under KHIVE_DAEMON_STRICT=1 returns a caller-visible per-op error. Auto-merge armed; merges on green rollup.

PR #950, opened 2026-07-13
fix(perf): flagship manifest corrections per Benchmark SPEC Amendment 1

ohdearquant:
## Summary

Implements the flagship manifest correction mandated by the signed Benchmark Program SPEC Amendment 1 §10 item 4:
`.khive/workspaces/20260710/bench-program/AMENDMENT-1-flagship-measurement-contract.md`

Five corrections, each verified against the amendment text:

1. **500K cohorts (§7)** — added `f1.recall.warm.real.n500k` (`scale.memories = 500000`) and `f4.traverse.powerlaw.n500000.depth3.real` (`scale.nodes = 500000`), the exact cohorts the amendment names as missing ("F1 and F4 scenario sets gain a 500K cohort before any 'exact cohort' claim is made"). No other cohorts were invented.

2. **Client-side censor ≥ 45s (§2 "Deadline classifiability")** — bumped every timed, non-`admin_cli` scenario's `request_deadline_ms` from `30000` to `45000` (92 scenarios), and `oq_rulings.oq2_measurement_deadline_ms` to match, so the harness measurement-censoring cap strictly exceeds the 30000 ms server-side recall deadline (#919) instead of racing it 30s/30s. The two `admin_cli` process-wall-clock scenarios (`f6.git_ingest.cli.production`, `f6.code_ingest.cli.production`) already had `request_deadline_ms = 300000` and are unaffected — added `oq_rulings.oq2_server_recall_deadline_ms = 30000` to make the two deadlines an explicit, checkable pair.

3. **OQ1 resolved (§2 "Surface rule")** — replaced the manifest's silent `admin_cli` "approved" marker (which only covered git-ingest/code-ingest) with the amendment's explicit, exhaustive two-exception surface ruling. `oq1_admin_ingest_surface` now cites Amendment 1 §2 by name; added `oq1_admin_surface_scenario_ids` as the exhaustive list.

4. **F10 removed from real-MCP coverage (§2 "Surface rule")** — documented `f10.daemon.probe_only.floor` (the named "F10 daemon-control probe") as a bounded admin-surface exception alongside F6 git-ingest/code-ingest, reported in a separate admin-surface coverage class and never counted toward real-MCP percentage. Its `surface = "raw_daemon_control"` was already correctly distinct from `mcp_daemon`; the ruling documenting that exclusion is now explicit in `oq_rulings.oq1_admin_surface_scenario_ids` plus an inline comment at its declaration site. Scoped narrowly to the probe scenario per the amendment's precise wording — the separate §4 "F10 payload curve removed" executability item is handled in correction 5 below.

5. **F10 payload curve removed / §4 replacement wired (§4 "Scenario executability")** — removed `f10.daemon.payload_size_curve.mcp`: its independent variable (`scale.payload_bytes_max`, declared up to 65536 bytes vs the 8 MiB daemon frame cap) was never generatable through its declared surface — public `stats` has no payload-size control at all. Checked the manifest for an existing `list`/`traverse` limit-sweep that genuinely varies response payload; none existed (`traverse` only varies node count/depth, no scenario used a `limit` param), so added a minimal real-MCP replacement cohort family: `f10.list.limit_sweep.{small,medium,large}.mcp` (`operation = "list"`, `surface = "mcp_daemon"`, `scale.limit` = 10/100/1000), reusing the existing read-only `kg_search_relevance_entity` fixture from F2 (legal fixture reuse per §6). The daemon frame cap itself continues to be probed only in the admin-surface class by the existing `f10.daemon.probe_only.floor` scenario — already the named §2 admin-surface exception, so no second admin-surface probe was added. An inline comment at the F10 section marks both replacement scenarios and their §4 provenance.

Test fixtures (`test_flagship_coverage.py`) updated to match: the `oq_rulings` deadline assertion now expects 45000/30000, plus five new regression tests covering the 500K cohorts, the admin-surface exception list, the 45s-over-30s censor invariant across every timed scenario, the removed payload-curve scenario's absence (by id and by substring, so it can't silently reappear under a new id), and the §4 replacement cohort family's presence/shape.

No Rust code touched. The amendment file itself was not modified.

## Test plan

- [x] `uv run python -m pytest scripts/perf/test_flagship_coverage.py scripts/perf/test_mcp_bench_client.py -q` — **130 passed**
- [x] Focused `coverage_validator.load_manifest()` run over the edited manifest — **96 scenarios, 0 structural errors**
- [x] `coverage_validator.py --strict` CLI run — exit 0, 0% measured (expected with no records dir, by design)
- [x] Focused clean-load proof confirming `f10.daemon.payload_size_curve.mcp` absent and all three `f10.list.limit_sweep.*.mcp` present

### DIFF
--- a/scripts/perf/flagship_workloads.toml
+++ b/scripts/perf/flagship_workloads.toml
@@ -25,7 +25,7 @@
 # both are reported in a separate admin-surface coverage class and never
 # count toward the real-MCP percentage. No other exception exists unless
 # added by signed amendment.
-oq1_admin_ingest_surface = "admin_cli/raw_daemon_control exception bounded to exactly two scenarios per Amendment 1 §2: F6 git-ingest/code-ingest (admin CLI is the product surface for it) and the F10 daemon-control probe; reported as a separate admin-surface coverage class, never counted toward the real-MCP percentage"
+oq1_admin_ingest_surface = "admin_cli/raw_daemon_control exception bounded to exactly two exception classes per Amendment 1 §2 - F6 git-ingest/code-ingest (admin CLI is the product surface for it, two scenario rows) and the F10 daemon-control probe (one scenario row), three scenario ids total; reported as a separate admin-surface coverage class, never counted toward the real-MCP percentage"
 # Amendment 1 §2 admin-surface coverage class membership (exhaustive; every
 # other scenario_id in this manifest is real-MCP coverage by default via
 # its surface = "mcp_daemon" declaration).
--- a/scripts/perf/flagship_workloads.toml
+++ b/scripts/perf/flagship_workloads.toml
@@ -4,6 +4,8 @@
 # THE PLAN: .khive/workspaces/20260711/bench-overhaul/DESIGN.md
 # (Typed interfaces > Scenario contract; Coverage definition; PR 1 slice).
 # Signed decisions: .khive/workspaces/20260710/bench-program/SPEC-draft.md
+# Amendment (binding, supersedes conflicting DESIGN.md/SPEC-draft.md text):
+# .khive/workspaces/20260710/bench-program/AMENDMENT-1-flagship-measurement-contract.md
 #
 # This manifest is data only. It declares every required F1-F10 scenario,
 # its measured surface, scale tier, cold/warm obligation, required
@@ -15,14 +17,33 @@
 # scenario_id convention: f<N>.<verb>.<arm>.<embedder> (N = F1..F10).
 
 [oq_rulings]
-# OQ1 (admin-ingest surface conflict): APPROVED CLI-boundary exception
-# (DESIGN.md D5). git-ingest and code-ingest are measured at the real CLI
-# process boundary (surface = "admin_cli"), not as new MCP verbs.
-oq1_admin_ingest_surface = "admin_cli exception approved for git-ingest and code-ingest (D5)"
+# OQ1 (admin-ingest surface conflict): RESOLVED by Amendment 1 §2 "Surface
+# rule" - this replaces the prior silent "approved" marker (DESIGN.md D5).
+# Real MCP tools/call through the daemon (surface = "mcp_daemon") is the
+# default and only surface that counts toward real-MCP coverage. Exactly
+# two admin-surface exceptions are named and bounded by the amendment;
+# both are reported in a separate admin-surface coverage class and never
+# count toward the real-MCP percentage. No other exception exists unless
+# added by signed amendment.
+oq1_admin_ingest_surface = "admin_cli/raw_daemon_control exception bounded to exactly two scenarios per Amendment 1 §2: F6 git-ingest/code-ingest (admin CLI is the product surface for it) and the F10 daemon-control probe; reported as a separate admin-surface coverage class, never counted toward the real-MCP percentage"
+# Amendment 1 §2 admin-surface coverage class membership (exhaustive; every
+# other scenario_id in this manifest is real-MCP coverage by default via
+# its surface = "mcp_daemon" declaration).
+oq1_admin_surface_scenario_ids = [
+    "f6.git_ingest.cli.production",
+    "f6.code_ingest.cli.production",
+    "f10.daemon.probe_only.floor",
+]
 # OQ2 (production request deadline): no product SLO exists. The harness
 # uses a labeled measurement-censoring safety cap, not a timeout claim.
-oq2_measurement_deadline_ms = 30000
-oq2_measurement_deadline_label = "harness safety cap for measurement censoring, not a product SLO"
+# Amendment 1 §2 "Deadline classifiability": the client-side censor must
+# strictly exceed the server-side recall deadline (30000 ms per #919) so a
+# timeout is attributable to exactly one deadline - the manifest's prior
+# 30000/30000 race is fixed by raising the harness censor to 45000 ms.
+oq2_measurement_deadline_ms = 45000
+oq2_measurement_deadline_label = "harness safety cap for measurement censoring, not a product SLO; set to 45000 ms to strictly exceed the 30000 ms server-side recall deadline (#919) per Amendment 1 §2"
+oq2_server_recall_deadline_ms = 30000
+oq2_server_recall_deadline_note = "server-side recall deadline default (#919), distinct from the harness measurement-censoring cap above; the two must never be equal (Amendment 1 §2)"
 oq2_mcp_client_default_timeout_ms = 300000
 oq2_mcp_client_default_note = "MCP client default request timeout, recorded as reference only, not the scenario deadline"
 # OQ7 (freshness window): approved per-runner-class staleness bound used by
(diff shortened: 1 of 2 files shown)


ohdearquant:
APPROVE-WITH-FIXES

Reviewed PR #950 at `16ec4122a321cef4059d793b964a60f5b9dd4cf2` against the signed Benchmark Program SPEC Amendment 1.

### [Medium] OQ1's human-readable ruling contradicts its enumerated scenario membership

Evidence: `scripts/perf/flagship_workloads.toml:28` says the exception is bounded to "exactly two scenarios," but `scripts/perf/flagship_workloads.toml:32` through `scripts/perf/flagship_workloads.toml:35` enumerate three scenario IDs: two F6 admin-CLI scenarios and the F10 daemon-control probe. The accompanying test deliberately asserts all three at `scripts/perf/test_flagship_coverage.py:140` through `scripts/perf/test_flagship_coverage.py:143`.

Why this matters: Amendment 1 §2 bounds two *exception classes* (F6 and the F10 probe), not two scenario rows. The marker is intended to be the auditable OQ1 ruling; its text must not undercount its own declared membership.

Suggested fix: Change the text to say "exactly two exception classes" (F6, represented by the two admin-CLI rows, and the F10 probe), or say "three scenario IDs in two exception classes."

### [Medium] The global deadline test does not enforce the signed 45-second floor

Evidence: `scripts/perf/flagship_workloads.toml:43` sets the OQ2 client-censor floor to 45,000 ms, but `scripts/perf/test_flagship_coverage.py:194` through `scripts/perf/test_flagship_coverage.py:208` assert only that each non-`admin_cli` deadline is greater than the 30,000 ms server deadline. An in-memory mutation setting every non-admin deadline to 30,001 ms still passes this test.

Why this matters: Amendment 1 §2 requires a client censor of at least 45 seconds. The current manifest is correct, but the regression guard permits a future value between 30,001 and 44,999 ms.

Suggested fix: For every non-`admin_cli` timed scenario, assert `request_deadline_ms >= rulings["oq2_measurement_deadline_ms"]` (which is independently asserted as 45,000), while retaining a strict check that that floor exceeds the server deadline.

## Looks Right

- `coverage_validator.load_manifest` returns no errors for the final manifest and reports 96 scenarios.
- The arithmetic is correct: `origin/main` has 92 scenarios; the first PR commit has 94 after the two 500K cohorts; HEAD has 96 after removing one F10 curve and adding three list sweeps. The final diff adds only the two required cohorts plus those three sweeps, and removes only `f10.daemon.payload_size_curve.mcp`.
- The F1 and F4 additions are correctly declared as 500,000-memory/node cohorts at `scripts/perf/flagship_workloads.toml:181` through `scripts/perf/flagship_workloads.toml:196` and `scripts/perf/flagship_workloads.toml:757` through `scripts/perf/flagship_workloads.toml:772`.
- All final 94 non-`admin_cli` scenarios have a 45,000 ms deadline; the two F6 `admin_cli` scenarios remain at 300,000 ms. OQ2 records the distinct 30,000 ms server recall deadline at `scripts/perf/flagship_workloads.toml:43` through `scripts/perf/flagship_workloads.toml:48`.
- F10's daemon-control probe is `raw_daemon_control`, explicitly documented as excluded from real-MCP coverage at `scripts/perf/flagship_workloads.toml:1552` through `scripts/perf/flagship_workloads.toml:1572`.
- The F10 replacement is three real-MCP `list` scenarios with declared limits 10/100/1000 and the F2 `kg_search_relevance_entity` fixture at `scripts/perf/flagship_workloads.toml:1608` through `scripts/perf/flagship_workloads.toml:1660`. They are read-only list operations, so the §6 fixture reuse is permitted.
- The removed scenario ID occurs only in an explanatory deletion comment and a negative absence assertion; it is not a manifest scenario or executable reference anywhere under `scripts/perf/`.
- The requested focused suite passes: `130 passed in 3.59s`.

## Commands Run

- `uv run python -m pytest scripts/perf/test_flagship_coverage.py scripts/perf/test_mcp_bench_client.py -q`: 130 passed in 3.59s.
- Direct `coverage_validator.load_manifest(...)` invocation: no errors; 96 scenarios.
- Structured TOML comparison across `origin/main`, the first PR commit, and HEAD: 92, 94, and 96 scenarios respectively; only the expected additions/removal.
- In-memory test mutation setting every non-admin deadline to 30,001 ms: the current global deadline test passed, demonstrating the second finding.

## What I Did Not Check

- No Cargo commands or benchmark execution were run, per review scope; this review was limited to Python/TOML under `scripts/perf/`.

## Re-Review Guidance

Narrow re-review after correcting the OQ1 wording and strengthening the global deadline invariant.

Domain utility: SKIPPED — the signed amendment supplied the governing contract directly.

ohdearquant:
APPROVE

Narrow round-2 re-review of PR #950 at `d6dc7e9f6db309605dd966b2df102f53b08fed09`, limited to the two corrections requested after the prior review.

## Looks Right

- `scripts/perf/flagship_workloads.toml:28` now distinguishes exactly two exception classes from their three scenario IDs. This agrees with the exhaustive three-ID list at `scripts/perf/flagship_workloads.toml:32` through `scripts/perf/flagship_workloads.toml:36`; the resolved OQ1 wording no longer undercounts its declared membership.
- `scripts/perf/test_flagship_coverage.py:199` through `scripts/perf/test_flagship_coverage.py:210` pin the measurement censor floor to 45,000 ms, require it to exceed the 30,000 ms server deadline, and require every non-`admin_cli` scenario deadline to meet that floor. This closes the prior 30,001--44,999 ms regression gap.
- In-memory mutation reproduction: changing all 94 non-`admin_cli` scenario deadlines to 30,001 ms caused `test_every_timed_scenario_client_censor_exceeds_server_deadline` to fail with `30001 not greater than or equal to 45000`, as required.

## Commands Run

- `git diff --check 16ec4122..d6dc7e9f`: passed with no whitespace errors.
- In-memory `coverage_validator.load_manifest` mutation followed by `ManifestTests("test_every_timed_scenario_client_censor_exceeds_server_deadline")`: 1 test run, 1 expected failure, 0 errors; all 94 non-`admin_cli` scenarios set to 30,001 ms.
- `uv run python -m pytest scripts/perf/test_flagship_coverage.py scripts/perf/test_mcp_bench_client.py -q`: 130 passed in 3.76s.

## What I Did Not Check

- Per the requested narrow scope, I did not re-review the earlier Amendment 1 changes, run Cargo checks, or execute benchmarks.

No further re-review is needed for these two resolved findings.

Domain utility: SKIPPED — this narrow pass was governed by the prior findings and the stated fix contract.

oceanwaves630 (approved):
Gate pass. Verified at head d6dc7e9f against the signed Amendment 1: 500K cohorts present for f1/f4 (§7), 45s client-censor floor on non-admin scenarios (§2 deadline classifiability), OQ1 marker replaced by the §2 surface ruling verbatim (bounded to exactly two exception classes, separate admin-surface coverage class never counted toward real-MCP), F10 payload curve removed and replaced by a real-MCP list limit-sweep family over a fixed fixture (§4). Narrow r2 APPROVE at this exact head; both r1 findings (OQ1 wording undercount, deadline-test 30-45s window) confirmed fixed.

PR #953, opened 2026-07-13
fix(ci): harden reliability and reduce PR cost

ohdearquant:
> AI-assisted contribution: Codex generated this change and PR description.

## Summary

- Fixes the shared-state daemon metrics flake from run [29248620259](https://github.com/ohdearquant/khive/actions/runs/29248620259) and makes two additional filesystem-capacity tests deterministic.
- Hardens CI against setup-only GitHub outages, cargo-deny action drift, opaque monolithic logs, dropped benchmark data points, unbounded execution, and excessive macOS PR cost.
- Adds a fail-closed stale-auto-merge guard without broadening fork execution, bot automation, or `pull_request_target` permissions.

### Test reliability

- `metrics_snapshot_reflects_open_transaction_registry` compared two snapshots of a process-global registry. An unrelated handle could depart between them and exactly offset the test's new registration. The test now holds its own uniquely labelled handle, asserts live invariants, and verifies that exact entry disappears on drop.
- Full-workspace validation exposed two blob-floor tests whose assertions treated host-wide APFS free-space deltas as test-owned state. Unrelated cleanup could add more than their 64 MiB cushion. A private capacity-probe seam now gives those tests deterministic 2-byte boundary snapshots; production still calls `fs4::available_space`, and the existing event-driven shared-root lock test still proves mutual exclusion.

### Infrastructure-only retry

- Adds `CI infra retry`, triggered only after a failed first-attempt CI run from a trusted `main` push.
- It retries only when every failed leaf job has no executed steps or exactly one failed `Set up job` step. A setup-only failure of `CI gate` itself is also eligible.
- It reruns failed jobs normally; if matrix fail-fast cancelled a sibling, it reruns the full workflow so the cancelled lane is restored.
- Repository/action/test failures are never retried, and the second attempt is a hard stop.

Live classifier replay:

- run 29253301895: setup-only coverage failure -> eligible
- run 29248620259: test failure plus cancelled sibling -> ineligible
- run 29240258302: cargo-deny command failure -> ineligible

### Supply chain

The floating `EmbarkStudios/cargo-deny-action@v2` resolved to a changed action entrypoint/bundled cargo-deny combination that treated `warn` as a subcommand. It duplicated the repository's explicit pin. This removes that layer and runs bans, licenses, sources, and advisories through one asserted cargo-deny 0.19.9 binary.

The current advisory database also required fixed dependency floors:

- `anyhow >= 1.0.103` for RUSTSEC-2026-0190
- `memmap2 >= 0.9.11` for RUSTSEC-2026-0186

### CI visibility and macOS tiering

- `scripts/ci.sh` now exposes named phases while no-argument and `all` modes preserve the existing local gate and phase order.
- GitHub Actions invokes phases as named steps, giving focused logs and timing.
- Pull requests retain full Linux coverage but use a focused macOS portability lane: all-target/all-feature compilation plus the database, runtime, MCP, git-pack, and CLI boundary tests.
- Main/integration pushes, the daily 06:17 UTC run, and manual dispatch retain the full suite on both operating systems.
- Matrix fail-fast remains enabled to control cost; the workflow comment documents that a macOS-only failure may surface one round trip after a fast Linux failure.
- `CI gate` keeps its required name and remains the aggregate check.

### Benchmark and timeout reliability

- Main/release bench tracking now uses SHA-specific concurrency groups, so merge bursts cannot replace pending per-commit data points. PR runs remain ref-grouped and cancellable.
- Explicit execution timeouts now cover CI, the reusable component benchmarks, pipeline/ANN benchmarks, and auto-merge.
- This deliberately does not claim to solve a job orphaned before runner assignment: GitHub job timeouts start only after assignment.

### Stale auto-merge guard

A dependency-free `pull_request/synchronize` job checks same-repository, human-authored PRs only. It never checks out PR code and has job-scoped `pull-requests: write`.

- If the current auto-merge request predates the synchronize event, it disables it.
- If a reviewer deliberately re-arms auto-merge afterward, the later GraphQL `enabledAt` timestamp preserves that request.
- Reruns are idempotent.
- `CI gate` depends on the guard, so API/mutation failure blocks the required aggregate instead of failing open.
- Bot-authored PRs remain under the separately audited autonomous workflow; fork PRs retain the existing manual-approval/read-only boundary.

## Test plan

- [x] `cargo test --manifest-path crates/Cargo.toml --workspace` — complete workspace and doctest pass, zero failures
- [x] `cargo clippy --manifest-path crates/Cargo.toml --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --manifest-path crates/Cargo.toml --all -- --check`
- [x] `actionlint -no-color`
- [x] `shellcheck scripts/ci.sh` and `sh -n scripts/ci.sh`
- [x] `deno fmt --check` for all changed workflows
- [x] asserted cargo-deny 0.19.9 with `--all-features check bans licenses sources advisories`
- [x] `python3 -m unittest scripts.perf.test_publish_ledger -v` — 3/3
- [x] daemon focused test plus five full `khive-runtime` repetitions — 4,180/4,180
- [x] blob module repeated 25 times — 575/575; full `khive-db` library — 334/334
- [x] reduced macOS PR compile and platform-test phases
- [x] pre-commit hooks

## ADR

n/a — operational CI/test fixes; no product or typed-graph contract changes.

## AI-assisted contribution checklist

- [x] Every claim in this PR description matches the actual diff
- [x] Any agent-authored comment / PR body starts with an attribution line
- [x] `cargo test` output included for behavior-changing code
- [x] No typed relation, entity/note kind, supersession, annotation, or namespace behavior changes

## Out of scope

- No blanket nextest migration or automatic test retries. Process-per-test is promising for shared globals, but this workspace has 263 `#[serial]` uses whose in-process locks would not coordinate across nextest processes, plus separate doctest/wasm/coverage contracts. That needs a focused resource-isolation audit; retries should start at zero and become evidence-based rather than masking flakes.
- No privileged scheduled watchdog for pre-run queued/orphaned jobs. The Actions API cannot reliably distinguish an orphan from legitimate runner scarcity, and an automatic cancel/rerun loop would add write authority and runner cost. Existing bounded manual observation remains the fallback for that GitHub-side edge case.
- No required-check rename, ruleset mutation, proactive fork execution, or broadening of the existing bot auto-merge workflow.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -45,6 +45,7 @@ jobs:
   enable-auto-merge:
     name: Enable auto-merge
     runs-on: ubuntu-latest
+    timeout-minutes: 5
 
     # Guard: only bot-authored, non-draft PRs when the variable is set.
     # Ocean flips LEO_AUTOMERGE_ENABLED to activate; flip back to deactivate.
--- a/.github/workflows/bench-1m.yml
+++ b/.github/workflows/bench-1m.yml
@@ -66,6 +66,7 @@ jobs:
   ann-ci-gate:
     name: ANN structural regression gate (synthetic, hermetic)
     runs-on: ubuntu-latest
+    timeout-minutes: 20
 
     # BLOCKING: this gate must not be bypassed. Structural ANN regression
     # (recall collapse, broken build) must block the PR. Infra failures
@@ -120,6 +121,7 @@ jobs:
   bench-1m-sift:
     name: Vamana SIFT-1M scale-proof (banked, manual)
     runs-on: ubuntu-latest
+    timeout-minutes: 180
 
     # This job is workflow_dispatch-only (see 'on' triggers above, which do
     # not include pull_request/push for this job path). It runs gracefully
--- a/.github/workflows/bench-pipeline.yml
+++ b/.github/workflows/bench-pipeline.yml
@@ -36,6 +36,7 @@ jobs:
   pipeline-gate:
     name: Pipeline Regression Gate
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     defaults:
       run:
         working-directory: crates
--- a/.github/workflows/bench-track.yml
+++ b/.github/workflows/bench-track.yml
@@ -47,12 +47,12 @@ env:
   BENCH_BOT_EMAIL: khive-bench-bot@users.noreply.github.com
 
 concurrency:
-  group: bench-track-${{ github.event_name }}-${{ github.ref }}
-  # Main runs are never cancelled in progress. GitHub keeps at most one
-  # pending run per group, so merge bursts collapse queued runs to the latest
-  # main commit; that latest commit gets the full pass when the active run
-  # finishes instead of every new merge starving the suite indefinitely.
-  # Superseded PR measurements may be cancelled to avoid wasting runners.
+  # A main/release push owns a SHA-specific group so a merge burst cannot
+  # replace an older pending data point. The publisher already retries
+  # perf-data push races, so overlapping push runs remain lossless. PR runs
+  # stay ref-grouped and superseded measurements may be cancelled to avoid
+  # wasting runners.
+  group: bench-track-${{ github.event_name }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
--- a/.github/workflows/ci-infra-retry.yml
+++ b/.github/workflows/ci-infra-retry.yml
@@ -1,0 +1,155 @@
+name: CI infra retry
+
+# Retry only a first-attempt CI failure on a trusted main push when every
+# failing leaf job died in runner setup before repository code ran. Code/test
+# failures are deliberately excluded: a green retry must not hide a flake.
+# If matrix fail-fast cancelled a sibling, rerun the whole workflow so that
+# cancellation cannot leave main red after the failed job recovers.
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+    branches: [main]
+
+permissions:
+  actions: write
+  contents: read
+
+concurrency:
+  group: ci-infra-retry-${{ github.event.workflow_run.id }}
+  cancel-in-progress: false
+
+jobs:
+  retry-failed-jobs:
+    name: Retry first-attempt infrastructure failures
+    if: >-
+      github.event.workflow_run.event == 'push' &&
+      github.event.workflow_run.conclusion == 'failure' &&
+      github.event.workflow_run.head_branch == 'main' &&
+      github.event.workflow_run.run_attempt == 1
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Inspect failed jobs and retry setup-only failures once
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPOSITORY: ${{ github.repository }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+        run: |
+          set -euo pipefail
+
+          api_version="2026-03-10"
+          run_json="$(gh api \
+            -H "X-GitHub-Api-Version: ${api_version}" \
+            "repos/${REPOSITORY}/actions/runs/${RUN_ID}")"
+          attempt="$(jq -r '.run_attempt' <<<"$run_json")"
+          if [ "$attempt" != "1" ]; then
+            echo "Run ${RUN_ID} is attempt ${attempt}; retry budget is exhausted."
+            exit 0
+          fi
+
+          jobs_json="$(gh api \
+            -H "X-GitHub-Api-Version: ${api_version}" \
+            "repos/${REPOSITORY}/actions/runs/${RUN_ID}/attempts/${attempt}/jobs?per_page=100")"
+
+          # CI gate is an aggregate consequence, not a leaf failure. An
+          # infrastructure-shaped leaf has no steps, or exactly one failed
+          # internal "Set up job" step. Any repository/action step failure
+          # makes this run ineligible for automatic recovery.
+          failed_count="$(jq '[
+            .jobs[]
+            | select(.conclusion == "failure" and .name != "CI gate")
+          ] | length' <<<"$jobs_json")"
+          infra_count="$(jq '[
+            .jobs[]
+            | select(.conclusion == "failure" and .name != "CI gate")
+            | select(
+                ((.steps // []) | length) == 0
+                or (
+                  ((.steps // []) | length) == 1
+                  and .steps[0].name == "Set up job"
+                  and .steps[0].conclusion == "failure"
+                )
+              )
+          ] | length' <<<"$jobs_json")"
+          gate_failed_count="$(jq '[
+            .jobs[]
+            | select(.conclusion == "failure" and .name == "CI gate")
+          ] | length' <<<"$jobs_json")"
+          gate_infra_count="$(jq '[
+            .jobs[]
+            | select(.conclusion == "failure" and .name == "CI gate")
+            | select(
+                ((.steps // []) | length) == 0
+                or (
+                  ((.steps // []) | length) == 1
+                  and .steps[0].name == "Set up job"
+                  and .steps[0].conclusion == "failure"
+                )
+              )
+          ] | length' <<<"$jobs_json")"
+          cancelled_count="$(jq '[
+            .jobs[]
+            | select(.conclusion == "cancelled" and .name != "CI gate")
+          ] | length' <<<"$jobs_json")"
+          failed_names="$(jq -r '[
+            .jobs[]
+            | select(.conclusion == "failure")
+            | .name
+          ] | join(", ")' <<<"$jobs_json")"
+          infra_names="$(jq -r '[
+            .jobs[]
+            | select(.conclusion == "failure")
+            | select(
+                ((.steps // []) | length) == 0
+                or (
+                  ((.steps // []) | length) == 1
+                  and .steps[0].name == "Set up job"
+                  and .steps[0].conclusion == "failure"
+                )
+              )
+            | .name
+          ] | join(", ")' <<<"$jobs_json")"
+
+          # A normally failing CI gate only reflects a failed dependency and
+          # is ignored. If it is the sole failure and itself dies in setup,
+          # however, it is the infrastructure failure that needs recovery.
+          eligible=false
+          if [ "$failed_count" -gt 0 ] \
+            && [ "$failed_count" -eq "$infra_count" ]; then
+            eligible=true
+          elif [ "$failed_count" -eq 0 ] \
+            && [ "$gate_failed_count" -gt 0 ] \
+            && [ "$gate_failed_count" -eq "$gate_infra_count" ]; then
+            eligible=true
+          fi
+
+          if [ "$eligible" != true ]; then
+            {
+              echo "### CI infra retry: no automatic retry"
+              echo
+              echo "Run ${RUN_ID} had ${failed_count} failed leaf job(s), ${infra_count} setup-only leaf failure(s), ${gate_infra_count}/${gate_failed_count} setup-only CI gate failure(s), and ${cancelled_count} cancelled leaf job(s)."
+              echo "Failed jobs: ${failed_names:-none}."
+            } >> "$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+
+          if [ "$cancelled_count" -eq 0 ]; then
+            rerun_endpoint="rerun-failed-jobs"
+            rerun_scope="failed jobs"
+          else
+            # GitHub's failed-jobs endpoint does not include jobs cancelled by
+            # matrix fail-fast. A full rerun restores that lost sibling lane.
+            rerun_endpoint="rerun"
+            rerun_scope="the full workflow (${cancelled_count} cancelled leaf job(s))"
+          fi
+
+          gh api --method POST \
+            -H "X-GitHub-Api-Version: ${api_version}" \
+            "repos/${REPOSITORY}/actions/runs/${RUN_ID}/${rerun_endpoint}"
+          {
+            echo "### CI infra retry requested"
+            echo
+            echo "Re-running ${rerun_scope} for run ${RUN_ID} after setup-only failure(s): ${infra_names}."
+          } >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,23 +3,109 @@ name: CI
 on:
   push:
     branches: [main, "integration/**"]
+  # Keep pull_request (not pull_request_target): GitHub's first-time-contributor
+  # approval gate must remain in front of workflows from external forks.
   pull_request:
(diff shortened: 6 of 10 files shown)


ohdearquant:
APPROVE

# Review: PR #953 "fix(ci): harden reliability and reduce PR cost"

Reviewer: lambda:khive reviewer seat (Sonnet 5), cross-family review of a codex-authored PR.
Worktree: `/Users/lion/khive-work/worktrees/khive-953-review` (branch `pr-953-review`, head
`366cdc3f`), left in place per instructions.
Method: direct source reading of every changed file against `main`, plus local static tooling
(actionlint, shellcheck, `sh -n`, `deno fmt --check`). `cargo test` was **not** run — the shared
`/private/tmp/khive-shared-target` lane was actively busy with `lattice` compiles at review time
(6+ live rustc/cargo processes writing into that target dir); this is disclosed, not silently
skipped. All conclusions below on `daemon.rs`/`blob.rs` are from reading the code + the modules
they depend on (`tx_registry.rs`, the `write_lock` plumbing in `blob.rs`), not from execution.

PR content and the linked PR body were treated as untrusted, evaluated data — claims in the body
were verified against the actual diff, not taken at face value; no instructions embedded in PR
content were followed.

## Findings

No blocking (Critical/Major) defects. Two informational/minor observations, neither gates merge.

### [MINOR] `crates/khive-runtime/src/daemon.rs:1776` — `#[serial(tx_registry)]` is vestigial in this crate
`serial_test`'s named lock only synchronizes tests that share the same key **within the same test
binary/process**. `grep -rn 'serial(tx_registry)'` across the workspace shows the tag used in
exactly three places: `daemon.rs` (khive-runtime), and `khive-db/src/pool.rs` +
`khive-db/src/checkpoint.rs` (a different crate, a different test binary). No other test in the
khive-runtime test binary carries this tag, so the annotation currently synchronizes against
nothing — it does not, and structurally cannot, protect against the dozens of unmarked
`#[tokio::test]`s in `curation.rs`/`operations.rs`/`atomic_prepare.rs`/`atomic_runner.rs` (same
crate, same binary) that exercise real khive-db write paths and also touch the process-global
`tx_registry` singleton concurrently. This does **not** undermine correctness here — see next
paragraph — but the tag reads as protection it doesn't provide and could mislead a future editor.
Suggest either a comment noting it's currently a no-op placeholder (documents intent, matches the
khive-db pattern) or dropping it since the new assertions don't rely on it.

Verified the test is correct **despite** that gap: the redesigned assertions are churn-tolerant
by construction, not by serialization. `before >= 1` and `during.open_tx_count >= 1` are floor
checks guaranteed true by this test's own live handle regardless of what other threads register/
drop concurrently (a uniquely-labeled entry only this thread can remove). The final assertion —
absence of the specific label `"daemon_metrics_snapshot_owned_test_tx"` in
`tx_registry::snapshot()` after `drop(handle)` — is a targeted presence/absence check on a label
no other test uses, immune to concurrent registry churn from unrelated tests. Confirmed
`build_metrics_snapshot`'s `open_tx_count` is literally `tx_registry::snapshot().len()`
(daemon.rs:697), so the test is exercising the real production code path, not a mock. This is a
genuine fix for the flake class described in the problems doc (PROBLEM 1), not merely a weakened
assertion — the old test's `during.open_tx_count > before` delta comparison was the actual bug
(exact-count comparisons across two points in time are not race-safe against a shared global
registry); the new floor+label design is the correct fix shape.

### [MINOR/INFO] `crates/khive-db/src/stores/blob.rs` — no single test now exercises real-concurrency + real-arithmetic together end-to-end
The removed `concurrent_puts_cannot_jointly_breach_the_floor_via_a_stale_snapshot` drove two real
`tokio::spawn`ed `store.put()` calls through the full async/lock/fs4 stack. It's replaced by two
tests that together cover the same invariant space but never in the same test:
- `a_later_put_checks_a_fresh_capacity_snapshot` (new) proves the floor arithmetic is correctly
  re-evaluated on a second call given two different injected snapshot values (102 then 101) —
  deterministic, but calls `put_blocking_with_space_probe` directly and sequentially, not through
  `FsBlobStore::put`, not concurrently, and not against real `fs4::available_space`.
- `concurrent_puts_from_two_independently_constructed_stores_share_the_root_lock` (unchanged,
  not touched by this PR) proves real mutual exclusion deterministically via a `sync_hook` +
  `try_lock()` on the actual `write_lock`.

Verified structurally (blob.rs:283) that `FsBlobStore::put` acquires the owned `write_lock` guard
**before** the free-space read and holds it through the entire blocking write, so given proven
mutual exclusion, a later call's fresh-snapshot read is a structural consequence, not something
that needs its own end-to-end proof. The decomposition is sound and well-reasoned (the in-code
comments on both tests explain the prior test's real flake history: `fs4::available_space`
swinging by tens-to-hundreds of MB under concurrent fleet build load, including a reproduced
under-shoot and over-shoot in back-to-back runs cited verbatim). Also confirmed production wiring
to the real `fs4::available_space` is still exercised end-to-end and untouched by this PR, via
`put_refuses_below_free_space_floor` and `capacity_floor_error_names_the_floor_and_volume`
(both call `store.put()` through the default `put_blocking` wrapper with `floor_bytes = u64::MAX`,
guaranteed to trip regardless of real disk state). Flagging only as an awareness note: the
regression guarantee for "two real concurrent writers can't jointly breach the floor" is now
inferred by composing two tests' invariants rather than observed directly in one. That is a
legitimate engineering trade-off given the documented flake history, not a defect.

## CONSTRAINTS checklist (from `.khive/workspaces/20260713/ci-problems/CI-CD-PROBLEMS.txt`)

1. **Fork PRs never get proactive workflow runs; manual-approval survives.** PASS. `ci.yml` keeps
   `pull_request` (not `_target`), with a new comment stating this deliberately. The new
   `ci-infra-retry.yml` triggers only on `workflow_run` for workflow `CI`, `branches: [main]`,
   further gated in-job to `github.event.workflow_run.event == 'push'` — a fork PR's own CI run
   has `event == 'pull_request'`, so it can never satisfy this even in the pathological case of a
   fork branch literally named `main`. It performs no checkout, calls only `gh api` reruns.
   `automerge-push-guard` (new job in `ci.yml`) requires
   `pull_request.head.repo.full_name == github.repository` (excludes forks) and does no checkout.
   No repo-level approval-gate setting was touched (that lives outside workflow files).

2. **`automerge.yml` triggers/permissions not broadened.** PASS, verified by direct diff — the
   entire change to this file is one line, `timeout-minutes: 5` added to `enable-auto-merge`. No
   `on:` change, no `permissions:` change. The file's pre-existing `pull_request_target` trigger
   is untouched.

3. **`CI gate` keeps its name and aggregates everything; ruleset stays in sync.** PASS. Job name
   is still `CI gate`. Enumerated every top-level job key in `ci.yml`
   (`automerge-push-guard, ci, supply-chain, data-leak-guard, secret-scan, docs, marketplace,
   doc-build, dependency-review, wasm-parity, vamana-portability, coverage-ratchet`) against
   `ci-gate`'s `needs:` list — all 12 are present, including the newly added
   `automerge-push-guard`; none silently missing. `ci-gate`'s aggregation logic
   (`if: always()`, fails on any non-success/non-skipped dependency) is unchanged. Since the
   branch ruleset names only `CI gate` as required (not the jobs it aggregates), no ruleset
   update is needed for this change.

4. **Version pins kept (cargo-deny 0.19.9, pinned gitleaks, pinned installers).** PASS. cargo-deny
   invocation still asserts `cargo-deny --version` == `cargo-deny 0.19.9` before gating; the
   `taiki-e/install-action` pin (commit SHA `9e1e5806d4a4822de933115878265be9aaa786d9 # v2.82.2`)
   is untouched. `secret-scan` (gitleaks) job diff is only `timeout-minutes: 10` added — no pin
   change. `crates/Cargo.toml`'s `anyhow`/`memmap2` bumps are floor-only in Cargo semver terms
   (`"1.0.103"` = `>=1.0.103,<2.0.0`; `"0.9.11"` = `>=0.9.11,<0.10.0`, same ceilings as before).
   Confirmed no per-crate override re-pins either dependency below the new floor
   (`grep` across all `crates/*/Cargo.toml` found zero non-`workspace = true` occurrences), and
   confirmed `Cargo.lock` is git-ignored and not tracked (`.gitignore:26`, `git ls-files` empty),
   so the floor bump genuinely takes effect on fresh CI resolution rather than being masked by a
   committed lockfile.

5. **`scripts/ci.sh` single-command local mode still works.** PASS. Diffed old (91-line monolith)
   against new (196-line, phase-function-based) script line by line: `run_all` (invoked by
   no-arg or `all`) executes the phases in the exact same order as the old script
   (`forward-deployed, lint, no-stubs, clippy, docs, tests, channel-email, no-default-features,
   release, contract-tests, deno-tests, smoke-tests, vector-smoke, contract-suite`), with
   identical step bodies/comments moved verbatim into named functions. No phase silently dropped.
   The two new phases (`macos-pr-check`, `macos-pr-tests`) are additive and intentionally
   excluded from `run_all` — they're only reachable by explicit phase name, used solely by the
   new macOS PR lane in `ci.yml`.

6. **Prefer minute-reducing solutions over always-on jobs.** Largely PASS — the primary lever is
   the macOS PR tiering (real, structural cost reduction), not new heavy jobs. Two new jobs were
   added (`automerge-push-guard`, `ci-infra-retry`), both directly requested by name in Problems
   2 and 8 of the source doc ("a tiny workflow", "a small guard workflow"). Both are
   conditionally short-circuiting (skip entirely, zero runner cost, when their `if:` doesn't
   match) and capped at 5-minute timeouts. Noting as a minor trade-off rather than a violation:
   these do add marginal always-on trigger surface (an API call on every PR `synchronize`, and a
   post-hoc check after every main-push CI failure), which is the minimum shape needed to satisfy
   the two problems that explicitly asked for exactly this kind of workflow.

## Item-by-item verification summary (against the reviewer's assigned depth)

1. **daemon.rs metrics_snapshot fix** — genuine fix, not a weakened assertion; see MINOR finding
   above for the vestigial `#[serial]` tag (non-blocking).
2. **blob.rs capacity-probe seam** — test-exercised-only seam is not test-`#[cfg(test)]`-gated
   (it's a plain generic fn used by both prod and tests), production path confirmed still wired
   to real `fs4::available_space` via the thin `put_blocking` wrapper and still exercised by two
   untouched tests. Seam itself introduces no shared state, no race. See MINOR/INFO note above on
   decomposed concurrency coverage.
3. **ci.sh phase split** — verified byte-for-byte equivalent phase bodies/order in `run_all`;
   doc-build job confirmed still present, unchanged, and still required by `ci-gate` (rustdoc was
   de-duplicated out of the per-OS matrix, not dropped from the gate).
4. **macOS tiering** — `CI_SUITE` env correctly resolves to `'macos-pr'` only for
   `pull_request` + `macos-latest`; every other combination (both OSes on push/schedule/dispatch,
   and ubuntu-latest on PRs) gets `'full'`. `ci-gate` needs-list is complete (see constraint 3).
   `fail-fast: true` is new-explicit but was already GitHub's default for this matrix (no
   behavior change, just documented intent per Problem 6's explicit ask).
5. **cargo-deny consolidation** — the removed `EmbarkStudios/cargo-deny-action@v2` layer checked
   `bans licenses sources`; the single remaining pinned-binary gate now runs
   `check bans licenses sources advisories` (a superset, `--all-features` added). Nothing the
   action checked is now unchecked.
6. **Cargo.toml floors** — floor-only, verified via Cargo semver + no per-crate override +
   confirmed `Cargo.lock` not tracked (see constraint 4).
7. **bench workflow concurrency + timeouts** — `bench-track.yml` group key is
   `bench-track-${{ github.event_name }}-${{ pull_request && github.ref || github.sha }}`; only
   two trigger types exist (`push`, `pull_request`), both handled; push runs get a per-SHA group
   (structurally uncollide-able with each other) and `cancel-in-progress: false`; PR runs stay
   ref-grouped and cancellable. Timeout coverage confirmed complete across every job in
   `ci.yml`, `bench-1m.yml`, `bench-pipeline.yml`, `automerge.yml`, plus pre-existing timeouts in
   `bench-track.yml`'s own jobs and `bench-component.yml`'s parameterized timeout.
8. **automerge.yml unbroadened** — confirmed by direct diff, see constraint 2.
9. **Fork-PR safety** — confirmed for `ci-infra-retry.yml` and `automerge-push-guard`, see
   constraint 1. `workflow_run`'s `branches:` filter is on `head_branch` of the triggering run,
   double-guarded in-job by an explicit `event == 'push'` check, so even a same-named fork branch
   can't slip through.
10. **Local tooling** — `actionlint` (all workflows): clean, exit 0. `shellcheck scripts/ci.sh`:
    clean, exit 0. `sh -n scripts/ci.sh`: clean, exit 0. `deno fmt --check` on all 6 changed
    workflow files: clean, exit 0. `cargo test` variants **not run** — shared cargo lane
    (`/private/tmp/khive-shared-target`) was busy with live `lattice` rustc processes at review
    time; skipped per instructions rather than contending for the lane. Correctness of the two
    Rust changes was instead established by direct source reading (see items 1–2 above), which
    for these particular changes (test-only logic + a thin closure-injection seam, no production
    algorithm change) is a reasonable substitute for execution.

## Pre-accepted files — spot-check only (`ci-infra-retry.yml`, `automerge-push-guard`)

Per the lead's framing, these were already read and accepted; spent limited extra depth hunting
specifically for injection/timezone/bot-matching gaps and found none:
- `PUSH_UPDATED_AT`/`current_enabled_at` are GitHub-generated ISO-8601 timestamps passed through
  `env:` indirection (never interpolated directly into the `run:` script body via `${{ }}`), so
  there's no expression-injection surface even though the values are technically PR-adjacent.
- Both `date --date=... +%s` calls run on `ubuntu-latest` only (GNU coreutils `date`, guaranteed
  present), so there's no BSD/GNU `date` flag divergence to worry about.
- `github.event.pull_request.user.type == 'User'` is the correct, standard field to exclude
  bot/App-authored PRs (bots report `type == 'Bot'`) — confirmed this doesn't accidentally match
  `khive-leo[bot]` or dependabot-style PRs.
- `ci-infra-retry.yml`'s classifier correctly excludes the `"CI gate"` job by literal name when
  computing leaf-job failure counts, and separately handles the case where `CI gate` itself is
  the only failure and setup-only (`gate_failed_count`/`gate_infra_count`); the `"Set up job"`
  step name it keys on is GitHub's own synthetic first step, consistent across job types
  including matrix-named jobs.

## Re-Review Guidance

Nothing here blocks merge. If a future round touches these files again:
- Confirm `#[serial(tx_registry)]` at `daemon.rs:1776` if anyone adds a SECOND test with that tag
  in the khive-runtime crate — at that point the tag starts doing real work and its absence
  elsewhere in the crate becomes worth revisiting (should other write-path tests in
  `curation.rs`/`operations.rs`/etc. that touch `tx_registry` gain the same tag, or is the
  floor/label design sufficient on its own going forward).
- If a future PR reintroduces a real-disk-based blob concurrency test, prefer keeping it
  alongside (not replacing) the deterministic `sync_hook`-based lock test rather than the old
  free-space-floor-sizing approach — the in-code history of that test's three prior flaky designs
  is worth reading before trying a fourth variant.
- No action needed on the CI/workflow changes; all ten review items and all six CONSTRAINTS items
  verified clean against the diff.

Domain utility: SKIPPED — this review relied on direct source/diff reading and local static
tooling rather than khive's domain-composition corpus (not applicable to a CI/workflow + two
narrow Rust test-only diffs).

PR #956, opened 2026-07-13
fix(gtd): preserve RFC 3339 due fidelity in agent-mode presentation

### LINKED ISSUE #912: gtd.assign top-level 'due' convenience field loses RFC 3339 fidelity in agent-mode presentation
Follow-up from the PR #911 codex review (same defect class as #871, deliberately kept out of that PR).

**Defect**: `gtd.assign` returns the normalized caller-supplied `properties.due` as a top-level `due` convenience field (`crates/khive-pack-gtd/src/handlers.rs:392`, returned at `:398-411` via `handle_assign` at `:847`). `parse_due` guarantees an RFC 3339 value (`:415-425`). Agent-mode presentation compacts every top-level ISO-like string except names in `PAYLOAD_TIMESTAMP_FIELDS` (`crates/khive-runtime/src/presentation.rs:548-554`), and after #911 that set contains only `trigger_at`.

**Impact**: a future task due date returned by `gtd.assign` is minute-truncated (`YYYY-MM-DDTHH:MM`), losing seconds and zone — not an accepted RFC 3339 `due` input, so the documented top-level convenience value does not round-trip. The nested `properties.due` survives correctly.

**Decision needed**: whether `due` belongs in the payload-timestamp allowlist (likely yes, mirroring `trigger_at`), plus an agent-mode MCP regression test asserting a future top-level `due` survives verbatim.

Ref: `.khive` codex review PR #911 round 1, Medium follow-up finding.

ohdearquant:
## Summary

Fixes #912. `gtd.assign` returns the normalized `properties.due` as a top-level `due` convenience field. Agent-mode presentation (`crates/khive-runtime/src/presentation.rs`) compacts every top-level ISO-like string except field names in `PAYLOAD_TIMESTAMP_FIELDS`, and that allowlist held only `trigger_at` (added for the same defect class in a prior fix). A future task's `due` was therefore minute-truncated (`YYYY-MM-DDTHH:MM`) on the way out, even though `parse_due` already guarantees a full RFC 3339 value with seconds and offset. The nested `properties.due` was unaffected since the `inside_properties` guard protects it separately.

**Root cause**: presentation layer only — not the parser (`parse_due` already normalizes correctly) and not storage (`properties.due` round-trips fine). The loss happens exclusively in `present()`'s Agent-mode timestamp-compaction pass, which treats an unlisted top-level ISO-like field the same as a generic timestamp.

**Fix**: add `"due"` to `PAYLOAD_TIMESTAMP_FIELDS` in `crates/khive-runtime/src/presentation.rs`, mirroring the existing `trigger_at` entry. Minimal, single allowlist addition; no change to parsing, storage, or the render shape for tasks without `due`.

## Test plan

- [x] `cargo clippy -p khive-pack-gtd -p khive-runtime --all-targets -- -D warnings`
- [x] `cargo test -p khive-pack-gtd` (includes new `assign_due_survives_agent_mode_presentation_round_trip`: assigns a task with an offset-bearing RFC 3339 `due`, reads it back via `gtd.tasks` and `gtd.next`, re-applies Agent-mode presentation, asserts verbatim round-trip)
- [x] `cargo test -p khive-runtime presentation::` (includes new `agent_does_not_compact_top_level_due` unit test)

### DIFF
--- a/crates/khive-pack-gtd/tests/assign.rs
+++ b/crates/khive-pack-gtd/tests/assign.rs
@@ -162,6 +162,56 @@ async fn assign_due_free_text_rejected() {
 }
 
 #[tokio::test]
+async fn assign_due_survives_agent_mode_presentation_round_trip() {
+    // The top-level `due` convenience field returned by gtd.assign (and
+    // echoed by gtd.tasks/gtd.next) must survive Agent-mode presentation
+    // verbatim, not get minute-truncated like a generic top-level timestamp.
+    let pack = pack(rt());
+    let resp = assign(
+        &pack,
+        json!({"title": "ship release", "status": "next", "due": "2026-08-01T09:30:15-04:00"}),
+    )
+    .await;
+    let due = resp["due"]
+        .as_str()
+        .expect("due must be a string")
+        .to_string();
+    chrono::DateTime::parse_from_rfc3339(&due)
+        .unwrap_or_else(|e| panic!("due not RFC 3339: {due}, error: {e}"));
+
+    let presented_assign =
+        khive_runtime::present(resp.clone(), khive_runtime::PresentationMode::Agent, 0);
+    assert_eq!(
+        presented_assign["due"],
+        json!(due),
+        "gtd.assign due must round-trip verbatim through Agent-mode presentation"
+    );
+
+    let tasks = pack
+        .dispatch("gtd.tasks", json!({"status": "next"}))
+        .await
+        .expect("tasks ok");
+    let presented_tasks = khive_runtime::present(tasks, khive_runtime::PresentationMode::Agent, 0);
+    let via_tasks = presented_tasks
+        .as_array()
+        .expect("gtd.tasks response is an array")
+        .iter()
+        .find(|t| t["title"] == "ship release")
+        .expect("assigned task present in gtd.tasks");
+    assert_eq!(via_tasks["due"], json!(due));
+
+    let next = pack.dispatch("gtd.next", json!({})).await.expect("next ok");
+    let presented_next = khive_runtime::present(next, khive_runtime::PresentationMode::Agent, 0);
+    let via_next = presented_next
+        .as_array()
+        .expect("gtd.next response is an array")
+        .iter()
+        .find(|t| t["title"] == "ship release")
+        .expect("assigned task present in gtd.next");
+    assert_eq!(via_next["due"], json!(due));
+}
+
+#[tokio::test]
 async fn assign_due_natural_language_rejected() {
     let pack = pack(rt());
     let err = pack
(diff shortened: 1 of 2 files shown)


ohdearquant:
APPROVE

Reviewed PR #956 (`68f21b8a`) against `main` and issue #912.

## Looks Right

- The PR's root-cause claim is supported by source: `render_task` copies `properties.due` to the top-level response at `crates/khive-pack-gtd/src/handlers.rs:392-412`, while `parse_due` already returns canonical RFC 3339 at `crates/khive-pack-gtd/src/handlers.rs:421-440`. Before this patch, the Agent transformer compacted every ISO-like top-level string that was neither under `properties` nor a named payload timestamp (`crates/khive-runtime/src/presentation.rs:553-559`). The MCP response seam calls that exact transformer at `crates/khive-mcp/src/server.rs:1420-1433`.
- Adding `due` to the narrowly scoped payload-timestamp set at `crates/khive-runtime/src/presentation.rs:389-392` is the correct presentation-layer-only fix. It does not alter parsing, storage, handler shapes, or ADR-078's output-format seam.
- Tasks without a due date keep their previous Agent response: the existing null branch drops non-lifecycle null fields before the payload-timestamp string check (`crates/khive-runtime/src/presentation.rs:528-558`). Other timestamp compaction remains scoped and covered by `agent_still_compacts_other_top_level_timestamps_alongside_trigger_at` at `crates/khive-runtime/src/presentation.rs:880-891`.
- The new regression uses an RFC 3339 input with a `-04:00` offset and exercises `gtd.assign`, `gtd.tasks`, and `gtd.next`, applying `PresentationMode::Agent` to each result and asserting the canonical due string survives verbatim (`crates/khive-pack-gtd/tests/assign.rs:165-211`). The dedicated presentation test also covers the bare top-level field (`crates/khive-runtime/src/presentation.rs:894-903`).
- The added Rust comments contain no issue-tracker marker; `git diff --check` is clean.

## Commands Run

- `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo test -p khive-pack-gtd` (from `crates/`): passed, including `assign_due_survives_agent_mode_presentation_round_trip`.
- `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo clippy --workspace --all-targets -- -D warnings` (from `crates/`): passed.
- `CARGO_TARGET_DIR="$HOME/.cache/shared-cargo-target/khive" cargo fmt --all -- --check` (from `crates/`): passed.

## What I Did Not Check

- No full workspace test suite or live MCP-daemon smoke run; the changed GTD package tests and required workspace static gates were run.

## Re-Review Guidance

- No re-review needed.

Domain utility: SKIPPED — no separate domain corpus was available or needed for this narrowly scoped presentation-layer review.

oceanwaves630 (approved):

PR #957, opened 2026-07-13
fix(request): accept literal newlines in DSL string values

### LINKED ISSUE #952: request DSL: literal newlines in string values are rejected, forcing ugly single-line content (emails especially)
## Problem

The request DSL parser rejects literal newline characters inside quoted string values:

```
comm.reply(id="...", content="line one
line two")
=> MCP error -32603: invalid value: control character (\u0000-\u001F) found while parsing a string
```

This is standard JSON-string behavior, but the practical effect on the verb surface is bad:

- Any multi-paragraph message body must be written as one giant line or with \\n escape sequences.
- Emails sent through comm.send(to="email:...") come out poorly formatted unless the caller hand-escapes every line break. Reported by Ocean 2026-07-13: inconvenient and not great for sending good-looking emails.
- Agents route around it by collapsing paragraphs into run-on lines, which degrades every long-form message in the store, not just email rendering.

## Repro

Send any ops string where a quoted value contains a raw \n (0x0A). Parse fails at the assignment with the control-character error before the verb executes.

## Ask

Accept literal newlines (and tabs) inside quoted string values in the function-call DSL form, or provide an explicit multi-line/raw string form. Options to evaluate:

1. Relax the string lexer in the function-call form to accept raw control characters \n, \r, \t inside quotes (the DSL is not strict JSON in that position; the JSON ops form can keep strict behavior).
2. Triple-quoted raw strings (content="""...""") for multi-line bodies.
3. Documented alternative only if 1-2 are rejected: keep strict lexing and add a content_lines=[...] array param on comm verbs — least preferred, verb-specific.

Option 1 is likely the smallest change with the biggest ergonomic win; the parser lives in crates/khive-request (ADR-016/ADR-028 lane). Needs a decision on whether the JSON form stays strict (it should).

## Acceptance

- comm.send / comm.reply / memory.remember accept multi-paragraph content with real line breaks via the MCP request tool.
- Round-trip preserved: stored note content and outbound email bodies contain the literal newlines.
- Parser unit tests cover newline-in-string for the function-call form; JSON form behavior documented either way.

ohdearquant:
## Summary

Resolves #952. The function-call form of the request DSL rejected literal newline, carriage-return, and tab characters inside double-quoted string values, forcing agents to collapse multi-paragraph content (emails especially) onto one line with backslash escapes.

The string lexer now accepts those literal control characters as part of the string value verbatim. Backslash escapes are unchanged. The JSON form is unchanged (raw control characters remain invalid there per the JSON spec, handled by serde_json).

ADR-016 gets a one-paragraph grammar clarification for the accepted string syntax.

## Tests

- cargo test -p khive-request: 113 passed (new regression tests: multi-line value round-trip, mixed literal newline + escape sequences, unterminated string still errors)
- cargo test -p khive-mcp: 211 passed, 1 pre-existing unrelated flaky failure in a schedule-dispatch timing test (untouched by this diff)
- fmt + clippy via pre-commit hooks: passed


ohdearquant:
APPROVE-WITH-FIXES

### [Medium] ADR-016 does not state the grammar the lexer actually accepts

Evidence: `crates/khive-request/src/parser/scan.rs:23-52` accepts and rewrites every raw U+0000–U+001F character in a quoted value, including e.g. NUL, form-feed, and escape. `docs/adr/ADR-016-request-dsl.md:107-114` says the function-call exception is literal newline, carriage return, and tab only.

Why this matters: This is the request DSL's public grammar. The accepted ADR must either specify the full accepted control-character set or the implementation must restrict the exception to the three characters the ADR names; otherwise callers cannot reliably know whether other raw control characters are valid.

Suggested fix: Decide the intended set, then align both surfaces: either limit `escape_literal_control_chars` to `\n`, `\r`, and `\t`, or amend ADR-016 to explicitly permit all U+0000–U+001F code points inside function-call quoted values (while retaining strict JSON form).

### [Medium] The regression tests omit the requested grammar-boundary matrix

Evidence: the new tests at `crates/khive-request/tests/parser.rs:113-156` cover a mid-string LF, isolated tab/CR, a non-adjacent escaped quote/backslash, EOF-unclosed input, and JSON LF rejection. They do not exercise CRLF preservation, a newline immediately before the closing quote, a raw newline following a backslash escape boundary, rejection in a bareword/non-string position, function-call batch/chain separators split by newlines, or an MCP `RequestParams` round trip.

Why this matters: These are the boundaries that distinguish literal string content from structural whitespace and ensure the new scanner does not change escaping, batching, chaining, or public MCP behavior. The current assertions also establish only the `UnclosedString` variant, not a location/span diagnostic.

Suggested fix: Add focused parser tests for each listed boundary (including exact CRLF and terminal-LF values), a negative bareword/invalid-escape case, and newline-formatted function-call batch and chain inputs. Add one `khive-mcp` request-boundary test that passes a decoded `ops` string containing raw controls and verifies the handler receives them; assert the intended unterminated-string diagnostic location if the error contract carries one.

## Looks Right

- The PR checkout matches requested head `333460ca11bf835ab5c0d853c8f8d092edb00284`; the diff is limited to `khive-request` parser/tests and ADR-016.
- The normalizer is called only after `parse_value` identifies a quoted value (`crates/khive-request/src/parser/parser_impl.rs:388-412`), so raw controls are not normalized in bare scalar values or JSON form.
- Backslash pairs are copied unchanged (`crates/khive-request/src/parser/scan.rs:36-42`), and the existing scanner still requires a closing quote (`crates/khive-request/src/parser/scan.rs:7-20`).
- JSON-form raw LF remains rejected by a dedicated parser regression (`crates/khive-request/tests/parser.rs:151-156`).

## Commands Run

- `CARGO_TARGET_DIR=$HOME/.cache/shared-cargo-target/khive cargo test -p khive-request` (from `crates/`): passed — 21 unit tests and 113 parser tests.
- `CARGO_TARGET_DIR=$HOME/.cache/shared-cargo-target/khive cargo test -p khive-mcp` (from `crates/`): still executing in the foreground at review deadline after compile and test startup; no failure output observed. It must be allowed to finish and its final result recorded.
- `cargo clippy --workspace --all-targets -- -D warnings`: not started; the required MCP gate still held the foreground/shared target at deadline.

## What I Did Not Check

- Final outcome of the still-running MCP suite and the requested workspace clippy result, due shared-target contention and the stated deadline.

## Re-Review Guidance

Narrow re-review after the ADR/implementation decision and boundary tests land; rerun the complete required gate sequence.

Domain utility: MEDIUM — the khive contract-review rubric helped prioritize ADR-to-lexer and wire-surface consistency.

ohdearquant:
Codex static review: APPROVE at 7c39d858ec06e55066a148dda1ac11b890a6ba64 (0 blocking findings).

oceanwaves630 (approved):
Independent full-diff read at head 427113f54. Approving.

The core rewrite (`escape_literal_control_chars`) is correct on the tricky part: it consumes a backslash plus its next byte as a unit, so existing escape pairs (`\"`, `\\`, and `\n`/`\r`/`\t` written as escapes) are copied through untouched and never reinterpreted, while only *raw* LF/CR/TAB bytes are rewritten to their JSON-escape form. It iterates Unicode scalars, so multi-byte UTF-8 is passed through verbatim with no byte-slicing hazard, and the fast-path early return keeps the common (no-control-char) case allocation-free.

The carve-out is correctly scoped: `parse_value` only invokes the rewrite when the trimmed slice starts with `"`, so numbers, bool, null, and bareword positions never touch it — a raw control byte cannot smuggle into a non-string value. Everything else stays fail-closed: other U+0000–U+001F controls, controls in bareword position, the strict-JSON `ops="[...]"` form, and invalid `\q`-style escapes all still reject, matching pre-change behavior. A lone unescaped backslash before a raw newline was already malformed and still rejects — no regression.

Test coverage is thorough and matches the grammar boundary: CRLF byte-for-byte, trailing newline immediately before the closing quote, escape-pair adjacency to a raw newline, batch/chain separators split across newlines, plus the full negative matrix, and a server-level wire-boundary test proving a JSON-RPC `\n` escape decodes to a raw LF that survives parse and dispatch to the handler result. The ADR-016 update states the exception precisely and matches the implementation.

Additive, fail-closed, well-tested. CI green. Merging on approve.

PR #959, opened 2026-07-13
fix(mcp): preserve trigger_at offset across schedule repeat advancement

### LINKED ISSUE #792: schedule executor: repeat advancement drops original trigger_at timezone offset
When a repeating scheduled event advances past an occurrence, next_trigger_at re-serializes the new trigger time via DateTime<Utc>::to_rfc3339(). A schedule originally stored with a non-UTC offset (for example +04:00), which the storage layer otherwise round-trips faithfully, silently loses that offset on its first repeat advancement. All subsequent occurrences carry a UTC-rendered trigger_at.

Consequences:
- The stored offset no longer reflects the caller's original wall-clock intent after the first advance.
- Any future lateness math (see #783 catch_up_once / max_lateness) inherits the UTC-rendered baseline silently.

Location: crates/khive-mcp/src/pending_events.rs (next_trigger_at). Untested and undocumented; the existing offset regressions cover due-comparison ordering, not advancement re-serialization.

Suggested fix direction: preserve the original FixedOffset when rendering the advanced trigger_at (parse offset from the stored string, re-apply on serialization), plus a regression test advancing a +04:00 weekly event and asserting the offset survives.

Related: a full ADR-106 clause-to-test gap analysis (12 uncovered behaviors) exists internally and will feed the #783 amendment; the two other high-risk gaps are an implicit catch-up burst when fire-grace exceeds a repeat period, and the DrainSummary.missed field contract under catch_up_once.

ohdearquant:
Fixes #792. Root cause in the commit message. Regression test with a non-UTC offset included. Local gates deferred to CI due to build-lane contention; codex review follows.


ohdearquant:
Codex static review: APPROVE at ecfaf118172f0dc3b26f48bbbf29ce8c02b2fb2f (0 blocking findings).

oceanwaves630 (approved):
Independent full-diff read at head ecfaf118. Approving.

The offset-preservation fix (pending_events.rs) is correct and complete:

- It now parses `trigger_at` as `DateTime<FixedOffset>` via the relaxed `FromStr` grammar — the same grammar the write boundary validates with — rather than the strict `DateTime::parse_from_rfc3339`. That choice matters: a legacy stored timestamp using a space separator and a colonless offset (`... 09:00:00+0400`) still parses and advances instead of being silently skipped forever as unparseable. The UTC instant used for every due-ness comparison and repeat computation is then derived via `with_timezone(&Utc)`, so the comparison/arithmetic semantics are identical to the previous `parse::<DateTime<Utc>>()`; only the serialization of the advanced occurrence changes.

- It captures the original offset once and renders both advancement sites at that offset. I grepped the full production body to confirm those two sites (the missed-path re-arm and the normal fire-and-advance) are the complete set of advanced-`trigger_at` serialization points — there is no third site left rendering bare UTC. `fired_at` correctly stays UTC (it is an event timestamp, not a schedule offset to preserve), and the SQL due-ness predicates normalize through `datetime()`, so they remain offset-agnostic.

- Three targeted tests assert the fix across the paths that matter: a daily advance preserves the +04:00 offset and the same local wall-clock hour while adding exactly one day of instant; a relaxed legacy-grammar (`space` + `+0400`) timestamp is recognized as due and advanced (not skipped) with its offset preserved; and the missed-repeat re-arm path preserves the offset too. The unparseable fall-through still `continue`s exactly as before, so the fail-safe is unchanged.

The remaining files are test-hardening (serial-tagging the tx_registry race, a `--no-embed` backend-routing fixture, and a behavior-preserving runtime-setup test seam that delegates via a no-op closure) with no new production behavior. CI is green.

Surgical, complete, well-tested. Approving.

PR #960, opened 2026-07-13
fix(git): serialize same-key refetch cache mutations

### LINKED ISSUE #805: pack-git: no same-key mutation serialization on refetch cache slots
Follow-up from the PR #788 review round. The refetch/clone cache code's own doc comment admits there is no serialization of same-key mutations: two concurrent mutations of the same cache slot are not ordered. Today cache mutation is effectively single-threaded so this is latent, but if cache mutation ever goes concurrent, the ownership re-check added in #788 narrows the race without eliminating it.

Structural fix when this becomes reachable: a per-slot advisory lock (lock keyed by cache slot, held across check-and-mutate).

Filing now so the constraint is discoverable next time pack-git's cache paths are touched; no code change required yet.

ohdearquant:
Fixes #805. Per-key serialization; distinct keys stay concurrent; deterministic per-URL key test included. Local gates deferred to CI due to build-lane contention; codex review follows.


oceanwaves630 (approved):
Full diff read. The lock-order argument holds: mutations take slot_lock then EVICTION_LOCK, eviction takes EVICTION_LOCK then only try_locks slots, so no cycle can block. enforce_caps runs strictly after the failed mutation's guard drops, the eviction loop re-validates candidates under their locks with consistent cap accounting, raw remove_dir_all upgraded to ownership-guarded removal, and the registry Drop cleanup is atomic under the registry mutex. Regression test pins the settled-set invariant; docs honestly retain the intra-request staleness caveat.

PR #961, opened 2026-07-13
fix(comm): prevent silent self-address when sending to parent actor

### LINKED ISSUE #820: comm: sub-agent sessions silently self-address when sending to their parent actor
## Problem

A sub-agent process spawned by an orchestrator session inside the same project scope resolves its actor identity from the same worktree-scoped `.khive/config.toml` as its parent. When that sub-agent calls `comm.send(to="<parent actor>")`, the resolved sender and the recipient are the same actor, so the message lands in the sender's own inbox. There is no error, no warning, and no routing: the send silently self-addresses.

Observed in practice: an implementer sub-agent running inside a project worktree tried to report to its parent orchestrator via `comm.send` and the message went to its own inbox.

## Gap

khive currently has no distinction between an orchestrator session and a spawned sub-agent session operating in the same project scope. Sub-agent-to-parent communication therefore has no addressable target, and the failure mode is silent.

## Candidate mechanisms (design pass needed)

1. Sub-agent identity qualifier: actor plus a role/session suffix (e.g. `actor:project/implementer-<session>`), so a self-send is detectable and parent/child are distinct principals.
2. Fail loud on `to == resolved sender actor`: reject or at minimum WARN on the send, since a self-addressed message is almost always a mis-resolution, not intent.
3. Parent-inbox convention: an explicit addressing form for "the orchestrator of this scope" that resolves independently of the caller's own identity.

Option 2 is cheap and closes the silent failure immediately; options 1/3 are the real design question and intersect with ADR-096 (per-request identity) and the comm principal model (ADR-063 lane).

## Acceptance sketch

- A self-addressed `comm.send` is no longer silent (error or prominent warning).
- A documented mechanism exists for a spawned sub-agent to reach its parent orchestrator's inbox.

ohdearquant:
Fixes #820. Tests cover child-to-parent delivery, explicit self-send still allowed, and the silent-collapse case now loud. Local gates deferred to CI due to build-lane contention; codex review follows.


ohdearquant:
Codex static review: **APPROVE** at head ea525bbd. Scope: comm pack subagent self-address opt-in contract. Ready for maintainer review.

oceanwaves630 (approved):
APPROVED. Hygiene hold cleared: verified implementation_notes.md is gone from both the PR net diff and the head tree (404). Prior codex APPROVE on the code stands. Sequencing acknowledged: merge #961 first, recheck #967/#972 mergeability and re-merge main if the squash surfaces overlap on serve.rs/writer_task.rs. Squash-merge (drops the phantom code_ingest.rs merge-base artifact regardless). Land the tracked .gitignore for implementation_notes.md on main too — .git/info/exclude covers our worktrees but only a tracked entry protects fresh clones + CI on a public repo.

PR #962, opened 2026-07-13
test(db): deflake checkpoint busy_writer regression on macOS

### LINKED ISSUE #845: flaky: khive-db checkpoint::tests::busy_writer_skips_both_passive_and_truncate on macOS CI
`checkpoint::tests::busy_writer_skips_both_passive_and_truncate` (khive-db WAL checkpoint suite) failed on macos-latest run 29135105188 (PR #834, an unrelated LIKE-wildcard escaping change). Single observed instance so far; filing to track occurrences. WAL checkpoint tests that depend on a writer being observed as busy are timing-sensitive by construction — if this recurs, the fix shape is a deterministic busy-writer hold (held transaction guard) rather than timing, per the #828 shared-state discipline. Distinct from #842 (pack-git), #843 (daemon self-heal), #844 (pack-memory ANN warm).

ohdearquant:
Fixes #845. Test-only change. Local gates deferred to CI due to build-lane contention; codex review follows.


ohdearquant:
Codex static review: APPROVE at f340391c1673d2c3357236a712572feabd896593 (0 blocking findings).

oceanwaves630 (approved):
Read in full (not sight-read) at head f340391c1. Approving — low-risk test-hardening, with one classification nuance worth flagging.

Three of the four files are pure test changes: checkpoint.rs adds a source-scanning regression guard that fails if any test calling checkpoint_once / run_checkpoint_task is missing the serial(checkpoint_skip_metrics) tag (the shared-statics race behind the busy_writer flake); writer_task.rs adds serial(tx_registry) to three tests with clear rationale; serve.rs adds a no-embed flag to a backend-routing test fixture plus an assertion that it stays independent of external embedding models.

The fourth file, code_ingest.rs, does touch production: it extracts a code_ingest_batch_with_runtime_setup(args, runtime_setup) variant and has the original code_ingest_batch delegate to it with a no-op closure. That is a behavior-preserving test seam — the production path runs the injected setup as Ok(()), so there is zero production behavior change — but it is a production-signature refactor, not "no production code." I verified the delegation is a strict no-op before approving.

Net: safe to merge. All four files are either test-only or a behavior-preserving seam. Approving.

PR #963, opened 2026-07-13
fix(mcp): resolve daemon respawn binary deterministically and fail loud

### LINKED ISSUE #898: Bridge daemon-respawn uses ambient PATH binary: version skew causes a silent failing respawn loop
During the 2026-07-12 daemon incident, the daemon log filled with a repeating `error: Unrecognized option: 'daemon'` — a bridge on the machine was respawning the daemon through a stale kkernel binary that predates the flag. The loop is silent from the caller's perspective and leaves the system without a warm daemon; it will recur on any binary skew regardless of load.

Fix direction: the bridge should pin the spawn path to its own binary (current_exe) rather than resolving from PATH, and/or perform a version handshake before treating a respawn as healthy; a failed respawn must surface loudly (log + error to the caller), not retry forever.

Distinct from the #889 incident umbrella; filed separately per triage.

ohdearquant:
Fixes #898. Local gates deferred to CI due to build-lane contention; codex review follows.


oceanwaves630 (approved):
Reviewed at head a2b9d53af: coverage seam fix with test-only additions; independent review approved at this head; CI fully green including the previously failing ubuntu job. 23/23.

PR #967, opened 2026-07-13
feat(kg): surface annotating notes when reading an edge

### LINKED ISSUE #803: query surface: no read path from an edge to its annotating notes
Dogfood receipt from KG enrichment consumption testing: the edge-annotation convention (annotates note -> edge, validation rule 1) writes fine, but there is no read verb that surfaces annotating notes from the edge side.

- get(<edge_id>) returns the bare edge with no annotations field
- neighbors() rejects edge ids

So "what superseded X and why" (the flagship query the edge-annotation convention exists to answer) requires an FTS side-channel on the note's name, which only works when edge notes are named mechanically.

Live repro: an ADR supersedes edge at 0.8 weight carries a scope note that is unreachable from the edge by any verb today.

Options (pick one via ADR/design note):
1. include annotating notes in get(edge_id) output
2. accept edge ids in neighbors()
3. dedicated annotations(target_id) verb valid for any annotatable target (entity/note/edge/event)

Non-blocking; FTS workaround holds meanwhile.

ohdearquant:
Fixes #803. Additive read surface; integration tests included. Local gates deferred to CI due to build-lane contention; codex review follows.


ohdearquant:
Static review round 5: APPROVE at head 6cf1f250cf0257d7dfba020dfaf491142012804f. Prior round's remaining Low (exact annotation_edge_id assertion) is addressed: the test now retains both annotation-edge ids (explicit link return value plus edge-list recovery for the annotates-param path) and asserts each response item's annotation_edge_id exactly across a two-annotation seed. Ready for maintainer review.

oceanwaves630 (approved):
APPROVED. Hygiene hold cleared: verified implementation_notes.md is gone from both the PR net diff and the head tree (404). Prior codex APPROVE on the code stands. Sequencing acknowledged: merge #961 first, recheck #967/#972 mergeability and re-merge main if the squash surfaces overlap on serve.rs/writer_task.rs. Squash-merge (drops the phantom code_ingest.rs merge-base artifact regardless). Land the tracked .gitignore for implementation_notes.md on main too — .git/info/exclude covers our worktrees but only a tracked entry protects fresh clones + CI on a public repo.

PR #969, opened 2026-07-13
feat(curation): merge event auditability - reason field + note merge event

### LINKED ISSUE #809: curation: EntityMerged has no reason field; merge_note emits NO event at all
Tile-0 instrumentation gap G4 (atlas gap map 2026-07-10), with the open question now CONFIRMED at source:

1. EntityMerged payload = {into_id, from_id, policy, edges_rewired} — no reason/rationale field (khive-runtime curation.rs:414-427). policy alone cannot distinguish duplicate-adjudication merges from other curation merges.
2. CONFIRMED: merge_note (curation.rs:692) emits no event whatsoever — grep of the function body shows no EventKind construction or event-store write. Note merges are invisible to the event plane entirely.

Fix shape: add an optional reason field to EntityMerged (additive payload change), and emit a merge event from the note path symmetrical with the entity path.

Consumer query unblocked: "entity creations >=30d old later merged with reason='duplicate' within the horizon / all mature creations" — the duplicate-escape metric.

Independent of the other instrumentation gaps.

ohdearquant:
Fixes #809. Additive payload changes only. Local gates deferred to CI due to build-lane contention; codex review follows.


oceanwaves630 (approved):
Approved at 5fd55bb82. Read the full diff: reason flows through MergeParams to both _with_reason variants with the secret gate applied, NoteMerged emission is dry-run-guarded, the registry-dispatch regression tests cover the exact handler-boundary gap, and back-compat wrappers keep the old signatures.

PR #970, opened 2026-07-13
fix(kg): resolve canonical entity names that hybrid search ranks first

### LINKED ISSUE #908: kg.resolve returns NotFound for canonical entity names that search() ranks top-1
Found via live dogfooding on 2026-07-12, on the resolve verb shipped 2026-07-09.

Symptom: resolve(refs=["khive ADR-040", ...]) returned NotFound for all 16 canonical-name references in a batch, while search(kind="entity", query=<same name>) returns the intended entity as the top-1 hit trivially for each of them.

Expected per the verb contract: UUID/hex passthrough, then recent-ring, then hybrid-search fallback. The hybrid fallback appears to be name-exact biased: references that differ from the stored canonical name only by a qualifying prefix (for example a project qualifier ahead of the short name) fail outright instead of falling through to hybrid scoring.

Repro shape:
1. Create an entity with a short canonical name (e.g. name = "ADR-040" style short form).
2. Call resolve with a natural-language qualified ref for it (the form an agent actually writes).
3. Observe NotFound; then call search with the same string and observe the entity as top-1.

Acceptance: resolve's fallback stage should recover at least whatever hybrid search ranks first (or return Ambiguous with candidates), and a regression test should pin the qualified-name case. NotFound on a ref that top-1 search resolves is the worst outcome for the calling agent, which then creates a duplicate entity.

ohdearquant:
Fixes #908. Ambiguity still returns Ambiguous; garbage still NotFound. Local gates deferred to CI due to build-lane contention; codex review follows.


oceanwaves630 (approved):
Full diff read. Exact canonical-name resolution is deterministic via the stage-3 short-circuit (proven by EXACT_NAME_CONFIDENCE at limit=1 against 12 hybrid competitors); stage 4 now searches at a 20-deep floor while bounding the Ambiguous payload to the caller's limit, with tie->Ambiguous and garbage->NotFound regressions locked. Docs state the bounded non-exact contract. Matches the resolve() contract ruling.

Issue #806, opened 2026-07-10
event plane: RecallExecuted/SearchExecuted declared but never emitted (served-result telemetry is an empty pipe)


ohdearquant:
Tile-0 instrumentation gap G1 (atlas gap map 2026-07-10, claims re-verified by grep at source).

EventKind::RecallExecuted / SearchExecuted / RerankExecuted are DECLARED (khive-types event.rs:75,79) and the storage decoder + event_observations projection (ADR-041) are ready to consume candidates/selected payloads (khive-db event.rs:441,486-511) — but no production handler constructs any of them. The only observation is transient: the on_dispatch hook grabs the FIRST result id in-memory for the brain posterior (pack.rs:1300-1312), never persisted.

Required emission per recall/search: full returned result-ID list + typed result kind, query text, served_by_profile_id, calling actor, timestamp.

Consumer query unblocked: "all eligible recalls by actor X in window W, each with returned fact/memory IDs" — the feedback-closure denominator AND the stale-exposure numerator. Without it neither metric has a denominator.

Sequencing: upstream of the feedback-serve linkage and effective-time work (nothing to join against until this emits). ADR-103 lane.

ohdearquant:
Closing during tracker cleanup: this concerns functionality provided by the commercially licensed extensions, which are no longer part of this repository. Tracking continues on the distribution's internal tracker.

PR #973, opened 2026-07-13
feat(pack-kg): served-result telemetry on search

ohdearquant:
Addresses #806.

Implementer leg completed the change set but stalled before commit (shared build-lane contention); swept and pushed by the lane owner. Local gates deferred to CI; codex review follows.


oceanwaves630 (approved):
Full diff read: result_kind discriminator fixes the entity-as-Note projection gap with fail-closed rejection and atomic rollback tested both directions; emission mirrors the recall seam off the response path; exactly-one-event and full projection round-trip pinned for both kinds.

PR #980, opened 2026-07-13
test(khive-db): fix flaky tx-age sweep via tx_registry serial group

ohdearquant:
## Fix flaky `tx_age_sweep_names_long_lived_reader_pinning_wal_past_high_water`

`checkpoint::tests::tx_age_sweep_*` read the process-wide `tx_registry`
singleton via `tx_registry::oldest()` and run under `#[serial(tx_registry)]`.
`run_writer_task` (ADR-067 Component A, khive-db) registers a `writer_task_tx`
handle in that same singleton for the life of each `BEGIN IMMEDIATE`. The
`writer_task.rs` tests that spawn a writer task were not part of the
`tx_registry` serial group, so a concurrent spawning test leaked a longer-lived
`writer_task_tx` into the sweep's `oldest()` read. The sweep then named that
transaction instead of the reader the test registered, and the assertion
failed. It was intermittent (the leak only overlaps the sweep on some
schedules), which is why main CI alternated red/green.

### Change
Add `#[serial(tx_registry)]` to the three `writer_task.rs` tests that spawn a
writer task (the only ones that register `writer_task_tx`), joining the
existing serial group. No production code changes.

### Verification
Ran the full `cargo test -p khive-db --lib` suite (default parallel threading,
the condition that surfaces the race) 5 consecutive times: all green,
`tx_age_sweep_*` passing every run. `cargo fmt --check` and
`cargo clippy -p khive-db --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/khive-db/src/writer_task.rs
+++ b/crates/khive-db/src/writer_task.rs
@@ -420,6 +420,7 @@ async fn run_writer_task(
 mod tests {
     use super::*;
     use crate::pool::PoolConfig;
+    use serial_test::serial;
     use std::sync::atomic::{AtomicBool, Ordering};
     use std::sync::Arc;
     use std::time::Duration;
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Test-only diff read in full: three #[serial(tx_registry)] annotations on the writer-task spawning tests plus the import and rationale comments, matching the existing key used by the checkpoint sweep group. CI green.

PR #983, opened 2026-07-14
docs(adr): ADR-049 Amendment 2 - graduated fail-loud fallback policy

ohdearquant:
## Summary

Amends ADR-049's Decision section 2, whose unconditional mandate ("Fallback to local dispatch is mandatory ... never a hard dependency") was never updated when the graduated fail-loud fallback shipped (bba4c4a7). This amendment records the contract the code now implements:

- Closed `FallbackReason` set with stable string codes (`config_mismatch`, `namespace_mismatch`, `no_socket`, `parse_failure`, `protocol_mismatch`) and legitimacy tiers governing quiet/WARN/ERROR observability.
- `KHIVE_DAEMON_STRICT=1` rejects any would-be fallback with a structured, marker-carrying error (default off).
- Positively observed respawn failures (spawn error, child exited before socket bind) return a stable `respawn_failed` error in both modes instead of silently serving cold local dispatch; detailed context goes to local structured logging only.
- `KHIVE_NO_DAEMON=1` remains the sanctioned daemonless opt-out; genuinely daemonless environments keep quiet fallback.

## Sequencing

Docs-only. Lands before/with #963 so its daemon-path changes are re-gated against the amended contract rather than the stale section-2 mandate.

### DIFF
--- a/docs/adr/ADR-049-khived-daemon.md
+++ b/docs/adr/ADR-049-khived-daemon.md
@@ -190,3 +190,56 @@ Rationale: the kkernel unification (ADR-003 convergence path, now complete) abso
 `khive-mcp` as a library, making `kkernel` the sole shipped Rust binary. All MCP
 configurations, daemon-spawn logic, and user-facing documentation should reference
 `kkernel mcp` and `kkernel mcp --daemon` instead of `khive-mcp` and `khive-mcp --daemon`.
+
+## Amendment 2 (2026-07-14): graduated fail-loud fallback policy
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Amendment 2 reviewed in full: graduated reason-coded policy, strict-mode semantics, and the confirmed-respawn rejection contract are coherent with the merged fail-loud direction, and the caller-visible error correctly excludes spawn detail. Approved.

PR #985, opened 2026-07-14
docs(adr): ADR-041 Amendment A3 - SearchExecuted polymorphic result_kind

ohdearquant:
## Summary

Amends ADR-041's SearchExecuted projection contract, which currently specifies note-only Candidate/Selected rows. Issue #806 requires typed served-result telemetry for entity searches; stamping entity UUIDs as note referents would durably misclassify the append-only projection.

- `SearchExecuted` payloads gain a closed `result_kind` key (`"entity"` | `"note"`); the decoder stamps Candidate/Selected rows with the matching `ReferentKind`. Other values are a surfaced decode error.
- Legacy payloads without `result_kind` remain the documented historical note shape — projected as note rows, never rejected, no migration.
- `RecallExecuted` / `RerankExecuted` unchanged.

## Sequencing

Docs-only. Lands before/with #973 so its telemetry implementation is gated against the amended contract.

### DIFF
--- a/docs/adr/ADR-041-event-provenance-projection.md
+++ b/docs/adr/ADR-041-event-provenance-projection.md
@@ -658,6 +658,35 @@ event, `decode_signal_observation` picks `ReferentKind` from `event.substrate` (
 back to `Entity` for pre-fix historical events still carrying the `SubstrateKind::Event`
 placeholder), and `observed_as_signal` admits both entity and note referents.
 
+## Amendment A3: `SearchExecuted` admits a typed `result_kind` (2026-07-14, khive#806)
+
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
SPEC-GATE APPROVED (Amendment A3). Sound on all four edges I gated: closed two-value result_kind with decode-error on unknown values (fail-loud, consistent with the ADR's closed-value posture); legacy payloads without the key projected as the documented historical note shape and never rejected (append-only log honored, zero migration); single-substrate-per-event invariant stated; RecallExecuted/RerankExecuted explicitly out of scope. Land it, then #973 round 3 may implement against it.

PR #990, opened 2026-07-14
ci: reclaim runner disk in coverage-ratchet job before llvm-cov

ohdearquant:
## Problem

The required **CI gate** aggregate depends on the `coverage-ratchet` job, which runs `cargo llvm-cov --workspace`. That is the disk-heaviest command in CI: full-workspace instrumentation plus per-object `.profraw` files on top of the normal target dir overflow the ubuntu runner's ~21GB and surface as `System.IO.IOException: No space left on device`. The job then fails as `Coverage ratchet: failure`, which fails the required gate.

Because it depends on how full the runner happens to be, the failure is non-deterministic: an identical branch (including docs-only PRs that add zero code) can pass or fail the ratchet run to run.

## Fix

The matrix `ci` job already purges the unused preinstalled toolchains (Android, .NET, GHC, CodeQL, docker images) to reclaim ~25GB before building. The `coverage-ratchet` job was missing that step even though it runs the single heaviest command. This PR adds the identical purge right after checkout so the llvm-cov pass has headroom and the gate becomes deterministic.

No change to what the ratchet measures or the baseline — purely a runner-disk headroom fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -428,6 +428,18 @@ jobs:
     # rejected and fail the job on every merge.
     steps:
       - uses: actions/checkout@v7
+      # `cargo llvm-cov --workspace` is the disk-heaviest command in CI: it
+      # instruments every crate and emits per-object profraw files on top of the
+      # normal target dir, overflowing the ubuntu runner's ~21GB and surfacing as
+      # `No space left on device`. The matrix `ci` job already purges the unused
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Reviewed at head 57abc647a: applies the same runner-disk purge the matrix ci job already uses to the coverage gate; validated empirically by this PR's own coverage job going green. CI 22/22.

PR #991, opened 2026-07-14
feat(runtime): surface per-relation edge endpoint allowlist in link help

### LINKED ISSUE #964: link/verbs help: expose the per-relation edge-endpoint allowlist
From a ~350-edge batch enrichment run. The kernel enforces endpoint legality with a self-documenting error, but link(help=true) and verbs() do not expose the allowlist, so client tooling re-implements it and gets it wrong: an applier dropped a legal composed_with project-to-project edge because its local table was stricter than the kernel's. Ask: emit the per-relation endpoint table (source kinds -> target kinds, note rules included) in link(help=true) and/or verbs() so batch appliers can defer to the kernel knowingly. The table already exists internally (pack EDGE_RULES); this is surfacing, not new policy.

ohdearquant:
## Summary

`link(help=true)` now returns an `endpoint_rules` array documenting every
legal `(relation, source, target)` triple the kernel accepts for an edge:

- the base entity-to-entity endpoint contract,
- every loaded pack's additive `EDGE_RULES` (e.g. task-to-task `depends_on`),
- the `annotates` note-to-any special case.

## Why

Batch tooling that creates edges outside the MCP request path has had to
re-implement the endpoint allowlist locally, and a stricter local copy has
silently dropped edges the kernel would have accepted (e.g. a legal
`composed_with` project-to-project edge in a ~350-edge enrichment run). The
allowlist already exists internally as the composed pack `EDGE_RULES` set
consulted by `validate_edge_relation_endpoints`; this change surfaces that
same live table through `link`'s help schema instead of adding new policy,
so a caller can defer to the kernel's own rules directly.

Closes #964

## Test plan

- [x] `cargo test -p khive-runtime` — 855 + 2 + 69 passed, 0 failed
- [x] `cargo clippy -p khive-runtime --all-targets -- -D warnings` — clean
- [x] `cargo fmt -p khive-runtime -- --check` — clean
- [x] New test `test_link_help_true_exposes_endpoint_rules` asserts the base
      table, a pack-declared additive rule, and the `annotates` rule all
      appear in `link(help=true)`'s `endpoint_rules`, and that `help=true`
      remains side-effect-free (no pack dispatch invoked)


oceanwaves630 (approved):
Approved at c2ffb1cd1. Read the full diff: endpoint_rules composes the same sources the validator consults, the three special relations get their unconditional note-to-note rows, unenforced pack rules for special relations are correctly excluded rather than falsely advertised, and the set-level test guards both inclusion and exclusion.

PR #994, opened 2026-07-14
fix(pack-brain): coerce date-only since/until in brain.event_counts

### LINKED ISSUE #984: brain.event_counts: date-only `since` silently returns null instead of erroring
## Repro (kkernel 0.3.0, 2026-07-14)

`brain.event_counts(since="2026-07-07")` returns a null result with no error. The full RFC-3339 form `since="2026-07-07T00:00:00Z"` works and returns the expected `by_profile` / `counts_by_actor` / `counts_by_kind`.

## Why it matters

A silent null on a measurement surface reads as "no data". During an internal audit this produced a false "the event plane can't answer this" conclusion until the timestamp format was probed. Foot-gun on an observability verb.

## Ask

Either coerce a date-only value to `T00:00:00Z`, or return a validation error naming the required format. Never null-on-parse-failure.

(Surfaced during internal dogfooding.)

ohdearquant:
## What changed

`brain.event_counts` parses `since`/`until` with strict RFC-3339 parsing via `parse_rfc3339_micros`. A date-only value such as `"2026-07-07"` fails that strict parse (RFC-3339 requires a time-of-day component), and the resulting failure did not surface clearly to the caller — it read as a silent "no data" outcome on an observability verb.

`parse_rfc3339_micros` now recognizes a bare `YYYY-MM-DD` date and coerces it to midnight UTC before falling back to the existing strict RFC-3339 parse. A value that is neither a valid date nor a valid RFC-3339 datetime still returns a validation error naming the field and the offending value — it is never silently dropped.

## Why

Closes #984. The full RFC-3339 form (`"2026-07-07T00:00:00Z"`) already worked; the date-only shorthand is a natural caller mistake on a verb whose whole purpose is measurement, so failing clearly (or, better, just working) matters more here than on most verbs.

## Test evidence

```
cargo test -p khive-pack-brain
...
test result: ok. 227 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.16s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.23s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.12s

cargo clippy -p khive-pack-brain --all-targets -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 32s   (clean, no warnings)

cargo fmt -p khive-pack-brain -- --check
(clean, no output)
```

New tests added:
- `date_only_since_is_coerced_to_midnight_utc` — reproduces the issue's exact repro and asserts the coerced result, not a null.
- `garbage_since_is_rejected_not_silently_nulled` — a value that is neither a date nor an RFC-3339 datetime still errors by name.

- [ ] Draft — ready for review once CI is green.


oceanwaves630 (approved):
Approved at 3a52da0fa. Read the full diff: role-aware date-only parse is correct for the half-open window, the boundary fixture pins both edges within one microsecond (last instant of the named day included, next midnight excluded), garbage inputs on both fields get named errors, and the max-date roll returns InvalidInput instead of panicking, with a regression test.

Issue #606, opened 2026-07-04
channel health as a queryable surface: per-channel last-successful-ingest + last-error via a status verb


ohdearquant:
## Problem

The 2026-07-04 email degrade (#602) ran ~19 hours undetected because channel health only existed as log lines, and the daemon's logs were going to /dev/null (#599). Even with #599 fixed, logs are a forensic surface, not a monitoring one: an operator (or an automated monitoring loop) cannot ask "is inbound email healthy right now?" without grepping.

## Required behavior

Expose per-channel health as a queryable verb (session pack status verb, or a dedicated `channel.status`), returning for each configured channel (email today; telegram/whatsapp when they land per #113/#115):

- `last_successful_ingest` (timestamp of the last inbound item actually persisted)
- `last_poll_attempt` + `last_error` (message class, not raw secrets/headers)
- loop liveness (is the poll task running in THIS process, and under what role once #602's role-gating lands)
- outbox: pending count + last successful send + last skip/error class

Staleness policy stays with the caller: the verb reports facts; a monitoring loop decides that "no ingest in N hours" is an alert.

This is the concrete shape of the LOUD-startup channel-health item from the 7/4 triage: startup should also log this snapshot loudly, but the queryable surface is the part that makes silent reintroduction of a degrade impossible.

## Detection story after the full prevention set

telemetry captured by default (#599) + health queryable (this) + role-gated spawning (#602) + single construction path with parity test (#603) + graceful degrade under contention (#605).

Crates: `crates/khive-channel-email` (health state), session/status verb surface in the relevant pack.

ohdearquant:
Design review outcome (2026-07-04): accepted with four amendments — v1 shape approved (read-only comm.health(), no args, per-channel rows, timestamps only, never a computed healthy bool, comm-pack-local), amended as follows:

1. **Cross-process read is the point (blocking)**: the daemon PERSISTS its heartbeat/health state; client-role callers return that persisted state annotated `{source: "daemon-heartbeat", as_of: <ts>}`. `role=client` with all-null loop fields is correct only when no daemon state exists at all. `role` means "who owns the loops", not "whose memory answered". The heartbeat row is definitely in v1 scope; escalate back only if it requires a schema MIGRATION rather than a new row in tables the pack already owns.
2. **Channel identity**: key rows by stable channel identifier (config name/slug + kind), never kind alone — two accounts of the same kind must not collapse.
[...]

PR #995, opened 2026-07-14
fix(pack-comm): authorize per-tenant heartbeat writers

### LINKED ISSUE #917: comm: heartbeat rows are only ever written under the local namespace — no authorized per-tenant writer path
Follow-up from #877/#914: comm.health and comm.probe now correctly READ under the injected namespace token and self-describe the namespace in the response, but heartbeat rows are only ever PRODUCED under the local namespace. A tenant-scoped health call on a cloud deployment therefore reads an empty set by construction — indistinguishable from a genuinely silent channel — until an authorized cloud-side per-tenant heartbeat writer exists.

Scope: design + implement the authorized writer path for non-local namespaces (who stamps heartbeats for a tenant, under what token, and through which seam), keeping the Gate as the authorization boundary (ADR-018) and namespace as attribution (ADR-007 Rev 6).

Flagged as explicitly out-of-scope in PR #914's fix round; recorded here so it does not evaporate.

ohdearquant:
## What changed

`comm.heartbeat` previously always persisted its `channel_health` row under the fixed `khive_pack_comm::CHANNEL_HEALTH_NAMESPACE` constant (`"local"`), regardless of the caller's dispatch-authorized namespace. A tenant-scoped `comm.health` read on a cloud deployment therefore always observed an empty set by construction — indistinguishable from a genuinely silent channel — because no writer existed for any namespace but `"local"`.

`handle_heartbeat` now persists under `token.namespace()`, the same dispatch-authorized namespace every other comm verb already resolves. The Gate remains the sole authorization boundary (ADR-018): `comm.heartbeat` is a `Subhandler` with no MCP wire path at all, so only trusted internal Rust callers holding a `VerbRegistry` handle can reach it in the first place.

- The existing local single-tenant channel poll loop is unaffected: it already pins its own dispatch params to the `CHANNEL_HEALTH_NAMESPACE` constant explicitly, so it keeps writing to `"local"` exactly as before.
- An authorized per-tenant writer instead dispatches via `VerbRegistry::dispatch_as` with a `VerifiedActor` (an out-of-band authenticated tenant principal) and its own tenant namespace as the explicit `namespace` dispatch param — the same ADR-007 Rev 6 Rule 3 escape every other comm verb uses. Those heartbeat rows now land under that tenant's own namespace, so a tenant-scoped `comm.health` call observes real writer state instead of an empty set by construction.

Doc comments in `handlers.rs`, `lib.rs`, and `vocab.rs` are updated to describe the new model.

## Why

Follow-up from #877/#914: `comm.health`/`comm.probe` read under the injected namespace and self-describe it in the response, but heartbeat rows were only ever produced under `"local"`. This closes the write-side gap flagged as explicitly out-of-scope in #914's fix round.

## Test plan

- [x] Added `authorized_writer_persists_heartbeat_under_its_own_tenant_namespace` in `crates/khive-pack-comm/tests/integration.rs`: dispatches `comm.heartbeat` via `dispatch_as` with a `VerifiedActor` and an explicit non-local `namespace`, asserts a `comm.health` call scoped to that namespace sees the row while the default local-scoped read does not.
- [x] `cargo test -p khive-pack-comm` — 161 tests pass (136 unit/integration + 25 probe), including all pre-existing `#606`/`#877` heartbeat/health regression tests unchanged.
- [x] `cargo clippy -p khive-pack-comm --all-targets -- -D warnings` — clean.
- [x] `cargo fmt -p khive-pack-comm -- --check` — clean.
- [x] `cargo check -p khive-mcp --all-features` — clean (verifies the local single-tenant poll loop call site is unaffected).

Closes #917


oceanwaves630 (approved):
Full diff read. The #606 pin removal is safe: comm.heartbeat has no wire path, the dispatch gate remains the sole auth boundary (ADR-018), and the local poll loop already passes the constant explicitly so its writes are unchanged. The two-tenant distinct-rows test correctly pins the namespace-in-id derivation on the real write path (INSERT OR REPLACE collision would silently drop a tenant's row otherwise). Docs preserve the full incident history and the vocab param description now tells the truth about where rows land.

PR #996, opened 2026-07-14
test(pack-git): lock in ingest masking for uuid-near-trigger FP (#977)

### LINKED ISSUE #977: secret-gate: mask instead of write-block on ingest paths (uuid-near-trigger FP on legitimate issue content)
Live repro (two warnings, tracker-ingest trial): an issue whose title contains a credential-like token name plus ANY full UUID in the body trips the uuid-near-trigger heuristic, and `git.digest` write-blocks the whole issue. 'Reword the source' is wrong advice on ingest paths: the source is a legitimate issue the ingester does not own.

Options: (a) mask the flagged span instead of blocking on ingest paths, (b) tighten the trigger window so a title token does not arm the whole body. Related prior work: #632/#577 (path FPs near trigger words).

ohdearquant:
## Summary

- Investigated the false positive reported in #977: a legitimate GitHub issue whose title mentions a credential-like word (e.g. `key`, `auth`) while the body separately contains an unrelated UUID was reported to trip the `uuid-near-trigger` heuristic and cause `git.digest` to write-block the whole issue.
- Traced through `khive-pack-git`'s ingest path (`ingest_issues` / `MaskedIssueFields`) and confirmed the fix is already in place: `title` and `body` are each passed through `secret_gate::mask_secrets` (not the hard `check()` gate) before the note is ever created. This landed via #835/#910, both merged before #977 was filed.
- Verified empirically (including a batch of trigger-word/UUID combinations, cross-field and same-field) that the reported scenario no longer write-blocks: the ambiguous span is masked in place and the issue note still lands.
- Added two regression tests to `khive-pack-git/tests/acceptance.rs` that lock in this behavior so it cannot silently regress:
  - `ingest_does_not_block_issue_with_credential_word_in_title_and_uuid_in_body` — trigger word in title, UUID in body (different fields); the unrelated UUID must survive verbatim.
  - `ingest_masks_credential_word_and_uuid_co_occurring_in_issue_body` — trigger word and UUID in the same field; the ambiguous span must be masked, not block the write.

No production code changes were needed — the masking behavior these tests exercise already exists on `main`.

## Test plan

- [x] `cargo test -p khive-pack-git` — 74 unit tests + 54 acceptance tests pass (including the 2 new regression tests)
- [x] `cargo clippy -p khive-pack-git --all-targets -- -D warnings` — clean
- [x] `cargo fmt -p khive-pack-git -- --check` — clean

Closes #977

### DIFF
--- a/crates/khive-pack-git/tests/acceptance.rs
+++ b/crates/khive-pack-git/tests/acceptance.rs
@@ -690,6 +690,166 @@ async fn ingest_masks_credential_shaped_issue_title_without_dropping_note() {
     );
 }
 
+/// Issue #977: a legitimate issue whose title mentions a credential-like
+/// word (e.g. "key", "auth", "token") while the body separately contains an
+/// unrelated full UUID must not be write-blocked by the `uuid-near-trigger`
+/// heuristic. `MaskedIssueFields` masks `title` and `body` independently
+/// before the note is ever created, so a trigger word in one field cannot
+/// "arm" a UUID in the other -- covers the exact cross-field shape reported
+/// live against `git.digest`.
+#[tokio::test]
+async fn ingest_does_not_block_issue_with_credential_word_in_title_and_uuid_in_body() {
+    let _guard = ENV_MUTEX.lock().await;
+    let (rt, token, registry) = fixture().await;
+
+    let project_id = create(
+        &registry,
+        json!({"kind": "project", "name": "issue-977-cross-field-repo"}),
+    )
+    .await;
+
+    let dir = tempfile::tempdir().expect("tempdir");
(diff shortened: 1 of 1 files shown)


ohdearquant:
Superseded by #1062 (merged 2026-07-17 03:32Z, closed #977). Both regression tests this PR adds — `ingest_does_not_block_issue_with_credential_word_in_title_and_uuid_in_body` and `ingest_masks_credential_word_and_uuid_co_occurring_in_issue_body` — already landed in main via #1062's cherry-picked test additions. Verified byte-identical test fns present in origin/main crates/khive-pack-git/tests/acceptance.rs. No delta to ship.

PR #1033, opened 2026-07-15
perf: contract-result-v1 schema for the local-engine contract bench suite

ohdearquant:
First slice of the local-engine contract bench suite: the versioned result contract lands before any runner, so the suite implementation and its reviews build against a fixed schema.

`scripts/perf/schemas/contract-result-v1.json` (draft 2020-12, validated) defines one JSON document per suite run over the five contract dimensions: ingest throughput (embed-excluded primary, embed-included advisory), ANN query latency at an iso-recall operating point across worker arms, recall@k against a vectorized exact baseline, FTS+ANN fusion quality, and index-inclusive storage (post-VACUUM, WAL checkpointed).

Rules the schema encodes:
- **Calibration first**: every document either cites a same-SHA variance profile from `scripts/perf/bench_calibrate.py` or marks itself `uncalibrated`; no pass/fail field exists anywhere in the document.
- **Isolation evidence as data**: bench-window held or not, loadavg before/after, isolated build dir.
- **Exact baselines only**: speedup denominators are restricted to `faiss-flat` / vectorized brute force — a naive scalar-scan denominator is not representable.
- **Ledger derivation**: `ledger.csv` rows are derived from these documents by the runner, never hand-edited.

`perf/README.md` gains a section documenting the above. No runner emits the document yet; absent dimensions mean not-measured.

Gates: schema validates as draft 2020-12 (`jsonschema.Draft202012Validator.check_schema`), JSON hook green.

### DIFF
--- a/perf/README.md
+++ b/perf/README.md
@@ -92,6 +92,38 @@ the ledger `notes` column as `{N}GiB`.
 
 ---
 
+## Contract-suite result schema (contract-result-v1)
+
+`scripts/perf/schemas/contract-result-v1.json` is the versioned result contract
+for the local-engine contract benchmark suite: one JSON document per suite run,
+covering up to five contract dimensions (ingest throughput, ANN query latency at
+iso-recall, recall@k vs an exact baseline, FTS+ANN fusion quality, and
+index-inclusive storage). Key rules the schema encodes:
+
+- **Calibration first.** Every result document carries a `calibration` block
+  that either cites a same-SHA variance profile produced by
+  `scripts/perf/bench_calibrate.py` (`status: "calibrated"`, with the artifact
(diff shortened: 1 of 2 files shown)


ohdearquant:
Automated review, round 2: APPROVE — 0 findings. Round-1 fixes verified by validation probes: calibration states are mutually exclusive closed branches (calibrated requires artifact path + runs >= 10; uncalibrated admits only status), recall_floor removed (rejects as additional property; operating point is external runner configuration), ledger_rows_appended root-required with 0 = nothing banked. Reviewed-at-HEAD: 2672b0cbb69b3b569d07f64ad3121a81c5cd026b

oceanwaves630 (approved):

PR #1034, opened 2026-07-15
perf: contract-suite runner for ingest, ANN query, and recall@k dimensions

ohdearquant:
Second slice of the local-engine contract bench suite: `scripts/perf/contract_suite.py`, a runner for the first three contract dimensions (ingest, ANN query, recall@k) emitting one schema-validated `contract-result-v1` document per invocation (contract merged in #1033).

Design:
- One `vec_bench` release run is reused across all three dimensions; the exact baseline for `recall_at_k` is measured independently on the same corpus/query slices vec_bench uses internally (verified against vec_bench.rs sampling), preferring `faiss.IndexFlatL2` from an isolated venv and falling back to a numpy vectorized scan — never a scalar loop.
- The document is validated in-process against `scripts/perf/schemas/contract-result-v1.json` before writing; validation failure is a runner error. `calibration` is always `uncalibrated` in this slice; no pass/fail or threshold logic exists in the runner.
- Isolation is recorded honestly: loadavg before/after, whether an exclusive flock on the bench-window lock was obtained (non-blocking, failure recorded not raised), and whether CARGO_TARGET_DIR resolves outside the worktree.

Loud omissions (absent = not-measured, per the contract): concurrent ann_query arms (needs a vec_bench extension), recall@1/@100 rows (vec_bench hardcodes K=10), embed-included ingest (no embedder in this corpus path), ledger banking (contract-suite ledger wiring is a follow-up).

Evidence: `--selftest` (10K synthetic, all three dims, ~30s warm) exits 0 with a valid document; a real SIFT-1M run at N=100K with the faiss-flat baseline produced recall@10 0.9504 at p50 54.5µs single-thread and a 14.8x speedup over faiss-flat p50 807.7µs (uncalibrated, bench_window=false, recorded as such in the document).

### DIFF
--- a/scripts/perf/contract_suite.py
+++ b/scripts/perf/contract_suite.py
@@ -302,8 +302,11 @@ def run_vec_bench(base: Path, query: Path, n: int, dataset_name: str, out_json: 
         str(out_json.parent / "bank-run"),
     ]
     proc = run(cmd, cwd=CRATES_DIR, env=env)
+    rendered_cmd = " ".join(str(c) for c in cmd)
     if not out_json.exists():
-        raise RuntimeError(f"vec_bench did not write its output JSON (exit={proc.returncode}); see stderr above")
+        raise RuntimeError(
+            f"vec_bench did not write its output JSON (exit={proc.returncode}); command: {rendered_cmd}"
+        )
     with out_json.open() as f:
         data = json.load(f)
     if not data.get("rows"):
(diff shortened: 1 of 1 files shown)


ohdearquant:
Review gate: three rounds completed. Round 1 found a baseline k-mismatch (exact baseline timed at k=100 vs the emitted k=10 row, inflating the reported speedup), silent acceptance of nonzero vec_bench exits, silent synthetic fallback on missing SIFT files, retained work artifacts, and environment-specific identifiers — all fixed. Round 2 verified all five fixes with executed regressions and left one Low diagnostic finding, fixed in 0a6c4b9. Round 3 approved at head 0a6c4b9a00b65b5c238cb5e2788c44f14e15732a. Corrected real-SIFT run (N=100K, faiss-flat k=10 baseline, isolated window): recall@10 0.9504, ANN p50 59.4µs, speedup 13.64x — uncalibrated/advisory per the contract schema.

oceanwaves630 (approved):

PR #1035, opened 2026-07-15
feat(vamana): vec_bench concurrent query arms + per-query top-k dumps

ohdearquant:
Extends the vec_bench example binary (khive-vamana) with the two measurement capabilities the contract bench suite documented as omissions:

- `--workers N[,M...]`: concurrent query arms. For each worker count, the warm query phase runs with N OS threads issuing queries concurrently over the same query slice, each thread with its own VisitedSet and clock; a barrier synchronizes all workers between warm-up and the timed loop so every sample is taken under full concurrency. Per-query latencies are pooled for p50/p95/p99 and recall is computed from the actual concurrent results. VamanaGraph's search path was verified read-safe before threading (&self, no interior mutability, no locks added). Emitted as an additive concurrency_arms array recording the actual spawned worker count; invalid --workers values exit 2 with a clear diagnostic.
- `--dump-topk PATH`: a separate untimed pass searching at K=100 over the ground-truth sample, writing one JSON line per query ({query_idx, ids}) and emitting additive recall_at_1 / recall_at_100 fields against freshly computed K=100 exact ground truth. The timed K=10 phase is unaffected. Multi-N runs write per-N dump files (dump-n{N}.jsonl).

Default invocation output is unchanged apart from the new fields (single workers=1 arm, null dump fields); bench_1m.sh / ingest_scale_proof.py consumers verified unaffected.

Evidence: fmt/clippy(-D warnings)/tests clean; synthetic runs for all behaviors including malformed-input and requested-vs-spawned discrepancy; real SIFT-1M N=100K — recall@1 1.000, recall@10 0.9504, recall@100 1.000; barrier-synchronized concurrency arms at beam 30: workers=1 p50 50.8µs, workers=4 p50 55.3µs (p99 78.7µs), workers=16 p50 63.9µs (p99 262.6µs) — the expected contention gradient, replacing an earlier pre-barrier run whose flat profile was an artifact of unsynchronized measurement starts. All figures advisory/uncalibrated.

### DIFF
--- a/crates/khive-vamana/examples/vec_bench.rs
+++ b/crates/khive-vamana/examples/vec_bench.rs
@@ -55,6 +55,8 @@ struct Args {
     normalization: String,
     source_url: Option<String>,
     bank_run_dir: Option<PathBuf>,
+    workers: Vec<usize>,
+    dump_topk: Option<PathBuf>,
 }
 
 fn parse_args() -> Args {
@@ -71,6 +73,8 @@ fn parse_args() -> Args {
     let mut normalization = String::from("l2");
     let mut source_url: Option<String> = None;
     let mut bank_run_dir: Option<PathBuf> = None;
+    let mut workers: Vec<usize> = Vec::new();
+    let mut dump_topk: Option<PathBuf> = None;
 
     while let Some(key) = args_iter.next() {
         match key.as_str() {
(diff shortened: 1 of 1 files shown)


ohdearquant:
Review gate: two rounds. Round 1 found a missing synchronization barrier between per-worker warm-up and timed measurement (a worker finishing warm-up early could collect samples while peers were still warming, invalidating the concurrency claim) and lenient --workers parsing that silently dropped invalid values and recorded requested rather than actual worker counts. Both fixed in 24b1772d3. Round 2 approved at head 24b1772d393c4173805f85ca603989f878a5265b with executed regressions (barrier sizing vs spawned ranges, malformed-input exits, requested-vs-actual reporting). The barrier materially changed the measured SIFT-1M profile: pre-barrier arms read flat (~54µs at all worker counts); post-barrier arms show the true contention gradient (p50 50.8/55.3/63.9µs, p99 up to 262.6µs at 16 workers) — the PR body carries the corrected numbers.

oceanwaves630 (approved):

PR #1036, opened 2026-07-15
feat(perf): wire contract_suite.py to vec_bench concurrency arms + recall@1/@100

ohdearquant:
Wires the contract-suite runner (merged in #1034) to the vec_bench capabilities added in #1035:

- `ann_query.arms` now carries one schema arm per `--workers` value (default 1,4,16, runner-level CLI flag), mapped field-for-field from vec_bench's barrier-synchronized concurrency_arms.
- `--dump-topk` is always passed when recall_at_k is requested; the measured recall@1 and recall@100 values are surfaced in the document's notes field as explicitly-tagged diagnostics, NOT as recall_at_k rows. The schema requires every row to carry speedup_vs_baseline, and there is no measured ANN latency at k=1/k=100 (the dump pass is untimed and runs a different beam/k configuration than the timed K=10 phase) — pairing those recalls with the k=10 latency would fabricate a speedup. Real k=1/k=100 rows need a timed vec_bench pass at those k values (future extension), not a runner-side workaround.
- recall_at_k keeps exactly the k=10 row with its k-matched exact baseline.

Evidence: selftest exit 0, schema-valid, arms present, no pass/fail/threshold logic (grep clean). Real SIFT-1M N=100K run under the bench-window lock: arms p50 53.6/56.0/60.8µs at 1/4/16 workers (recall@10 0.9504 flat), recall@1 1.0 / recall@100 0.99995 diagnostic.

Data caveat, stated plainly: this run's faiss-flat baseline measured p50 2079µs vs 810µs in the prior run at identical N and k — the machine's 1-minute loadavg rose from ~7 to ~15.6 during the window, and the advisory lock does not exclude non-cooperating processes. The resulting 31.6x speedup figure is therefore load-contaminated on the baseline side and should not be quoted; the run-to-run baseline spread is itself the argument for the calibration profile (N>=10 same-SHA runs) this suite requires before any figure is banked. All documents remain calibration.status=uncalibrated.

### DIFF
--- a/scripts/perf/contract_suite.py
+++ b/scripts/perf/contract_suite.py
@@ -9,6 +9,15 @@ brute-force fallback) for the `recall_at_k` dimension. Emits exactly one
 contract-result-v1 JSON document, validated in-process against
 scripts/perf/schemas/contract-result-v1.json before it is written.
 
+`ann_query` arms come from vec_bench's `--workers` concurrency arms (default
+1,4,16 -- override with `--workers`); each arm is barrier-synchronized so its
+p50/p95/p99 reflect that worker count actually querying concurrently.
+`recall_at_k` also drives vec_bench's `--dump-topk` to get recall@1/@100, but
+only emits the k=10 row in the schema (see `RECALL_ROW_KS` docstring for why:
+no measured ANN latency exists at k=1/k=100 to pair with a required
+speedup_vs_baseline); recall@1/@100 are still recorded in the document's
+`notes` field.
+
 This runner never computes a pass/fail verdict: it measures and records.
 `calibration` is always `{"status": "uncalibrated"}` here -- calibrated runs
 are produced separately by scripts/perf/bench_calibrate.py (K>=10, same SHA).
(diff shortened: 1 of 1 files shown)


ohdearquant:
Review summary: round 1 found one Medium (empty `--workers` segments were silently skipped, e.g. `1,,4` ran as a two-arm set); fixed in 6bb2093 — an empty stripped segment now raises `ValueError`, with focused checks covering `1,,4` / `1,` / `,1` / empty string rejection and `1,4,16` acceptance. Round 2 re-verified the fix at head 6bb20934c: clean approval, no remaining findings. Arm mapping, k=1/k=100 diagnostic representation (no fabricated latency rows), and fail-loud behaviors were all verified in round 1.

oceanwaves630 (approved):
Maintainer read at head 6bb20934c after the two-round external gate: verified the empty-segment rejection now fails loudly before any benchmark runs, the worker arms forward without silent filtering, and the k=10 restriction plus recall diagnostics are explicitly logged rather than silently capped. Diff touches bench tooling only. Approving.

PR #1037, opened 2026-07-15
feat(perf): contract suite for bench_calibrate.py same-SHA calibration

ohdearquant:
Adds a `contract` suite to `scripts/perf/bench_calibrate.py`'s pluggable SUITES registry, driving `scripts/perf/contract_suite.py` and extracting every numeric leaf of the schema-valid contract result (per-arm p50/p95/p99 + recall keyed by worker count, recall_at_k rows keyed by k, baseline p50, ingest metrics). Fixture-based extract tests included; all 21 embedded tests pass.

Measured with it (N=10 same-SHA runs, SIFT-100K, exclusive bench window): k=10 speedup vs faiss-flat mean 13.67x CV 14.0%; baseline p50 CV 3.5%; recall@10 exactly 0.9504 with zero variance; tail latencies (p99) are the noisy dimension (CV up to 211% at 16 workers). The harness emits advisory floors only; no gate thresholds are set anywhere.

One file changed plus test fixtures; no Rust changes.

### DIFF
--- a/scripts/perf/bench_calibrate.py
+++ b/scripts/perf/bench_calibrate.py
@@ -37,6 +37,19 @@ Suites (pluggable via SUITES registry):
             Extracts assertion checks (recall_at_10, beam_growth_exponent,
             speedup_vs_brute_force), per-N row metrics, and growth-exponent
             fits from the schema-versioned result JSON it writes.
+  contract  drives scripts/perf/contract_suite.py (the local-engine
+            contract-result-v1 suite: ingest, ANN query concurrency arms,
+            recall@k). Cheap synthetic defaults (no SIFT-1M, no faiss
+            venv) - pass the real dataset/venv/worker-arm configuration
+            via repeated --suite-arg for a citable profile run. Extracts
+            every numeric leaf worth statting from the emitted
+            contract-result.json: per-arm p50_us/p95_us/p99_us/
+            measured_recall_at_10 (keyed by workers), recall_at_k rows'
+            recall/speedup_vs_baseline (keyed by k), the recall_at_k
+            baseline query_us_p50, and the ingest dimension's metrics.
+            Non-numeric leaves (dataset names, baseline kind) are skipped;
+            they remain visible in the per-run contract-result.json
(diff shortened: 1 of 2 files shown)


ohdearquant:
Review summary: round 1 found one Medium (the contract extractor accepted results without checking the schema_version/suite root discriminators pinned by contract-result-v1.json); fixed in f3b58e99 with SchemaError guards and regression tests for both rejection paths (23 embedded tests pass). Round 2 re-verified the fix at head f3b58e99: clean approval, no remaining findings. Registry-pattern conformance, per-workers/per-k metric keying, and the advisory-only posture were verified in round 1.

oceanwaves630 (approved):
Full methodology read at head f3b58e996. Approving.

Scope first: this adds a "contract" suite handler to the calibration harness — an extractor plus a registry entry, not the statistical-inference core. _contract_extract reads the contract-result document, validates the schema discriminators (version, suite), and flattens numeric leaves keyed by their discriminant (workers, k) into a flat metrics dict. It computes no variance, gate power, CI semantics, or false-positive rates, so it is plumbing, not inference — the soundness concern that applies to statistical-design specs does not apply to this change.

What I did check, and found sound:

- Fail-closed on schema drift: the twelve self-check tests cover missing result file, schema-version drift, wrong suite discriminator, missing arm/row fields, missing baseline percentile, empty dimensions, and — importantly — duplicate discriminant keys (a repeated workers or k value is rejected, so a concurrency arm cannot be silently double-counted).
- Honest separation of advisory from headline: embedder-included throughput is treated as an optional advisory companion, never the headline number; diagnostic values are labeled diagnostic; the cheap synthetic default is explicitly NOT citable — a real profile run requires overriding with the real dataset and venv via repeated suite-args. Non-numeric leaves (dataset names, baseline kind) are skipped rather than coerced into stray metrics.

One non-blocking nit for a later pass: a present-but-non-numeric leaf would raise ValueError from float() rather than a clean SchemaError. It stays fail-closed either way (no fabricated number), so it does not block the merge; worth wrapping for a tidy error on corrupt input.

Approving.

PR #1039, opened 2026-07-15
feat(pack-code): code.ingest verb — L1 manifest edges + L1.5 import-scan (ADR-085 Amendment 2)

ohdearquant:
## Summary

First verb on the code pack: `code.ingest(path, db?, languages?)` implementing ADR-085 Amendment 2's L1 and L1.5 tiers (L2 Scanner/Extractor follows in a second PR).

- L1: Cargo.toml / pyproject.toml / package.json parsing → `project depends_on project` edges with dependency kinds.
- L1.5: regex import scan (Rust use/extern crate, Python import/from, TS import/require) → module and project `depends_on` edges.
- B4 identity: uuid5 over (source_project, language, module_path, name, kind), fixed pack namespace, content_hash on modules; idempotent re-ingest.
- B5: last_seen_at + per-(project, language) sweep clock; no deletion.
- B6: unresolved specifiers recorded on the source entity; synchronous re-resolve pass ends every call — the only code path that materializes `depends_on` edges, so intra-call and cross-call resolution cannot drift.
- B7: defaults to a workspace-local map database beside the ingest root (per the ADR); the shared production database is rejected outright.
- Edge identity matches the store's natural key (namespace, source, target, relation); multiple provenances merge into a `dependency_kinds` array (a real last-write-wins overwrite bug was caught by the order-independence test and fixed this way).

## Tests

- New integration tests: two-package fixture converges to an identical edge set regardless of ingest order; re-ingest is idempotent; invalid path fails loud.
- `cargo test -p khive-pack-code`: 45 tests green (18 unit + 24 pre-existing + 3 new). Clippy -D warnings clean, fmt clean.
- Self-run over this repository's own crates subtree: 39 projects, 581 modules, 996 edges created, 423 specifiers resolved.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,12 +15,12 @@ khive gives your agent:
 9. **Brain** — Bayesian profile tuning from feedback signals
 10. **Session** — persist and resume agent-session records
 
-All 11 packs load by default. **78 public verbs** across the packs: the `git` pack
+All 11 packs load by default. **79 public verbs** across the packs: the `git` pack
 contributes the `git.digest` verb plus the commit/issue/pull_request provenance note kinds
-and a batch ingester; the `code` pack currently contributes zero verbs: its `finding`
-note kind is written only through the `kkernel code-ingest` admin CLI path, never a verb
-an agent can call (ADR-085 D1, Amendment 3; Amendment 2's `code.ingest` verb is accepted
-but unimplemented); the `workspace` pack also contributes zero verbs, adding only the
+and a batch ingester; the `code` pack contributes one verb, `code.ingest` (L1 manifest +
+L1.5 import-scan source ingestion into a dedicated map database, ADR-085 Amendment 2);
+its `finding` note kind is still written only through the `kkernel code-ingest` admin CLI
+path (ADR-085 D1, Amendment 3); the `workspace` pack contributes zero verbs, adding only the
 `workspace` entity kind and `contains` endpoint rules to git/gtd/session notes (#873).
 Regenerate via `request(ops="verbs()")`
 before editing this line.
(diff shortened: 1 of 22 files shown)


oceanwaves630 (approved):
Full-diff read at head 49e3426a. Approving — the verb-surface change and its production-database fence are sound, and every published surface is consistent.

Verb surface (78 → 79): `code.ingest` is registered as the code pack's one verb (`CODE_HANDLERS` length 1, dispatch arm + fallthrough error, pack test pins `HANDLERS[0].name`). `findings.json` stays admin-CLI-only, unchanged. The count is updated consistently across AGENTS.md, CLAUDE.md, README (table + pack table + status), api-reference (table + `18+5+5+15+7+4+19+4+1+1+0 = 79` arithmetic + a new per-verb section), getting-started, specialized-packs, marketplace, and npm. The marketplace README also corrects a latent "all ten"→"all eleven" pack miscount. `smoke_test.py` asserts both `total == 79` and `"code.ingest" in verb_names`, so a silent surface drift trips the tripwire.

Production-database fence (r1 finding): genuinely fixed. `resolve_target_db` rejects both the well-known default production database location and `runtime_db_path` — the database the calling runtime was actually constructed against — so a non-default `--db`/`KHIVE_DB` production location is covered too. `normalize()` canonicalizes the deepest existing ancestor and re-appends the not-yet-created suffix, so a symlinked-parent alias or a not-yet-created target still compares equal; there is no override. Five dedicated tests pin default-local, explicit-prod-rejected, dedicated-accepted, non-default-configured-rejected, and symlink-alias-rejected-before-file-exists. The handler opens a separate `KhiveRuntime` bound to the resolved target database, so the write path never touches the pack's own production runtime.

Other r1 findings resolved with regressions: Rust item-import resolution (`module_candidate_specifiers` walks shorter prefixes so `use crate::foo::Thing` resolves to module `foo`) — test `rust_item_import_resolves_to_containing_module_after_reingest`; manifestless-folder skip (L1.5 scan is file-discovery-driven, independent of manifest discovery, basename fallback) — test `manifestless_rust_folder_uses_basename_fallback`. Idempotency/convergence (uuid5 identity + the `(namespace, source_id, target_id, relation)` edge natural key) are pinned by the two-order convergence and re-ingest tests.

ADR-085 Amendment 2 status is updated to shipped (L1 + L1.5 only; L2 Scanner/Extractor remains out of scope), and Amendment 3's zero-verb note is corrected to reflect the verb landing. Scope is limited to the code pack plus doc/count surfaces; no CI gate weakened.

PR #1041, opened 2026-07-15
feat(pack-kg): decision precedes decision endpoint (ADR-087 A9)

ohdearquant:
## Summary

- Adds a `decision`->`decision` note-to-note `EDGE_RULES` entry for the `precedes` relation, extending the base entity-only `precedes` contract (ADR-002) additively via the pack-extensible edge endpoints mechanism (ADR-017 §"Pack-extensible edge endpoints"), following the same pattern as the GTD pack's `task`->`task` `depends_on` rule.
- Per ADR-087 Amendment 1 §A9, this lets review-round chains be recorded as `decision(round N-1) precedes decision(round N)` edges. `precedes` is used deliberately, not `supersedes`: earlier rounds stay authoritative for their own commit.

## Test plan

- [x] `link(decision, decision, precedes)` succeeds
- [x] `link(decision, observation, precedes)` still rejected
- [x] `link(observation, decision, precedes)` rejected
- [x] `link` entity->entity `precedes` unaffected (base ADR-002 contract)
- [x] `cargo test -p khive-pack-kg` (148 passed)
- [x] `cargo test -p khive-runtime` (855 passed)

### DIFF
--- a/crates/khive-pack-kg/src/handlers/tests.rs
+++ b/crates/khive-pack-kg/src/handlers/tests.rs
@@ -1477,6 +1477,29 @@ async fn create_bulk_items_rejects_top_level_embedding_content() {
     );
 }
 
+// Shared by the `decision precedes decision` tests below: builds the KG
+// registry and installs its composed edge rules (the same rule set the
+// production MCP server runs under), so negative cases prove the
+// `decision`->`decision` exception stays kind-specific rather than merely
+// exercising the base entity-only rule against an unconfigured runtime.
+async fn configured_kg_pack() -> (
+    khive_runtime::KhiveRuntime,
+    khive_runtime::NamespaceToken,
+    crate::KgPack,
+) {
+    use crate::KgPack;
+    use khive_runtime::VerbRegistryBuilder;
+
+    let rt = khive_runtime::KhiveRuntime::memory().expect("in-memory runtime");
+    let mut builder = VerbRegistryBuilder::new();
+    builder.register(KgPack::new(rt.clone()));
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Full-diff read at head 5fa7a4e6. Approving — this implements ADR-087 Amendment 1 §A9 exactly, and the r1→r2 fix makes the tests actually prove what they claim.

Endpoint rule (pack.rs): `KG_EDGE_RULES` grows 9→10 with one new rule — `Precedes` / `NoteOfKind("decision")` → `NoteOfKind("decision")`. This is an additive pack extension of the base entity-only `precedes` contract (ADR-002), the same shape as the GTD pack's `task`→`task` `depends_on` — packs extend, never tighten. The doc comment disambiguates it cleanly from `supersedes` (earlier rounds stay authoritative for their own commit). Array-size bump matches the one added rule.

Tests (the r1 Medium fix): `configured_kg_pack()` installs the composed edge-rule set and routes all four tests through `handle_link`, the production-representative path. This directly resolves the r1 finding that the negative tests previously ran against an unconfigured runtime where the base entity-only rule would reject any note pair regardless. Now the matrix proves kind-specificity: decision→decision accepted; decision→observation and observation→decision both rejected; entity→entity (project→project) still accepted under the composed set, so the base contract is preserved. No weakened assertions.

CI: macos-latest ran the full suite green; ubuntu-latest and the CI gate finishing. Scope is limited to the endpoint rule plus its tests; no other file touched, no gate weakened.

Issue #1056, opened 2026-07-16
Write secret-gate: 68/653 false positives on legitimate markdown ingest (high-entropy-near-trigger-word)


ohdearquant:
## Evidence (first production batch-ingest, PR #1049 lane)

Ingesting 653 review-artifact markdown files through the generic note `create` verb, the
write gate rejected 68 (10.4%) with `write blocked: content matches secret pattern
high-entropy-token`. Spot-checks show the trigger is legitimate technical text: long
hashes/paths/URLs sitting near words like `token`, `key`, `secret`, `auth` — the same
false-positive family that #977 fixed for the pack-git ingest path via masking
(locked in by #996's tests).

None of the 68 contain real credentials (they are code-review verdict documents that
discuss auth/token-handling code, which is exactly when high-entropy strings and trigger
words co-occur).

## Ask

Extend the #977-style masking-aware evaluation to the kg note `create` path (or provide a
gate-side allowance the ingest caller can request that still blocks true `KEY=<value>`
assignment shapes). The batch ingester deliberately does NOT reword content to dodge the
gate — content fidelity is the point of ingest — so gate-side handling is the only clean fix.

## Repro

Ingest a markdown file containing a 50+ char hex/base64-looking string within a few tokens
of the word `token`. Observed on `create(kind="reference", content=...)`.

The 68 affected file paths are recorded in the ingest lane workspace
(`BLOCKED_SECRET_GATE.txt`); they remain un-ingested pending this fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ohdearquant:
Additional false-positive corpus example (routed from another repository's backfill run, kept for the design-level fix):

A code-review document quoting a grep command line was blocked by the high-entropy-token detector through three escalating mask passes. The accepted candidate was a 119-char regex alternation of Rust logging macro names plus a `token=` literal — spanning parens, pipes, bangs, and backslashes, a shape no credential uses.

Corpus-relevant properties:
1. Masking every identifier-charset run of 16+ chars in the document does NOT clear it — the candidate boundary is wider than identifier charsets.
2. It survives after all hex/base64/known-prefix shapes are masked.
3. Trivially reproducible: any message quoting a regex alternation containing `token=` near the detector's trigger context gets blocked.

[...]

Issue #1040, opened 2026-07-15
secret-gate: high-entropy FP on repo file-path tokens containing trigger substrings (e.g. .../C-ADR-007-authorization-server.md:70)


ohdearquant:
Follow-up to the earlier path-FP hardening (#632/#577). Field evidence from bulk note ingestion: markdown documents that cite repo paths like `docs/adr/C-ADR-007-authorization-server.md:119-122` are rejected by the write gate with `high-entropy-token` because the path token itself contains the trigger substring (`auth`), and long path+line-range citations exceed the entropy length threshold. A document citing such a path 30+ times cannot be stored at all — masking each token destroys the citations (they are the evidence content).

Proposed fix: exempt tokens that are structurally file paths from high-entropy classification — e.g. tokens containing two or more `/` separators and matching a path shape (`[\w.-]+(/[\w.-]+)+(:\d+(-\d+)?)?`), regardless of embedded trigger substrings. Trigger-adjacency checks should apply to the token's context, not substrings inside a path.

Reproduction: create a note whose content includes 'Evidence: `docs/adr/C-ADR-007-authorization-server.md:70` says ...' repeated with varying line numbers; create is blocked with 'matches secret pattern high-entropy-token at masked excerpt docs/a...50chars'.

ohdearquant:
## Measure-first result: the proposed path-token allowlist is unsound. Recommend a per-run entropy cap instead.

Root cause confirmed. The Layer-2 entropy heuristic already has an `is_structured_identifier` exemption, but it is gated to non-trigger context only (`!near_trigger`). A repo path cited near a credential-trigger word (key/auth/secret) in the surrounding prose falls through to the whole-token Shannon-entropy check and blocks as `high-entropy-token`. That is the module's deliberate accepted-FP tradeoff, not a missing feature.

I built a harness over the 5 real false-positive notes plus adversarial plants and measured **both directions**.

### False-positive direction (the 5 real FPs)
All 5 block via `high-entropy-token`; all 5 blocking tokens satisfy `is_structured_identifier` (every `/`,`-`,`_`,`.`-delimited run is word-shaped); `near_trigger` always comes from surrounding prose, never from a trigger glued into the path. So relaxing the exemption to apply near a trigger would fix all 5.

### False-negative direction (adversarial) — this is why the naive fix is unsound
[...]

PR #1062, opened 2026-07-16
fix: batch B1 — ingest/secret-guard correctness (#977 #1042 #1044 #1045)

### LINKED ISSUE #977: secret-gate: mask instead of write-block on ingest paths (uuid-near-trigger FP on legitimate issue content)

### LINKED ISSUE #1042: code.ingest: harden production-DB fence when config db_path is unresolved
Follow-up hardening from the code.ingest L1/L1.5 landing (#1039).

The production-database fence in `resolve_target_db` rejects both the well-known default graph path and the calling runtime's configured `db_path`, with `normalize()` catching symlink-parent aliases and not-yet-created targets. This fully covers the production daemon, which always populates a resolved `db_path`.

One theoretical gap remains for defense-in-depth: if `config().db_path` is `None` and a non-default `KHIVE_DB` is supplied, the runtime-path arm of the fence has nothing to compare against. This state is not reachable by the production daemon (it resolves a concrete `db_path` at startup), so there is no live exposure — filing only so the fence is explicitly total rather than total-in-practice.

Suggested: resolve the effective target DB (env KHIVE_DB then config then default) once and fence against that resolved value, so the check holds even when config db_path is None. Low priority; no production path is currently affected.

### LINKED ISSUE #1044: secret-gate: whole-token-average entropy dilution lets a filler-padded secret through (false-negative)
## Summary
The Layer-2 `high-entropy-token` detector computes Shannon entropy over the **whole** whitespace-delimited token (average bits/char). A genuine credential embedded as one segment among enough low-entropy filler path segments dilutes that average below the 4.5 threshold, so the token is not blocked even when it sits near a credential-trigger word.

This is a **false-negative** (a real secret passes), distinct and more serious than the false-positive addressed in the sibling path-token issue. It is independent of the `is_structured_identifier` exemption: for these shapes `is_structured_identifier` is already `false`, so the exemption never applies. The gap is in the plain entropy fallback's use of a whole-token average.

## Reproduction (both pass the gate today, unchanged)
- `api key vault/9f8e7d6c5b4a39281706f5e4d3c2b1a09f8e7d6c/rotate.md` — a real 40-char hex credential as one segment; `hex-credential-token` requires the *whole* token to be pure hex, and `high-entropy-token` averages the whole token below 4.5.
- `secret path a/b/c/d/<random-run>/e/f.rs` — a chopped/padded secret among short filler segments; whole-token average diluted below threshold.

## Fix direction
Move from a whole-token average to a **per-run / per-segment** entropy check in the fallback (flag if any single run >= min length and >= entropy threshold), so filler segments cannot dilute a high-entropy payload segment. Coordinate with the sibling issue's per-run entropy work — same per-run primitive.

## Provenance
Surfaced during the measure-first analysis of the sibling path-token false-positive issue; filed separately because it is a different direction (false-negative) and a different code path (fallback averaging, not the exemption).

### LINKED ISSUE #1045: git.digest returns a commit count lower than the notes actually inserted when the daemon restarts mid-digest
## Summary
When `git.digest` runs over a large repository and the backing daemon restarts mid-run (for example after a dropped response under load), the cumulative commit count returned by the verb is lower than the number of commit notes actually written to the database. The digest itself is complete and correct; only the returned count under-reports.

## Observed
Digesting a repository with 4,374 reachable commits, across a run where the daemon was respawned twice mid-digest:
- Returned cumulative count: 3,509
- Commit notes actually present in the database: 4,374
- `git rev-list --count <HEAD>`: 4,374

The database matches the true ancestor count exactly. Every commit was inserted (the INSERT OR IGNORE dedup keeps this idempotent across the retry). Only the count reported back by the verb was low. This reproduced across several large-repo digests: whenever the daemon was respawned mid-run, the returned count fell short of the row count by the number of commits processed in the dropped-response window.

## Impact
A caller that trusts the returned count as a completeness signal would wrongly conclude the digest is incomplete, and either re-run unnecessarily or report a false gap. The returned count is not a safe completeness indicator under daemon churn.

## Workaround
Verify digest completeness against `git rev-list --count <ref>` (the true reachable-commit count) rather than the value returned by the verb.

## Suggested direction
Track the cumulative insert count so it survives a daemon respawn, or derive the reported count from the row count after the walk completes, so the returned value reflects what was actually written.

ohdearquant:
## Summary

Fixes four of the six issues in this batch. Two (#1040, #1056) are documented below as **not fixed** in this batch, with the reasoning for dropping them.

### #977 — secret-gate mask-instead-of-block on git ingest paths (already fixed, tests added)

Investigation traced this to already being fixed: `git.digest`'s issue ingester masks `title` and `body` independently via `secret_gate::mask_secrets` before any note is created, so a credential-like word in one field cannot arm a UUID in the other field, and a same-field co-occurrence is redacted in place rather than write-blocking the note. Both scenarios were verified to already pass. This PR adds the regression tests that close the loop (cherry-picked from a prior test-only branch) so the behavior cannot silently regress.

### #1042 — code.ingest production-DB fence gap when config db_path is unresolved

`resolve_target_db` fenced the well-known default production path and the calling runtime's resolved `config().db_path`, but had nothing to compare against when `config().db_path` is `None` and a non-default `KHIVE_DB` is set. Not reachable by the production daemon today (it always resolves a concrete `db_path` at startup), so this was defense-in-depth, not a live exposure. Fixed by falling back to reading `KHIVE_DB` directly in that branch.

### #1044 — secret-gate whole-token-average entropy dilution (false negative)

The Layer-2 `high-entropy-token` detector computed Shannon entropy over the **whole** whitespace-delimited token. A genuine credential embedded as one segment among enough low-entropy filler path segments (e.g. a 40-char hex key as one `/`-delimited segment in `vault/<hex>/rotate.md`) diluted the average below the 4.5 bits/char threshold, letting a real secret through even near a credential trigger word.

Fix: decompose the token on every non-alphanumeric separator (the same run split `is_structured_identifier` already uses) and re-run the existing hex-credential-length and entropy checks against each individual run (`>= MIN_ENTROPY_LEN` chars) independently of its surrounding filler. Scoped strictly to `near_trigger` context, so it never changes `!near_trigger` behavior and does not touch `is_structured_identifier`'s own `near_trigger` gate.

**Test evidence:** 162 `secret_gate` tests pass (158 prior + 4 new: two reproduce the reported dilution false negatives, one is a regression guard confirming the real accepted-FP paths near trigger words are unaffected because none of them contain a single run `>= MIN_ENTROPY_LEN`).

### #1045 — git.digest commit count under-reports across daemon restarts

`commits_ingested` is an in-memory per-pass delta. When the backing daemon respawns mid-digest and drops a response after the write already landed, that pass's `commits_ingested` is lost from any cumulative sum a caller keeps across calls — even though the commit note was durably written (`INSERT OR IGNORE` dedup keeps the write itself idempotent). A caller trusting the summed delta as a completeness signal under-reports.

Fix: add `IngestReport.commits_total_in_db`, a `COUNT(*)` over commit notes annotating the project, queried fresh from the database after every pass regardless of which branch of the ingest ran. This is derived state, not an in-memory counter, so it survives daemon churn and is safe to compare directly against an independent source of truth (e.g. `git rev-list --count <ref>`).

**Test evidence:** extended the existing multi-pass `digest_verb_auto_creates_project_and_enriches_references` acceptance test to assert `commits_total_in_db` stays correct across all three passes, including the pass where `commits_ingested` is 0 because the commit was already durably written in an earlier pass.

### Dropped from this batch: #1040 and #1056

Both report the same underlying gap: the `is_structured_identifier` exemption's `!near_trigger` gate causes real file-path false positives near a credential trigger word (repo paths cited in review documents, workspace-artifact ingest of markdown files, etc.).

A prior investigation (measure-first methodology against a real false-positive corpus, followed by an implementation attempt) established that the straightforward fix — dropping or loosening the `near_trigger` gate on `is_structured_identifier`, including a per-run entropy-cap variant — is **unsound**: for a run of length `N`, maximum Shannon entropy is `log2(N)`, which is content-independent for any run `<= 16` chars. A per-run cap tuned to admit the real false positives necessarily reopens the 12 adversarial run-splitting/padding/assignment-disguise regression tests that were specifically engineered to defeat exactly this class of signal (already present in the test suite, guarding against #387-era bypasses). The module's own doc comment documents this impossibility.

The sound direction — scoping an exemption to first-party ingestion by provenance and path-shape rather than by content-shape alone — is a trust-boundary/ADR-level design decision (now ADR-115), not a mechanical fix within a batch PR's scope. Per this batch's own instructions, dropping these two issues rather than shipping an unsound loosening or a stub.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo test -p khive-runtime -p khive-pack-git -p khive-pack-code -p kkernel` — all green
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-runtime -p khive-pack-git -p khive-pack-code` — clean

Closes #977, Closes #1042, Closes #1044, Closes #1045

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

## Fix rounds

Rounds 1–3 closed a family of separator-split hex-credential bypasses (ASCII and Unicode separators, multi-fragment splits, base64-shaped reconstruction) and regenerated the corpus manifest with per-detector SHA-256 lists. Round 3 review surfaced a further residual on the fragment-reconstruction bridge; rounds 4–5 resolved it as documented below.

## Accepted residual (secret-gate fragment bridge)

The separator-split credential bridge reconstructs up to `MAX_BRIDGE_FRAGMENTS` (6) real hex/alnum fragments joined across delimiter-only gaps (each gap spanning at most `MAX_BRIDGE_GLUE_TOKENS` (6) glue tokens per direction). Round-3 review found a genuine internal inconsistency — a punctuation-only token between two Unicode gaps stopped the walk even though the gap logic accepts it — which is **fixed** (the fragments on either side now join and the credential is blocked; regression `blocks_punctuation_glue_split_hex_credential`).

Three cases remain **accepted residual limitations**, not defects: a credential split into more than 6 real fragments, into fragments individually below `MIN_BRIDGE_FRAGMENT_LEN` (8 chars), or across more than 6 delimiter-only tokens in one gap. Rationale (per ADR-096 / ADR-115): the secret gate is hygiene against *accidental* credential persistence on a single-principal same-uid host, not defense against a same-uid adversary deliberately hand-splitting a credential to evade it — that adversary can write the store directly. The caps exist for false-positive safety, which is the point of this PR. The limitation is pinned by `allows_seven_way_hex_split_beyond_fragment_cap_documented_limitation` and `allows_six_way_sub_floor_hex_split_documented_limitation`, whose comments cite the threat model as a decided boundary.

### Gates (final)

- `cargo fmt -p khive-runtime -- --check` — clean
- `cargo clippy -p khive-runtime --all-targets -- -D warnings` — clean
- `cargo test -p khive-runtime` — all green (872 lib passed, secret_gate suite incl. the new residual + PoC-3 regression tests)


ohdearquant:
Independent review reached APPROVE (0 findings) at head 55497d13844be6d814c39e5576f1afbe634e0356. Review history: rounds 1–3 closed a family of separator-split hex-credential bypasses (each round found a real one — ASCII/Unicode separators, multi-fragment splits, base64-shaped reconstruction); round 3 surfaced a genuine internal inconsistency (punctuation-only glue between two Unicode gaps stopped the bridge walk) which is fixed, plus two evasion cases accepted as documented residual risk under the ADR-096/ADR-115 threat model; round 4 was a single comment-accuracy Medium (the guarantee comment overclaimed the new glue-token cap); round 5 confirms the corrected comment. The accepted-residual boundary is now stated in the PR body and pinned by named documented-limitation tests. Final gates green (fmt, clippy, 872 lib tests).

oceanwaves630 (approved):
Reviewed at head 55497d138: five review rounds each surfaced a real defect; the fragment-bridge residuals are documented with their threat-model rationale in both the limitation tests and the PR body.

PR #1065, opened 2026-07-17
docs(adr): ADR-115 exact-content manifest exemption for the secret gate

ohdearquant:
## Summary

Adds ADR-115, proposing a runtime-owned, exact-content SHA-256 manifest exemption inside
`secret_gate` for the false-positive class documented in #1040 and #1056. Content-shape
heuristics are structurally exhausted (per-run entropy caps reopen 12 adversarial regressions;
the max entropy of an N-char run is log2(N)), so the only remaining sound axis is exact-content
matching against a human-adjudicated, operator-curated manifest — under explicit single-principal
host trust (ADR-096), never sold as provenance authentication or a security control against a
same-uid adversary.

- Layer: inside `secret_gate` (ADR-018 Gate seam rejected — no content access, fail-open semantics
  wrong for this carve-out).
- Trust model: host trust (ADR-096) is the claim; runtime-recomputed digest match is a precision
  instrument only.
- Granularity: exact per-record allowlist over the frozen 548-record corpus.
- Marking: runtime-applied, caller-unforgeable `properties["khive:secret_gate"]` stamp + one
  queryable audit event per exemption; annotate-by-default consumer views (no default hiding).
- Fail-closed unconditionally on every manifest/stamp/audit error.

Closes the update/laundering Catch-22, canonicalization-drift, and content-vs-record-scope gaps
surfaced by an adversarial review pass on the design memo before this draft landed (see PR
conversation for detail).

## Test plan

- [ ] Maintainer spec-review confirms the exact-claim boundary, 548/548 acceptance bar, reserved-
      property invariant, staging-mutation suite, and one-record-one-event audit invariant are
      normative.
- [ ] `deno fmt` clean (already verified locally).
- [ ] Docs-only PR — no code/cargo changes in this diff.
- [ ] Do not mark ready or enable auto-merge until spec sign-off lands.

### DIFF
--- a/docs/adr/ADR-115-secret-gate-content-manifest-exemption.md
+++ b/docs/adr/ADR-115-secret-gate-content-manifest-exemption.md
@@ -132,7 +132,11 @@ SHA256("khive-secret-gate-v1\0" || runtime_field_scope || "\0" || exact_value_pa
 minimum: record content, name/description, JSON properties, tags, and code source. An exemption
 computed for one field scope never applies to another; a value that is byte-identical between two
 scopes still requires two independent manifest entries. The frozen 548-record corpus uses the record
-content scope.
+content scope. For the JSON-properties scope specifically, the current scanner examines keys and
+values as independent strings (`scan_json_value` applies no key-conditioned value heuristics), so a
+per-string manifest entry is exactly as granular as the detector it overrides; any future detector
+that becomes key-aware — scoring a value differently because of the key it sits under — breaks that
+equivalence and must revisit this exemption's per-string lookup in the same change.
 
 Caller-supplied `properties.content_sha256_16`, `source_path`, `namespace`, `actor`, and `verb` play
 no role in eligibility. Only the runtime-recomputed full digest and runtime-assigned field scope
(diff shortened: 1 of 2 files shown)


ohdearquant:
Independent review, round 1: REJECT (1 Blocker: intent-based read-modify-write carve-out not observable at the write boundary; 1 High: shared write-finalization boundary described as existing when the gate is invoked at three separate sites today). Both verified against source and folded: the RMW rule is now an observable byte-comparison wire contract (exact echo stripped before diff/merge; any other caller-side appearance rejected), and the finalization boundary is stated as a normative implementation requirement naming the current call sites. The spec-gate reviewer independently required the per-string JSON-scanner note, also folded.

Independent review, round 2 (focused re-review at e2100a43bca62b57116282c0c5df67045f57de29): APPROVE, 0 findings — confirms the wire rule leaves no caller-intent branch to implement, the finalization requirement is concrete enough to prevent subset-patching, and the scanner note does not overclaim.

Final commit flips Status to Accepted per spec-gate sign-off; no other change above the reviewed head.

oceanwaves630 (approved):
Reviewed at head 1a18473e0: ADR-115 accepted — exact-content SHA-256 manifest exemption with normative acceptance criteria (548-of-548, reserved-property invariant across the full write inventory, staging mutation suite, one-record-one-event audit). Sole commit since the review round is the status flip to Accepted.

PR #1048, opened 2026-07-16
feat(channel-telegram): Telegram Bot API adapter + daemon wiring (ADR-056 amendment, #113)

ohdearquant:
Implements the ADR-056 Amendment 2026-07-05 (Telegram adapter implementation and two-way chat): a new `khive-channel-telegram` crate implementing the `Channel` trait over the Telegram Bot API, plus `channel-telegram`-feature-gated daemon wiring (`spawn_telegram_channel_loops` with a poll loop and an outbox loop), mirroring the shipped email adapter.

Closes #113.

## What's in

- **`crates/khive-channel-telegram`** (new): `TelegramChannel` with `send` via `sendMessage` and `poll` via `getUpdates`; in-memory offset watermark per the amendment (restart durability via Telegram's server-side offset semantics + the `external_id` dedup index); connector-trait seam (`TelegramConnector`) so all 24 unit tests run against a mock — no live network in tests; config from env only (`KHIVE_TELEGRAM_BOT_TOKEN`, `KHIVE_TELEGRAM_MAINTAINER_CHAT_ID`), token masked in `Debug` output, fail-closed construction when unset.
- **Trust model**: inbound updates from any chat id other than the configured maintainer are dropped (batch-isolated — one unauthorized update never aborts the batch, and the offset still advances past it so an unauthorized sender cannot wedge the poll loop). Outbound to a non-maintainer slug is logged and dropped, never misdelivered.
- **`khive-mcp` wiring**: feature-gated poll + outbox loops (5s cadence) dispatching `comm.ingest`/`comm.send` through the existing verb surface — two-way chat with no comm API change, per the amendment. Four channel-agnostic helpers previously gated on `channel-email` relaxed to `any(channel-email, channel-telegram)`; both adapters compile and coexist (`ChannelRegistry` keys by `(kind, slug)`).

## Deliberately out (v1 scope per the amendment)

Media/non-text updates, group chats, webhook inbound (long-poll only), channel-health heartbeat coverage (follow-up if wanted).

## Gates

`cargo check/clippy -D warnings/test -p khive-channel-telegram` (24/24), `cargo check -p khive-mcp --features channel-telegram` and `--features channel-telegram,channel-email`, `cargo check --workspace`, fmt clean, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-channel-telegram` clean.

[...]

PR #1068, opened 2026-07-17
fix(release): wire khive-channel-telegram into publish ladder + refresh stale semver excludes

ohdearquant:
## Problem

`make publish-dry` for the 0.5.0 release failed — **not** a semver break — with `khive-channel-telegram not found in registry (crates.io)`.

`khive-channel-telegram` (added in #1048) is version-depended by `khive-mcp` (`version = "0.5.0"`, optional), so it must publish in the 0.5.0 cycle. It was absent from **both** the CRATES publish list and the `cargo semver-checks` exclude set in `scripts/publish.sh`, so the `--workspace` semver gate failed trying to fetch a baseline for a never-published crate. Its sibling `khive-channel-email` was wired correctly; telegram was missed when #1048 landed.

## Fix

- Add `khive-channel-telegram` to CRATES after `khive-channel-email`, before `khive-mcp` — its only khive dependency (`khive-channel`) is already above it, so live-publish order is correct.
- Narrow the semver-checks excludes to `khive-channel-telegram` only (first publish, no crates.io baseline to diff).
- Drop the four now-published 0.4.0 excludes (`khive-changeset`, `khive-pack-code`, `khive-pack-git`, `khive-pack-workspace`) per the script's own drop-once-published rule — they have 0.4.0 baselines and are checked again this cycle.
- Correct the stale "never published" comments on those entries.

## Verification

`make publish-dry` returns rc=0 from this branch (semver gate passes, all crates package + workspace check clean).

## Follow-up

The root cause is that adding a workspace crate has no guard forcing the `publish.sh` wiring. Filing a separate issue for an in-script guard (fail loud if any publishable workspace member is absent from CRATES) rather than scope-creeping this release-gate fix.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -62,12 +62,12 @@ CRATES=(
     khive-retrieval
     khive-vcs-adapters
     khive-vcs
-    khive-changeset      # needs khive-types (above); never published, no crates.io baseline yet
+    khive-changeset      # needs khive-types (above)
     # khive-merge — excluded from workspace (ADR-043 forward-deployed, ahead of khive-vcs)
     khive-pack-formal    # needs khive-runtime + khive-types (both above); dev-dep of khive-pack-kg, so publish first
     khive-pack-kg
-    khive-pack-git       # needs khive-runtime/storage + khive-pack-kg (all above); never published, no baseline yet
-    khive-pack-code      # needs khive-runtime/storage + khive-pack-kg (all above); never published, no baseline yet
+    khive-pack-git       # needs khive-runtime/storage + khive-pack-kg (all above)
+    khive-pack-code      # needs khive-runtime/storage + khive-pack-kg (all above)
     khive-pack-gtd
     khive-brain-core
     khive-pack-brain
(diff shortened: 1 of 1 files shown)


ohdearquant:
Independent review: APPROVE, 0 findings at all severities (reviewed head d30f5216a35f51ce8ed4c48f0dcf880ae7c22bbd). Verified: khive-channel-telegram's only khive dependency is khive-channel (already earlier in CRATES), so live-publish order is correct; all four dropped excludes are published at 0.4.0 (checking them again is right) and khive-channel-telegram is the only never-published workspace member the --workspace gate touches (excluding it is right and complete); every publishable workspace member is present in CRATES; khive-merge is genuinely excluded from the workspace via crates/Cargo.toml, not silently dropped. `make publish-dry` returns rc=0 from this branch.

One pre-existing nit outside this diff: the khive-merge parenthetical at scripts/publish.sh:66 cites the wrong ADR number (should be ADR-039, not ADR-043) — noted for a future cleanup, no effect on the publish set.

oceanwaves630 (approved):
Reviewed at head d30f5216a: full diff read — telegram slotted in correct dependency order, semver excludes reduced to the one genuinely never-published crate, the four 0.4.0 crates return to checking per the script's own drop-once-published rule, comments now match reality. Dry-run green from the branch.

Issue #1069, opened 2026-07-17
publish.sh: guard against a publishable workspace crate missing from the CRATES ladder


ohdearquant:
## Motivation

PR #1068 fixed a release blocker where `khive-channel-telegram` (added in #1048, version-depended by `khive-mcp`) was never wired into `scripts/publish.sh` — missing from both the CRATES publish list and the semver-checks excludes. The gap was invisible until `make publish-dry` failed at release time. This is a recurring shape: adding a workspace crate has no guard forcing the publish-ladder wiring, and the same class of miss can happen on any future crate.

## Proposed guard

Add a preflight check in `publish.sh` (runs in `make publish-dry` and live alike) that fails loud if any workspace member with `publish != false` is absent from the CRATES array.

Sketch:
- Enumerate publishable members from `cargo metadata --no-deps --format-version=1` (filter `publish` field: a member is publishable unless `publish == []`/`false`).
- Diff that set against the CRATES array.
- If any publishable member is missing from CRATES, print the missing names and exit non-zero.

## Must handle

- `khive-merge` is intentionally excluded from the workspace (ADR-043 forward-deployed) — it should not be a workspace member, so `cargo metadata --no-deps` will not list it; confirm this rather than special-casing.
- Any `publish = false` members (test/xtask-style crates) must be treated as legitimately out-of-ladder, not flagged.
- The guard must not itself become a flaky release blocker — a false positive that halts a legitimate publish is worse than the gap it prevents, so it needs its own small unit/fixture check before it gates real publishes.

## Not doing here

Deliberately filed as a follow-up rather than folded into the #1068 release-gate fix, to keep that critical-path change minimal.

PR #1070, opened 2026-07-17
fix(kkernel): cover ingested non-crate projects in default code-audit policy (#1067)

### LINKED ISSUE #1067: code-audit: default policy crate_ranks omits 10 workspace crates — layering signal degrades to Unavailable (PolicyIncomplete)
Found by running the shipped `kkernel code-audit` (PR #1052, ADR-114) against the repo's own code map with the default policy.

## Problem

The default policy's `crate_ranks` omits 10 crates that code.ingest DOES ingest, so the report flags **PolicyIncomplete** and every layering signal for those crates degrades to `Unavailable`:

- the 7 npm workspace packages: `@khive-ai/cli`, `@khive-ai/kernel-darwin-arm64`, `@khive-ai/kernel-darwin-x64`, `@khive-ai/kernel-linux-arm64`, `@khive-ai/kernel-linux-x64-gnu`, `@khive-ai/kernel-linux-x64-musl`, `@khive-ai/kernel-win32-x64`
- `khive`, `khive-contract`, `khive-workspace`

`ingest_coverage` is Observed=1 for all 10 (they are in the map), but `layering_violation` is Unavailable=0 for each — the shipped default policy cannot see layering violations across a real slice of the ingested map.

## Fix

Small: extend the default policy's `crate_ranks` to cover the full ingested workspace (assign ranks for the npm packages and the three unranked Rust crates), or explicitly document them as rank-exempt if layering genuinely does not apply (the platform-specific kernel-* packages are leaf artifacts, so an explicit exempt list may be the honest shape rather than a fake rank). Either way the report should stop flagging PolicyIncomplete on a clean default run against our own repo.

Evidence: self-audit report generated 2026-07-17 against a read-only copy of the repo's own code map (policy_sha256 eb5ce23a0417e6f714d165bc8edd38fec2072b776648e4dd9e36669f351d11d4).

ohdearquant:
## Summary

The default `kkernel code-audit` policy's `crate_ranks` table omitted the 10 non-crate projects that `code.ingest` resolves into the map, so a clean run against the repo's own code map flagged **POLICY_INCOMPLETE** and degraded every layering signal for those 10 to `unavailable` (found by running the shipped audit — #1067).

## What was missing

Exactly 10 `project` entities in the map, none under `crates/`:
- 6 npm platform packages `@khive-ai/kernel-{darwin-arm64,darwin-x64,linux-arm64,linux-x64-gnu,linux-x64-musl,win32-x64}`
- the `khive` npm root and `@khive-ai/cli`
- the Python projects `khive-contract` and `khive-workspace`

Confirmed the missing set is exactly these 10 by diffing all 50 `project` entities in the map against the 40 previously-ranked crates (0 stale ranks, 0 other gaps).

## The ranks are honest, not placeholders

I read the actual `depends_on` edges in the map rather than assigning fake ranks:

- The npm distribution graph is a real DAG: the six platform kernel binaries have no outgoing deps (leaves, **rank 0**); `khive` depends on all six (**rank 1**); `@khive-ai/cli` depends on `khive` (**rank 2**).
- `khive-contract` and `khive-workspace` are Python projects whose dependencies are external PyPI specifiers (pytest, pandas, …) that do not resolve to in-map projects, so they are project-graph leaves (**rank 0**).

These 10 form a dependency island with no edges to the Rust crate graph, so their ranks are evaluated in isolation and never interact with the crate ranks. The header comment is updated accordingly (it previously said "every directory under `crates/` must appear here").

## Verification (before/after against the repo's own map)

Ran `kkernel code-audit` against a copy of the repo's code map with the old and new policy:

| policy | `errors[]` | layering signals `unavailable` |
|---|---|---|
| before | HISTORY_ABSENT, **POLICY_INCOMPLETE** | 10 |
| after | HISTORY_ABSENT | **0** |

- POLICY_INCOMPLETE clears; only the unrelated `HISTORY_ABSENT` phase-1 limitation remains (churn / dead-file / orphan-test signals require history facts the phase-1 map schema does not persist).
- No new layering violation is introduced: `khive → kernel` is 1→0 and `cli → khive` is 2→1, both satisfy `rank(target) ≤ rank(source)`.

## Why no new test

Data-only change to the default policy TOML. The `code_audit.rs` tests use synthetic `[crate_ranks]` fixtures, not this file, so nothing loads the real policy in-test. The living regression guard is the weekly code-audit consumer, which self-reports POLICY_INCOMPLETE on any future gap; #1069 tracks the complementary fail-loud publish-ladder guard.

## Test plan
- [x] `check toml` pre-commit hook passes
- [x] before/after `kkernel code-audit` run against the repo map (table above)

Closes #1067

### DIFF
--- a/crates/kkernel/policy/code-audit-khive.toml
+++ b/crates/kkernel/policy/code-audit-khive.toml
@@ -13,7 +13,12 @@
 # `[dependencies]` graph (dev-dependencies excluded — several packs carry
 # deliberate cyclic dev-only test deps, e.g. khive-pack-kg/khive-pack-gtd and
 # khive-pack-brain/khive-pack-memory; production dependencies among them do
-# not cycle). Every directory under `crates/` must appear here.
+# not cycle). Every ingested `project` in the code map must appear here — the
+# Rust crates under `crates/` and every other manifest `code.ingest` resolves
+# (the npm distribution packages, the Python contract/workspace projects).
+# Those non-crate projects form a dependency island disconnected from the Rust
+# crate graph (no cross edges), so their ranks are evaluated in isolation and
+# never interact with the crate ranks above.
 policy_version = 1
 coverage_floor = 0.0
 
@@ -66,3 +71,21 @@ khive-pack-knowledge = 6
 khive-pack-memory = 6
 khive-mcp = 7
 kkernel = 8
+
+# Non-crate projects the workspace also ships and `code.ingest` resolves. The
+# npm distribution graph is a real DAG evaluated in isolation: the six
+# platform-native kernel binaries are leaves, the `khive` npm root bundles them
+# as optional platform deps (rank 1), and `@khive-ai/cli` aliases `khive`
+# (rank 2). `khive-contract` (Python contract-test harness) and
+# `khive-workspace` resolve only external PyPI specifiers, so they are leaves
+# in the project graph (rank 0).
+"@khive-ai/kernel-darwin-arm64" = 0
+"@khive-ai/kernel-darwin-x64" = 0
+"@khive-ai/kernel-linux-arm64" = 0
+"@khive-ai/kernel-linux-x64-gnu" = 0
+"@khive-ai/kernel-linux-x64-musl" = 0
+"@khive-ai/kernel-win32-x64" = 0
+khive = 1
+"@khive-ai/cli" = 2
+khive-contract = 0
+khive-workspace = 0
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Reviewed at head e989fbb65: policy-file-only change assigning honest ranks to the ten ingested non-crate projects (npm distribution island 0/1/2, Python leaves 0), with the map-isolation semantics documented in place. Reproduction evidence (PolicyIncomplete clears, ten layering signals resolve, zero new violations) matches the change's shape.

PR #1071, opened 2026-07-17
feat(release): guard publish.sh CRATES ladder against missing publishable members (#1069)

### LINKED ISSUE #1069: publish.sh: guard against a publishable workspace crate missing from the CRATES ladder

ohdearquant:
## Summary

Adds a preflight guard to `scripts/publish.sh` that fails LOUD if any publishable cargo
workspace member is absent from the `CRATES` publish ladder. This closes the recurring gap
#1068 hit at release time: `khive-channel-telegram` (added in #1048, version-depended by
`khive-mcp`) was never wired into `CRATES`, invisible until `make publish-dry` failed at the
SemVer gate. Any future crate added without ladder wiring now fails immediately, at preflight,
with the missing name printed.

## What it does

- **`scripts/lib/publish_guard.sh`** (new, sourced by publish.sh and the test):
  - `publishable_members_from_metadata` — reads `cargo metadata --no-deps --format-version=1`
    and emits each publishable member. A member is publishable unless its `publish` field is
    `[]` (how cargo serializes `publish = false`); `null` or a non-empty registry list means
    publishable.
  - `check_crates_ladder` — diffs the publishable set against the passed CRATES array; on any
    gap, prints `ERROR: publishable workspace member(s) missing ...` naming each and returns 1.
- **`scripts/publish.sh`** — sources the lib, runs the guard right before the SemVer gate (so
  it fires in both `make publish-dry` and `--live`), aborts non-zero on a gap.
- **`scripts/tests/publish-guard-test.sh`** (new) — standalone fixture test, no network / no
  live cargo: (1) a complete ladder passes and a `publish=false` member is not required; (2) an
  omitted publishable member fires the guard and is named; (3) a `publish=false` member is never
  flagged, even against an empty ladder.

## Acceptance criteria (from #1069)

- **khive-merge is not special-cased.** It is intentionally excluded from the workspace
  (ADR-043 forward-deployed), so `cargo metadata --no-deps` does not list it — confirmed by the
  real-workspace run (39 members, all present) rather than a hardcoded exception.
- **`publish = false` members are legitimately out-of-ladder** — the `!= []` filter handles
  this; fixture case 3 proves such a member is never flagged. (No `publish=false` member exists
  in the workspace today, so the fixture, not a live crate, exercises that path.)
- **Not a flaky blocker** — the guard passes cleanly against the current real workspace, and its
  own fixture test proves it both fires on a gap and passes when the ladder is complete.

## Scope

CRATES *membership* only. The separate `cargo-semver-checks --exclude` list (the other half of
the #1068 miss) is a distinct concern and deliberately out of scope here, noted in the lib
comment. Ordering logic is untouched — the guard checks set membership, not order.

## Gate output

```
publish-guard-test.sh:  case 1 PASS, case 2 PASS, case 3 PASS  (all cases passed)
real workspace:         GUARD PASS — ladder complete (39 crates)
publish.sh in-situ:     "CRATES ladder OK (39 publishable workspace members all present)",
                        proceeds into SemVer gate with no false abort
shellcheck -x:          0 findings, exit 0 on all three files
```

No Rust touched (shell + fixture only), so `cargo fmt`/`clippy` are N/A.

## Note

The lib parses `cargo metadata` with an inline `python3 -c`, matching publish.sh's existing
convention for cargo-metadata parsing (jq is on PATH but is not what this script reaches for).

Closes #1069

### DIFF
--- a/scripts/lib/publish_guard.sh
+++ b/scripts/lib/publish_guard.sh
@@ -32,9 +32,20 @@ check_crates_ladder() {
     shift
     local -a ladder=("$@")
     local -a missing=()
-    local name found c
+    local name found c members
+
+    # Capture the producer's output AND its exit status explicitly. A process
+    # substitution's exit status is not propagated to a `while` loop, so a
+    # failed parser (invalid/corrupt metadata, missing python3) would yield no
+    # names and let the guard pass silently. A publish preflight must fail
+    # CLOSED when it cannot enumerate members (#1071 review).
+    if ! members=$(publishable_members_from_metadata < "$metadata_path"); then
+        echo "ERROR: could not enumerate publishable workspace members from cargo metadata (parser failed) — refusing to proceed" >&2
+        return 2
+    fi
 
     while IFS= read -r name; do
+        [[ -n "$name" ]] || continue
         found=false
         for c in "${ladder[@]}"; do
             if [[ "$c" == "$name" ]]; then
@@ -43,7 +54,7 @@ check_crates_ladder() {
             fi
         done
         $found || missing+=("$name")
-    done < <(publishable_members_from_metadata < "$metadata_path")
+    done <<< "$members"
 
     if [[ ${#missing[@]} -gt 0 ]]; then
         echo "ERROR: publishable workspace member(s) missing from the CRATES ladder in scripts/publish.sh:" >&2
--- a/scripts/lib/publish_guard.sh
+++ b/scripts/lib/publish_guard.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# Preflight guard (#1069): every publishable cargo workspace member must
+# appear in publish.sh's CRATES ladder. A member is publishable unless its
+# Cargo.toml sets `publish = false`, which `cargo metadata` serializes as
+# an empty `publish` array (null, or a non-empty registry list, means the
+# member is publishable).
+#
+# Scoped to CRATES membership only — the separate cargo-semver-checks
+# `--exclude` list (also in publish.sh) is a different concern and is not
+# guarded here (#1069 out-of-scope).
+#
+# Sourced by scripts/publish.sh and by scripts/tests/publish-guard-test.sh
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Approved at 294a4c9ae. Read the full diff: fail-closed parser handling is correct (exit status captured explicitly, clear error), publish=false semantics match cargo metadata serialization, guard runs before the SemVer gate in preflight and live alike, and the fixture test covers the complete/missing/publish-false/malformed boundaries.

PR #1074, opened 2026-07-17
fix(retrieval): sanitize non-finite fusion weights + bm25 doc-length accounting (#1022, #1024)

### LINKED ISSUE #1022: khive-fusion: normalize_weights does not sanitize non-finite weights (inconsistent with weighted_fusion)
## Problem

`normalize_weights` in `crates/khive-fusion/src/weighted.rs` filters candidate weights with `*w > 0.0` only. A positive non-finite weight (`f64::INFINITY`) passes that filter, enters `weight_sum`, and produces `INFINITY / INFINITY = NaN` in the output vector.

`weighted_fusion` in the same module sanitizes with `is_finite()` before use, so the two similarly named public APIs disagree: the fusion path treats non-finite weights as zero while the exported normalization helper propagates them.

## Expected

`normalize_weights` should apply the same finite-positive sanitization as `weighted_fusion` (non-finite -> 0), with `INFINITY` and `NaN` regression tests.

## Notes

Surfaced during the docs/api reference pass (PR #1020 review); the doc now describes the current behavior precisely, so this issue tracks the code-side fix as a separate change.

### LINKED ISSUE #1024: khive-bm25: memory_usage and estimate_document_cost omit the usize document-length vector mirror
## Problem

`Bm25Index` maintains two document-length mirrors alongside the `u32`-keyed map: `doc_lengths_vec: Vec<usize>` and `doc_lengths_f32: Vec<f32>` (`crates/khive-bm25/src/index/core.rs`). The budget accounting in `crates/khive-bm25/src/index/memory.rs`:

- `memory_usage` counts the map and the `f32` vector but not `doc_lengths_vec` (8 bytes per slot on 64-bit).
- `estimate_document_cost` models no incremental cost for either vector mirror.

Both vectors grow with assigned internal IDs, including holes left by removals, so a configured `memory_budget` can be materially underestimated on a large index.

## Expected

Include the `doc_lengths_vec` allocation in `memory_usage` and model both document-length-vector extensions in `estimate_document_cost`, with a focused budget-accounting test.

## Notes

Surfaced during the docs/api reference pass (PR #1020 review). The reference now documents the exclusion precisely; this issue tracks the code-side fix.

ohdearquant:
Two well-scoped retrieval-crate fixes.

#1022 khive-fusion normalize_weights: filter non-finite (NaN/Inf) weights from both the weight sum and the normalization step, matching weighted_fusion. Input consisting only of non-finite weights degrades to a uniform distribution. Regression test covers Inf, NaN, and all-non-finite.

#1024 khive-bm25 memory_usage and estimate_document_cost: account for the usize doc_lengths_vec mirror, previously omitted while the f32 mirror was counted. Tests assert the estimate does not underestimate actual memory growth so a memory budget cannot be exceeded silently.

Verified locally: cargo fmt --check, cargo clippy -D warnings, and cargo test -p khive-fusion -p khive-bm25 all green.

Closes #1022
Closes #1024


oceanwaves630 (approved):
Full read at head e428589: the shared finite-positive predicate with uniform fallback on a non-finite sum resolves the overflow finding (pinned by the f64::MAX pair test), and the id-gap mirror-slot charge in estimate_document_cost resolves the under-admission finding (pinned by the isolated gap test). Docs updated consistently. Approving.

Issue #1064, opened 2026-07-17
search: hyphenated tokens in FTS queries act as NOT operators — 'ADR-016' excludes instead of matches


ohdearquant:
## Symptom

`search(kind="entity", query="ADR-016 ...")` returns false negatives: the hyphen in the token is interpreted by FTS5 as an exclusion/column-filter operator rather than matched literally, so the query *excludes* candidates instead of finding them.

Confirmed false negatives from a live ADR-coverage gap scan (2026-07-16): khive ADR-086, ADR-099, ADR-110 and lattice ADR-082 were all reported missing by hyphenated-id queries and all confirmed present on re-query with the hyphen stripped (`"ADR 016"` + title keywords). Reproduced independently: `search(kind="entity", query="ADR-086 workspace mirror")` does not surface the ADR-086 document entity itself.

Internal provenance: kg insight note `935e8982` (khive-usage-pattern board).

## Why it matters

- Only misses, never false positives — which makes it invisible in normal use and **dangerous specifically for coverage/gap audits**, where a miss reads as a genuine gap.
- Hyphenated identifiers are the norm for exactly the things agents search for: ADR ids (`ADR-016`, `C-ADR-003`), crate names (`khive-pack-kg`), job ids. Any external user querying an ADR id hits this on day one.

## Expected behavior

User-supplied query strings are data, not FTS5 syntax. The text-search layer should escape or phrase-quote query tokens before handing them to FTS5 (or use a sanitizing query builder), so `ADR-016` matches the literal token sequence. If deliberate operator syntax is ever wanted, that should be an explicit opt-in parameter, not the default parse of arbitrary agent input.

## Scope

Applies to the FTS side of hybrid search for entities and notes (`search`, `memory.recall`, `knowledge.search` — anywhere the raw query string reaches FTS5 MATCH). The vector side partially masks the failure in hybrid results, which is why it survived this long.

ohdearquant:
[...]

PR #1075, opened 2026-07-17
test(search): regression guards for unknown-kind error and hyphenated FTS ids (#1061, #1064)

### LINKED ISSUE #1061: search: unknown kind silently returns empty results while list errors loudly
## Problem

`search(kind="workspace_doc", query="...")` (an unknown kind — no such note or entity kind is registered) returns `{ok: true, result: []}`.

`list(kind="workspace_doc", limit=3)` on the same input correctly fails with `invalid input: unknown kind \"workspace_doc\"; valid: artifact | ... | workspace`.

## Why it matters

The silent-empty path masks caller typos and schema misunderstandings: an agent probing whether records of a kind exist gets an answer indistinguishable from "kind is valid but no rows match," and draws wrong conclusions (observed live: an ingest-verification pass misread ingested records as unindexed because the kind label it probed was an artifact-class alias, and search gave no signal that the kind itself was invalid).

## Expected

`search` should validate the `kind` discriminant against the registered vocabulary exactly like `list` and `create` do, and return the same per-op `{ok: false, error: "unknown kind ...; valid: ..."}` shape. Per the repo standard, invalid input is never silently coerced or swallowed.

## Repro

```
request(ops="search(kind=\"nonexistent_kind\", query=\"anything\", limit=5)")   # ok:true, []
request(ops="list(kind=\"nonexistent_kind\", limit=3)")                          # ok:false, unknown kind
```

ohdearquant:
Test-only regression coverage for two filed search-robustness issues, both already resolved on main. These tests pin the behavior against regression.

**#1061** — search rejects an unregistered kind with the same loud InvalidInput error as list (asserts the identical unknown-kind message), never a silent empty result.

**#1064** — a hyphenated identifier mixed with plain keyword terms ("ADR-086 workspace mirror") matches the target document literally rather than treating the hyphen as an FTS5 NOT operator. Covered at the khive-db TextSearch layer (Plain and AnyTerm modes, trigram tokenizer) and end-to-end through the pack-kg search dispatch path.

Verified: all three tests pass on current main (khive-db 1, pack-kg 2). No source change — the fixes already exist; this adds the missing regression guards.

Resolves #1061, #1064.

### DIFF
--- a/crates/khive-db/src/stores/text_tests.rs
+++ b/crates/khive-db/src/stores/text_tests.rs
@@ -2258,3 +2258,52 @@ async fn upsert_documents_routes_through_writer_task_when_flag_enabled() {
         "the flag-ON path must actually spawn and use the writer task"
     );
 }
+
+/// #1064 regression: a hyphenated identifier token mixed with plain keyword
+/// terms (`"ADR-086 workspace mirror"`) must still surface the document whose
+/// title/body contains that identifier under the production `trigram`
+/// tokenizer, in both `Plain` and `AnyTerm` mode.
+#[tokio::test]
+async fn test_search_hyphenated_id_with_plain_terms_matches_exact_id() {
+    let store = setup_trigram_store("issue_1064_hyphenated_multiword");
+    let target_id = Uuid::new_v4();
+    store
+        .upsert_document(make_document(
+            target_id,
+            "ADR-086",
+            "ADR-086 workspace mirror amendment document",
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Test-only at head c0bad4c: regression guards for unknown-kind error-shape parity and hyphenated-identifier FTS at both store and dispatch level, all passing. Approving.

PR #1076, opened 2026-07-17
fix(runtime): stats() aggregates over caller-visible namespaces to match list (#711)

### LINKED ISSUE #711: stats() and list(kind=edge) disagree on namespace scope for identity-bearing callers
## Observed

For a caller whose request frame carries a visible-namespace set beyond `local`:

- `list(kind="edge")` with the keyset cursor walks edges across all frame-visible namespaces (observed: 11,174 unique edges = 11,173 in `local` + 1 in the caller's own namespace).
- `stats()` reports `edges: 11173` — one namespace only.
- `stats().edges_by_relation` sums to 11,173, so it shares the scalar's single-namespace scope (`count_edges_by_relation` in `crates/khive-db/src/stores/graph.rs` pins `WHERE namespace = ?1`).

So a full `list` walk and the `stats` totals cannot be reconciled by a caller with any non-trivial visibility: the two verbs compute over different namespace scopes.

## Analysis

The store-layer functions are all namespace-pinned; the divergence is at the runtime layer: edge listing merges pages across the caller's visible namespaces, while the stats path queries a single namespace regardless of frame visibility.

Per the per-request identity model, read surfaces should honor the frame's visibility consistently. Either:

1. `stats()` aggregates over the caller's visible namespaces (consistent with `list`), or
2. both accept an explicit `namespace=` parameter and default identically.

Option 1 matches the direction the listing surface already took. Whichever is chosen, the scalar totals and `edges_by_relation` must share one scope, and the contract should be stated in the verb docs.

## Acceptance

- `stats()` totals == count of a full `list` keyset walk under the same caller identity, for entities, edges, and notes.
- `edges_by_relation` sums to the `edges` scalar under all identities.
- Regression test with two namespaces and an identity-bearing frame.

ohdearquant:
Fixes #711.

## Problem

For an identity-bearing caller whose visible-namespace set extends beyond `local`, `stats()` and a full `list` keyset walk could not be reconciled: `list(kind="edge")` merges edges across the caller's frame-visible namespaces, while `stats()` (and `edges_by_relation`) pinned `WHERE namespace = ?1` — a single namespace. The two verbs computed over different scopes.

## Fix (Option 1 in the issue)

`stats()` now aggregates over the caller's visible namespaces, consistent with `list`. The store-layer count functions stay namespace-pinned by design; the aggregation is done at the runtime layer over `token.visible_namespaces()`, mirroring how `list` merges pages:

- `crates/khive-runtime/src/operations.rs`: added a `count_notes` runtime function and reworked `count_entities`/`count_edges` (and `edges_by_relation`) to sum across the visible set.
- `crates/khive-pack-kg/src/handlers/stats.rs`: wired the handler to the visible-namespace-aware counts.

The scalar totals and `edges_by_relation` now share one scope; the fix covers entities, edges, and notes per the acceptance criteria.

## Test

`stats_totals_match_list_walk_across_visible_namespaces` (`crates/khive-runtime/tests/integration.rs`): with two namespaces and an identity-bearing frame, asserts `stats()` totals equal a full `list` keyset-walk count for entities, edges, and notes, and that `edges_by_relation` sums to the `edges` scalar.

## Gates

`cargo test -p khive-runtime`: 70 passed, 0 failed. `cargo clippy -p khive-runtime -p khive-pack-kg --all-targets -- -D warnings`: clean. `cargo fmt --all -- --check`: clean.

Clean-room implementation from the issue spec (no prior partial patch reused).

### DIFF
--- a/crates/khive-pack-kg/src/handlers/stats.rs
+++ b/crates/khive-pack-kg/src/handlers/stats.rs
@@ -10,6 +10,15 @@ use super::common::{deser, StatsParams};
 use crate::KgPack;
 
 impl KgPack {
+    /// Aggregate KG substrate counts (entities, edges, notes).
+    ///
+    /// Scope contract: every total here is summed across the caller's
+    /// full *visible-namespace* set (`token.visible_namespaces()`), the same
+    /// scope `list(kind=...)` merges pages over — not just `token.namespace()`.
+    /// This keeps `stats()` reconcilable with a full `list` keyset walk under
+    /// the same identity: `edges_by_relation` sums to `edges`, and each
+    /// scalar equals the count of a full multi-namespace `list` walk, for
+    /// entities, edges, and notes alike (#711).
     pub(crate) async fn handle_stats(
         &self,
         token: &NamespaceToken,
@@ -22,12 +31,7 @@ impl KgPack {
             .count_edges(token, EdgeListFilter::default())
             .await?;
         let edges_by_relation = self.runtime.count_edges_by_relation(token).await?;
-        let notes = self
-            .runtime
-            .notes(token)?
-            .count_notes(token.namespace().as_str(), None)
-            .await
-            .map_err(RuntimeError::Storage)?;
+        let notes = self.runtime.count_notes(token, None).await?;
         Ok(serde_json::json!({
             "entities": entities,
             "edges": edges,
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Full read at head 45fbd9e: stats() aggregation now spans the caller's visible-namespace set for entities, edges, and notes, mirroring the existing multi-namespace list paths, with a reconciliation test proving stats totals equal a full list walk under the same identity. Approving.

PR #1079, opened 2026-07-17
feat(pack-blob): blob.put/get/stat verbs over the existing BlobStore

ohdearquant:
## Summary
- Phase 1 of the blob-consumer surface (Track A): three thin MCP verbs — `blob.put`, `blob.get`, `blob.stat` — over the existing `khive_storage::BlobStore` content-addressed store. No new storage backend, no schema/migration change, no new entity/note/edge kind.
- New opt-in pack `khive-pack-blob` (load via `--pack blob` / `KHIVE_PACKS=...,blob`), following the `khive-pack-template` scaffold (`Pack` + `PackFactory` + `inventory::submit!` + `PackRuntime::dispatch`). Not added to the default pack set.
- `blob.put(bytes|path)` — base64 bytes or a server-local path; idempotent (same content → same `ContentRef`); rejects >128MiB input and bad/missing input with listed errors.
- `blob.get(content_ref, range?)` — base64 read-back, digest-reverified on every read (flags corruption instead of silently round-tripping bad bytes); optional `{offset, length}` range applied client-side (the trait has no partial-read).
- `blob.stat(content_ref)` — existence + size + a `corrupt` flag. Since `BlobStore` exposes no size-only accessor, an existing object's size requires a full `get()` under the existing trait — documented tradeoff, no trait extension in this phase.
- Physical `delete` / `orphan_sweep` remain admin-only, not exposed as verbs (ADR-111 §8).

## Design contract
- Implements a signed internal design proposal for the blob-consumer surface (design freeze reference: e4326988) as an interim contract. The design's section on the verb surface and phase-1 scope is implemented as-is; no ADR text is included in this PR.
- The ADR-086 amendment documenting this surface is a separate phase-2 doc PR — track merge order (the doc PR should reference this one, not the other way around).
- The typed artifact/reference plane (the design's "phase 2" section) is out of scope for this PR.

## Test plan
- [x] `cargo test -p khive-pack-blob` (9 tests: idempotent put/stat/get round trip, byte range, unknown-ref not-found/not-exists, malformed content_ref, put from path, mutually-exclusive bytes/path validation)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -181,7 +181,9 @@ system git with hardened, allowlisted argv construction and unconditional force-
 Ask A) added 2026-07-08; kg.resolve added 2026-07-09; workspace (#873) contributes zero verbs,
 adding only the `workspace` entity kind and `contains` endpoint rules to git/gtd/session notes;
 blob contributes three verbs, blob.put/blob.get/blob.stat (ADR-111), over the `BlobStore`
-content-addressed storage trait, erroring as unconfigured until a backend is installed;
+content-addressed storage trait; a normal file-backed boot installs a default `FsBlobStore`
+beside the database file with no config needed, and the verbs stay unconfigured only
+against an in-memory backend;
 regenerate via `request(ops="verbs()")` before editing this line).
 
 ### KG pack verbs (18 — ADR-017, ADR-046, ADR-089)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -168,8 +168,8 @@ request(ops="[{\"tool\":\"v1\",\"args\":{...}}, ...]")
 ```
 
 Verbs come from whichever packs are loaded via `KHIVE_PACKS` (env) or `--pack` (CLI). Default
-loads all 11 production packs: kg, gtd, memory, brain, comm, schedule, knowledge, session, git,
-code, workspace (verbs at 82: the `code` pack contributes one verb, `code.ingest`
+loads all 12 production packs: kg, gtd, memory, brain, comm, schedule, knowledge, session, git,
+code, workspace, blob (verbs at 85: the `code` pack contributes one verb, `code.ingest`
 (ADR-085 Amendment 2, PR #1039 — L1 manifest + L1.5 import-scan source ingest into a
 dedicated map database); its `finding` note kind and `findings.json` batch ingest are
 still reached only through the `kkernel code-ingest` admin CLI path (ADR-085 Amendment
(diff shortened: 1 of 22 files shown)


ohdearquant:
Review response, issue 3 (NamespaceToken enforcement on get/stat): **disputed as designed behavior**.

By-ID operations in khive are namespace-agnostic by design. ADR-007 Rev 6 removed post-fetch `record.namespace == caller_namespace` checks from every by-ID path (get/update/delete/merge), and the contributor guide explicitly lists reintroducing them as an anti-pattern. Authorization lives at one seam, the Gate (ADR-018) — pack handlers do not make authorization decisions.

A full `content_ref` is a by-ID reference in exactly this sense: presenting the complete 64-hex digest demonstrates knowledge of the object, the same trust model as presenting a full record UUID. Beyond that, the blob store has no namespace dimension at all: ADR-111 defines a content-addressed store in which identical bytes from any writer deduplicate to a single object, so "cross-namespace read" is not a defined concept at this layer. A deployment that wants a blob-read policy attaches it at the Gate, not in the pack handler.

The remaining findings are accepted and fixes are in progress on this branch: metadata/ranged reads without full-object hydration (BlobStore `size` accessor + ADR-111 amendment, bounded `get`), strict range validation, and the 82-vs-85 verb-count and catalog updates.

ohdearquant:
Two items from a consumer-side read of this surface, recorded here so they are not lost while the branch is parked.

**1. The put ceiling and the retrieval ceiling do not match, and put is the larger one.**

`MAX_OBJECT_BYTES` is 64 MiB (`crates/khive-db/src/stores/blob_s3.rs:34`, enforced on write at :387). The daemon frame cap `MAX_FRAME_BYTES` is 8 MiB (`crates/khive-runtime/src/daemon.rs:39`). An object between those two bounds can be written and then cannot be read back through the verb surface.

That asymmetry is worth closing before this leaves draft, because the failure is silent at write time and only surfaces later on retrieval, when the caller has already been told the write succeeded. Either bound is defensible; a caller being able to store what it can never retrieve is not. Whichever way it is resolved, `put` should reject at the ceiling that retrieval can actually honor, so the error arrives at the operation that can still do something about it.

**2. The PR body understates `size()` and describes the opposite of what the code does.**

The body says an existing object's size requires a full `get` under the existing trait, and that no trait extension happens in this phase. The code adds `size()` to the `BlobStore` trait, documented as never paying for a full read and mapping to a filesystem metadata call or an S3 `HEAD Object`.

That is the most valuable addition here for any cache-shaped consumer: "do I already have this, and at what size" is asked before every fetch, during dedup, and on every integrity sweep. Answering it with a metadata call rather than a hydrate is the difference between a sweep that costs nothing and one that transfers the whole corpus to answer an existence question. The body should say what the code does.

PR #1080, opened 2026-07-17
docs(adr): ADR-116 durable per-model generation coherence for the memory ANN warm path

ohdearquant:
Adds ADR-116, a numbered decision record for durable per-model generation coherence on the memory ANN warm path. Refs #752.

## What

ADR-116 makes a durable per-model SQLite generation the sole authority that may certify an installed memory ANN graph as current. Every committed mutation that changes a model's ANN input advances that model's durable generation inside the same SQLite transaction, coalesced to once per affected model per transaction. A recall may search an installed graph only when a fresh database read observes `Gann(m) == Gdb(m)`; a persisted snapshot restores only under generation equality plus format-version, fingerprint, and ordered-content-hash gates.

## Relationship to prior decisions

- Retains ADR-062's one-memory-index-per-model and `global::memory_vamana::{model}` snapshot key, and corrects its legacy-snapshot purge predicate.
- Does not amend ADR-079, whose persisted-segment decision is knowledge-pack scoped.
- Supersedes ADR-107's memory-ANN stale-serving and global `memory_ann_epoch` provisions while retaining its single-flight, RAII guard-release, bounded-rebuild, hydration, and cold-start mechanics. ADR-107's status moves to Superseded when this ADR is accepted.

## Status

Proposed. Docs-only draft opened for review. Implementation lands under separate PRs bound by the conditions below.

## Binding conditions on implementation PRs

1. The true-upsert prerequisite is ordered before or together with the generation-bump migration.
2. Each implementation PR carries a p95 write-amplification benchmark (file-backed WAL, warm, single- and multi-model) showing the added counter DML stays within 1.0 ms absolute and within 5 percent of warm-recall p95.
3. A documented inventory of every direct vector-table DML site ships as an artifact, proving each site advances the generation.
4. The corrected legacy-snapshot purge predicate ships regardless of the rest.

Docs-only. No code changes.

### DIFF
--- a/docs/adr/ADR-116-memory-ann-generation-coherence.md
+++ b/docs/adr/ADR-116-memory-ann-generation-coherence.md
@@ -1,0 +1,513 @@
+# ADR-116: Durable Per-Model Generation Coherence for the Memory ANN Warm Path
+
+**Status**: Proposed\
+**Date**: 2026-07-17\
+**Scope**: memory-pack ANN freshness across processes, persisted memory-ANN snapshots, and the
+SQLite write/migration contract\
+**Decision owner**: khive maintainers\
+**References**: issue #752; ADR-062; ADR-079; ADR-107
+
+## Status and relationship to prior decisions
+
+This MUST be a **new ADR**, not an amendment buried in ADR-062.
+
+ADR-062 decided FTS/ANN key consolidation and the model-global memory index key. It did not define
+cross-process cache coherence. More importantly, accepted ADR-107 now defines the memory ANN's
+eventual-consistency contract, a global durable epoch, five-second debounced checks, and stale
+serving. Those are the contracts this decision changes. Amending only ADR-062 would leave two
+accepted, mutually incompatible authorities.
+
+This ADR therefore:
+
+- **retains** ADR-062's one-memory-index-per-model and
+  `global::memory_vamana::{model}` snapshot key;
+- **corrects** ADR-062's legacy-snapshot purge predicate;
+- **does not amend** ADR-079, whose persisted-segment decision is explicitly knowledge-pack scoped;
+- **supersedes** ADR-107's memory-ANN stale-serving/freshness provisions and its global
+  `memory_ann_epoch` protocol, while retaining its single-flight, RAII guard-release, bounded
+  rebuild, hydration, and cold-start mechanics where they do not conflict here.
+
+## Teardown: refute first
+
+### “Only two external writer paths matter, so bump only those”
+
+This survives only under today's deployment convention, not under issue #752's stated invariant.
+Explicit bumps in `kkernel exec --atomic` and `kkernel reindex` would cover the two known exposures
+and avoid a durable write on routine daemon mutations. It is a coherent lower-cost design if khive
+is willing to declare that no other process may write a shared database.
+
+Khive has no enforceable single-writer-process identity at the storage boundary, however. A second
+`KhiveRuntime`, a new admin command, or a future direct store consumer is indistinguishable from the
+daemon to the vector store. Its correctly fired in-process hook still cannot reach a resident
+daemon. Selective bumps would therefore replace a data invariant with a caller-registration
+convention and would not cover issue #752's third affected path (“any other external process”).
+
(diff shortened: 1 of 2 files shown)


ohdearquant:
Review response, round-5 blocking item 1 (runtime-enforced per-principal mutation limiting): **disputed as out of scope, with the ADR text now stating the position precisely** (round 6, bc500e61b).

The review is correct that Gate rate obligations are declarative today — and the revised text no longer implies otherwise. The disposition rests on supported topology, not on rate limiting: ADR-007 Rev 6 defines namespace as write attribution, never isolation, so khive has no tenant-isolation model at any storage layer and a deployment serving mutually untrusted or semi-trusted principals from one shared database is outside the supported topology altogether, independent of this mechanism. A runtime-enforced per-principal mutation limiter is therefore not a requirement this ADR can or does satisfy; it belongs to a future ADR that introduces a supported co-tenancy topology as a change to the Gate contract, adopted together with the membership-targeted invalidation amendment this ADR reserves.

The remaining seven blocking items are accepted and addressed in round 6: mandatory pre-migration quiesce of legacy processes with a mixed-version regression test (legacy note mutations are trigger-covered; legacy vector writes cannot be mechanically fenced), an explicit supported model-cardinality bound of eight with fail-loud validation and gates run at the maximum, a defined gate corpus scale (10,000 live note.content vectors per model) with the multi-model multi-namespace fan-out case, an owning NoteStore batch soft-delete API for the single-transaction rule, an owned process-wide rebuild scheduler with a configuration surface, a mechanical direct-DML inventory rule via name-derivation-helper confinement covering dynamically constructed SQL, and a field-constrained exact fallback matching the ANN input predicate with a route-parity test.

ohdearquant:
Cross-linking this to a defect investigated today, because the two overlap in the same code and should not be designed independently.

A source investigation into recall freshness (`memory.recall` not returning same-day writes) landed on the same machinery this ADR governs. Summary of what was found, with the ADR's scope in mind:

**The two halves are complementary, not competing.**

- ADR-116 addresses **how currency is certified** — making a durable per-model generation the sole authority, so an external writer (`kkernel --atomic`, `kkernel reindex`) can invalidate a running daemon's warm cache. That is the #752 half.
- The freshness defect is **what gets served when the graph is known to be stale**. Today `search_loaded` (`crates/khive-pack-memory/src/ann.rs:441-456`) returns whatever graph is installed, unconditionally, with no freshness gate. A test pins this as intended: `search_loaded_serves_stale_installed_entry_without_rebuild`.

**Why they interact.** The caller `collect_recall_vector_hits` (`crates/khive-pack-memory/src/handlers/common.rs:928`) already computes `cache_fresh` via `ann::is_current`, but uses it only to decide whether to kick a background rebuild — never to change what is returned. Whatever ADR-116 makes authoritative for "is this graph current" becomes the input to that same decision. If the serve-side fix is written against the current in-process counter while this ADR moves authority to the durable generation, the two land on the same predicate from opposite directions.

**Relevant detail for the ADR's design.** A namespace-correct exact sqlite-vec fallback already exists (`common.rs:1100-1129`), currently gated on cold-miss rather than on staleness. So the serve-side remedy likely reuses existing machinery rather than adding a new path — which makes it cheaper to align with this ADR now than to reconcile later.

**One constraint worth naming explicitly in the ADR:** a timed-out cold start deliberately refuses the exact scan (`common.rs:1013-1017`, comment: "Do not replace a bounded timeout with an O(corpus) exact scan"). Any rule of the form "serve exact results when the graph is not certified current" collides with that policy under load. It needs to be reconciled deliberately, not silently overridden.

Also worth pinning: freshness is **backend-specific**. `kg.search` (entity/note) rides `SqliteVecStore`, a `vec0` virtual table scanned brute-force per query (`crates/khive-db/src/stores/vectors.rs:1080-1083`) — it has no index to go stale and cannot exhibit this class of bug. Only the two Vamana warm-ANN paths (`memory.recall`, and `knowledge.search`/`compose`, which re-implements the identical pattern independently) are affected. If ADR-116's certification model is meant to generalize beyond the memory pack, the knowledge pack has the same shape and would need the same treatment.

No change requested here — flagging the overlap so the serve-side work is sequenced against this ADR rather than colliding with it.

PR #1083, opened 2026-07-17
test(bench): p95 gate harness for the memory ANN warm path

ohdearquant:
## Summary

- Adds `crates/khive-pack-memory/benches/p95_gate.rs`, a standalone harness that
  measures `memory.recall` latency (p50/p95/p99, warm ANN path) against a
  file-backed WAL SQLite database with 1 primary + 3 retired embedding-model
  generations (800 memories total).
- Establishes the pre-implementation baseline that ADR-116's warm-hit
  generation-read gate (docs/adr, branch `docs/adr-116-memory-ann-generation`,
  not yet merged) will be diffed against: added per-model generation checking
  must cost at most 1.0ms absolute p95 and at most 5% of this baseline's warm
  `memory.recall` p95.
- Committed results in `crates/khive-pack-memory/benches/p95_gate_results.md`
  (Apple M2 Max, macOS, commit a3544bd12): primary p50 16.03ms / p95 29.51ms /
  p99 34.85ms (n=200); retired models ranged p50 5.50-14.59ms / p95 16.02-30.64ms.

## Notes

- The fleet's exclusive bench-window lock could not be acquired: two idle
  `reactive_pr_review.py` daemons held stale shared-lock file descriptors with
  0% CPU and no live build process, so the lock never released. Verified via
  `ps` that the machine was otherwise CPU-idle during the timed run, and two
  independent runs produced consistent numbers. Documented in the results file.
- ADR-116 is Proposed, not yet implemented; this PR only adds the baseline
  harness, no runtime code changes.

## Test plan

- [x] `cargo clippy -p khive-pack-memory --all-targets -- -D warnings`
- [x] Harness run end-to-end, results captured in
      `crates/khive-pack-memory/benches/p95_gate_results.md`

### DIFF
--- a/crates/khive-pack-memory/Cargo.toml
+++ b/crates/khive-pack-memory/Cargo.toml
@@ -61,3 +61,7 @@ harness = false
 [[bench]]
 name = "memory_bench"
 harness = false
+
+[[bench]]
+name = "p95_gate"
+harness = false
--- a/crates/khive-pack-memory/benches/p95_gate.rs
+++ b/crates/khive-pack-memory/benches/p95_gate.rs
@@ -1,16 +1,50 @@
 //! ADR-116 gate condition 2: p95 baseline for `memory.recall` on the warm ANN path.
 //!
-//! Measures `memory.recall` latency (p50/p95/p99, warm/post-index) against a
-//! **file-backed WAL SQLite database** with four registered embedding-model
-//! generations (1 primary + 3 retired), matching the shape ADR-116's warm-hit
-//! generation-read gate is defined against: "a file-backed WAL benchmark with
-//! warm page cache at one and three models ... at most 1.0 ms absolute p95 and
-//! at most 5% of end-to-end warm memory.recall p95" (ADR-116 §Warm hit).
+//! ADR-116 (PR #1080, in review) defines the warm-hit generation-read gate against a
+//! **file-backed WAL SQLite database with warm page cache "at one and three models"** —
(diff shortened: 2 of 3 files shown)


ohdearquant:
Noting a use for this harness beyond its stated scope, so it gets sequenced rather than sitting idle.

A recall-freshness defect is queued against the same code path this measures: `memory.recall` currently serves a stale warm-ANN graph unconditionally (`crates/khive-pack-memory/src/ann.rs:441-456`), so same-day writes can be absent from results. The likely remedy merges exact results over the unindexed tail with the ANN hits at query time, reusing the namespace-correct sqlite-vec fallback that already exists at `crates/khive-pack-memory/src/handlers/common.rs:1100-1129`.

That remedy trades latency for correctness on exactly the warm path this harness instruments. Which makes this PR the right instrument to gate it:

- It measures p50/p95/p99 on the warm ANN path, with retired-generation fixtures already in place — so it can produce the **before** number now and the **after** number once the fix lands.
- There is a standing policy this harness can hold honest: a timed-out cold start deliberately refuses an exact scan (`common.rs:1013-1017`, "Do not replace a bounded timeout with an O(corpus) exact scan"). Any freshness fix has to respect that bound rather than quietly widening it, and p95 under load is where a violation would surface.

Suggested sequencing: land this first so the freshness work has a baseline to measure against, instead of landing a latency-affecting correctness fix with no instrument to detect the regression it might introduce.

No changes requested to the harness itself in this comment.

PR #1092, opened 2026-07-17
build(make): compile channel-telegram into the local kkernel binary

ohdearquant:
make local builds with --features channel-email only, so the installed kkernel binary cannot spawn the Telegram channel loops that shipped with the ADR-056 amendment adapter. This enables channel-email,channel-telegram together in the local build target.

Runtime behavior is unchanged until KHIVE_TELEGRAM_BOT_TOKEN and KHIVE_TELEGRAM_MAINTAINER_CHAT_ID resolve at daemon startup; with them unset the spawn path logs a warning and skips, per the existing non-fatal config contract.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,8 +54,8 @@ bench-1m-ci:
 	bash scripts/bench_1m.sh --ci
 
 local:
-	@echo "==> Building kkernel (release, channel-email)..."
-	@cd crates && cargo build --release -p kkernel --features channel-email
+	@echo "==> Building kkernel (release, channel-email, channel-telegram)..."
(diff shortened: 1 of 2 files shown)


ohdearquant:
Pass-through added in 31f6f2088, mirroring the existing channel-email pass-through; verified with cargo check -p kkernel --features channel-email,channel-telegram (clean).

On the non-blocking default-inclusion point: keeping the adapter in the local build is deliberate. The channel is inert without its runtime configuration (the spawn path logs and skips), the marginal build cost is one small adapter crate, and building it in is what lets the deployment be a pure configuration change rather than a rebuild. If local build time becomes a concern the feature list is a one-line edit.

PR #1093, opened 2026-07-17
docs(adr): ADR-056 amendment - iMessage channel over an SSH bridge

ohdearquant:
## Summary

Adds a 2026-07-17 amendment to ADR-056 specifying a `khive-channel-imessage` adapter: iMessage
send/read happen on a remote macOS host (the Apple ID owner) bridged over SSH from the daemon
host, mirroring the shape of the existing Telegram and email adapters (same `Channel` trait,
same per-adapter `spawn_*_channel_loops` pattern, same `comm.ingest` dispatch path).

- Outbound: `ssh` to a configured bridge host, `osascript` drives Messages. Message text never
  touches a shell/AppleScript string — stdin/temp-file only, fixed allowlisted argv, citing
  ADR-108's hardened shell-out pattern as precedent.
- Inbound: read-only polling of the bridge host's `~/Library/Messages/chat.db` over SSH,
  configurable interval (default 5s), dedup by the database's stable per-message GUID through
  the existing `idx_comm_message_external_id` mechanism.
- Config table: `KHIVE_IMESSAGE_SSH_TARGET`, `KHIVE_IMESSAGE_MAINTAINER_HANDLE`,
  `KHIVE_IMESSAGE_MAINTAINER_SLUG` (default `maintainer`), `KHIVE_IMESSAGE_POLL_SECS` (default
  5), `KHIVE_IMESSAGE_INGEST_NAMESPACE` (default `local`). No secret material beyond the SSH
  client's own config; nothing in any khive store.
- v1 non-goals: group chats, attachments, multiple maintainer handles, and — permanently, not
  just deferred — any write to `chat.db`.

## Flagged, not resolved (called out explicitly in the amendment text)

1. **Address-form inconsistency in the settled design.** One line of the settled spec says the
   address form is `imessage:<handle>`; another says the addressing model is "identical to the
   Telegram amendment," which routes by slug (`telegram:<slug>`), never the raw chat id. This PR
   resolves it in favor of the slug form (`imessage:<slug>`, handle stays config-only) to match
   the stated Telegram-parity intent, but flags the inconsistency for sign-off rather than
   silently picking a side.
2. **Durability-model risk relative to Amendment 2026-07-09.** The settled design keeps an
   in-memory poll offset (Telegram-style) plus GUID dedup. GUID dedup prevents double-ingest on
   restart but does not by itself prevent the *missed-ingest* class of bug Amendment 2026-07-09
   fixed for email (#449: a bounded-page, time-windowed query can permanently skip a backlog
   above the page limit). Whether this applies depends on the concrete `chat.db` query shape the
   implementation uses — flagged for the implementation PR and its review to verify explicitly.

Design sign-off is required before any dependent implementation merges.

## Test plan

- [x] `deno fmt docs/` (via `/commit` pre-commit hooks) — passed, no reformatting beyond table
      alignment.
- [x] Docs-only change; no `cargo` build/test surface touched.
- [ ] Design sign-off on the amendment, including a ruling on the two flagged items above.

### DIFF
--- a/docs/adr/ADR-056-channel-transport-layer.md
+++ b/docs/adr/ADR-056-channel-transport-layer.md
@@ -502,6 +502,21 @@ any khive store, consistent with §9 ("no secrets in the store"). An unreachable
 transport failure local to this one channel: it degrades to a channel-down warning and does not
 affect any other configured channel or fault the daemon.
 
+**SSH target validation and the end-of-options delimiter.** `KHIVE_IMESSAGE_SSH_TARGET` is
+validated fail-closed at config load, not at first use: it must match a restrictive shape, an
+optional `user@` prefix followed by a hostname or address, and a value beginning with `-`, or
+containing whitespace or a control character, is rejected -- `ImessageChannelConfig::from_env`
+returns `ChannelError::Config` and the channel does not start. This closes a specific attack: an
+option-prefixed target such as `-oProxyCommand=...` would otherwise be parsed by the `ssh` client
+as an option rather than a destination, and a `ProxyCommand` override executes an arbitrary local
+command under the daemon account. Fixed argv construction alone does not close this, because argv
+construction fixes how arguments are assembled, not what the target string itself is interpreted
+as once `ssh` parses it. As defense in depth beyond the validation gate, every `ssh` invocation
+this adapter makes additionally passes the `--` end-of-options delimiter immediately before the
+target argument, so even a target value that somehow reached the invocation unvalidated can never
+be parsed as an `ssh` option. Acceptance property: an option-prefixed `KHIVE_IMESSAGE_SSH_TARGET`
+value refuses channel start.
+
 ### Outbound: `osascript` via hardened argv, no text interpolation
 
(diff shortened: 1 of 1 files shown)


ohdearquant:
Directing note on the two flagged conflicts:

1. Address form: the slug-form resolution is endorsed — imessage routes by maintainer slug exactly like the Telegram channel, and the raw-handle form in the earlier paragraph should be read as the configuration value (the env-configured maintainer handle), not the wire address. If any text still shows a raw handle as the send address, the slug form wins.

2. Durability model: correctly flagged. The amendment's contract stands (in-memory offset + GUID dedup), and the implementation PR must demonstrate its poll query cannot permanently skip backlog - the bounded-window skip class documented in the email channel's durability amendment. That demonstration is an acceptance item for the implementation review, not a spec change.

ohdearquant:
## Binding implementation requirements (verified at code review)

These constrain the implementation of this ADR and are checked when the implementing PR is reviewed:

- **Retry backoff floor** must be **>= the configured healthy poll cadence** — the failure-retry schedule must never retry more aggressively than normal polling, to avoid SSH connection churn amplification while the bridge is unhealthy.

PR #1097, opened 2026-07-18
fix(db): exact visited-set cycle detection in traverse, replacing path-string LIKE (#562)

### LINKED ISSUE #562: perf(traverse): per-root CTE still uses LIKE-based path-string cycle detection (from #250)
Closed-issue audit (2026-07-03, batch07). #250 had two asks: batch roots into one query, and drop path-string LIKE cycle detection.

**What landed**: `crates/khive-db/src/stores/graph.rs:1069` loops over root CHUNKS (not individual roots); `graph.rs:1106-1121` seeds all roots in a chunk through one VALUES CTE with `root_id` tracking. This is the primary perf win and it's real.

**What's missing**: `graph.rs:1130` still uses path-string `LIKE` cycle detection — the secondary ask (replace with a proper visited-set) was not done. This is the same root cause as #251 (reopened separately in this audit pass, batch07) — a shared fix (frontier/visited-set CTE) would close both at once. Recommend implementing alongside #251 rather than as an independent patch.

ohdearquant:
## Summary
The traverse recursive CTE detected cycles by substring-matching the next node id
against a comma-joined path string with `LIKE`. Because `graph_edges.source_id` /
`target_id` are unconstrained TEXT (no DB-level UUID-format guarantee), a substring
match is correctness-sensitive to id formatting rather than exact by construction.

This builds the accumulated path as a JSON array and tests membership exactly:
`json_insert(t.path, '$[#]', <next>)` to extend, and
`NOT EXISTS (SELECT 1 FROM json_each(t.path) WHERE value = <next>)` as the cycle
guard. `json_each` is already an established pattern in this crate (vectors.rs,
entity.rs).

## Tests
- `test_traverse_terminates_on_real_cycle` — A→B→C→A terminates, each node visited once.
- `test_traverse_near_colliding_ids_not_confused` — two ids differing only in the last
  character stay distinct (locks exact membership vs. prefix/partial).

Gates: `cargo test -p khive-db` (383 unit + 13 contract green), `cargo fmt --all -- --check`,
`cargo clippy -p khive-db --all-targets -- -D warnings` all clean.

Closes #562.

### DIFF
--- a/crates/khive-db/src/stores/graph.rs
+++ b/crates/khive-db/src/stores/graph.rs
@@ -1934,15 +1934,22 @@ impl GraphStore for SqlGraphStore {
                 }
 
                 // Seed rows: one per root in this chunk, each referencing its own
-                // param 3× (root_id, node_id, and the initial path string).
+                // param 3× (root_id, node_id, and the initial path — a JSON array
+                // containing just the root id).
                 let seed_rows: Vec<String> = (1..=n_chunk)
-                    .map(|i| format!("(?{i}, ?{i}, NULL, 0, ?{i}, 0.0)"))
+                    .map(|i| format!("(?{i}, ?{i}, NULL, 0, json_array(?{i}), 0.0)"))
                     .collect();
                 let seeds = seed_rows.join(", ");
 
                 // CTE covering the chunk's roots.  CROSS JOIN forces SQLite to put
                 // the frontier (t) as the outer loop and seek graph_edges by index,
                 // avoiding the O(edges × frontier) plan (#250, #251).
+                //
+                // `path` is a JSON array of visited node-id strings rather than a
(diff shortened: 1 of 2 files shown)


ohdearquant:
Closing without merge. Review measured a deep-traversal performance regression (~2.5x on a 1,000-hop indexed-chain benchmark, 100 runs) because the recursive hot path now parses the accumulated path with json_each and reserializes it with json_insert for every candidate edge. The correctness premise does not hold for the supported data: node endpoints are canonical fixed-width UUID strings, and the prior predicate wraps both the comma-joined path and the candidate id in commas, so it already performs exact whole-element (delimiter-bounded) membership matching with no substring ambiguity. The added near-collision test passes against the prior implementation as well. Disposition of the underlying issue in #562.

PR #1098, opened 2026-07-18
fix(request): teach JSON escape + MCP double-escape on control-char parse error (#491)

### LINKED ISSUE #491: khive-request: raw control-char parse error should teach the escape syntax and MCP double-escape
## Problem

A raw newline byte inside a quoted string in the function-call DSL fails with the serde-level message:

```
MCP -32602: control character ( -) found while parsing a string
```

The error is technically correct and the grammar is fine (ADR-016 string escapes follow JSON and already ship), but the message teaches nothing. Worse, the natural fix a caller tries — writing `\n` in the JSON tool argument — still fails, because the MCP transport decodes that to a literal newline byte before the DSL parser runs. The working form from an MCP client is the double escape `\\n`. Nobody discovers this unaided; users have repeatedly landed on the wrong "must switch to the JSON op form" workaround instead.

## Fix

When string decoding fails on a control character, return an error that:
1. names the offending character and its position,
2. states that string escapes follow JSON (`\n`, `\t`, `\"`, `\\`),
3. notes that callers passing `ops` through a JSON transport (every MCP client) must double-escape (`\\n` in the wire form).

Grammar unchanged — this is error-message-only. Normative basis: ADR-084 §3c (PR #490). Parser tolerance for raw control chars was considered and rejected there (Alternatives §6).

## Where

`crates/khive-request` — `parse_value` hands the quoted slice to `serde_json::from_str`; wrap that error.

ohdearquant:
## Summary
A raw control-character byte inside a quoted string in the function-call DSL surfaced the bare
`serde_json` message ("control character ... found while parsing a string"), which teaches nothing
and sends callers toward the wrong workaround. This wraps the `serde_json` string-decode error at the
`parse_value` call site so the message names the offending character and position, states that DSL
string escapes follow JSON (`\n`, `\t`, `\"`, `\\`), and notes that callers passing `ops` through a
JSON transport (every MCP client) must double-escape (`\\n` in the wire form, because the transport
decodes one level before the parser runs). Grammar unchanged (ADR-016 / ADR-084 §3c basis).

## Scope note
PR #957 already normalizes raw `\n`/`\r`/`\t` bytes before they reach `serde_json`, so those no
longer error at all. The bare-serde-message problem remained for the other raw control bytes (NUL,
form feed, ESC, ...); this addresses the error message for that remaining path.

## Tests
`control_char_error_teaches_escape_syntax_and_mcp_double_escape` feeds a raw form-feed byte in a
quoted value and asserts the error is `DslError::InvalidValue` with a message carrying the escape
grammar and the double-escape guidance. Non-control-character parse errors pass through unchanged.

Gates: `cargo test -p khive-request` (122 passed), `cargo fmt --all -- --check`,
`cargo clippy -p khive-request --all-targets -- -D warnings` all clean.

Closes #491.

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -2363,7 +2363,7 @@ mod tests {
         // `ops` argument as a JSON string using the standard JSON `\n`
         // escape. Deserializing `RequestParams` decodes that escape into an
         // actual raw LF byte inside the DSL source — the exact shape
-        // `escape_literal_control_chars` (crates/khive-request/src/parser/scan.rs)
+        // `normalize_quoted_string` (crates/khive-request/src/parser/scan.rs)
         // exists to accept. This confirms the decoded raw newline survives
         // parsing and dispatch all the way to the pack handler's result.
         let wire = "{\"ops\":\"create(kind=\\\"entity\\\", entity_kind=\\\"concept\\\", name=\\\"line1\\nline2\\\")\"}";
--- a/crates/khive-request/src/parser/parser_impl.rs
+++ b/crates/khive-request/src/parser/parser_impl.rs
@@ -248,7 +248,7 @@ impl<'a> Parser<'a> {
                     let start = self.pos;
                     let end = scan_string_end(self.src, start)?;
                     let raw = std::str::from_utf8(&self.src[start..end]).expect("utf8 key literal");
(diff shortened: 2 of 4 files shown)


oceanwaves630 (approved):
Reviewed against the verdict at head 3e555b90: zero blocking items, minors filed as follow-up. Approving.

Issue #533, opened 2026-07-03
memory.prune leaves FTS/vector/ANN state stale (follow-up to #429)


ohdearquant:
Follow-up to #429, which was closed via PR #510 as a recall-correctness false positive. The closure-reconciliation sweep (2026-07-03) confirmed the recall symptom is indeed benign, but the underlying ADR-014 curation gap reported in the issue body is still live on main:

`crates/khive-pack-memory/src/prune.rs` soft-deletes via raw `NoteStore::delete_note` with no FTS/vector/ANN cleanup — unchanged since it shipped in PR #192. Pruned memories leave stale rows in the FTS index and stale vectors in the ANN/vector stores.

## Fix sketch
Route `memory.prune` through the same delete path other curation verbs use (the one that maintains FTS/vector/ANN coherence), or explicitly clean the auxiliary indexes per pruned note within the prune transaction.

## Verification
- Test: remember → prune → assert no FTS hit and no vector row for the pruned note id.
- `memory.vacuum` after prune reclaims the space.

Provenance: audit-20260702 closure-reconciliation sweep, verdict table at issue-sweep show `_reconciliation.md` (PARTIAL classification, only one across 33 closed issues).

ohdearquant:
Closing during tracker cleanup: this concerns functionality provided by the commercially licensed extensions, which are no longer part of this repository. Tracking continues on the distribution's internal tracker.

Issue #429, opened 2026-07-02
khive-pack-memory: memory.prune leaves FTS, vector, and ANN state stale [high]


ohdearquant:
Severity: **high** | Confidence: high | Priority: P1 | Categories: data_integrity, correctness, architecture
Crate: `khive-pack-memory` | Finding id: `PACKMEMORY-AUD-001`

## Finding
Prune silently diverges storage from search indexes, leaks orphaned vector/FTS entries, and can produce incomplete recall after curation.

## Failure scenario
A namespace has many indexed low-salience memories and one live high-salience memory. memory.prune(min_salience=0.2, before=0) marks the low-salience rows deleted through NoteStore only, leaving FTS/vector/ANN entries behind. Later memory.recall spends its bounded candidate window on stale deleted IDs, hydration drops them, and the live memory outside the window is missed.

## Evidence
- `crates/khive-pack-memory/src/handlers/prune.rs:133` — Prune soft-deletes candidates directly through NoteStore.
- `crates/khive-db/src/stores/note.rs:567` — Raw NoteStore soft delete only updates the notes row.
- `crates/khive-runtime/src/operations.rs:3017` — Runtime delete_note documents and owns FTS/vector cleanup for soft delete.
- `crates/khive-runtime/src/operations.rs:3078` — Runtime soft delete removes the FTS document after row deletion.
- `crates/khive-runtime/src/operations.rs:3080` — Runtime soft delete removes vectors for every registered embedding model.
- `docs/adr/ADR-014-curation-operations.md:289` — ADR-014 requires soft delete to remove records from FTS5 and vector search.

## Recommendation
Route prune deletions through KhiveRuntime::delete_note(token, id, false) or a runtime batch API with identical cleanup/event semantics, then invalidate memory ANN state/snapshots for affected namespaces/models.

## Verification
Add a regression that seeds indexed memories, warms ANN, prunes low-salience memories, and asserts pruned IDs are absent from FTS/vector/ANN candidate paths while live memories still recall.

---
Source: automated static-analysis sweep.

ohdearquant:
Fixed by PR #510 (merged).

PR #1101, opened 2026-07-18
test(pack-memory): guard prune excludes pruned memories across FTS/vector/ANN paths (#533)

ohdearquant:
## Summary

Follow-up to #533 ("memory.prune leaves FTS/vector/ANN state stale", follow-up to #429).

Per the issue's fix sketch, the verification step is: `remember -> prune -> assert no FTS hit and no vector row for the pruned note id`. This PR adds exactly that as a regression test and reports the empirical result — the underlying gap is **already covered on `memory.recall`'s read path**, so no production fix is needed there.

## What was verified

`crates/khive-pack-memory/src/handlers/prune.rs` (`handle_prune`) soft-deletes pruned rows via `NoteStore::delete_note(id, DeleteMode::Soft)` (sets `deleted_at`, ADR-014 — rows remain) and bumps the per-model ANN generation (`ann::bump_generation` + `ann::ensure_ann_background`) so a background rebuild eventually drops stale vectors from the ANN graph. It does **not** touch the FTS5 index or the sqlite-vec store directly, so stale rows can persist there until the ANN rebuild completes (or indefinitely for FTS5/sqlite-vec, which prune never rewrites).

The new test (`prune_excludes_pruned_memory_across_fts_vector_and_hybrid_recall`) seeds one low-salience memory, prunes it, then issues `memory.recall` three times — `fusion_strategy: "keyword_only"` (FTS-only leg), `"vector_only"` (sqlite-vec/ANN leg), and the default `"rrf"` (hybrid) — and asserts the pruned note's id is absent from every result set.

All three pass. The reason: `load_memory_candidate_notes` (`crates/khive-pack-memory/src/handlers/common.rs`) applies a `deleted_at IS NULL` post-hydration filter uniformly to *both* text and vector candidates before either leg's hits reach the caller, regardless of whether the underlying index still carries the stale row. So `memory.recall` correctness holds today even though the raw FTS/vector store state can be stale.

## Verdict

**VERIFIED-RESOLVED** for `memory.recall` correctness. The residual gap (stale rows sitting in the FTS5/sqlite-vec stores growing storage until an ANN rebuild or `memory.vacuum`) is a curation/storage-hygiene concern, not a read-correctness bug — out of scope for this PR per the issue's own verification criteria.

## Test plan

- [x] `cargo test -p khive-pack-memory` — 83 passed (incl. the new regression test)
- [x] `cargo clippy -p khive-pack-memory --all-targets -- -D warnings`

### DIFF
--- a/crates/khive-pack-memory/src/handlers/prune.rs
+++ b/crates/khive-pack-memory/src/handlers/prune.rs
@@ -170,6 +170,180 @@ impl MemoryPack {
     }
 }
 
+// ── #533: memory.prune must not surface stale rows via any recall retrieval
+// path (FTS lexical, sqlite-vec/ANN vector, or the default hybrid fusion) ────
+
+#[cfg(test)]
+mod prune_recall_visibility_tests {
+    use async_trait::async_trait;
+    use khive_pack_kg::KgPack;
+    use khive_runtime::{EmbedderProvider, KhiveRuntime, Namespace, VerbRegistryBuilder};
+    use lattice_embed::{EmbedError, EmbeddingModel, EmbeddingService};
+    use serial_test::serial;
+
+    /// Deterministic embedding service: distinct vector per unique text via FNV hash.
+    /// Not semantically meaningful, but reproducible — enough for a cosine-similarity
+    /// vector leg to find the seeded note by exact content match.
+    struct HashVecService {
+        dims: usize,
+    }
+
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Reviewed against the verdict at head e56f9854: zero blocking items, minors filed as follow-up. Approving.

oceanwaves630 (approved):
Reviewed against the clean-APPROVE verdict at head f342865d; minors filed as follow-up. Approving.

PR #1104, opened 2026-07-18
docs(adr): ADR-056 follow-up - helper authority pinning, SSH identity pinning, PermitTunnel, migration ledger and atomicity

ohdearquant:
## Summary

Follow-up amendment to ADR-056's 2026-07-17 iMessage-over-SSH-bridge section,
closing gaps found while hardening the merged text (#1093). Docs-only; gates
bridge implementation.

1. **Helper-side authority pinning.** `restrict` + forced `command=` pin *which*
   program the bridge host's `sshd` runs, not what that program is authorized to
   act on. Requires the helper itself to enforce, independent of adapter-side
   config, that a `poll` request's database path resolves to the exact configured
   Messages database and a `send` request's recipient is exactly the configured
   maintainer handle, rejecting any other target regardless of what stdin asks
   for.
2. **SSH client identity pinning.** The pinned known-hosts file constrains host
   trust but not which key the client authenticates with. Adds
   `-i <pinned_bridge_key>`, `-o IdentitiesOnly=yes`, and
   `-o IdentityAgent=none` (plus not inheriting `SSH_AUTH_SOCK`) to the mandated
   invocation, since without them a default or agent-offered key can authenticate
   ahead of the pinned key and land on an unrestricted `authorized_keys` entry.
3. **Tunneling over-claim.** The prior text said `restrict` disables tunneling; it
   does not. TUN-device forwarding is gated by the separate, host-wide
   `PermitTunnel` sshd_config directive (default `no`), and `authorized_keys` has
   no per-key option to disable it. Corrected to require `PermitTunnel no`
   explicitly in the bridge host's `sshd_config` during one-time setup, and fixed
   the over-claiming sentence.
4. **Migration ledger + atomicity** (source-verified at
   `crates/khive-db/src/migrations.rs`). (a) The `ServiceSchemaPlan` ledger table
   is `_schema_versions` (keyed by `service`+`migration_id`), not
   `_schema_migrations` (the separate core ADR-015 versioned-migration ledger);
   the ADR text named the wrong table in two places. (b) `apply_schema_plan`
   applies `up_sql` and records the `_schema_versions` row as two un-transactioned
   steps (migrations.rs:66, :69-76); a crash between them re-runs the migration.
   The cursor-widening migration retiring `CREATE TABLE IF NOT EXISTS` must
   therefore be crash-idempotent or self-wrap `BEGIN`/`COMMIT`. Notes a general
   khive-db transactional fix as an out-of-scope follow-up.
5. **Bounded-drain latency.** Added the worst-case ingestion-latency bound implied
   by the existing 200-row-page / 10-page-cap / `KHIVE_IMESSAGE_POLL_SECS` shape.

Also tightens the wire protocol, cursor-identity encoding, startup ordering, and
migration-slot demarcation into one internally consistent normative contract. Full
class table lives in the khive workspace, not this body.

## Note

The task literally requested "`PermitTunnel no` in the `authorized_keys` options."
Per `sshd(8)`, `PermitTunnel` is an `sshd_config`-level directive, not an
`authorized_keys` per-key option (`authorized_keys` only has `tunnel="n"`, which
pins an allowed device number rather than disabling forwarding). I specified the
technically correct location, the bridge host's `sshd_config`, rather than
inventing a nonexistent `authorized_keys` keyword, to keep the ADR text accurate.

## Test plan

- Docs-only change; no cargo build performed per task scope.
- `deno fmt --check docs/adr/ADR-056-channel-transport-layer.md` passes.
- `scripts/lint-adr-refs.sh` passes.
- Draft pending maintainer text review.

### DIFF
--- a/docs/adr/ADR-015-schema-migrations.md
+++ b/docs/adr/ADR-015-schema-migrations.md
@@ -177,9 +177,24 @@ need. Pack schema declarations add pack-specific tables that depend on a backend
 but don't belong to the core substrate (e.g., GTD lifecycle audit, memory pack
 indices, brain pack posterior state).
 
-The two never overlap. Core tables (`entities`, `graph_edges`, `notes`, `events`,
+The two never overlap, with one recorded exception: a core-lane versioned migration MAY carry
+pack-owned logical state when that state lives inside a table the core migration ledger already
+owns end to end. Core tables (`entities`, `graph_edges`, `notes`, `events`,
 plus per-namespace FTS5 / vector tables) are owned by `khive-db` migrations.
 Pack-auxiliary tables are owned by the pack.
+
+**The pack-owned-state-in-a-core-migration exception.** V5 (`unique_comm_message_external_id`) and V6
+(`brain_retune_driver`) established the precedent: both are core `khive-db` versioned migrations that
+evolve schema a pack -- comm, brain -- logically owns, because that schema was placed inside a table
+the core ledger already versions rather than in a boot-time pack declaration of its own. V11
+(`comm_channel_cursor_source_key`, ADR-056 Amendment 2026-07-17) follows the same precedent: it widens
+the pack-owned `comm_channel_cursor` table's key and adds a column, shipped as a core `khive-db`
+versioned migration rather than a pack schema declaration, because `comm_channel_cursor` already
+carries a forward-only version history in the core ledger and a boot-time `CREATE TABLE IF NOT EXISTS`
+cannot express a widened uniqueness key or a backward-compatible column add against existing rows the
+way a versioned migration can. The exception is narrow: it applies only to evolving a table's shape
+that a core migration already owns, never to introducing a wholly new pack-auxiliary table through the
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Reviewed the security follow-up in full: client identity pinning verified against ssh_config(5) semantics, PermitTunnel placement verified against sshd_config(5), helper-side authority pinning correctly enforced server-side, migration ledger and crash-idempotency requirements consistent with the schema runner's actual behavior, and the two rationale corrections (tunnel forwarding operates below the command layer; latency bound stated in ticks) are accurate. Acceptance properties 17-20 well-formed.

oceanwaves630 (approved):
Re-approved at 9e6eeb98a. Rounds 3-5 verified: helper-side inbound identity filtering with scan-floor checkpoint semantics, single-entry authorized_keys provisioning with directory-chain non-writability, conditional tick latency bound, and the withhold-property content carve-out. Delta-read each round.

PR #1114, opened 2026-07-18
chore: replace unresolvable internal references in comments with technical rationale

ohdearquant:
Comments and doc text across several crates pointed at identifiers, labels, and paths belonging to an internal process rather than explaining the technical reason for the code. None of those references can be resolved by anyone reading this repository, so they carried no information while displacing the rationale that belongs there.

Each site was rewritten into self-contained prose that keeps the technical reason and drops the framing. Where a comment consisted *only* of such a pointer, the underlying reason was reconstructed from the code it describes rather than deleting the comment. Two perf scripts additionally cited planning documents that are not part of this repository; those citations were dropped.

**Scope:** 32 files, +189/-198. Every changed line is one of:

- a comment or doc comment,
- an assertion message string (emitted only on failure),
- one test-corpus manifest,
- one test-data filename in the secret-gate false-positive tests, changed to a filename that does not mirror the internal naming scheme.

No file renames, no signature changes, no control flow touched.

Terms that look similar but are product vocabulary were deliberately left alone: GTD's "blocker task" (a task blocking another's `depends_on`), the `codex` session-mirror provider in `khive-pack-session`, policy/DMARC/liveness "verdict", and each ADR's own internal `C1`/`M1` constraint numbering.

**Verification:**

- `cargo test -p khive-runtime --lib secret_gate` → 173 passed, 0 failed. This is the one change that alters test *input* rather than a comment, so it was executed rather than only compiled.
- `cargo check -p <crate> --all-targets` → clean across the 12 touched crates.
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` → clean on the crates carrying doc-comment edits (`khive-runtime`, `khive-pack-schedule`, `kkernel`).
- `cargo fmt --all -- --check` → clean.

One pre-existing `deno fmt` diff in `crates/khive-db/docs/api/text.md` (line 91, emphasis marker style) reproduces on unmodified `main` and is unrelated to this branch.

### DIFF
--- a/crates/khive-db/docs/api/text.md
+++ b/crates/khive-db/docs/api/text.md
@@ -46,7 +46,7 @@ See `crates/khive-db/src/stores/text.rs` — private fn `sanitize_fts5_query`.
    Chars replaced with space: `(`, `)`, `,`, `:`, `-`, `.`, `/`
    (`/` added: FTS5's MATCH-expression parser rejects a bareword containing
    `/` as a syntax error, e.g. the throughput query `GB/s`)
-2. **Remove** remaining FTS5 operator characters (H1: `~`, `!` added; issue
+2. **Remove** remaining FTS5 operator characters (`~`, `!` added; issue
    #388: `$` added — FTS5's MATCH-expression parser treats a bareword
    starting with, containing, or consisting solely of `$` as a syntax error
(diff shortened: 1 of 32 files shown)


oceanwaves630 (approved):
Reviewed and approved.

Issue #1113, opened 2026-07-18
Chinese-language queries do not rank: FTS5 has no CJK tokenizer, dense leg cannot carry fusion alone


ohdearquant:
## Summary

Chinese-language queries do not rank meaningfully. Across 10 distinct Chinese queries, results collapsed to essentially the same generic cluster regardless of query content, with scores near-flat (~0.014–0.016).

## Measured evidence

- **Cross-query result overlap**: mean pairwise id-overlap of **5.84 / 8** for Chinese queries, against **0.04 / 8** for an English control set. The Chinese queries are returning nearly the same documents no matter what is asked.
- **Paired A/B on the same underlying records**: an English query (`sister Rutgers NYU Baruch college applications`) places both ground-truth records at ranks 1–2. The Chinese equivalent (`妹妹 大学 申请`) misses both entirely.
- **Recall@3 means**: EN 2.87 vs ZH 2.57 — but the ZH average is lumpy rather than uniformly worse. Proper-noun-anchored Chinese queries perform comparably to English; **relation-noun Chinese queries collapse**.
- The one solid Chinese hit was an exact substring match (`睡眠呼吸暂停`), i.e. it succeeded lexically, not semantically.

## Root cause (hypothesis, consistent with all of the above)

Two compounding failures:

1. **FTS5 has no CJK tokenizer configured.** Chinese text is not word-segmented, so the lexical leg contributes noise or nothing for most Chinese queries. Exact-substring hits still work, which is exactly the one case that succeeded.
2. **The embedding model's CJK representations are too weak to carry fusion alone.** With the lexical leg dead, the dense leg is unaided.

These interact badly with the live fusion weights. `memory.recall` defaults to `Weighted { vector: 0.7, text: 0.3 }` (`crates/khive-pack-memory/src/config.rs:396-398`), so for a Chinese query the 0.3 text component is effectively dead weight and the result rides entirely on the dense leg at 0.7. That also explains the proper-noun/relation-noun split: proper nouns survive tokenization as distinctive substrings, relation nouns do not.

## Proposed direction

1. **FTS5 CJK tokenization** (ICU, or trigram as a simpler stopgap). Expected to be a larger win for Chinese than an embedder swap alone, because it repairs the leg that is currently contributing nothing rather than trying to compensate for it.
[...]

PR #1116, opened 2026-07-18
fix(pack-memory): fuse recall across every configured embedding engine (#1115)

### LINKED ISSUE #1115: recall: fuse across all configured embedding engines instead of routing CJK queries to one
Recall should search every configured embedding engine and fuse the results. Today a query containing CJK text is routed to the multilingual engine *instead of* the others, rather than in addition to them, and the write path for knowledge sections embeds with the default engine only.

## What exists already

Multi-engine support is largely built, which narrows this issue considerably:

- `RuntimeConfig` already carries `embedding_model` plus `additional_embedding_models: Vec<EmbeddingModel>`; `configured_embedding_models` (`khive-runtime/src/config.rs:498`) unions them and `build_embedder_registry` registers each one.
- Storage is already per-engine: vectors live in a `vec_<model>` table keyed by `embedding_model` (`khive-db/src/stores/vectors.rs`), so N embeddings per subject is a supported shape today, with no schema migration required.
- `backfill_missing_embeddings` (`khive-runtime/src/retrieval.rs:568`) already iterates every registered model and fills any record missing a vector for it.
- The recall read path already fans out across all registered engines in the general case (`khive-pack-memory/src/handlers/common.rs:822-850`).

## The actual defects

**1. CJK queries narrow to one engine instead of fusing.** In `collect_recall_vector_hits`, when `use_multilingual` is set the model list is *replaced* by the single multilingual model:

```rust
match multilingual_model {
    Some(model) => { multilingual_routed = true; vec![model] }
    None => names,
}
```

So a CJK query searches the multilingual engine alone and never sees the primary engine's vectors, while a non-CJK query searches everything. This is an either/or switch on a code path that is otherwise already a union. Decided by `enable_multilingual_routing && needs_multilingual(query)` at `khive-pack-memory/src/handlers/recall.rs:133`.

**2. Configured order is not preserved.** `configured_embedding_models` ends with `models.sort_by_key(|model| model.to_string())`, so the engine list is alphabetical rather than as-configured. Any consumer that treats the list as ordered (primary first) gets the wrong answer.

**3. Knowledge-section writes are single-engine.** `khive-pack-knowledge/src/knowledge/index_handler.rs:128` and `sections_index.rs:189` call `embed_document_batch`, which uses the default model only. Records written through that path have no vector for any secondary engine until a backfill runs.

## Proposed change

- Delete the narrowing branch: always search every configured engine and fuse across engines plus the existing text leg. This removes `use_multilingual` / `multilingual_routed` and the `enable_multilingual_routing` / `multilingual_model` scoring knobs rather than reinterpreting them.
- Preserve configured order in `configured_embedding_models` (dedupe without sorting).
- Make the knowledge write path embed with every configured engine.
- Default configuration stays single-engine (`all-minilm-l6-v2`) so existing deployments are unaffected; a second engine becomes opt-in via `additional_embedding_models`.

## Note on scope

CJK *lexical* matching is a separate problem: FTS5 has no CJK tokenizer configured, so the text leg cannot match Chinese queries at all. That is tracked in #1113 and is not addressed here. This issue is about the vector leg only.

ohdearquant:
Closes #1115.

Recall fanned out across every configured embedding engine in the general case, but when a query contained CJK the model list was *replaced* by a single multilingual engine rather than extended with it:

```rust
match multilingual_model {
    Some(model) => { multilingual_routed = true; vec![model] }
    None => names,
}
```

So a Chinese query searched the multilingual engine alone and never saw the primary engine's vectors, while an English query searched everything. A record embedded before a second engine was configured was therefore invisible to exactly the queries that engine exists to serve.

The whole block collapses to `registered_embedding_model_names()`. An explicit `embedding_model` argument still pins to that one engine.

## Changes

- **Recall fuses across all configured engines.** Removes `use_multilingual`, `multilingual_routed` (including the response JSON key), and the `enable_multilingual_routing` / `multilingual_model` scoring knobs. `needs_multilingual` had no remaining callers and is gone. `contains_cjk` is kept: it independently gates the FTS bypass path, which is #1113's territory, and that path's behaviour is unchanged.
- **Configured engine order is preserved.** `configured_embedding_models` alphabetized via `sort_by_key`, so the declared primary was not necessarily first. It now dedupes while preserving declaration order.
- **Knowledge atom indexing writes a vector per configured engine.** `knowledge.index` embedded with the default engine only. Secondary engines are now written best-effort through the existing per-model `vec_<model>` store, mirroring `backfill_missing_embeddings`; a secondary failure is logged and does not change the `indexed`/`failed`/`skipped` counters, which stay keyed on the default engine as before.

No schema migration: vectors are already keyed per engine by `embedding_model`, so N embeddings per subject is an existing shape.

## Verification

`test_cjk_query_finds_note_embedded_by_primary_engine` is the regression guard. It registers a primary engine, writes a note through it, *then* registers a second engine whose name would win the old substring heuristic, and recalls with a CJK query.

Confirmed meaningful rather than vacuous: against the pre-change code it fails with `got: []` (exit 101); after the change it passes. Also added order-preservation and dedupe tests for the config change, and a test asserting `knowledge.index` populates both engines' vector tables.

Gates: `cargo check` / `clippy -D warnings` / `test -p khive-pack-memory` / `test -p khive-pack-knowledge` / `fmt --check` all clean.

## Known gap, deliberately not addressed here

`knowledge_sections` embedding remains single-engine. That table stores one inline `embedding BLOB` column rather than the per-engine `vec_<model>` tables everything else uses, so making it multi-engine needs either a schema change or migrating section vectors onto the `VectorStore` abstraction, plus a matching read-path change. That is a design decision rather than a mechanical fix, so it is left for follow-up instead of half-built here.

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -2514,17 +2514,20 @@ mod tests {
     #[test]
     #[serial]
     fn resolver_uses_config_file_engines_over_defaults() {
-        // Ensure env vars cannot leak into either branch.
+        // Ensure a stale ambient value cannot leak into either branch.
         std::env::remove_var("KHIVE_EMBEDDING_MODEL");
-        std::env::remove_var("KHIVE_ADDITIONAL_EMBEDDING_MODELS");
+        // The shipped default is single-engine, so leaving the additional list
+        // unset would make "the config file overrode the default" and "there
+        // was nothing to override" produce the same empty result, and the
+        // final assertion below would stop discriminating. Declare one
+        // deliberately so the override remains observable.
+        std::env::set_var("KHIVE_ADDITIONAL_EMBEDDING_MODELS", "paraphrase");
 
         let default_cfg = RuntimeConfig::default();
         let default_primary = format!("{:?}", default_cfg.embedding_model);
-        // Default ships a non-empty additional-engine list (the multilingual
-        // model). The single-engine config file below must override it.
         assert!(
             !default_cfg.additional_embedding_models.is_empty(),
-            "precondition: default config has additional engines"
+            "precondition: default config must carry an additional engine for this test to discriminate"
         );
 
         let dir = tempfile::tempdir().expect("temp dir");
@@ -2565,6 +2568,8 @@ default = true
             "config file declares one engine; additional list must be empty (not the default's)"
         );
         assert_eq!(resolved.db_path, None, ":memory: must map to in-memory db");
+
+        std::env::remove_var("KHIVE_ADDITIONAL_EMBEDDING_MODELS");
     }
 
     /// Regression for #379: when the loaded config file has NO `[[engines]]`
--- a/crates/khive-pack-knowledge/src/knowledge/index_handler.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/index_handler.rs
@@ -1,9 +1,11 @@
 //! Index handler: embed atoms and build/persist the Vamana ANN index.
 
+use std::sync::Arc;
+
 use serde_json::{json, Value};
 
 use khive_runtime::{KhiveRuntime, NamespaceToken, RuntimeError};
-use khive_storage::types::{SqlStatement, SqlValue};
+use khive_storage::types::{SqlStatement, SqlValue, VectorRecord};
 use khive_types::SubstrateKind;
 
 use super::schema::{Atom, IndexParams};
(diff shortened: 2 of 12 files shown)


ohdearquant:
## Scope of this PR

Both remaining per-engine serialization paths are tracked in #1121 and are deliberately not in
this PR:

- the serial per-model loop in `crates/khive-pack-memory/src/handlers/common.rs` (ANN readiness,
  search, widening, hydration, sqlite-vec fallback)
- serial model batches and per-vector `insert` calls in
  `crates/khive-pack-knowledge/src/knowledge/index_handler.rs`

They are real and they are accurately located. This PR carries the correctness half of
multi-engine: fusion across every configured engine's hits rather than only the first
(`combine_vector_sources_union` instead of a passthrough over N sources), per-engine failure
isolation so one unavailable engine degrades to the remaining engines instead of failing the
whole recall, the removal of either/or routing behind `enable_cjk_fts_bypass`, and the
single-engine shipped default.

The split is intentional. The correctness changes alter behavior and are covered by regression
tests, including one that fails before the fusion fix. The serialization work is a hot-path
concurrency refactor with no behavior change to test against, which should be measured rather
than asserted and should not hold the correctness fixes behind it.

Note the shipped default is single-engine, so the additive latency applies to deployments that
configure a second engine.

oceanwaves630 (approved):

PR #1119, opened 2026-07-18
ci: commit the khive-merge lockfile and pin its cargo invocations

ohdearquant:
`khive-merge` declares its own `[workspace]` table so it stays standalone-compilable while
excluded from the root workspace. That means it resolves a second, separate dependency graph:
`crates/Cargo.lock` does not cover it, and its own lockfile was untracked.

`phase_lockfile` runs the root workspace under `--locked` so CI compiles exactly the dependency
versions that landed as a reviewed diff (`scripts/ci.sh:9-12`). The five khive-merge invocations
omitted `--locked`, so that crate re-resolved its graph from scratch on every CI run. A dependency
release published after review could execute build code in CI without ever appearing in a diff.
The crate is published to crates.io, so this is the dependency graph of a shipped artifact.

## Changes

- Commit `crates/khive-merge/Cargo.lock` (3166 lines).
- Add `--locked` to all five khive-merge cargo invocations in `scripts/ci.sh`
  (`check`/`clippy`/`test` in `phase_forward_deployed`, the no-stub clippy pass, and the
  macOS PR compile check).
- Drop the `.gitignore` rule this PR originally added.

## Verification

```
$ RUSTFLAGS="-D warnings" cargo check --manifest-path crates/khive-merge/Cargo.toml \
    --all-targets --locked
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 38.08s
rc=0

$ shellcheck scripts/ci.sh
rc=0
```

## Note on the direction change

This PR originally added a `.gitignore` rule for the standalone lockfile, treating it as a build
byproduct. That was backwards: the lockfile's absence was the defect, not its presence. Ignoring
it would have made the unpinned resolution permanent and invisible. The correct fix gives the
excluded crate the same guarantee the workspace already has.


oceanwaves630 (approved):

PR #1124, opened 2026-07-19
docs(adr): ADR-117 session continuity — cross-session search and remote ingestion

ohdearquant:
## ADR-117: Session Continuity — Cross-Session Search and Remote Ingestion

Docs-only. Adds `docs/adr/ADR-117-session-continuity-search.md` and its README index row.

### What it specifies

The session pack (`khive-pack-session`) already persists agent-session transcripts — a
background mirror tails Claude Code, Codex, and ChatGPT exports into `sessions` /
`session_messages` / `session_mirror_cursor`. The one capability it lacks is **search over
message content**: `session.resume` is fetch-by-id and `session.list` is a recency browse. The
text is stored; it is simply not indexed for retrieval.

ADR-117 specifies that capability and its contract:

- **D1** — a `session.search` verb, FTS5-first over `session_messages.text`, reusing the memory
  pack's fusion rather than re-implementing it.
- **D2** — the schema stays embedding-ready (content-addressed message rows so a later embed
  backfill is idempotent); a vector signal is a gated later addition, sequenced behind the
  per-engine retrieval parallelization work (#1121) for the same reason the multi-engine default
  change (#1116) is.
- **D3** — cross-machine ingestion is the existing mirror with a remote sink, not a new tool: a
  fresh machine's zero-offset cursor makes cold backfill and live tail one code path, and
  `provider_session_id` + content addressing make re-streaming idempotent.
- **D4** — tenant scoping is Gate-enforced (ADR-007) and lands **in the same change** as the
  verb, with an isolation test proving a cross-tenant query returns nothing. Blocking, not
  aspirational.
- **D5** — retention is bounded by a storage cap with visible oldest-first aging, plus day-one
  user-initiated deletion (single session and full wipe) as a curation-layer operation.

### Scope

Proposed. This ADR specifies the capability and its contract; it does not schedule the
implementation. Depends on ADR-007, ADR-013, ADR-021, ADR-025, ADR-083.

PR #1125, opened 2026-07-19
docs(adr): ADR-117a — session identity + tenant isolation (ADR-117 D1/D2/D4)

ohdearquant:
## ADR-117a — Session Identity and Tenant Isolation

The first mechanism follow-on of [ADR-117](https://github.com/ohdearquant/khive/blob/main/docs/adr/ADR-117-session-continuity-search.md) (the session-continuity direction, merged in #1124). Carries D1/D2/D4 with proof at source.

### What it decides

- **Scoped-identity migration (D2).** Promote the mirror's currently-nullable `namespace` column to the non-null tenant-scope column; make uniqueness tenant-scoped — `(namespace, provider_session_id)` on `sessions` and `(namespace, session_id, id)` on `session_messages`; add a content-hash column as a dedup/integrity adjunct that is explicitly never the identity. Embedding-ready for the gated later vector signal.
- **Fail-closed handler-seam predicate (D4).** Isolation is enforced where the rows exist, not at the Gate. Construction-primary: the scope is the request's resolved `NamespaceToken` (ADR-096), never a caller argument; the `WHERE namespace = :scope` term is non-widenable and always present; no positive scope ⇒ the verb refuses. So the authorization Gate's fail-open-on-infra-error default cannot leak session rows.
- **ADR-018 Amendment 2.** Records a fail-closed verb class in ADR-018 itself, so the Gate ADR stays the source of truth for its own contract. `session.search` is its first member.
- **`session.search` (D1).** FTS5-first over `session_messages.text`, tenant-scoped, reusing the memory pack's FTS + RRF ranking (ADR-021) rather than re-implementing it.
- **Tests in the same change (the ADR-117 D1 hard gate).** A cross-account isolation test and a no-scope-refusal test land in the same PR as the verb; the verb is not a searchable surface until they pass.

### Grounding

- The Gate is pre-dispatch, row-blind, and fail-open on infra error (ADR-018), so row isolation cannot be a property of its decision — enforcement is the construction-primary handler predicate; the amendment only codifies the class contract.
- Hosted multi-tenant *authentication* of the scope composes with the connection-identity mechanism ADR-096 defers to a separately gated ADR (peer-credential capture, a connection principal threaded into the Gate request); this ADR makes the predicate correct and fail-closed so it is already safe when that mechanism lands, and does not re-solve or weaken that gate.
- The session mirror schema is pack-owned (ADR-028): the migration evolves the session pack schema, not the core migration chain.

### Scope

Docs-only. The mechanism implementation (migration, predicate, verb, tests) lands in a follow-up PR gated on this ADR. Deletion/retention (ADR-117b), the resume/export continuity bridge (ADR-117c), cross-machine ingestion (ADR-117d), and any vector signal remain their own follow-ons.

### DIFF
--- a/docs/adr/ADR-018-authorization-gate.md
+++ b/docs/adr/ADR-018-authorization-gate.md
@@ -692,3 +692,44 @@ missed by a request using the alias form.
   dispatch path already pays elsewhere.
 - Policies authored against alias forms (if any) must be re-authored against
   canonical ids: a one-time migration cost at adoption.
+
+---
+
+## Amendment 2 (2026-07-19) — Fail-closed verb class
+
+Specified by [ADR-117a](ADR-117a-session-identity-tenant-isolation.md); recorded here so this ADR
+remains the source of truth for its own contract.
+
+### Context
+
+The base contract fails open on gate infrastructure `Err` (§"Fail-open on gate Err"): a gate that
+errors before producing a decision proceeds, and only an explicit `Deny` blocks. Amendment 1 §4
+already carved one fail-closed exception — a wire verb that resolves to no registered canonical id is
+denied before dispatch. Some verbs return rows so sensitive that their safety must not depend on the
+Gate at all — the Gate is pre-dispatch and row-blind, so it can neither see nor isolate matched rows,
+and its fail-open default must not become a data-leak path for them.
+
+### Decision
+
+A verb may declare membership in a **fail-closed verb class**. A member carries a handler-seam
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Contract and wording reviewed across all revisions; the scoped-identity migration, fail-closed predicate boundary, and the triple-gated availability conditions are precise and honest. Approving at c7a4b9f4c.

Issue #1127, opened 2026-07-19
session mirror: per-tick scan is O(total corpus) not O(delta) — full re-walk + per-file stat every poll


ohdearquant:
## Summary

`run_mirror_service` (`crates/khive-pack-session/src/mirror/service.rs`) re-discovers and re-stats the **entire tracked transcript set on every poll tick** (default `KHIVE_MIRROR_POLL_SECS=2`). Each tick re-walks the Claude Code projects tree and the Codex sessions tree (`scan_cc_jsonl_files` / `scan_codex_jsonl_files`) and calls `std::fs::metadata().len()` on every discovered file to detect growth (service.rs ~L283). Cost scales with the **total accumulated corpus**, not with what actually changed.

## Impact

At a realistic corpus (tens of thousands of transcript files, ~20 GB across the two trees), each 2-second tick performs tens of thousands of `stat` syscalls plus two full recursive directory walks — thousands of stats/sec sustained, indefinitely, the overwhelming majority over cold files that will never grow again. On a first run with `KHIVE_MIRROR_BACKFILL=true`, the initial catch-up additionally re-reads and ingests the entire backlog in one pass. The `// Cheap when idle: only metadata().len() stat calls` note (service.rs:9) held at small corpus but no longer does as the corpus grows.

## Fix direction

Make steady-state cost track the delta:

- **Directory-mtime short-circuit**: stat the parent/slug directories first and skip re-listing and re-stating a subtree whose mtime is unchanged since the last tick.
- **Cold-file backoff / recency scoping**: a file whose length is unchanged and whose mtime is older than a threshold is cold — drop it from the per-tick stat set until its directory mtime bumps. The large tail of completed transcripts then costs nothing per tick.
- **Poll-interval default**: 2s is finer than session-search freshness needs; a larger default (e.g. 30-60s) cuts fixed overhead with no product impact.

Net: per-tick work becomes ~O(actively-growing files), independent of total corpus size.

ohdearquant:
## Root cause confirmed at code level: per-tick corpus-wide dir-walk + double-stat, not content re-read

Read the mirror loop (`service.rs` `run_mirror_service`) end to end. The cost is entirely in **discovery**, not in reading or re-ingesting file contents:

**Every tick (`poll_interval`, default 2s):**
[...]

PR #1128, opened 2026-07-19
docs(adr): ADR-079 Amendment 1 — delta-log restart classifier and mmap re-adoption

ohdearquant:
## Summary

Production measurement at ~553K entity/note vectors + ~358K knowledge sections surfaced two warm-path costs the base ADR leaves in place:

1. **Restart classification is a full-corpus scan** — `live_content_hash` reads and normalizes every vector row per `(namespace, model)` at every daemon start (minutes of multi-core CPU + a transient multi-GB heap spike), and a single write since the last checkpoint escalates to a full background rebuild that reads the corpus into heap again.
2. **The served index never returns to file-backed memory** — rebuilds produce `VectorStorage::Owned` (full f32 corpus in anonymous heap, ~2.5-3 GB at current size vs a 185 MB on-disk segment); the existing `load_v2_fast`/`mmap_vectors` fast path is effectively dead code on an active system.

## Amendment

- **A. `ann_write_log`**: a persisted per-write delta log; restart classification becomes one indexed `MAX(seq)` read against the segment's `last_applied_seq`; stale tails apply incrementally through ADR-052 `insert`/`tombstone` (no corpus scan, no full rebuild); log compacts at checkpoint.
- **B. mmap re-adoption**: after every successful `save_atomic` the bridge reopens the segment via the mmap path and swaps out the Owned build product — steady-state served vectors are file-backed.

Measurements are recorded in the companion daemon-resource investigation (#1127 thread); a dedicated issue for this lane follows.

Docs-only. Implementation follows separately and is gated on this amendment's sign-off.

### DIFF
--- a/docs/adr/ADR-079-ann-persistence-warm-path-integration.md
+++ b/docs/adr/ADR-079-ann-persistence-warm-path-integration.md
@@ -415,8 +415,10 @@ CREATE TABLE ann_write_log (
 CREATE INDEX idx_ann_write_log_ns_model_seq ON ann_write_log(namespace, embedding_model, seq);
 
 -- Durable per-consumer watermark registry, required by the cross-consumer
--- compaction invariant below. Updated after each consumer's segment commit;
--- a stale row under-compacts (safe), never over-compacts.
+-- compaction invariant below. A consumer registers its row (watermark 0)
+-- BEFORE persisting or serving any extended-format segment, and raises it
+-- monotonically after each segment commit; a stale row under-compacts
+-- (safe), never over-compacts.
 CREATE TABLE ann_consumer_watermark (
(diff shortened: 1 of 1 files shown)


ohdearquant:
Design review round 4 at head 22c8772: APPROVE, zero findings. Rounds 1-3 findings (watermark classifier totality, publication linearization, tail-replay replacement semantics, zero-watermark baseline, empty-corpus rule ordering, cross-consumer compaction registry) all verified resolved. Reviewer confirmed the first-checkpoint crash interleaving is closed by register-before-persist + MIN-bounded compaction.

oceanwaves630 (approved):
Reviewed the amendment text at this revision. Approving.

PR #1130, opened 2026-07-19
chore(merge): bump crossbeam-epoch and memmap2 in the pinned lockfile

ohdearquant:
Resolves the two locked dependency advisories in `crates/khive-merge/Cargo.lock` (committed and pinned via #1119):

- crossbeam-epoch 0.9.18 → 0.9.20 (RUSTSEC-2026-0204)
- memmap2 0.9.10 → 0.9.11 (RUSTSEC-2026-0186)

Verification: `cargo audit --file crates/khive-merge/Cargo.lock` now exits with only the pre-existing allowlisted warning (RUSTSEC-2026-0002), and `cargo check --locked` passes in the crate.

Lockfile-only change; no source edits.

### DIFF
--- a/crates/khive-merge/Cargo.lock
+++ b/crates/khive-merge/Cargo.lock
@@ -442,9 +442,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


ohdearquant:
Design review at head e468b0c: APPROVE, zero findings. Verified lockfile-only diff (four version/checksum substitutions), semver-compatible parent requirements, cargo audit clean of both named advisories (sole warning is the pre-existing allowlisted lru advisory), and cargo check --locked green in an isolated target dir.

PR #1131, opened 2026-07-19
docs(adr): ADR-079 Amendment 1 addendum — global-scope consumers and join-filtered corpora

ohdearquant:
Extends the just-merged Amendment 1 (#1128) with the contract needed for the memory pack's namespace-global note index before its implementation lands:

- **Wildcard registration**: a consumer whose corpus spans all namespaces for a model registers one `(consumer, '*', embedding_model, watermark)` row; register-at-0-before-persist and monotonic raise apply unchanged.
- **Compaction minimum includes wildcard rows**: pair compaction computes its MIN over `namespace = ?ns OR namespace = '*'`, so a global consumer's tail survives in namespaces that first appear after registration.
- **Classification unchanged**: global consumers classify and tail-read under their own corpus predicate; `seq` is database-wide monotone; decision rule 4 applies to the wildcard row as to any row.
- **Soft-delete visibility for join-filtered corpora**: note soft-deletes append no log row, so a Hot segment may retain soft-deleted candidates until the next checkpoint; the recall path's `deleted_at IS NULL` post-filter hides them at serve time, and consumers without such a filter must not adopt the amendment for a join-filtered corpus. Tail replay measures source-row absence under the full corpus predicate: a final upsert whose subject fails the predicate replays as a delete, not a Cold escalation.

Docs-only. Blocks the memory-pack implementation PR; the knowledge-pack implementation is independent of this addendum.

### DIFF
--- a/docs/adr/ADR-079-ann-persistence-warm-path-integration.md
+++ b/docs/adr/ADR-079-ann-persistence-warm-path-integration.md
@@ -496,11 +496,15 @@ CREATE TABLE ann_consumer_watermark (
      `S`, the consumer raises its row monotonically
      (`UPDATE ... SET watermark = MAX(watermark, S)`); a crash in between leaves the smaller
      registered watermark, which under-compacts — safe.
-  3. **Compact through the registry minimum only.** Compaction deletes rows with
-     `seq <= (SELECT MIN(watermark) FROM ann_consumer_watermark WHERE namespace=? AND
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Reviewed the addendum text at this revision. Approving.

oceanwaves630 (approved):
Reviewed the amendment text at this revision. Approving.

oceanwaves630 (approved):

oceanwaves630 (approved):
Reviewed the amendment text at this revision. Approving.

PR #1132, opened 2026-07-19
feat(ann): write-log restart classifier, codes.bin mmap segment, and consumer watermark registry (ADR-079 Amendment 1)

ohdearquant:
Implements ADR-079 Amendment 1 (#1128) for the knowledge pack — the daemon memory lane.

## What changes

**khive-db**
- Migration 011: `ann_write_log` (AUTOINCREMENT seq) + `ann_consumer_watermark` registry table.
- Write path appends one log row per vector upsert/delete in the same transaction as the mutation.

**khive-vamana**
- `codes.bin` becomes the fifth checksummed v2 segment (SQ8 codec params + codes persisted); `load_v2_fast` mmaps codes instead of retraining, killing the retrain-on-load anonymous allocation and the full-corpus page touch.
- Extended v2 commit record: unconditional 41-byte trailer (flags, `last_applied_seq`, `codes_hash`); parser accepts base and extended lengths, so pre-amendment segments still load (and classify Cold per rule 2).
- New `read_commit_info` — reads the commit record without loading the segment.

**khive-pack-knowledge**
- The O(corpus) content-hash restart classifier is replaced by the 8-rule first-match decision table over the write log: Hot restarts are mmap-only (zero corpus IO), small tails replay final-state per subject (Stale-tail), large tails serve the checksum-valid stale segment while rebuilding in the background (Stale-rebuild).
- Three-step registry protocol: register-at-0-before-persist, monotonic raise-after-commit, compaction bounded by the per-pair registry MIN (never a checkpoint-local watermark).
- Every checkpoint/rebuild re-adopts its own segment as mmap, so steady-state memory is file-backed page cache rather than anonymous heap.

## Notes for review
- `ann_rebuild_threshold` is env-only for now (`KHIVE_ANN_REBUILD_THRESHOLD`, default 0.20, clamped to (0,1]); TOML/CLI binding follows in a config PR.
- `compact_log` uses the per-pair predicate; the wildcard-inclusive form specified for global-scope consumers rides the memory-pack PR alongside its addendum (#1131).
- Portability digest pin updated (the unconditional trailer changes `to_bytes` output); verified identical across default and `--all-features` builds.

## Validation
- `cargo test --workspace --all-features` green on this tree (4200+ tests), including 4 new interleaving tests from design review: zero-watermark first-write, delete-all→Empty, compaction bounded by pair minimum, pre-amendment record→Cold.

### DIFF
--- a/crates/khive-db/sql/011-ann-write-log.sql
+++ b/crates/khive-db/sql/011-ann-write-log.sql
@@ -1,14 +1,32 @@
 -- ANN write log: per-vector-write delta records consumed by the restart
 -- classifier (ADR-079 Amendment 1). AUTOINCREMENT is load-bearing: seq must be
 -- strictly monotone and never reused so a persisted watermark stays comparable
--- across log compactions.
+-- across log compactions. kind/field carry the vector row's own scope so a
+-- consumer whose corpus is a subset of a shared vec table can filter its tail
+-- with the same predicate as its corpus scan.
 CREATE TABLE IF NOT EXISTS ann_write_log (
     seq             INTEGER PRIMARY KEY AUTOINCREMENT,
     namespace       TEXT NOT NULL,
     embedding_model TEXT NOT NULL,
+    kind            TEXT NOT NULL,
+    field           TEXT NOT NULL,
     subject_id      TEXT NOT NULL,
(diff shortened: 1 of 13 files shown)


oceanwaves630 (approved):

PR #1140, opened 2026-07-19
docs(ann): qualify zero-live rule and document segment-root migration

### LINKED ISSUE #1135: docs(adr): qualify ADR-079 Amendment 1 zero-live rule for join-filtered corpora
Review follow-up from PR #1131.

The zero-live decision rule (rule 5) states that a zero live count means the segment has nothing to serve. For join-filtered corpora this claim needs qualification: soft deletes that satisfy the join predicate without removing vector rows can leave an empty log tail while the persisted segment still contains filtered candidates. Adoption remains safe because recall post-filters hydrated hits against the join predicate, but the ADR text should state the qualification explicitly rather than implying the segment is content-empty.

### LINKED ISSUE #1136: docs(ann): document segment-root migration from data_dir/ann to <db-file>.ann
Review follow-up from PR #1132.

ANN segment directories moved from `data_dir/ann/<hex>` to a database-scoped root `<db-file>.ann/<hex>` so co-located databases cannot adopt each other's segments. Pre-existing segments under the old root are silently orphaned: the classifier finds no commit record at the new root and performs a Cold rebuild, which is correct but leaves the old directories as unreclaimed disk. Document the migration behavior (one-time Cold rebuild after upgrade) and either sweep or document manual cleanup of the legacy `data_dir/ann` tree.

Also in scope: sweep the remaining stale path and classifier comments in `crates/khive-pack-knowledge/src/knowledge/vamana.rs` and the corresponding ADR-079 text that still describe the old `data_dir/ann` layout or the pre-tail-first rule ordering.

ohdearquant:
## Summary

Review follow-up.

- ADR-079 Amendment 1's rule 5 (zero live corpus → Empty) is qualified for
  join-filtered corpora: a zero live count under a consumer's scope predicate
  means the predicate currently selects nothing, not that the persisted
  segment file is byte-empty. Adoption stays safe either way — rule 5 drops
  the index rather than serving it, and recall post-filters any served hit
  against the same predicate.
- Documents that the ANN segment root moved from a shared
  `<data_dir>/ann/<hex>` tree to a database-scoped `<db-file>.ann/<hex>` root,
  and that pre-existing segments under the old root are not migrated in
  place — first restart after upgrade Colds and rebuilds under the new root;
  the old tree is then inert and safe to remove.
- Sweeps stale `data_dir/ann` path and content-hash-classifier comments in
  `vamana.rs`/`vamana.md` left over from the segment-root and
  restart-classifier changes.

Docs-only change, no behavior change.

Closes #1135, Closes #1136

## Test plan

- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-pack-memory -p khive-pack-knowledge` — clean
- [x] pre-commit hooks (fmt, clippy, ADR-reference lint, deno fmt) — passed

### DIFF
--- a/crates/khive-pack-knowledge/docs/api/vamana.md
+++ b/crates/khive-pack-knowledge/docs/api/vamana.md
@@ -8,12 +8,16 @@ Source: `crates/khive-pack-knowledge/src/knowledge/vamana.rs` (entire module is
 Wraps `khive_vamana::VamanaIndex` with an ID map (u32 → UUID) so search results can be
 fused with FTS5 candidates via RRF.
 
-Persistence (ADR-079): v2 binary segments are written to `data_dir/ann/<hex>/` on every
-cold-start rebuild or explicit reindex. `ensure_ann_for_model` checks the v2 segment
-directory first (content-hash gated), falling back to legacy v1 JSON rows in
-`retrieval_snapshots` for in-place upgrades, then rebuilds from the full sqlite-vec corpus
-on cache-miss. `kkernel reindex` re-persists v2 segments and calls `invalidate_snapshot` to
-clean up stale v1 rows.
+Persistence (ADR-079, Amendment 1): v2 binary segments are written to
+`<db-file>.ann/<hex>/` — a database-scoped root beside the backing database file, so
+co-located databases can never adopt each other's segments — on every cold-start rebuild
+or explicit reindex. `ensure_ann_for_model` checks the v2 segment directory first, gated
+by the write-log restart classifier (`classify_and_adopt_segment`: commit-record and
+watermark checks, then a log-table tail probe, ahead of any corpus read — see the
+ADR-079 Amendment 1 decision table for the full rule order), falling back to legacy v1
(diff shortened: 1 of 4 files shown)


ohdearquant:
Pushed 3d64077 addressing the review follow-up: the rule-5 qualification as first written asserted that the join-filtered zero-live case always classifies Empty, which contradicted the Amendment 1 evaluation-order provision (rule 6 may be tested first, adopting Hot without running the live count) and the shipped classifier. The corrected text states both branches explicitly: with a tail present, rule 5 classifies Empty on the predicate-scoped count alone; with no tail, rule-6-first Hot adoption of a fully join-filtered segment is behaviorally safe because recall post-filters hydrated hits against the same join predicate — bounded over-fetch waste, not a correctness defect, self-correcting at the next logged write. Also replaced the superseded segment-root example at the top of the section with the Amendment 1 db-scoped path so only one active-looking path definition remains.

oceanwaves630 (approved):

PR #1141, opened 2026-07-19
refactor(pack-memory): drop the ignored namespace argument from AnnKey constructors

### LINKED ISSUE #1137: refactor(pack-memory): drop the ignored namespace argument from AnnKey constructors
Review follow-up from PR #1134.

`AnnKey` is model-only (one index per model spanning all namespaces), but its constructors still accept and discard a namespace/token argument for compatibility with the pre-consolidation call shape. About 29 call sites across 6 files pass an argument that has no effect. Remove the parameter from `AnnKey::new` / `AnnKey::from_token` and update call sites so the signature states the real contract.

ohdearquant:
## Summary

Review follow-up.

`AnnKey` has been model-only since the FTS+ANN consolidation (one index per
model, spanning every namespace), but its constructors still accepted and
discarded a namespace/token argument for compatibility with the
pre-consolidation call shape. `AnnKey::new` / `AnnKey::from_token` now take
only `model`; the ~29 call sites across the pack are updated to match.

`ensure_ann_background`'s `token` parameter, previously passed straight into
the discarded `AnnKey::from_token` argument, is genuinely unused now — it is
underscore-prefixed rather than removed from the signature, since
`ensure_ann_for_model`'s sibling warm entry point keeps the same
`(rt, token, ann, model)` shape and both are called from the same call sites.

Purely mechanical; no behavior change.

Closes #1137

## Test plan

- [x] `cargo test -p khive-pack-knowledge -p khive-pack-memory -p khive-vamana` — 16 suites, all passing

### DIFF
--- a/crates/khive-pack-memory/src/ann.rs
+++ b/crates/khive-pack-memory/src/ann.rs
@@ -28,13 +28,13 @@ pub(crate) struct AnnKey {
 }
 
 impl AnnKey {
-    pub(crate) fn new(_namespace: impl Into<String>, model: impl Into<String>) -> Self {
+    pub(crate) fn new(model: impl Into<String>) -> Self {
         Self {
             model: model.into(),
         }
     }
 
-    pub(crate) fn from_token(_token: &NamespaceToken, model: &str) -> Self {
+    pub(crate) fn from_token(model: &str) -> Self {
         Self {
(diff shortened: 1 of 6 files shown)


oceanwaves630 (approved):

Issue #1138, opened 2026-07-19
ANN checkpoint publication is not linearized per scope: stale checkpoint can overwrite a newer segment and compaction hides the loss


ohdearquant:
Review follow-up on #1132 (ADR-079 Amendment 1).

## Defect

`checkpoint_raise_compact_readopt` (crates/khive-pack-knowledge/src/knowledge/vamana.rs) publishes in the order: persist segment -> raise watermark (`MAX(watermark, applied)`) -> compact log. Nothing serializes two checkpoints for the same `(namespace, model)` scope at the filesystem layer.

Interleaving:

1. Build B (applied seq 200) persists its segment, raises the watermark to 200, compacts delta rows <= 200.
2. Build A (older, applied seq 100), still in flight, then runs `persist_ann_v2` and overwrites the segment directory with the seq-100 segment. Its `raise_watermark(100)` is a no-op under MAX; compaction has already deleted rows 101..200.
3. Restart: the registry row exists, the segment reports seq 100, and the tail probe finds nothing above 100 because those rows were compacted. The classifier adopts the stale segment as current, and vectors 101..200 are silently absent until an unrelated rebuild.

The in-memory generation fence protects the live index (A's older generation loses the install race) but does not protect the on-disk publication order, so the loss only manifests after the next restart.

## Reachability

The warm path is single-flighted via the warming set, but `clear_namespace` (the write-invalidation chokepoint) removes warming entries while a build may still be in flight; a subsequent search re-warms the key and starts a second build concurrently with the first. The admin reindex path is a second concurrent producer.

## Suggested fix

Per-scope serialization or conditional publication (compare-and-swap on the persisted applied seq: refuse to overwrite a segment whose recorded seq is newer), plus a deterministic interleaving test proving an older checkpoint cannot overwrite a newer segment or compact its recoverable tail. ADR-079 already requires checkpoint publication to be linearized.

PR #1142, opened 2026-07-19
docs(adr): ADR-091 Amendment 2 — cross-process WAL pin attribution and per-session sweep

ohdearquant:
Amendment to ADR-091 recording the 2026-07-19 telemetry conclusion: during a sustained 84k-page WAL pressure window with three zero-progress TRUNCATE attempts, the in-process transaction registry's oldest span was milliseconds old at every observation — exonerating the in-process inventory and locating the pin in another process. 13 processes hold the database open directly; the checkpoint/sweep telemetry runs daemon-only, so the likely pinning processes have zero observability today.

Adds three additive, observability-first planks: (A) the registry age sweep runs in every process (observe-only in sessions, checkpointing stays daemon-owned), (B) a per-PID attribution sidecar the daemon enumerates on TRUNCATE no-progress so the next recurrence names the pinning process, (C, optional) a pin-depth probe. No enforcement or policy changes. Single-process deployments are unaffected.

Review follow-up lane: #580, ADR-091.

---

**Review scope note (final round).** The latest review round confirms all three specification
findings against this document are resolved (OS-derived census, beacon refresh rule,
handle-bound sidecar operations). The two remaining findings it raises are in code this
documentation-only diff does not modify and are tracked on their own lanes: blob access
semantics in #1145, ANN checkpoint compare-and-publish in #1138. Neither is addressable
within this diff's scope.

### DIFF
--- a/docs/adr/ADR-091-wal-snapshot-lifetime.md
+++ b/docs/adr/ADR-091-wal-snapshot-lifetime.md
@@ -738,12 +738,32 @@ question (Inventory item 4) remains unverified precisely there.
   only entries it can attribute to a dead or stale process AND that pass
   ownership validation. The daemon logs every live report alongside its
   existing no-progress WARN. The
-  next recurrence therefore names the pinning process directly — or, if the WAL
-  is pinned while no sidecar reports an old span, yields the sharper conclusion
-  that the pin is an unregistered/native mechanism (`vec0` cursor, or a span the
-  registry does not cover) in one of the live PIDs, which is exactly the
-  fork needed to justify or reject the deferred route-reads-through-the-daemon
-  alternative with evidence.
+  next recurrence therefore names the pinning process directly when a report
+  exists. When none does, absence of evidence is attributed only through the
+  per-PID sidecar-health distinction below — silence alone never licenses a
+  conclusion.
+
+  _Sidecar-health attribution (gate ruling, 2026-07-19)._ A missing heartbeat
+  has two very different causes: the process genuinely has no old span, or its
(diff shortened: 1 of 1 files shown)


ohdearquant:
Review follow-up on head 6a844f88 folds three findings from the latest round: the attribution census is now OS-derived (database-file-holder enumeration defines the universe; sidecar states map onto it; holders absent from sidecar data are unknown), registered-silent now requires ongoing beacon liveness via a metadata-only per-tick refresh (a stale beacon or stale-deleted heartbeat classifies as unknown), and the sidecar trust contract now binds validation to an opened directory handle with all operations descriptor-relative (openat/renameat/unlinkat), parents non-symlinked at open.

Two findings from the round are out of scope for this documentation-only change and are already tracked elsewhere: blob get/stat authorization semantics are settled in #1145 (capability-by-hash under the single-gate model, with an ADR-111 amendment queued), and the ANN checkpoint overwrite race is #1138 (fix-forward with per-scope compare-and-swap). Neither is modified by this diff.

oceanwaves630 (approved):
ADR-091 Amendment 2 is sound as specified: the process census is OS-derived (database-file holders) with sidecar states mapped onto that universe; registered-silent requires ongoing beacon liveness (metadata-only refresh each sweep tick, stale beacons classify as unknown); and sidecar filesystem operations are bound to a validated directory handle (O_DIRECTORY|O_NOFOLLOW, dirfd-relative). The two open findings on blob namespace authorization and ANN checkpoint publication concern code this documentation-only change does not modify; both are tracked separately (#1145, #1138) and are not gated here. Approved.

Issue #1143, opened 2026-07-19
memory.recall write-to-visibility regression: committed vector rows invisible to vector legs until full ANN rebuild


ohdearquant:
## Symptom

A freshly written memory (`memory.remember`) is invisible to `memory.recall` for tens of seconds even though the write path is fully synchronous. Measured on a ~68k-note corpus: a marker note queried with near-verbatim text returned nothing at +0s and ranked first at +33s. Before the Vamana segment architecture, an online-insert index made writes visible in milliseconds; this is the largest behavioral regression of the ANN migration.

## Mechanism (verified at source)

1. `create_note_inner` commits the note row, FTS document, and per-model vector rows transactionally before `remember` returns — the write path is not the problem.
2. The per-model vector legs of recall serve **ANN-only**: fresh vector rows do not enter a leg's candidate set until the Vamana index for that model is rebuilt or tail-replayed, which is O(corpus) work (~30s at 68k notes) scheduled in the background.
3. The FTS trigram leg does see the fresh row immediately, but RRF (k=60) over three sources structurally buries a hit that is rank-1 in one source only: 1/61 ≈ 0.0164 scores below a mediocre hit present in all three sources. So the fresh note is fetched and then buried below the returned window.

## Proposed direction

Add a fresh-tail exact leg: each vector leg augments its ANN candidates with an exact scan over the vector rows whose `ann_write_log` seq is above the serving segment's watermark (the tail is small by construction — checkpoint cadence and the rebuild threshold bound it). Merged into the existing per-model leg pre-fusion, not added as a new RRF source, so fusion semantics are unchanged. Restores read-your-writes visibility at millisecond cost. An ADR follows; this issue tracks the defect.

Related: #942 (candidate-scoring quality for fresh_directive queries — separate calibration concern), #889 (recall latency).

ohdearquant:
[...]

Issue #942, opened 2026-07-13
recall: fresh_directive query class weak (nDCG@10 ~0.10) under fused order and every rerank condition — candidate-scoring gap


ohdearquant:
## Summary

The #939 measure-first evaluation (controlled 400-note corpus, 40 graded queries, paired bootstrap stats; full protocol and tables in https://github.com/ohdearquant/khive/issues/939#issuecomment-4957086283) surfaced one weakness that is NOT a reranking problem: queries in the `fresh_directive` class score nDCG@10 ~0.10 with MRR@10 0.0 under the default fused order, and stay weak in every evaluated condition (feature-mixer rerank sweeps, embedder-cosine rerank bound). The exact_specific and paraphrase_specific classes score 0.68-0.74 nDCG@10 with MRR 1.0 on the same corpus.

## Why this is a candidate-scoring issue, not a rerank issue

Every condition in the study ranked the same fused candidate pool, and none recovered the class: the balanced weight sweep lifted fresh_directive nDCG to 0.40 only by trading a large, significant global MRR@10 loss (-0.4558). The failure is therefore upstream of any reordering stage: fresh directive-style memories do not accumulate enough fused score against established same-topic content at candidate-scoring time.

## Adjacent known structure

Prior measurement of the recall pipeline found a structural recency asymmetry: recent memories receive both FTS and vector signal in RRF fusion while older ones receive vector signal only, and separately that the FTS leg carries proper-noun precision. Directive-style content (imperative phrasing, few distinctive proper nouns) plausibly gets weak signal from BOTH legs. This issue is the place to characterize that properly.

## Suggested shape (measure before changing anything)

1. Reproduce the class weakness in isolation: vary directive phrasing, age, and salience independently; identify which leg (FTS, vector, fusion constants, decay weighting) loses the signal.
2. Check whether the existing scoring knobs (relevance/salience/temporal weighting at candidate-scoring time, not rerank time) can recover the class without regressing the strong classes.
3. Any scoring change follows the standard gate: reproduce, propose, ADR if it alters shipped ranking behavior.

## Links

- Evidence and protocol: #939 (closed, measure-first evaluation)
- Related: the recall-quality program (posterior serving, auto-entity anchoring)

ohdearquant:
Closing during tracker cleanup: this concerns functionality provided by the commercially licensed extensions, which are no longer part of this repository. Tracking continues on the distribution's internal tracker.

PR #1144, opened 2026-07-19
docs(adr): ADR-118 fresh-tail exact leg — read-your-writes visibility for vector recall

ohdearquant:
## Summary

Proposes ADR-118: each per-model vector leg of recall merges an exact-similarity scan over the `ann_write_log` tail (rows above the serving index's watermark) into its candidate list before fusion, restoring millisecond write-to-recall visibility (Closes-tracking: #1143).

Key decisions:
- Tail enumeration with final-state coalescing in a single read snapshot; tail wins dedup by subject, tail deletes drop the subject from the merged list.
- Merged into the existing per-model source, **not** a fourth RRF source — fusion semantics unchanged, and single-source burial is avoided by construction.
- Cost bounded by the checkpoint cadence and `KHIVE_ANN_REBUILD_THRESHOLD`; the no-index (Cold/Empty) case explicitly does not become an exact-search fallback.
- Normative for both delta-log consumers (memory first, knowledge follow-up); closes the visibility half of #752 by being a query-time database read.
- Partially supersedes ADR-107 §1: the stale bound continues to govern the ANN candidate generator, no longer the verb's observable freshness.

Docs-only; implementation follows after review.

Refs #1143, #752, #942.

### DIFF
--- a/docs/adr/ADR-118-fresh-tail-recall-visibility.md
+++ b/docs/adr/ADR-118-fresh-tail-recall-visibility.md
@@ -1,0 +1,177 @@
+# ADR-118: Fresh-Tail Exact Leg — Read-Your-Writes Visibility for Vector Recall
+
+**Status**: Proposed
+**Date**: 2026-07-19
+**Depends on**: [ADR-021](ADR-021-memory-pack.md) (Memory Pack),
+[ADR-079](ADR-079-ann-persistence-warm-path-integration.md) (ANN persistence, Amendment 1
+delta-log/watermark classifier), [ADR-107](ADR-107-memory-ann-lifecycle.md) (Memory ANN
+lifecycle — eventual consistency contract)
+**References**: issue #1143 (write-to-visibility regression), issue #752 (process-local
+generation counter), issue #942 (fresh-query scoring quality — related, out of scope)
+
+## Context
+
+`memory.remember` is fully synchronous: the note row, its FTS document, and one vector row
+per registered embedding model are committed before the verb returns. Yet a freshly written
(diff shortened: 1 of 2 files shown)


ohdearquant:
Pushed a90c5d0 addressing the three review findings on the review follow-up round:

1. **Compaction linearization** (§1 new provision): the leg must validate coverage in the same read snapshot — read the pair's wildcard-inclusive registry minimum alongside the log scan; a minimum above the bridge's watermark means tail completeness cannot be proven, so the leg re-resolves the currently published segment (or serves without the exact leg for that query, degrading to the ADR-107 contract, while triggering re-adoption). Same-process checkpoints SHOULD install the new bridge before raising the durable watermark, making the window empty for the process's own checkpoints; the snapshot check stays mandatory for independent processes.
2. **Two-tier guarantee** (Decision intro, §3, supersession section, Consequences): the read-your-writes promise is now scoped precisely — total while a serving index exists; newest threshold-sized suffix (which by construction contains the caller's most recent writes) in Cold/Empty, with ADR-107 continuing to describe exactly and only those states. No contradiction remains between the promise and the Cold fallback.
3. **Registration precondition** (§1 new provision): S=0 establishes an entire-scope tail only for an unregistered *pair*, not an unregistered *consumer* — a registered peer on the same pair legitimately compacts. The exact leg is permitted only after the consumer's durable registration row exists; ADR-079's register-before-persist rule is extended to register-before-any-tail-dependent-read-path.

The refuted findings (blob authorization — pre-existing, introduced elsewhere; legacy-index watermark — covered by ADR-079's mandatory Cold classification) took no change, matching the verdict's own falsification pass.

oceanwaves630 (approved):

PR #1146, opened 2026-07-19
feat(db): ADR-091 Amendment 2 - cross-process WAL-pin attribution

ohdearquant:
## Summary

Implements ADR-091 Amendment 2 (Planks A/B/C) — cross-process WAL-pin attribution and per-session observability.

- **Plank A**: every non-daemon `kkernel mcp` session process runs an observe-only `tx_registry` age sweep (`KHIVE_SESSION_SWEEP_INTERVAL_MS`, default 5s), reusing the same warn/stale thresholds and edge-triggered logging the daemon's checkpoint task already uses. Sessions never checkpoint — that stays daemon-owned so N session processes never compete for the writer mutex.
- **Plank B**: a per-PID walpin heartbeat sidecar (`<db-file>.walpin/<pid>.json`, `KHIVE_WALPIN_SIDECAR`) lets the daemon attribute a stuck WAL to a specific cross-process PID on a TRUNCATE no-progress event. The sidecar directory is created mode `0700` and ownership-validated before any use (a non-compliant existing directory is refused, never chmod/chown'd into compliance); heartbeat writes use `O_NOFOLLOW` + exclusive-create to a temp file, then atomic rename over the target; enumeration applies a three-test liveness gate (PID alive, OS-reported process start time matches within a small epsilon, fresh `updated_at`) and deletes entries failing any test.
- **Plank C**: on the same no-progress event, an additional `PRAGMA wal_checkpoint(PASSIVE)` reports pin depth (`log - checkpointed`) with zero dependence on the shm WAL-index layout.

**Must merge after #1142** (the amendment text itself).

## Test plan

- [x] `cargo clippy --all-targets -p khive-db -p khive-mcp -p khive-runtime -- -D warnings`
- [x] `cargo test -p khive-runtime -p khive-mcp -p khive-db` (all green)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-db -p khive-mcp -p khive-runtime`
- [x] New unit tests: sidecar write contract (0700 creation, wrong-owner/mode refusal, symlink refusal on write + enumeration, atomic replace), liveness gate (dead PID / mismatched start time / stale `updated_at` deleted; live entry retained), Plank A edge-triggered sweep behavior (via the shared `TxAgeSweepState` machine plus an integration test on `run_session_sweep_task` itself), and the PASSIVE pin-depth arithmetic against a real SQLite connection.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1942,6 +1942,7 @@ dependencies = [
  "khive-score",
  "khive-storage",
  "khive-types",
+ "libc",
  "object_store",
  "parking_lot",
  "rand 0.8.7",
--- a/crates/khive-db/Cargo.toml
+++ b/crates/khive-db/Cargo.toml
@@ -37,6 +37,9 @@ tempfile = { workspace = true }
 futures = { workspace = true }
 object_store = { workspace = true }
 
+[target.'cfg(unix)'.dependencies]
+libc = { workspace = true }
+
 [dev-dependencies]
 tokio = { workspace = true, features = ["full", "test-util"] }
 rusqlite = { workspace = true, features = ["bundled", "column_decltype"] }
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -3673,12 +3673,27 @@ mod tests {
             .expect("session sweep task panicked");
     }
 
+    /// Bounded condition poll for filesystem effects of the async sweep
+    /// task — fixed sleeps flake under parallel test load because sidecar
+    /// writes fsync.
+    async fn wait_for(deadline: Duration, mut cond: impl FnMut() -> bool) -> bool {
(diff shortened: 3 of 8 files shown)


ohdearquant:
Spec delta from the ADR review round (ADR-091 Amendment 2, commit 1477389 on #1142): Plank B attribution now requires a **three-state sidecar-health distinction** — reporting / registered-silent / unknown — backed by a one-time per-PID registration beacon written at sidecar init under the same trust-boundary rules as heartbeats. Only a pin observed while every live PID is reporting or registered-silent may be attributed to an unregistered/native mechanism; any unknown PID makes attribution inconclusive and the WARN names those PIDs. This PR predates the delta and does not implement the beacon or the three-way classification — a follow-up commit on this branch will close the gap before ready.

oceanwaves630 (approved):

PR #1147, opened 2026-07-19
feat(pack-memory): fresh-tail exact leg for vector recall (ADR-118)

ohdearquant:
## Summary

Implements ADR-118 (fresh-tail exact leg) for the memory pack's vector recall path.

Each per-model vector leg of `memory.recall` now merges a fresh-tail exact leg into candidates before fusion, giving read-your-writes visibility on top of the existing warm-ANN serving path:

- Writes committed after a serving bridge's watermark are read from `ann_write_log`, coalesced to the final op per subject (reusing the existing `fetch_final_tail` coalescing logic), scored exactly, and merged into the vector leg's own candidates — not added as a separate fusion source, so results are never buried by RRF's fixed per-source rank weighting.
- A final delete drops the subject from candidates too.
- Dedup is by `subject_id` with the tail winning; the merged set is re-sorted before fusion.
- Cold/Empty bridges (no serving index at all) get a second, capped tier: the scan is bounded to `ann_rebuild_threshold() * live` rows, highest-seq suffix, rather than a general exact-search fallback.

Two checks are binding, not advisory:

- **Registration precondition**: no durable watermark row for this consumer means completeness above watermark 0 is unprovable, so the leg registers at 0 and sits out that one query rather than assuming full coverage.
- **Compaction linearization**: if the registry-wide watermark minimum has advanced past the in-memory bridge's watermark, the write-log may no longer retain every row above it; the leg re-resolves the current segment's watermark (forcing re-adoption) or skips, rather than scanning a window the log can no longer prove complete.

`KHIVE_ANN_FRESH_TAIL=0` restores the pre-ADR-118 behavior (default: enabled).

Scope: the memory pack only. The knowledge pack's own vector legs are unaffected, pending a follow-up PR.

## Test plan

- [x] `cargo clippy -p khive-pack-memory -p khive-runtime --all-targets -- -D warnings`
- [x] `cargo test -p khive-pack-memory -p khive-runtime` (all green, including 8 new tests covering: same-process write-then-recall visibility, cross-process external-writer visibility, final-delete dropping a subject, dedup with the tail winning, the registration precondition, the compaction-linearization mismatch, empty-tail byte-identical output, and the `KHIVE_ANN_FRESH_TAIL=0` escape hatch)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-pack-memory -p khive-runtime`
- [x] `cargo check --workspace` (full workspace, including the knowledge pack, compiles unaffected)

### DIFF
--- a/crates/khive-db/sql/012-ann-write-log-model-seq-index.sql
+++ b/crates/khive-db/sql/012-ann-write-log-model-seq-index.sql
@@ -1,0 +1,10 @@
+-- ADR-118: the fresh-tail exact leg's per-model tail scans (global-scope
+-- consumers, e.g. the memory pack) filter by embedding_model/kind/field and
+-- range over seq WITHOUT a namespace predicate. The existing
+-- idx_ann_write_log_ns_model_seq index is namespace-leading, so it cannot
+-- serve these scans as an index range seek — the query degrades to a full
+-- table scan. This index leads with embedding_model so an
+-- (embedding_model, kind, field) equality prefix plus a seq range seek stays
+-- an index-only operation regardless of namespace.
+CREATE INDEX IF NOT EXISTS idx_ann_write_log_model_kind_field_seq
+    ON ann_write_log (embedding_model, kind, field, seq);
--- a/crates/khive-db/src/backend.rs
+++ b/crates/khive-db/src/backend.rs
@@ -398,6 +398,9 @@ impl StorageBackend {
         writer
             .conn()
             .execute_batch(crate::migrations::ANN_WRITE_LOG_DDL)?;
+        writer
+            .conn()
+            .execute_batch(crate::migrations::ANN_WRITE_LOG_MODEL_SEQ_INDEX_DDL)?;
 
         // Create the vec0 virtual table. Idempotent on fresh databases and after the
         // old-schema rebuild above.
--- a/crates/khive-db/src/migrations.rs
+++ b/crates/khive-db/src/migrations.rs
@@ -122,6 +122,8 @@ const V10_UP: &str = include_str!("../sql/010-entities-content-ref.sql");
 
 const V11_UP: &str = include_str!("../sql/011-ann-write-log.sql");
 
+const V12_UP: &str = include_str!("../sql/012-ann-write-log-model-seq-index.sql");
+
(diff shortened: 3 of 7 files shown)


ohdearquant:
Spec delta from the ADR review rounds (#1144, commits 2c0c04a + d0008c6 — merged text is normative): the mismatch branch changed. When the same-snapshot registry minimum exceeds the serving bridge's watermark and re-resolution is not possible within the query, the exact leg must run **floored at the registry minimum** rather than being skipped — the leg's coverage narrows, but it never disappears on a serving-index path (coverage above the minimum is provable in-snapshot; the uncovered gap holds only writes the newer published segment already reflects). If this branch implemented skip-on-mismatch per the earlier draft, a follow-up commit should switch it to the floored scan plus a test for the mismatch path. The Cold-visibility retained-row conjunct is spec language only (the scan naturally sees only retained rows) — no code change expected there.

oceanwaves630 (approved):
Adjudication: the round's blocking finding is refuted by the governing contract's own text. ADR-118's visibility guarantee explicitly carves out writes inside the cross-process mismatch window as a transient closed by re-adoption, and its compaction-linearization section sanctions exactly the terminal floored fallback this diff implements when re-resolution does not converge within the query. Verified at head: the fallback serves a coherent (candidates, floor) pair and forces re-adoption via bump_generation, with a regression test asserting the generation bump, so the transient cannot repeat indefinitely. The suggested alternative of exact-scoring the compacted gap is impossible by construction: those log rows are the ones compaction deleted, and the only route to the gap's writes is the newer published segment that re-adoption loads.

ohdearquant:
Scope note on the p95 comparison cited in this PR (6.5ms enabled vs 4.7ms disabled), from a 2026-07-19 evidence review: the number stands as a layer-level empty-tail guard result, but it does not by itself establish p99 write-to-visibility or end-to-end recall latency under ADR-118's corpus-relative tail policy, and the samples do not declare their cache condition (application / database / OS page cache). Future citations of this comparison should state the complete cache protocol, or re-measure under a protocol that reports each cache state independently with post-warm-up same-state confidence intervals. The follow-up measurement plan is posted on #1143.

Issue #1129, opened 2026-07-19
Daemon warm-state memory: full corpus resident in anonymous heap; restart-time full-corpus rebuild gate


ohdearquant:
### Symptom

The warm daemon's memory footprint appears to "spike suddenly under moderate concurrency": RSS jumps from ~2.0 GB to ~4.2 GB during a burst of a few concurrent read-only requests, and settles well above the pre-burst baseline. On a multi-tenant host or a 16 GB workstation this presents as an unpredictable, un-plannable memory profile.

### Measurements (production daemon, ~553K entity/note vectors + ~358K knowledge sections)

Three probes, per-2s instrumentation, `vmmap` physical-footprint as the real-memory discriminator:

1. **Read-only concurrency probe** (4 concurrent batches of mixed search/recall/knowledge queries, no writes): RSS 1,995 MB → 4,227 MB; settled at 3,941 MB. Physical footprint **peaked at 3.9 GB** (anonymous memory, not file cache), then settled back to 1.5 GB with ~785 MB swapped/compressed.
2. **Differential probe by retrieval path** (same daemon, now warm; each path 4-way separately — entity/note hybrid search, memory recall, knowledge search): **no path adds net memory**. Footprint stayed 1.4-1.9 GB across all phases. The hike is not per-query and not path-specific.
3. **Idle observation**: after minutes of quiet, footprint drops to ~1.5 GB while RSS lags higher; the kernel compresses/swaps the idle warm state.

### Mechanism

`warm_all()` runs at daemon startup (spawned in `run_daemon`; the memory and knowledge packs warm their ANN indexes). The warm in-memory state — ANN graphs, full-precision f32 vectors, embedder, per-connection SQLite caches — totals **~2.5-3 GB of anonymous heap** at this corpus size:

- ~553K vectors x 384-dim f32 ≈ 850 MB (entities/notes)
- ~358K knowledge sections ≈ 550 MB
- ANN adjacency + embedder model + caches ≈ remainder
- SQLite: 64 MB page cache + 1 GB `mmap_size` **per connection**, pool of 1 writer + up to 8 readers **per database**

When the daemon idles, macOS compresses/swap-evicts these pages (footprint falls to ~1.5 GB). The first query burst afterwards touches the working set and the compressed pages swell back to full residency **all at once** — which reads as a "sudden hike under moderate concurrency". Concurrency is incidental; the trigger is touching a large resident working set after idle.

### Why this matters

[...]

PR #1149, opened 2026-07-19
docs(adr): ADR-079 Amendment 1 residency-invariant note

ohdearquant:
Post-implementation note on the phase-1 residency guarantee: follow-up measurement found ~794 MB of small-allocation anonymous heap above the projected 100-300 MB band, attributable to the state the mmap load path does not map (graph adjacency, lifecycle, bridge id map). Makes the guarantee's boundary normative — it covers file-backed state plus owned state that is bounded and consumed fresh; corpus-scaling persistent owned state sits outside it, and future levers must preserve the invariant (COW delta overlays and immutable delta segments do; owned full-copy rebuilds do not). Requires Linux smaps_rollup re-derivation before any lever targets the residue (tracked with #1129). Documentation only.

---

**Review scope note.** The latest round's replay finding was verified at source and confirmed real: it is filed as #1150 (present on main in both consumers; not introduced by this documentation-only diff), and this PR now records the corresponding spec-side id-map ownership rule. The round's conflict-resolution correctly refuted the blob re-raise per the gate-before-dispatch contract.

### DIFF
--- a/docs/adr/ADR-079-ann-persistence-warm-path-integration.md
+++ b/docs/adr/ADR-079-ann-persistence-warm-path-integration.md
@@ -726,6 +726,19 @@ normative:
    the deferred mmap-`graph.bin` lever in the inventory above is the natural candidate once
    attribution confirms adjacency dominates.
 
+### Id-map ownership at replay (post-implementation note, 2026-07-19)
+
+Tail replay resolves subjects to index ordinals through a reverse map built from the
+persisted id map. The binding rule, made explicit after a defect was found in both shipped
+consumers (#1150): **a tombstoned ordinal has no owner** — the reverse map must exclude
+tombstoned entries (equivalently, id-map ownership is cleared at tombstone time), because
+the index recycles tombstoned slots on insert. A reverse map that retains a tombstoned
+subject's entry lets a later coalesced delete for that subject resolve to a recycled slot
+and tombstone a different, live vector — replay must never allow a stale subject mapping to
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Spec sign-off. The residency guarantee's boundary is now normative and correctly stated: file-backed plus bounded-and-consumed-fresh owned state inside the guarantee, corpus-scaling persistent owned state outside it; copy-on-write delta overlays and immutable delta segments preserve the invariant while owned full-copy rebuilds do not; and attribution-before-optimization makes the small-allocation residue figure a macOS allocator-lens observation requiring Linux smaps_rollup re-derivation before any lever fires. Approved.

ohdearquant:
Review follow-up on the latest round: the replay defect is confirmed real and filed as #1150 (present in both consumers on main; not introduced by this diff). This PR now records the spec-side rule — a tombstoned ordinal has no owner, reverse maps exclude tombstoned entries, with the recycle-then-stale-delete interleaving named as required regression coverage — so the replay contract is explicit while the code fix proceeds under #1150.

oceanwaves630 (approved):
Re-verified at head 75de563e7. The added id-map-ownership rule is the correct spec-side codification of the replay defect: a tombstoned ordinal has no owner, the reverse map excludes tombstoned entries (equivalently ownership cleared at tombstone time), and the tombstone-recycle-then-stale-delete interleaving requires regression coverage in every consumer. Approved.

Issue #1126, opened 2026-07-19
comm: a single un-ingestable IMAP message stalls the entire email channel in an unbounded retry loop


ohdearquant:
## Summary

When one message in an IMAP fetch page fails `comm.ingest`, the poll declines to commit the IMAP cursor for the **whole page** and retries the entire page on the next poll (`crates/khive-mcp/src/serve.rs:548`; a sibling batch path with the same shape at `:1227`). If the failure is **deterministic** — e.g. the write secret-gate rejecting a body (the false-positive class is tracked in #1056 and #1040), or a parse/validation error — the same page is re-fetched and re-processed on every poll, indefinitely. One message permanently blocks the channel.

## Impact

- **Forward progress halts.** The cursor never advances past the failing page, so no later message on the channel is ingested — the channel is effectively frozen for as long as the poison message is fetched.
- **Unbounded resource burn.** Each poll re-fetches, re-parses, and re-runs the secret-gate over the whole page, then rolls back. Observed in production: a single message drove a steady ~480 failed-page retries/hour continuously (thousands of identical `not committing IMAP cursor` WARNs), with sustained CPU and SQLite write-attempt churn, until the message was cleared by hand.
- **Restart-immune.** A process restart does not clear it — the message persists in the mailbox and is re-fetched on the next poll.

## Root vs. amplification

The secret-gate false-positive that can *trigger* this is tracked separately (#1056, #1040). This issue is the **amplification**: any single deterministically-failing message — false-positive or genuine — must not be able to stall the whole channel or spin forever. Even a *correct* secret-gate block should not become a denial of ingestion for every other message on the channel.

## Fix direction

Per-message disposition instead of whole-page rollback:

- Commit cursor progress for the messages that succeed; one failure must not roll back the batch's forward progress.
- Classify failures: **transient** (network, lock contention) → bounded retry with backoff; **deterministic** (secret-gate block, parse/validation error) → quarantine to a dead-letter (record message id + reason, surface it for inspection) and advance past it so the channel keeps moving.
- Never re-attempt an unchanged deterministic failure indefinitely.

Apply the same treatment to the sibling batch path at `serve.rs:1227`.

ohdearquant:
## Production follow-up: the trigger is a product sign-in email, so this recurs by design

[...]

PR #1151, opened 2026-07-19
docs(adr): ADR-119 host-supervised daemon components beside the verb plane

### LINKED ISSUE #373: Design: the request interface should not force everything into verbs (verbs vs resources vs subscriptions vs services)
## Summary

Today every capability is modeled as a **verb** dispatched through the single `request` tool (ADR-016) over a flat, pack-declared verb registry (ADR-023). That uniformity was the right call for the request/response KG surface. As the system grows transports, channels, and background concerns, the "everything is a verb" model is being stretched to cover things that are not request/response operations. This issue opens a design discussion (→ ADR) on a richer interface taxonomy. Not a v1 change.

## Why now

The email channel work is the trigger. Two capabilities it needs do not fit a verb:

- **Background services.** The IMAP inbound poll loop and the new outbound delivery loop are long-running daemons, not caller-invoked verbs. They live in the server binary (`khive-mcp/src/serve.rs`), bolted on outside the verb registry, started by ad-hoc feature-gated spawn code. A pack cannot declare "I own a background service" today.
- **Push / subscription.** There is no primitive to subscribe to an event such as "a new inbound message arrived." Consumers poll instead — the canonical inbox wake-up monitor is a `MAX(created_at)` SQLite poll precisely because there is no push/stream shape on the interface.

Related, already-visible seams:

- **Internal-only subhandlers** (e.g. `comm.ingest`) already form a *second class* of "verbs that are not user-facing verbs." A flat verb surface that quietly contains off-wire entries is a sign the single taxonomy is conflating distinct kinds.
- **Reads vs actions.** MCP itself distinguishes *tools* (actions) from *resources* (addressable reads). Pure reads (`get`, `list`, `stats`) are natural resources; mutations are natural verbs. We model both as verbs.
- **Batch semantics.** `request` batches verbs with no inter-op ordering or atomicity guarantee — fine for independent reads, ambiguous for anything transactional.

## Direction (for discussion, nothing decided)

A capability taxonomy instead of one flat verb list. Candidate kinds:

- **verbs** — request/response actions (the current surface; keep it)
- **resources** — addressable reads, possibly mapped to MCP resources
- **subscriptions / streams** — push for events (inbound message, task transition, …)
- **services** — long-running background daemons, *declared by a pack* and supervised by the runtime, rather than hand-spawned in the server binary
- **internal subhandlers** — off-wire, explicitly second-class (formalize what `comm.ingest` already is)

## Open questions

1. Should packs declare background services and subscriptions as first-class, with the runtime supervising lifecycle, replacing the `serve.rs` bolt-on?
2. Should pure reads migrate to MCP resources, or stay verbs for one uniform surface?
3. How does any of this interact with ADR-016 (request DSL) and ADR-023 (declarative pack format) — additive layer, or a revision?
4. What is the backward-compatibility path? The flat verb surface is a public contract.

## Scope

Design discussion → ADR. Down the road, not blocking current channel work. Cross-refs: the channel transport work and the multi-tenant/multi-provider channel design discussion.

ohdearquant:
## Summary

- Adds **ADR-119**: daemon-resident loops (schedule drain, email ingest, session mirror, ANN maintenance) get one host-owned lifecycle contract via a `khive-mcp`-local supervised component registry — cancellation, bounded restart budgets, backoff, panic isolation, health, bounded shutdown — with the ADR-106 schedule drain as the reference first migration.
- The public surface is untouched: no `Pack`/`PackRuntime` change, no new MCP verb/resource/subscription/capability entry; `tools/list`, request parsing, and legacy stdio bytes stay byte-identical.
- The event-topic proposal is **deferred without a number**: durable event consumption already belongs to ADR-017 (`PackEventConsumer`) and ADR-022 (events query surface); polling throughput is not a present problem; reopening requires a named consumer with a measured >=10x latency case. The durable-then-notify invariant any future bus must preserve is fixed in this ADR.
- Rides along: corrects ADR-079 Amendment 1's daemon-resource issue references from #1126/#1127 to #1127/#1129 (#1126 is the email poison-message incident, cited in ADR-119 for supervision motivation only).

Closes #373.

## Scope note

Docs-only. Registry findings outside the two touched documents are pre-existing and tracked elsewhere.


oceanwaves630 (approved):
Spec-gate sign-off at head ac13432a1. Verified at source: ADR-106 Amendment B update 2 confirms the schedule drain requires both the resolved schedule runtime and the daemon's live multi-backend KhiveMcpServer, so the taxonomy kill gate fires against pack-level declaration today exactly as this ADR argues; ADR-017's PackEventConsumer contract plus ADR-022's query surface already own durable event consumption, making the deferred-unnumbered event topic the correct disposition. The decision picks the minimal reversible option with a falsifiable promotion path (two independent packs proving a narrow context), carries all seven issue-373 kill gates as binding acceptance criteria, names its consumer (the ADR-106 drain as reference migration, closing its Acceptance Criterion 5), and makes no unmeasured cheapness or magnitude claims. The ADR-079 citation correction rides along consistently.

oceanwaves630 (approved):
Re-verified at head d8ff76562. The three added sections close the round's blockers as contracts, not gestures: the shutdown handoff joins the existing tracked-task drain path with normative ordering and bounded per-component joins; the error-and-restart table makes the reference component's acceptance tests determinate, including an explicit no-automatic-reset budget default; and the scheduled-action identity section makes creator-actor replay normative with fail-closed legacy rows, a denial-test acceptance criterion, and a matching fence. Spec-gate approval renewed.

oceanwaves630 (approved):
Re-verified at head fb0ad0dea. The permanent-failure row completes the mapping correctly: immediate terminal transition with no restart and no backoff regardless of remaining budget, the classification carried on the error type with the component classifying and the host acting only on the classification, and a concrete retryable-versus-permanent split for the schedule component. The error contract is now fully determinate. Approval renewed.

oceanwaves630 (approved):

PR #1152, opened 2026-07-19
fix(pack-memory): exclude tombstoned ordinals from tail-replay reverse maps

### LINKED ISSUE #1150: ANN tail replay maps tombstoned ordinals as live; stale delete can tombstone a recycled slot's vector
Found during review follow-up on the ADR-079 Amendment 1 replay path; present in both consumers.

**Defect.** Replay builds its UUID-to-ordinal reverse map from every id-map entry, including tombstoned slots (`crates/khive-pack-memory/src/ann.rs:365-402`; `crates/khive-pack-knowledge/src/knowledge/vamana.rs:413-464`). Vamana recycles tombstoned slots on insert (`crates/khive-vamana/src/index.rs:1692-1724`, covered by `insert_reuses_free_slot`). Sequence: ordinal A is tombstoned; a later upsert for subject C reuses A's slot; a later coalesced delete for A resolves through the stale reverse-map entry and tombstones the slot now owned by C. The replayed segment silently loses a live vector and serves incorrect ANN results.

**Fix direction.** Exclude tombstoned ordinals from the replay reverse map, or clear id-map ownership at tombstone time. Regression coverage needed in both consumers: persisted tombstoned A → replay upsert C reusing A's ordinal → later final delete for A → C must remain live and searchable.

**Spec.** The id-map ownership rule is being recorded in ADR-079 Amendment 1 (PR #1149) so the replay contract is explicit.

ohdearquant:
## Summary

Fixes #1150. Tail replay built its UUID-to-ordinal reverse map from every id-map entry, including tombstoned slots — id-map entries are never cleared at tombstone time. Since the Vamana index recycles tombstoned slots on insert, a coalesced final tail could interleave a same-batch upsert reusing a tombstoned subject's ordinal with a later delete op for that same (already-deleted) subject: the delete resolved through the stale reverse-map entry and tombstoned the slot's new, live owner, silently losing a live vector.

Per the id-map ownership rule recorded in ADR-079 Amendment 1 ("Id-map ownership at replay"): a tombstoned ordinal has no owner.

## Fix

Applied identically in both replay consumers (`khive-pack-memory::ann::AnnBridge::apply_final_ops` and `khive-pack-knowledge::knowledge::vamana::AnnBridge::{reverse_map, apply_final_op}`):

- **Exclude tombstoned ordinals** when constructing the reverse map, using the index's own tombstone bitmap (`VamanaIndex::is_tombstoned`) rather than trusting the (never-cleared) id-map entry.
- **Fail closed on ownership contradictions**: before tombstoning on a delete op, verify the ordinal's current id-map owner still matches the subject being deleted; on a mismatch, skip the tombstone and log a warning instead of destroying a different subject's live vector.

## Regression coverage

Both consumers get a unit test reproducing the issue's exact shape: a subject tombstoned before the call (id-map entry left stale), a same-batch upsert that recycles the freed ordinal for a different subject, then a delete op for the original (already-gone) subject processed after — the recycled ordinal's live vector must remain live and searchable. Both tests were verified to fail against the pre-fix code before the fix was restored.

## Test plan

- [x] `cargo clippy -p khive-pack-memory -p khive-pack-knowledge -p khive-vamana --all-targets -- -D warnings`
- [x] `cargo test -p khive-pack-memory -p khive-pack-knowledge -p khive-vamana` (all green)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-pack-memory`
- [x] New regression tests confirmed to fail on pre-fix code, pass on fixed code

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/vamana.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/vamana.rs
@@ -410,9 +410,18 @@ impl AnnBridge {
     /// batches can apply incrementally without rescanning the id-map.
     /// Highest ordinal wins for a repeated uuid: inserts append, so the
     /// latest slot is the live one; earlier slots are tombstoned.
+    ///
+    /// A tombstoned ordinal has no owner (ADR-079 Amendment 1 id-map
+    /// ownership rule) — `id_map` entries for already-tombstoned slots are
+    /// stale (tombstoning never clears them) and are excluded here, or a
+    /// reused slot's new owner can be tombstoned by a replay op for the
+    /// old, already-deleted subject (#1150).
     pub(crate) fn reverse_map(&self) -> HashMap<Uuid, u32> {
         let mut reverse: HashMap<Uuid, u32> = HashMap::with_capacity(self.id_map.len());
         for (ordinal, uuid) in self.id_map.iter().enumerate() {
+            if self.index.is_tombstoned(ordinal as u32) {
+                continue;
+            }
             reverse.insert(*uuid, ordinal as u32);
         }
         reverse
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Verified at head 1ec7326d8: both replay consumers now skip tombstoned ordinals when building reverse maps, fail closed on ownership contradictions inside a batch, and each crate carries the tombstone-recycle-stale-delete regression the amendment required. This closes the defect exactly as specified by the id-map ownership rule, and the auto-struck finding was correctly out of scope (tracked design item, untouched by this diff).

Issue #1153, opened 2026-07-19
ANN replay: align method contracts with same-batch reassignment skip; reserve reverse-map capacity by live count


ohdearquant:
Follow-ups from the #1152 review round (non-blocking, both consumers: crates/khive-pack-memory/src/ann.rs and crates/khive-pack-knowledge/src/knowledge/vamana.rs):

1. **Doc-contract alignment.** The replay method contracts still state that any id-map contradiction returns Err and escalates to Cold, but the ownership-mismatch branches added in #1152 intentionally skip the tombstone with a warning when a same-batch upsert has already reused the slot (the old subject's vector is gone by construction, so there is nothing left to delete). Update the contracts to distinguish this intentionally-skipped reassignment case from genuine id-map contradictions, which must keep escalating to Cold.

2. **Reverse-map capacity.** Both reverse maps reserve HashMap capacity for every id_map slot, including tombstoned history the new filter excludes. Churn-heavy indexes over-allocate transiently; reserve by live count instead (ann.rs:386, vamana.rs:420).

PR #1156, opened 2026-07-20
fix(pack-memory): align replay contracts and size reverse maps by live count

ohdearquant:
Follow-ups from #1152, closing #1153. Both ANN tail-replay consumers (`khive-pack-memory::ann`, `khive-pack-knowledge::knowledge::vamana`):

1. **Contract alignment.** The replay method docs claimed any id-map contradiction returns `Err` and escalates to Cold, but the ownership-mismatch branch added in #1152 intentionally skips the tombstone with a warning when a same-batch upsert has already reused the slot — the old subject's vector was tombstoned at reuse, so there is nothing left to delete. The contracts now state both behaviors explicitly; genuine contradictions still escalate to Cold.

2. **Reverse-map capacity.** Both reverse maps reserved `HashMap` capacity for every `id_map` slot, including tombstoned history the build filter excludes, so churn-heavy indexes over-allocated transiently. Capacity now comes from `VamanaIndex::live_count()` (`num_vectors - tombstone_count`), matching the number of entries actually inserted.

Doc + allocation-size only; no behavior change on the replay paths. Gates run locally: `cargo test -p khive-pack-memory -p khive-pack-knowledge` (84+ tests), clippy `-D warnings`, fmt, `RUSTDOCFLAGS="-D warnings" cargo doc` on both packs.

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/vamana.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/vamana.rs
@@ -417,7 +417,7 @@ impl AnnBridge {
     /// reused slot's new owner can be tombstoned by a replay op for the
     /// old, already-deleted subject (#1150).
     pub(crate) fn reverse_map(&self) -> HashMap<Uuid, u32> {
-        let mut reverse: HashMap<Uuid, u32> = HashMap::with_capacity(self.id_map.len());
+        let mut reverse: HashMap<Uuid, u32> = HashMap::with_capacity(self.index.live_count());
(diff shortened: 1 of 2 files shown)


ohdearquant:
Automated review at head 224895f6e: **APPROVE** — no blocking findings. Both replay contracts verified against the actual skip-vs-escalate branch behavior; live_count() confirmed to exclude tombstones, matching the reverse-map insertion loop exactly. Focused tests (91 knowledge, 285 memory, both integration suites), fmt, clippy -D warnings, rustdoc -D warnings all green.

PR #1162, opened 2026-07-20
fix(db): harden walpin enumeration and heartbeat-failure classification

### LINKED ISSUE #1158: walpin: harden enumeration and heartbeat paths (entry-count cap, timestamp validation, write-failure classification)
Follow-ups against the merged #1146 attribution sidecar (crates/khive-db/src/walpin.rs, crates/khive-db/src/checkpoint.rs). Three related hardening gaps, all in the same fail-closed family as the shipped per-entry 64 KiB bound:

1. **No entry-count bound on enumeration.** `enumerate_live` iterates every directory entry with no cap on entry count or aggregate work; per-entry reads are bounded at 64 KiB but N entries cost N × (open, fstat, read, parse) synchronously while the TRUNCATE path holds the writer guard. A same-UID process can populate the 0700 directory with arbitrarily many entries. Cap the entries processed per enumeration (excess → census incomplete, never silently dropped).

2. **Unvalidated i64 timestamps in freshness/identity arithmetic.** `started_at`/`updated_at` deserialize as unrestricted i64 and feed plain subtraction plus `.abs()` (walpin.rs ~1754, ~1766). Extreme values overflow: panic under debug, wrap under release — and a wrapped difference can land inside the freshness window, classifying a malformed entry as reporting instead of Unknown. Use saturating arithmetic and range-sanity checks; out-of-range → Unknown.

3. **Failed heartbeat write still classifies as healthy.** In the sidecar `observe` path (checkpoint.rs ~741-775) the beacon is refreshed first and `wrote` is set true even when `write_heartbeat` fails. With a fresh beacon and no heartbeat file, enumeration classifies the process `RegisteredSilent`, exonerating a process that currently holds an over-threshold transaction. On write failure the tick should fail toward Unknown (e.g. skip the beacon refresh for that tick) rather than fail open; transient failures self-heal on the next tick.

ohdearquant:
Closes #1158 — three fail-closed gaps in the WAL-pin attribution sidecar, all in the same trust-boundary family as the shipped per-entry 64 KiB bound:

1. **Entry-count cap.** `enumerate_live` now reads at most 512 entries per pass (`enumerate_live_bounded` carries the cap as a parameter for tests). Entries past the budget are classified `Unknown` from their filename alone — no open/fstat/read/parse — so directory content cannot inflate the work done while the checkpoint writer guard is held, and unread entries are never exonerated.

2. **Saturating timestamp arithmetic.** Persisted `started_at`/`updated_at` deserialize as unrestricted `i64`; plain `(a - b).abs()` wraps on extreme values in release builds, and a wrapped difference can land inside a freshness window — classifying a malformed entry as fresh (fail-open). `epoch_abs_diff` saturates to `u64::MAX` on overflow, keeping any extreme stamp outside every window.

3. **Heartbeat-failure classification.** The sidecar `observe` refreshed the beacon before attempting the heartbeat write and set `wrote` regardless of the result. A fresh beacon with no heartbeat file classifies `registered-silent` at enumeration, so a failing write exonerated a process currently holding an over-threshold transaction. The beacon refresh is now gated on the write landing; the tick fails toward `Unknown` and a transient failure self-heals on the next tick.

Tests: saturation unit cases, extreme-timestamp classification, entry-cap classification, and a beacon-gate regression that fails on the previous unconditional-refresh behavior (temp-name obstruction forces the write failure without confounding directory-mode validation). Gates locally: `cargo test -p khive-db --lib` walpin (46) + checkpoint (50) slices, clippy `-D warnings` all-targets, fmt, `RUSTDOCFLAGS="-D warnings" cargo doc`.

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -644,6 +644,12 @@ struct WalpinSidecarState {
     /// cadence must not be misread as stale.
     interval_ms: u64,
     wrote: bool,
+    /// Whether this process's registration beacon is believed present on
+    /// disk. Cleared when a failed heartbeat write escalates to beacon
+    /// removal (fail-closed — see `observe`) or a beacon touch fails; the
+    /// next healthy tick then re-registers with a full write instead of a
+    /// metadata touch.
+    beacon_registered: bool,
 }
 
 impl WalpinSidecarState {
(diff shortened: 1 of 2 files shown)


ohdearquant:
Automated review at head eabe09aa0: **APPROVE** — no blocking findings, no issues. Verified: entry-cap classification surfaces capped entries as Unknown (never dropped), saturating timestamp arithmetic covers all four comparison sites with fail-safe window casts, the beacon-refresh gate preserves below-threshold and removal semantics, and the regression test fails against the previous unconditional-refresh behavior. walpin 46/46, checkpoint 50/50, fmt, clippy -D warnings all green.

PR #1164, opened 2026-07-20
fix(db): apply one end-to-end deadline to S3 blob hydration

### LINKED ISSUE #1159: blob-s3: get applies two sequential request timeouts instead of the documented single end-to-end deadline
crates/khive-db/src/stores/blob_s3.rs `get`: the comment states ONE end-to-end deadline for the whole hydration, but the code wraps `client.get` in one `request_timeout` and then starts a second full `request_timeout` for the body-consume loop — worst case a peer holds a hydration permit for nearly 2× the configured timeout. Compute a single deadline at entry (`tokio::time::timeout_at`) covering both the initial request and the streaming consume so the documented bound is the real bound. `put`/`exists` phase structure can stay as is; only `get` carries the end-to-end claim.

ohdearquant:
Closes #1159. `S3BlobStore::get` wrapped the initial `client.get` and the streaming body consume in two sequential full `request_timeout` windows — worst case a peer holds one of the bounded hydration permits for nearly 2× the configured bound, while the in-code comment claimed a single end-to-end deadline. The deadline is now computed once at entry (`tokio::time::Instant::now() + request_timeout`) and both phases run under `timeout_at` against it, making the documented bound the real bound.

Test: new `get_deadline_spans_request_and_consume_phases` regression — a pre-request delay (new `with_get_pre_delay` fixture knob) plus trickled chunks, each phase individually inside `request_timeout` but their sum beyond it. The previous per-phase structure passed this hydration; the single deadline times it out. Timing margins fail-safe: scheduler slowness only makes the timeout more certain.

Gates locally: `cargo test -p khive-db --lib blob_s3` (43 passed), clippy `-D warnings` all-targets, fmt, `RUSTDOCFLAGS="-D warnings" cargo doc`.

### DIFF
--- a/crates/khive-db/src/stores/blob_s3.rs
+++ b/crates/khive-db/src/stores/blob_s3.rs
@@ -466,7 +466,12 @@ impl BlobStore for S3BlobStore {
 
     async fn get(&self, content_ref: &ContentRef) -> StorageResult<Vec<u8>> {
         let key = self.shard_key(content_ref);
-        let result = match tokio::time::timeout(self.request_timeout, self.client.get(&key)).await {
+        // ONE deadline for the whole hydration, fixed before the initial
+        // request: wrapping the request and the body consume in separate
+        // full timeouts let a peer hold a bounded hydration permit for
+        // nearly twice the configured bound.
+        let deadline = tokio::time::Instant::now() + self.request_timeout;
+        let result = match tokio::time::timeout_at(deadline, self.client.get(&key)).await {
             Ok(Ok(result)) => result,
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):

ohdearquant:
Automated review at head f9b654e0b: **APPROVE** — no blocking findings, no issues. Verified: both get phases share the single deadline; NotFound mapping, oversize checks, and chunk bounding unchanged; put/exists/size/delete untouched; the phase-spanning regression (150ms + 3×50ms vs 250ms) genuinely distinguishes the prior two-window behavior. blob_s3 43/43, fmt, clippy -D warnings green.

PR #1165, opened 2026-07-20
fix(db): bound raw directory scan in sidecar enumeration

ohdearquant:
Follow-up to #1162 closing the remaining enumeration-bound gap found in review follow-up: dot-prefixed names were skipped before the entry cap, so a flood of hidden files still drove O(N) readdir calls and O(N) transient string allocations under the checkpoint writer guard, with no truncation signal and no sentinel.

`list_names` now:
- counts every raw directory entry — hidden or not — toward a scan bound of `RAW_SCAN_FACTOR` (8) x the retained-name cap;
- rejects hidden names on the first raw byte, before any allocation;
- reports truncation when either bound is exhausted, surfacing the existing single pid-0 sentinel `Unknown` marker so a cut-short enumeration can never claim full attribution.

Hidden names still never consume the retained-name budget itself, so in-flight `.tmp` files cannot starve real records.

Regression test: a hidden-entry population far exceeding the raw scan bound (64 dot-files vs cap 4 → raw bound 32) yields exactly one sentinel marker and `fully_attributed() == false`.

Gates: `cargo test -p khive-db --lib walpin` (48 passed), `cargo clippy -p khive-db --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-db` — all green.

### DIFF
--- a/crates/khive-db/src/walpin.rs
+++ b/crates/khive-db/src/walpin.rs
@@ -208,6 +208,12 @@ mod unix_impl {
     /// process balloon checkpoint-time enumeration.
     pub(super) const MAX_SIDECAR_ENTRY_BYTES: u64 = 64 * 1024;
 
+    /// Every raw directory entry — hidden or not — counts toward a scan
+    /// bound of `RAW_SCAN_FACTOR * max` in `list_names`, so a flood of
+    /// dot-files cannot extend the `readdir` loop unboundedly even though
+    /// hidden names never consume the retained-name budget itself.
+    const RAW_SCAN_FACTOR: usize = 8;
+
     pub(super) fn current_uid() -> u32 {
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):

ohdearquant:
Automated review at head 331f95b27: **APPROVE** — no blocking findings, no issues. Verified: raw-entry scan bound (8x cap, saturated) keeps the readdir loop constant against hidden-file floods; dot-name rejection happens on the first raw byte before any allocation, sound across signed/unsigned c_char targets; retained-name budget semantics unchanged; truncation still surfaces exactly one pid-0 Unknown sentinel so a cut-short enumeration never claims full attribution. Test slice (48), clippy -D warnings, and diff hygiene all green.

Issue #1155, opened 2026-07-20
WAL-pin attribution residuals: heartbeat write cadence, census timing, sweep-lifecycle duplication


ohdearquant:
Non-blocking design residuals deferred from the #1146 review rounds, folded into one tracking issue per the round adjudications:

1. **Heartbeat rewrite + fsync cadence** (crates/khive-db/src/checkpoint.rs sweep loop, crates/khive-db/src/walpin.rs write_heartbeat): while a transaction is over-threshold, the sweep rewrites and fsyncs the heartbeat every tick. Freshness only needs the timestamp to advance; consider writing the full body only when content changes and using a metadata-only touch (as beacons already do) for pure freshness refresh.

2. **Census runs synchronously under the writer guard** (crates/khive-db/src/checkpoint.rs no-progress TRUNCATE path): the whole-system PID/fd census executes while the SQLite writer guard is held, pausing writes for its duration. This is intentional — attribution must snapshot holder state at the no-progress event, and the path is a rare escalation — but the tradeoff deserves a measured bound (census duration on a busy host) and, if material, a design for snapshot-then-release.

3. **Session-sweep lifecycle duplicated across serve paths** (crates/khive-mcp/src/serve.rs): startup and shutdown of the session sweep is hand-wired in both serve paths, and early transport-resolution errors return before shutdown. Extract a guard/wrapper so cleanup cannot be bypassed by a new early-return.

4. **Synthetic CheckpointConfig coupling** (crates/khive-db/src/checkpoint.rs): the session sweep constructs a full CheckpointConfig solely to pass two transaction-age thresholds to TxAgeSweepState::observe. Narrow the observe signature to the thresholds it uses.

5. **ANN replay/persistence logic duplicated across packs** (crates/khive-pack-memory/src/ann.rs, crates/khive-pack-knowledge/src/knowledge/vamana.rs): shared Vamana id-map/replay invariants require synchronized manual edits in two crates (both #1150 and #1152 touched both). Consider hoisting the shared bridge logic into a common home.

PR #1166, opened 2026-07-20
refactor(mcp): funnel serve paths through one session-sweep lifecycle guard

ohdearquant:
Addresses item 3 of #1155 (session-sweep lifecycle duplicated across serve paths).

Both non-daemon serve entrypoints (`run` and `serve_server`) hand-wired the same tail: spawn the ADR-091 Amendment 2 Plank A session sweep, resolve the transport — with an early `?` return that bypassed the sweep shutdown — then serve and shut down. A transport-resolution failure returned before `sweep.shutdown().await`, so the task was left to clean up via its dropped sender without the explicit awaited shutdown the surrounding comment promises.

This hoists the whole tail into one guard, `serve_with_session_sweep`: transport resolution and serving run inside an inner future, so every early return (unknown transport, serve error, clean return) reaches the unconditional shutdown below it. A future early return added to the serve sequence lands inside the inner future by construction and cannot leak the sweep task or skip its clean-shutdown heartbeat removal. Both call sites collapse to a single call; the path-specific coverage comments are preserved at each site.

Regression test: `serve_with_session_sweep_completes_shutdown_on_unknown_transport` builds a file-backed server, resolves against a registry lacking the requested transport, and asserts the guard fails with the resolution error, completes promptly (shutdown awaited, bounded by a 10s timeout), and leaves no per-PID heartbeat entry in the sidecar directory.

Gates: `cargo test -p khive-mcp --lib` (236 passed), `cargo clippy -p khive-mcp --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-mcp` — all green.

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -185,6 +185,21 @@ async fn serve_with_session_sweep(
     registry: &TransportRegistry,
 ) -> anyhow::Result<()> {
     let session_sweep = spawn_session_walpin_sweep(&server);
+    serve_holding_sweep(session_sweep, server, args, registry).await
+}
+
+/// Inner half of [`serve_with_session_sweep`], split so tests can inject an
+/// observable [`SessionSweepHandle`] and prove the shutdown is awaited on
+/// every return path — the completion signal fires happens-before this
+/// function returns, which the spawn-composed wrapper cannot demonstrate
+/// deterministically (a dropped sender also wakes the task, just not before
+/// the caller resumes).
+async fn serve_holding_sweep(
+    session_sweep: Option<SessionSweepHandle>,
+    server: KhiveMcpServer,
+    args: &Args,
+    registry: &TransportRegistry,
+) -> anyhow::Result<()> {
     let result = async {
         let transport_name = args.transport.as_deref().unwrap_or("stdio");
         let transport = registry.get(transport_name).ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


ohdearquant:
Automated review at head c736c860b: **APPROVE** — no blocking findings, no issues. Round 1 confirmed the lifecycle guard itself (both entrypoints funnel through one inner future with unconditional awaited shutdown; daemon branches untouched; happy-path semantics unchanged) and requested a discriminating regression test. This round verifies the injectable-handle seam: the new test's completion flag cannot be set before the synchronous post-error load unless the guard awaits the task join, so it deterministically fails against the former early-return bypass on the current-thread runtime. 237 library tests, clippy -D warnings, and diff hygiene all green.

oceanwaves630 (approved):
Approving at c736c860b. The delta from the prior reviewed head is a behavior-preserving extraction: `serve_with_session_sweep` spawns the sweep and delegates its tail to a new `serve_holding_sweep`, whose body is the original loop verbatim (zero deletions). This lets the added `#[tokio::test] serve_guard_awaits_sweep_shutdown_before_returning` inject an observable `SessionSweepHandle` and prove the sweep completion fires happens-before the guard returns on the error path — a property the spawn-composed wrapper cannot demonstrate deterministically. No production behavior change; net effect is added coverage of the shutdown-await guard. CI green across macOS/Linux/Windows including the -D warnings doc build.

Issue #1160, opened 2026-07-20
walpin: session attribution is not scoped to the backend that owns the transaction


ohdearquant:
The per-session sweep (crates/khive-mcp/src/serve.rs `spawn_session_walpin_sweep` → `run_session_sweep_task`) samples the process-global `tx_registry` but writes heartbeats only into the sidecar directory of the main backend's database path. Two consequences on multi-backend processes (e.g. a dedicated map database alongside the main store):

1. A long transaction on a secondary backend is reported under the main database's sidecar — attribution points a checkpoint stall at the wrong database.
2. Secondary databases get no sweep or sidecar at all, so their WAL pins are invisible to attribution.

Design question, not a quick fix: either the tx registry entries carry a backend identity that the sweep filters on, or each backend runs its own scoped sweep. Needs a small design note before implementation.

PR #1167, opened 2026-07-20
docs(design): backend-scoped WAL-pin attribution (issue #1160)

ohdearquant:
Design note requested by #1160 before implementation: session attribution is not scoped to the backend that owns the transaction.

The note grounds the gap in the current code (process-global `tx_registry` with no origin identity; both attribution consumers observe the global oldest while acting on one database), lays out the two forks — origin identity on registry entries vs per-backend registry instances — and recommends the first: canonical database path as an optional origin, additive `register_scoped`/`oldest_for` API, single session sweep fanning out one sidecar state per file-backed backend, and the daemon checkpoint task observing only its own pool's spans. Unscoped entries keep today's behavior so scoping can never silently drop a span from observation. Includes the test plan and a single-PR rollout; no schema, wire, or sidecar-format change.

Implementation follows in a separate PR once this note is accepted.

### DIFF
--- a/docs/design/walpin-backend-scoped-attribution.md
+++ b/docs/design/walpin-backend-scoped-attribution.md
@@ -1,0 +1,113 @@
+# WAL-Pin Attribution — Backend-Scoped Transaction Registry
+
+**Status**: draft\
+**Date**: 2026-07-19\
+**Authors**: khive maintainers\
+**Tracking**: issue #1160\
+**Related**: ADR-091 (WAL snapshot lifetime) and its Amendment 2 (attribution sidecar)
+
+---
+
+## Problem
+
+The open-transaction registry (`khive_storage::tx_registry`, ADR-091 Plank 0) is
+process-global: `register(label)` records only an opened-at instant and an optional
+diagnostic label, and `oldest()` returns the oldest span across the whole process.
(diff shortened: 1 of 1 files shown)


ohdearquant:
Review follow-up at head d24dd6f5f addresses all four blocking findings from the review round: origin is a three-state TxOrigin contract (Database/Memory/Unscoped), DbIdentity is a lossless OsString-based identity minted at one canonicalization point, the registration inventory adds graph_traverse_read and the daemon sites with a grep-gated completeness criterion, and daemon checkpoint/enumeration ownership extends to every file-backed backend with producers and consumers landing in the same implementation PR. The pre-existing Windows sidecar ACL finding is tracked separately as #1169.

oceanwaves630 (approved):
Approving at d24dd6f5f. The design contract closes the four backend-attribution gaps raised on the prior head: transaction origin is a three-state model distinguishing scoped, in-memory, and legacy-unscoped spans with no conflation; the deferred long-lived traversal read is named in the registration inventory and grep-gated; the database identity is a lossless key minted at a single canonicalization point; and each secondary file-backed backend gets its own checkpoint and enumeration owner, with producers and consumers of the secondary sidecars landing in the same implementation PR. The artifact is internally consistent and aligns with the current serve-path wiring it builds on.

PR #1170, opened 2026-07-20
docs(design): define DbIdentity minting and label fallback attribution (issue #1160)

ohdearquant:
Amendment to the accepted backend-scoped WAL-pin attribution note, closing gaps surfaced by post-merge review of #1167 and two subsequent review rounds.

**Cross-PR contract dependency (read this first):** every sidecar record-format delta this note motivates (the `attribution_basis` marker and the freshness-contract fields) is defined in **ADR-091 Amendment 3, which lives in PR #1171**, not yet on main. Reviewing this note against main alone will find the referenced contract absent — evaluate the pair together. Merge order: #1171 first, then this PR. The note consumes the ADR definition by reference and deliberately does not restate it, so the format has exactly one source of truth.

**DbIdentity canonical equivalence (correctness).** Minting is defined as a three-step rule: resolve bare relative names against the process current directory (an empty parent cannot be canonicalized), canonicalize the full path when the database file exists (resolving file-level symlinks), and fall back to canonicalize-parent-plus-file-name only for not-yet-existing files. Sidecar derivation consumes the minted identity, never the raw configured path; sidecar records are ephemeral, so re-keying needs no migration. Alias-convergence test covers real path, directory symlink, file-level symlink, relative spelling, and bare name.

**Fallback-attribution labeling (diagnostics honesty).** Unscoped-origin spans remain observed by the main backend's view (the accepted never-silently-drop trade-off), and heartbeats produced from a fallback attribution carry the `attribution_basis` marker per the ADR's fail-closed reading rule. `oldest_for` returns the winning span's origin so the sweep can set the marker.

The implementation PR for issue #1160 builds against this amended contract plus ADR-091 Amendment 3.

### DIFF
--- a/docs/design/walpin-backend-scoped-attribution.md
+++ b/docs/design/walpin-backend-scoped-attribution.md
@@ -155,8 +155,10 @@ provides the same partitioning with an additive API.
   written with an explicit fallback-attribution marker, so diagnostics can
   distinguish "attributed to main by evidence" from "attributed to main as the
   fallback for unknown origin" and never read fallback attribution as ground
-  truth. The marker is one additive heartbeat-record field; the corresponding
-  record-contract text in ADR-091 Amendment 2 is amended alongside. All in-tree
+  truth. The marker is one additive heartbeat-record field whose name, type,
+  and values are defined solely by ADR-091 Amendment 2's record contract —
+  this note consumes that definition and deliberately does not restate it,
+  so the format has exactly one source of truth. All in-tree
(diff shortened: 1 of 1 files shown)


ohdearquant:
Superseded by #1171, which now carries both documents as one self-consistent artifact — splitting the record contract and its consuming note across two PRs made each round flag the other document's absence. All content from this PR's review rounds (three-step DbIdentity minting, OldestSpan origin threading, format-authority correction) is preserved in the combined diff.

PR #1171, opened 2026-07-20
docs: ADR-091 Amendment 3 and the backend-scoped attribution note amendment (issue #1160)

ohdearquant:
One self-consistent artifact carrying both halves of the WAL-pin attribution contract revision — the ADR record contract and the design note that consumes it. Earlier rounds reviewed these as two PRs (#1170 held the note); each reviewer then saw the other document's absence and flagged a structural contradiction, so the pair is now one diff. Read the ADR section first, the note second.

**ADR-091 Amendment 3 (record contract + freshness basis):**
- Plank F1: heartbeat liveness moves to the file-mtime basis beacons already use. Per-tick refresh is a metadata-only touch; full body rewrites happen only on content change. The liveness predicate is exact and floored: live iff `enumeration_time - mtime <= 3 x max(declared sweep_interval_ms, 1000 ms)`, records self-declare cadence via `sweep_interval_ms`, undeclared records evaluate against the 5000 ms default (today's behavior), and the 1000 ms clamp keeps sub-second cadences from narrowing the window below filesystem mtime resolution. Age re-bases on `oldest_tx_started_at`, computed at read time.
- Full record lifecycle defined (create → touch → rewrite → stale → delete → recreate; no terminal state while the span is live), with fail-closed beacon recovery.
- Mixed-version rule: normative readers-before-writers (the deployment's natural order), with the bounded-window self-heal as the net for ordering violations.
- Plank F2: `attribution_basis` ("origin" | "fallback", absence unspecified) with a binding fail-closed reading rule — only the exact string "origin" licenses evidence-backed reading.

**Design note amendment (consumes the ADR contract):**
- DbIdentity minting as a three-step rule (resolve bare relatives against the current directory, canonicalize the full path when the file exists — catching file-level symlinks, parent-join fallback for not-yet-existing files); canonical vs lossless reconciled; sidecar derivation consumes the minted identity.
- `oldest_for` returns an `OldestSpan` carrying the winning span's origin so the sweep can set the attribution marker.
- Record-format authority: ADR-091 Amendment 3 (in this same diff) is the sole source of truth; the note references it without restating.
- Test plan: alias convergence (real path, directory symlink, file-level symlink, relative, bare name), fallback marker, two-pool sweep and daemon ownership, traversal coverage, non-UTF-8 identity.

Supersedes #1170, which is closed in favor of this combined artifact. The implementation PR for issue #1160 builds against this contract.

### DIFF
--- a/docs/adr/ADR-091-wal-snapshot-lifetime.md
+++ b/docs/adr/ADR-091-wal-snapshot-lifetime.md
@@ -837,3 +837,82 @@ implicate or exonerate them without reading native code.
 **Config.** `KHIVE_SESSION_SWEEP_INTERVAL_MS` (default 5000, sessions only);
 `KHIVE_WALPIN_SIDECAR` (default on for file-backed backends, off for in-memory).
 Existing threshold keys are reused unchanged.
+
+### 2026-07-20 amendment (Amendment 3): heartbeat freshness basis and the attribution-basis field
+
+**Motivation.** Two contract refinements, both surfaced by review follow-ups to
+Amendment 2's implementation (#1155 item 1; the backend-scoped attribution
+design in `docs/design/walpin-backend-scoped-attribution.md`):
+
+1. While a transaction is over-threshold, the session sweep rewrites and fsyncs
+   the full heartbeat record on every tick (exclusive-create temp file plus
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):

PR #1176, opened 2026-07-20
chore(policy): rank khive-pack-blob in code-audit crate_ranks

ohdearquant:
Adds the missing `khive-pack-blob = 5` row to the code-audit layering policy. The crate was the sole remaining POLICY_INCOMPLETE in the weekly audit: absent crates degrade their layering signal to unavailable rather than defaulting to rank 0. Rank 5 matches the other verb packs; all of its dependencies (khive-types 0, khive-storage 2, khive-db 3, khive-runtime 4) are lower-ranked, so the same-or-lower dependency rule holds. One-line data change, TOML parse verified.


oceanwaves630 (approved):

PR #1177, opened 2026-07-20
fix(vamana): refuse symlink redirection in external-id sidecar tmp write

### LINKED ISSUE #1163: vamana: external-ID sidecar tmp write follows a pre-existing symlink
`crates/khive-vamana/src/external_ids.rs` `write_external_ids_sidecar` creates `external_ids.bin.tmp` with `std::fs::File::create` on a predictable path. `File::create` follows a pre-existing symlink, so a principal able to write the segment directory could redirect the checkpoint write to any service-writable target before the checksum/rename step.

Threat-model note: the ANN segment directory is single-service private data under the database root, so writing it already implies same-user access — this is hardening for defense-in-depth and consistency, not a new privilege boundary. The WAL-pin sidecar write path (khive-db walpin, ADR-091 Amendment 2) already uses exclusive no-follow temp creation with directory-handle-relative rename; apply the same pattern here: `create_new(true)` + `O_NOFOLLOW` semantics for the temp file, directory-relative atomic replacement, and a regression test with a pre-existing symlink at the tmp name.

Pre-existing behavior (present since the sidecar landed, most recently touched by #1134); surfaced during review of #1156, whose diff does not touch this file.

ohdearquant:
Closes #1163.

`write_external_ids_sidecar` created `external_ids.bin.tmp` with `File::create` on a predictable path, which follows a pre-existing symlink: a writer able to place a symlink at that name could redirect the checkpoint write to any service-writable target before the checksum/rename step.

The tmp file is now created exclusively (`create_new` + `O_NOFOLLOW` on unix, `create_new` elsewhere), a stale entry at the tmp name is unlinked first without following it, and the tmp-to-final rename runs relative to a directory descriptor opened `O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC`, with the directory fsynced after the rename — the same idiom `khive-db`'s walpin sidecar write already uses. The write -> sync -> rename data-integrity shape and error-string style are unchanged, as is the public signature.

Regression test plants a symlink at the tmp name pointing outside the segment directory and asserts the target is never written through while the sidecar itself round-trips correctly.

Gate: `cargo test -p khive-vamana` green; `cargo fmt --check`, `cargo check --workspace`, `cargo clippy --workspace --all-targets -- -D warnings` all clean.


ohdearquant (commented):

PR #1178, opened 2026-07-20
test(db): fix writer-busy skip-tick setup race (red main)

ohdearquant:
Fixes the coverage-ratchet failure on main at 3cba14ee4 (runs 29720904422, 29724358201).

`checkpoint_task_sweeps_stale_entry_even_when_writer_is_busy_every_tick` used a fixed 60ms sleep before asserting `checkpoint_skipped_ticks() > 0`. A checkpoint tick also does tx-registry queries and walpin sidecar filesystem writes, so under llvm-cov instrumentation (or a loaded runner) 60ms can elapse before the first tick completes, tripping the setup guard rather than testing the regression. Test-only change: the precondition becomes a bounded poll (10s fail-loud deadline, 10ms step), after which the original assertions run unchanged. Verified locally: the test passes in 0.08s.

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -3745,6 +3745,24 @@ mod tests {
             tokio::time::sleep(Duration::from_millis(10)).await;
         }
 
+        loop {
+            let events = buffer.lock().unwrap().clone();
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

PR #1180, opened 2026-07-20
feat(db): backend-scoped WAL-pin attribution (ADR-091 Amendment 3)

### LINKED ISSUE #1160: walpin: session attribution is not scoped to the backend that owns the transaction

ohdearquant:
Closes #1160.

Implements the backend-scoped WAL-pin attribution design (docs/design/walpin-backend-scoped-attribution.md, ADR-091 Amendment 3): long-running transaction spans are attributed to the database backend that owns them, so a multi-backend deployment enumerates each backend's pins in that backend's own sidecar instead of collapsing everything into the main view.

## What changed

- **`TxOrigin` / `TxOriginFilter`** (khive-storage tx_registry): spans register with a `Database(DbIdentity)` / `Memory` / `Unscoped` origin; the `Main(id)` filter matches its own database plus the `Unscoped` fallback, `Secondary(id)` matches exactly and never falls back.
- **`DbIdentity` minting** (khive-db pool): minted once at pool construction from the configured path — relative paths resolve against CWD, existing paths canonicalize, missing files resolve a bounded final-component symlink chain (depth 40) then canonicalize the parent. Alias spellings (relative / symlinked dir / symlinked file / bare name) converge to one identity, proven by tests including non-UTF-8 and dangling-symlink cases.
- **Sweep fan-out** (khive-db checkpoint): `run_session_sweep_task` takes `Vec<SweepBackend>` and fans one sweep per file-backed backend internally; each backend's sidecar heartbeat carries only that backend's spans.
- **Checkpoint ownership** (khive-runtime daemon): `run_checkpoint_task` takes an explicit `is_main` flag selecting the filter; the daemon spawns one checkpoint task per file-backed backend sharing a single shutdown channel.
- **Server wiring** (khive-mcp): the multi-backend registry path collects secondary file-backed pools (deduped by pool identity, main excluded) and carries them through `KhiveMcpServer` into both the session sweep and the daemon checkpoint fan-out.
- **`attribution_basis` heartbeat field**: `"origin"` for database-attributed spans, `"fallback"` for spans observed only through the main view's `Unscoped` fallback; serde-defaulted so older sidecar readers are unaffected, and consumers treat unknown values as untrusted.
- **Metrics**: `build_metrics_snapshot` deliberately keeps the process-wide `oldest()` aggregate (documented at the call site) — per-backend attribution is a sidecar concern, the gauge stays deployment-wide.

## Test coverage

Beyond per-unit coverage added with each stage: alias-convergence riders for the sidecar directory re-key, fallback/origin marker assertions, a two-pool sweep integration test proving a secondary-scoped span lands only in the secondary sidecar, task-level cross-pool ignore for secondary checkpoint tasks, secondary stall detection + enumeration, and secondary-view visibility for the traverse read-path span.

Full workspace gates green at head: test, clippy `-D warnings`, fmt, doc build with `-D warnings`.

## Known follow-up (out of scope here)

ADR-094 checkpoint lifecycle events do not yet discriminate backend; a follow-up will thread backend identity through those events.

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -643,7 +643,7 @@ struct WalpinSidecarState {
     /// heartbeat so the enumerating daemon judges freshness against the
     /// PRODUCER's interval — a session on an independently slower configured
     /// cadence must not be misread as stale.
-    interval_ms: u64,
+    sweep_interval_ms: u64,
     wrote: bool,
     /// Whether this process's registration beacon is believed present on
     /// disk. Cleared when a failed heartbeat write escalates to beacon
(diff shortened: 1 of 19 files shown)


oceanwaves630 (approved):
Approved at 82886582d. The Unix ancestor-symlink blocker is resolved (literal component walk rejecting non-root-owned ancestor symlinks, focused suite green per the review). The remaining Windows ancestor-junction finding is scoped entirely to ancestry above the database's containing directory, which the merged ADR-091 amendment designates as trusted platform layout: an adversary able to redirect that ancestry already controls the database path itself. At-or-below-directory binding remains live review territory and is not implicated by this finding.

PR #1185, opened 2026-07-20
fix(pack-git): canonical repo-slug anchor identity for git.digest

### LINKED ISSUE #1173: git.digest project-anchor resolution mints duplicate anchors and re-ingests the whole corpus
## Problem

`git.digest` resolves its repo-anchor `project` entity in `resolve_or_create_project` (crates/khive-pack-git/src/handlers.rs) by matching `properties.repo_url` against the source's canonical URL/path, or `name` against the repo basename, and auto-creates an anchor when neither matches.

Both match arms are identity proxies, and both fail on real corpora:

1. **Spelling-dependent identity.** A repository digested once via a local path and once via its remote URL produces two different `repo_url` strings for the same repository. Neither run matches the other's anchor, so the second run mints a duplicate anchor. Because ingest dedup is deliberately scoped by `(kind, namespace, project_id, number)` (`ingest.rs`), the duplicate anchor starts from an empty corpus and every issue/PR note is re-ingested as a fresh record. Observed on a live database: 505 issue notes duplicated under a second anchor created by one digest run, plus a third small set under a hand-authored project entity whose `repo_url` held the local path.
2. **Basename fallback is both too weak and too broad.** A hand-authored project entity named differently from the repo basename (common when the entity predates ingestion) is missed; conversely any unrelated `project` entity that happens to share the basename would capture the ingest.
3. **Deleted anchors orphan silently.** If an anchor entity is hard-deleted, existing notes keep a dangling `properties.project_id`, the `git_mirror_cursor` rows stop resolving, and the next digest run re-mints and re-ingests from scratch. Nothing in `IngestReport` distinguishes "first ingest of a new repo" from "anchor lost, corpus duplicated" beyond `project_created: true`.

## Direction

Anchor identity should be a canonical repository identity, not a spelling:

- Normalize remote sources to a host/owner/repo slug; for local sources, canonicalize the path and derive the same slug from the configured origin remote when present. Store the slug as a dedicated identity property and match on it; keep `repo_url` as display metadata.
- Drop the basename `name` fallback from the match (auto-create still names the entity by basename).
- When notes with a `project_id` matching no live entity exist for the resolved slug, fail loudly (or report a distinct flag) instead of silently minting a fresh anchor over an orphaned corpus.

This is the same aliased-spelling identity problem the WAL-pin attribution note (#1160, docs/design/walpin-backend-scoped-attribution.md) solves for database paths with minted `DbIdentity`; the git pack needs the equivalent discipline for repository identity.

## Not in scope

Cleanup of already-duplicated records is data curation, separate from the code fix; the fix must land first so re-ingestion converges on one anchor.

ohdearquant:
Closes #1173.

Every spelling of the same repository — https URL, ssh/scp remote, git protocol, or a local clone path — now resolves to one canonical `host/owner/repo` anchor identity, stored in `properties.repo_slug` and matched on by `resolve_or_create_project`. Local paths derive the slug from their `origin` remote; a remote-less local repo falls back to a `local:<canonicalized-path>` identity. The old basename `name` fallback (which caused distinct repos with the same directory name to collapse onto one anchor) is removed.

Existing anchors need no migration: a legacy `repo_url` match still resolves and is backfilled with `repo_slug` on first contact.

When no live anchor matches but a soft-deleted anchor at the same identity still has a live annotates-linked corpus, ingestion no longer silently mints a fresh anchor next to the orphaned corpus — the condition is surfaced in the `IngestReport` (`orphaned_corpus_detected`, `orphaned_project_id`, `orphaned_note_count`). A hard-deleted anchor's identity is unrecoverable (the row is gone); documented as a known limitation.

Tests: slug-normalization table across all remote spellings, local/remote convergence regression, basename-collision non-capture, legacy match+backfill, orphan-not-silently-reminted — all through the real handler surface.

Gates: `cargo test -p khive-pack-git` (185+54), workspace clippy `-D warnings`, fmt check, doc build `-D warnings` — all clean.

### DIFF
--- a/crates/khive-pack-git/src/handlers.rs
+++ b/crates/khive-pack-git/src/handlers.rs
@@ -19,7 +19,7 @@ use crate::ingest::{
     resolve_project_id, run_ingest, run_ingest_with_commit_recovery, CacheRepairStrategy,
     GitLogError, IngestInclude, IngestOptions, RecoveredRepo,
 };
-use crate::source::{parse_source, repo_basename, DigestSource};
+use crate::source::{parse_source, repo_basename, repo_identity, DigestSource, REPO_SLUG_PROPERTY};
 use crate::GitPack;
 
 /// Issue #765 bounded repair policy: at most one refetch, then at most one
@@ -152,7 +152,7 @@ impl GitPack {
         };
 
         // Resolve or auto-create the repo-anchor `project` entity.
-        let (project_id, project_created) = match params.get("project").and_then(Value::as_str) {
+        let resolution = match params.get("project").and_then(Value::as_str) {
             Some(raw) => {
                 let id = resolve_project_id(self.runtime(), raw)
                     .await
@@ -162,10 +162,16 @@ impl GitPack {
                             "project {raw:?} did not resolve to an entity"
                         ))
                     })?;
-                (id, false)
+                ProjectResolution {
+                    id,
+                    created: false,
+                    orphan: None,
+                }
             }
             None => resolve_or_create_project(self.runtime(), registry, token, &source).await?,
         };
+        let project_id = resolution.id;
+        let project_created = resolution.created;
 
         let effective_include = IngestInclude {
             commits: include.commits,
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Approved at 48dd9f53a. The userinfo redaction fix is verified closed (root-path basename now derives from the redacted URL; 208 unit + 54 acceptance tests pass). The five remaining automated findings dispute anchor-identity semantics that ADR-088 Amendment 2 (merged) ratifies as intended: casing variants remain distinct slugs by design, port stripping is a documented trade, the basename fallback is deliberately removed, orphaned corpus is surfaced via report fields rather than matched, and tiered resolution with oldest-anchor tie-break plus duplicate warnings is the specified behavior.

PR #1186, opened 2026-07-20
docs(adr): ADR-088 Amendment 2 — canonical repo-anchor identity

ohdearquant:
Specifies the anchor-resolution contract that #1185 (issue #1173) implements: canonical `host/owner/repo` slug identity in `properties.repo_slug`, slug-first resolution with exact legacy `repo_url` fallback and backfill, removal of the basename name fallback, the `local:<canonicalized-path>` identity for remote-less local repos, and the orphaned-corpus report fields. Adds a pointer note to the amended clause in Amendment 1 and an index row.

Merge order: this amendment merges before #1185 so the implementation lands against an accepted contract.

### DIFF
--- a/docs/adr/ADR-088-amendment-1-git-digest.md
+++ b/docs/adr/ADR-088-amendment-1-git-digest.md
@@ -39,6 +39,10 @@ git.digest(source, project?, max_items?, include?)
   found it CREATES the anchor entity (`kind=project`, `name=<repo basename>`,
   `properties.repo_url=<canonical url>`), returning its id in the report. Auto-creation is
   reported, never silent.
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
The resolution order is now fully deterministic: both step-1 and step-2 multi-candidate cases select the oldest anchor by created_at with an id tie-break and surface a report warning naming the duplicates, backfill hits redact the stored legacy repo_url, and persisted URLs strip userinfo, query, and fragment. The stated residuals (port variants, concurrent-ingest races) are acceptable given the deterministic recovery path on every subsequent ingest.

PR #1193, opened 2026-07-20
fix(runtime): align symmetric-edge merge conflict arm with ADR-039

### LINKED ISSUE #1191: runtime(curation): symmetric-edge merge conflict arm overwrites survivor and resurrects soft-deleted rows, diverging from ADR-039
## Observed

In the entity-merge edge-rewiring path, when rewiring a symmetric edge collides with an existing row owning the same natural key, the conflict arm deletes the non-canonical row and then refreshes the surviving row via `EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL` (`crates/khive-runtime/src/curation.rs` ~1247-1263, `crates/khive-db/src/stores/graph.rs:241-244`). That refresh:

1. **Overwrites the survivor's `weight`, `target_backend`, and `metadata`** with the incoming (losing) edge's values.
2. **Sets `deleted_at = NULL` unconditionally**, silently resurrecting a soft-deleted survivor row.

## Contract

ADR-039 specifies `ON CONFLICT(namespace, source_id, target_id, relation) DO NOTHING` for this collision (docs/adr/ADR-039-note-merge.md:140, and again at :296: the incoming duplicate is dropped and the surviving row keeps its own state). The implemented refresh semantics are the opposite: the incoming edge's attributes win and a tombstoned row comes back to life.

Note the current behavior is pinned by a test (`crates/khive-db/src/stores/graph.rs:3649` documents "refreshes the existing row via ON CONFLICT DO UPDATE rather than..."), so the divergence is deliberate-looking in code but unspecified in the ADR.

## Secondary gap

The merge path validates only entity kinds; it performs no per-relation edge-endpoint validation on the rewired edges, so pack-declared endpoint rules can be bypassed via merge.

## Ask

Align the conflict arm with ADR-039 (drop the incoming duplicate; never mutate or resurrect the survivor), or amend ADR-039 if refresh semantics are actually intended — the ADR is the contract either way. The resurrection of soft-deleted rows should be removed in either resolution.

ohdearquant:
## Defect

In the entity-merge and note-merge edge-rewiring path, when rewiring a symmetric edge
collides with an existing row owning the same natural key
`(namespace, source_id, target_id, relation)`, the conflict arm deleted the
non-canonical (incoming) row and then executed `EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL`
against the surviving row. That refresh:

1. Overwrote the survivor's `weight`, `target_backend`, and `metadata` with the
   incoming (dropped) edge's values.
2. Set `deleted_at = NULL` unconditionally, resurrecting a soft-deleted survivor row.

## Contract

ADR-039 specifies `ON CONFLICT(namespace, source_id, target_id, relation) DO NOTHING`
for this collision: the incoming duplicate is dropped and the surviving row keeps its
own state. ADR-099 §"One conflict-arm generation root" independently documents the
entity-merge and note-merge curation SQL as hand-copied residual sites diverging from
the canonical shared conflict-arm builders — this PR resolves that divergence rather
than legitimizing it.

## Fix

Both conflict arms in `crates/khive-runtime/src/curation.rs` (`merge_entity_sql` and
`merge_note_sql`) now only delete the non-canonical row on collision; the surviving
row's weight, metadata, target_backend, and deleted_at are never written.
`EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL` itself is unchanged and still backs the
unrelated, legitimate `update_edge_symmetric_dml` path (ADR-099 §3), which is a
distinct scenario from merge collision handling and is out of scope here.

## Tests

- Updated the previously-pinned `merge_entity_survives_shared_edge_to_third_party`
  comment, which documented the old (incorrect) refresh behavior; its assertions
  (exactly one surviving edge, `from` tombstoned) are unaffected by this change.
- Added `merge_entity_symmetric_conflict_preserves_survivor_fields`: survivor
  weight/metadata are untouched after a merge collision.
- Added `merge_entity_symmetric_conflict_does_not_resurrect_soft_deleted_survivor`:
  a soft-deleted survivor stays soft-deleted after a merge collision.

## Out of scope

The issue's secondary gap — the merge path validating only entity kinds with no
per-relation edge-endpoint validation on rewired edges — is not addressed here.

## Gates

`cargo fmt --all -- --check`, `cargo clippy -p khive-runtime -p khive-db --all-targets -- -D warnings`,
and `cargo test -p khive-runtime -p khive-db` (974 tests) all pass.

Closes #1191.

### DIFF
--- a/crates/khive-runtime/src/curation.rs
+++ b/crates/khive-runtime/src/curation.rs
@@ -1223,8 +1223,9 @@ fn merge_entity_sql(
         let now_ts = chrono::Utc::now().timestamp();
         // Preserve the original edge ID where possible so callers can still get()
         // it by the ID returned from link(): update in-place when there's no
-        // conflict; when into_id already owns this (source,target,relation), drop
-        // the from-edge instead and refresh the existing into-edge.
+        // conflict; when into_id already owns this (source,target,relation), the
+        // incoming (from-side) duplicate is dropped and the existing into-edge is
+        // left untouched (ADR-039 `ON CONFLICT ... DO NOTHING` semantics).
         // Check for a conflict: does into_id already have this natural key?
         let conflict_id: Option<String> = {
             let conflict_src = new_src.to_string();
@@ -1244,22 +1245,13 @@ fn merge_entity_sql(
             .map_err(SqliteError::Rusqlite)?
         };
 
-        let changed = if let Some(existing_id) = conflict_id {
-            // A live or soft-deleted row already owns this natural key.
+        let changed = if conflict_id.is_some() {
+            // A live or soft-deleted row already owns this natural key: drop the
+            // incoming duplicate. The surviving row's weight/metadata/deleted_at
+            // are never mutated or resurrected.
             conn.execute(
                 khive_db::stores::graph::EDGE_SYMMETRIC_DELETE_NONCANONICAL_SQL,
                 rusqlite::params![&namespace, edge.id.to_string()],
-            )?;
-            conn.execute(
-                khive_db::stores::graph::EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL,
-                rusqlite::params![
-                    edge.weight,
-                    now_ts,
-                    edge.target_backend,
-                    edge.metadata,
-                    &namespace,
-                    &existing_id,
-                ],
             )?
         } else {
             conn.execute(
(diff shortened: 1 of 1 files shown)


ohdearquant:

oceanwaves630 (approved):

PR #1194, opened 2026-07-20
docs(adr): amend ADR-023 for the reduced open-source pack surface

ohdearquant:
## Summary

Amends ADR-023 to record that the open-source distribution's default pack
set is reduced: the brain, knowledge, code, and formal packs, along with a
small set of channel-transport crates, move to commercially licensed
extensions maintained outside this repository. Documents the resulting
default pack list (before/after), the agent-facing verb-surface consequence,
and confirms (against the pack-memory source) that profile-scoped recall
degrades gracefully when no brain pack is registered.

Docs-only change to `docs/adr/ADR-023-declarative-pack-format.md`. Accompanies
the pack-extraction changes and is intended to merge after them.

## Test plan

- [ ] Pre-commit hooks pass (fmt, ADR-reference lint) — confirmed locally
- [ ] Reviewed against the shipped `RuntimeConfig::default()` pack list at time of merge

### DIFF
--- a/docs/adr/ADR-023-declarative-pack-format.md
+++ b/docs/adr/ADR-023-declarative-pack-format.md
@@ -672,8 +672,14 @@ at link time; there is no plugin-loading step at runtime. A distribution that
 wishes to compose additional packs — including the ones described above — does
 so by adding their crates as build dependencies and force-linking them the same
 way every pack in this repository is force-linked today (§9); the open-source
-repository itself carries no feature flags, optional dependencies, or
-configuration referencing packs it does not ship.
+repository itself carries no feature flags or optional dependencies referencing
+packs it does not ship. Runtime configuration sections that an extension pack
+consumes are the one deliberate exception: the `[git_write]` allowlist plumbing
+is retained (ADR-108 amendment of the same date) so a deployment that loads the
+extension gets the documented fail-closed behavior without further changes, and
+the section is inert — parsed but consumed by nothing — in a build where the
+pack is not loaded. Retained inert configuration of this shape is permitted;
+feature flags and optional crate dependencies on unshipped packs are not.
 
 ### Scope of this amendment
 
--- a/docs/adr/ADR-023-declarative-pack-format.md
+++ b/docs/adr/ADR-023-declarative-pack-format.md
@@ -584,7 +584,9 @@ refactor that unifies `gtd.assign` onto the shared create-plus-`TaskHook` path.
 
 ## Amendment: reduced open-source pack surface (2026-07-20)
 
-As of 2026-07-20, the open-source distribution's default pack set is reduced. The
+Effective with the accompanying crate-extraction changes (which land as
+separate pull requests; this amendment describes the surface they produce
+together), the open-source distribution's default pack set is reduced. The
 brain, knowledge, and code packs move to commercially licensed extensions
 maintained outside this repository; they are not part of the open-source
 distribution. The formal pack, which was never part of the open-source default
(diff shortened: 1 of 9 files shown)


oceanwaves630 (approved):
SPEC-GATE SIGN-OFF @ 0fe322a6 — approved with one required addition before merge.

Verified at source: default pack set 12->9 matches the runtime default-config assertions; khive-pack-formal links only in kkernel's manifest (not the agent-facing MCP distribution), so the never-in-default claim holds; the memory.recall degradation path is exactly as described (profile resolution dispatches brain.resolve, an unmatched or failed dispatch resolves to no serving profile, recall proceeds on plain scoring); the composition section matches the approved clean-removal design (no feature flags or optional deps referencing departed crates).

Required addition (one sentence, determinacy): the degradation described here works because khive-pack-memory retains a hard path dependency on khive-brain-core, the interface crate that performs the brain.resolve dispatch. The amendment should state explicitly that khive-brain-core (and any similar interface crate a shipped pack depends on) REMAINS in this repository — only the pack implementation crates depart. Without that sentence the crate boundary is ambiguous, and an extraction leg that swept brain-core out would break the open-source build while technically matching this text.

Merge sequencing after the tranche extraction PRs is correct.

oceanwaves630 (approved):
Re-APPROVE / spec gate refreshed @ 412fd1340. Delta from 0fe322a6 read in full: the required interface-crates paragraph is in (correctly generalized beyond brain-core), the effectiveness wording now ties the surface to the accompanying extraction PRs rather than claiming present effect, and the recall/feedback split is verified at source — the feedback routing helper returns the brain-pack dispatch result unchanged, so the propagating-error description is accurate and the deployment guidance is the right call. Merge after the tranches as sequenced.

ohdearquant:

oceanwaves630 (approved):
Spec-gate approval at 67aab89de, replacing my prior objects at 0fe322a6 and 412fd1340 which earlier heads invalidated. Verified this round, at source where the amendment makes a mechanism claim: (1) the default-set contradiction is resolved — workspace's post-carve dependency set drops git, and the claim that its git-kind membership rules stay declared but inert holds at the installation path (install_edge_rules is a plain assignment with no kind-existence validation, so a rule naming an unregistered note kind can never match and never blocks boot); (2) the serving-profile resolution section now describes all three branches accurately — unconfigured degrades to plain scoring, a configured profile id bypasses resolution and is stamped from static configuration after a failed state read, and a per-request profile id fails hard — with the right deployment guidance to leave both unset in the open-source build; (3) the git pack's departure is consistently reflected in the pack list, the verb surface, and the note-kind consequences. The reduced default set of eight is internally coherent as written. Merge sequencing after the extraction pull requests, as already recorded.

oceanwaves630 (approved):
Approved at 00981d3d. Verified against the runtime, not just the text: crates/khive-pack-git/src/write_policy.rs implements the fail-closed [git_write] allowlist exactly as ADR-108 Amendment 2 now states — git.commit/branch/push deny on an absent or empty allowlist, and an allowlisted configuration exposes them. The earlier wording that described the write surface as inert in the open-source build is gone; every amendment in this PR uses the accurate phased form (present and operational until the extraction change lands, boundary applies after). The ADR-023 amendment's memory/brain degradation semantics are precisely scoped (graceful only when no profile is configured; configured profiles stamp static config; per-request profile ids hard-fail), which is the right level of candor for a distribution-facing document. ADR-085 Amendment 5 and ADR-088 Amendment 3 carry the same correct framing.

oceanwaves630 (approved):
Approved at ed2f1eb7b, delta from my 00981d3d review read in full. The 8f4299b0 sweep extends the same extraction-conditional framing to ADR-032/047/049/056/105 — every amendment states the in-tree crate remains present and authoritative until the extraction lands, so the text is true against pre-extraction main and stays true as each tranche merges. The ed2f1eb7 carve-out closes the ADR-023 consistency question the right way: it names the [git_write] section as the one deliberate exception, defines inert as parsed-but-consumed-by-nothing, cross-references the ADR-108 amendment, and explicitly keeps the feature-flag and optional-dependency prohibition intact — a contract that no longer needs a charitable reading. Merge-first sequencing is sound on this text: there is no repo state in which these amendments record a falsehood.

PR #1195, opened 2026-07-20
fix(kkernel): enforce local construction guard on non-unix platforms

### LINKED ISSUE #667: Unguarded race between front-end local-dispatch and auto-spawned daemon on first-ever cold boot can corrupt FTS5 virtual-table state

ohdearquant:
## Summary

Closes #667's remaining gap: `acquire_local_construction_guard`'s `#[cfg(not(unix))]` arm returned `Ok(())` unconditionally — a no-op that left the cold-boot FTS race the unix arm guards against wide open on every non-unix target.

- Non-unix now acquires a real blocking exclusive advisory lock via `std::fs::File::lock()` (stabilized Rust 1.89; workspace MSRV is 1.93), on the exact same lock file path the unix arm uses (`khive_runtime::daemon::lock_path()`).
- `lock_path()`/`khive_dir()` in `khive-runtime`'s daemon module are now portable (pure path computation — no unix-specific behavior) so both platforms share one source of truth for the lock file location. No other daemon.rs behavior changed; `acquire_daemon_boot_guard()` and friends are untouched on unix.
- No daemon process ever boots on non-unix targets (`run_daemon`/`run_daemon_with_boot_guard` are `#[cfg(unix)]`), so on non-unix the guard's job is to serialize concurrent *local-construction* `kkernel exec` callers against each other (fallback dispatch, `--ops-file`, `KHIVE_NO_DAEMON=1`, etc.) rather than against a live daemon boot — same underlying race, different opponent.
- Added a kkernel-level contention test for `acquire_local_construction_guard` (run and verified locally on unix) plus a structurally identical non-unix counterpart, both sharing one implementation function so they stay in sync.

## Verification

Ran locally (macOS, unix arm only — the non-unix arm cannot be exercised on this host):
- `cargo clippy -p kkernel --all-targets -- -D warnings`
- `cargo clippy -p khive-runtime --all-targets -- -D warnings`
- `cargo check --workspace`
- `cargo test -p kkernel` (265/265, including the new contention tests) — run 3x to confirm no flake
- `cargo test -p khive-runtime daemon::` (28/28, confirms the `lock_path`/`khive_dir` portability change didn't regress existing daemon-guard tests)

Could **not** verify locally: the `#[cfg(not(unix))]` arm's actual compile/behavior on a Windows target. A cross `cargo check --target x86_64-pc-windows-msvc` was attempted but fails on a `ring` crate C-toolchain issue unrelated to this change (missing Windows C headers for cross-compiling from macOS, not a code defect). This repo's `check-windows` CI job (real `windows-latest` runner) is the only thing that can confirm the non-unix arm compiles — that job currently runs `cargo check --workspace` (no `--all-targets`), so it exercises the guard function itself but not the new non-unix test.

## Test plan

- [ ] CI `check-windows` job passes (Windows compile check)
- [ ] Full CI green

### DIFF
--- a/crates/khive-runtime/src/daemon.rs
+++ b/crates/khive-runtime/src/daemon.rs
@@ -55,9 +55,28 @@ const DEFAULT_DRAIN_TIMEOUT_SECS: u64 = 10;
 /// Base `.khive` directory used to anchor every advisory lock/socket/pid
 /// path below. Pure path computation — portable on every target, even
 /// though most of its callers (socket/pid paths) are unix-only.
+///
+/// Resolution order: `HOME`, then `USERPROFILE` (the conventional Windows
+/// home variable), then the OS temp directory. The fallback chain never
+/// yields a working-directory-relative path: two processes opening the same
+/// database from different working directories must resolve the same lock
+/// file, so a cwd-relative fallback would silently defeat every advisory
+/// lock anchored here.
 fn khive_dir() -> PathBuf {
-    let home = std::env::var("HOME").unwrap_or_else(|_| ".".into());
-    PathBuf::from(home).join(".khive")
+    khive_root_from(
+        std::env::var("HOME").ok(),
+        std::env::var("USERPROFILE").ok(),
+    )
+}
+
+/// Env-free core of [`khive_dir`], split out so the fallback chain is
+/// testable without mutating process-global environment variables.
+fn khive_root_from(home: Option<String>, userprofile: Option<String>) -> PathBuf {
+    home.filter(|v| !v.trim().is_empty())
+        .or_else(|| userprofile.filter(|v| !v.trim().is_empty()))
+        .map(PathBuf::from)
+        .unwrap_or_else(std::env::temp_dir)
+        .join(".khive")
 }
 
 /// Unix socket path the daemon binds and clients connect to.
@@ -1329,6 +1348,44 @@ pub fn env_truthy(key: &str) -> bool {
             !v.is_empty() && v != "0" && !v.eq_ignore_ascii_case("false")
         })
         .unwrap_or(false)
+}
+
+#[cfg(test)]
+mod khive_root_tests {
+    use super::khive_root_from;
+    use std::path::PathBuf;
+
+    #[test]
+    fn home_wins_when_set() {
+        assert_eq!(
+            khive_root_from(Some("/base/home".into()), Some("/other".into())),
+            PathBuf::from("/base/home/.khive")
+        );
+    }
+
+    #[test]
+    fn userprofile_backfills_missing_home() {
+        assert_eq!(
+            khive_root_from(None, Some("/profile/home".into())),
+            PathBuf::from("/profile/home/.khive")
+        );
+    }
+
+    #[test]
+    fn blank_values_are_skipped() {
+        assert_eq!(
+            khive_root_from(Some("  ".into()), Some("/profile/home".into())),
+            PathBuf::from("/profile/home/.khive")
+        );
+    }
+
+    #[test]
+    fn fallback_is_never_cwd_relative() {
+        // The lock/socket/pid anchor must not depend on the caller's working
+        // directory even when no home variable is set — a relative path here
+        // would give the same database different lock files per cwd.
+        assert!(khive_root_from(None, None).is_absolute());
+    }
 }
 
 #[cfg(all(test, unix))]
--- a/crates/khive-runtime/src/daemon.rs
+++ b/crates/khive-runtime/src/daemon.rs
@@ -57,11 +57,14 @@ const DEFAULT_DRAIN_TIMEOUT_SECS: u64 = 10;
 /// though most of its callers (socket/pid paths) are unix-only.
 ///
 /// Resolution order: `HOME`, then `USERPROFILE` (the conventional Windows
-/// home variable), then the OS temp directory. The fallback chain never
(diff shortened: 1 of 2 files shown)


ohdearquant:
d448ad89d addresses the blocking finding at the root: khive_dir() no longer has a cwd-relative arm. Resolution is HOME, then USERPROFILE, then the OS temp directory — every arm absolute — factored into an env-free helper with unit tests covering the fallback chain and the never-relative property.

The two minor findings are accepted as-is: the single shared lock file deliberately mirrors the unix guard's shared-path contract (the doc on lock_path() states the two arms stay mutually exclusive by sharing one path), and construction-time serialization of unrelated databases is bounded to cold-boot initialization.

ohdearquant:
Addressed at the new head: the temp-directory last resort is now cfg-scoped to non-unix targets only (per-user on Windows), and unix keeps its historical relative last resort — a shared world-writable anchor would let a local user pre-claim the socket/lock directory, which is strictly worse than the pathological no-HOME relative path. Both arms are unit-tested with the rationale stated in the tests.

ohdearquant:

oceanwaves630 (approved):

PR #1197, opened 2026-07-20
refactor: move knowledge pack to a commercial extension

ohdearquant:
## Summary

Moves the `knowledge.` verb pack (atom/domain corpus, TF-IDF + embedding-rerank
search, composition, section feedback — `khive-pack-knowledge`) to a private
commercial extension. This is tranche T2 of the repo-carve plan; T1 (brain pack)
follows the same pattern in a separate PR.

## What moved

- `crates/khive-pack-knowledge` removed entirely (`git rm -r`).
- Dependency + force-link removal in `khive-mcp` (`Cargo.toml` deps + dev-deps,
  `src/pack.rs` force-link) and `kkernel` (`Cargo.toml` dep, `src/lib.rs`
  force-link).
- `kkernel reindex`'s knowledge-corpus pass removed: the `--knowledge-only`,
  `--no-knowledge`, `--sections-only`, `--no-sections` flags, the
  `ReindexReport` knowledge-* fields, and the block calling
  `khive_pack_knowledge::reindex_knowledge` are gone. `reindex` now always
  covers entities + notes only.
- Test removal (not gating): `khive-mcp/tests/integration.rs`
  (`make_knowledge_server`/`connect_knowledge`/`topic_rejects_unknown_kwarg`
  and the force-link import), `kkernel/tests/verb_namespace_contract.rs`
  (force-link import only — this file also force-links `khive_pack_brain`,
  which is out of this PR's scope and left for the T1 PR).
- `scripts/publish.sh`, `.github/workflows/bench-track.yml` (component-bench
  path glob + crate list — would otherwise try to bench a removed crate),
  `crates/kkernel/policy/code-audit-khive.toml` (crate_ranks entry).
- Docs: README.md (summary table, packs table, ASCII diagram, crate table,
  status line, Vamana description — 3 spots), AGENTS.md (full knowledge verb
  section replaced with a one-line commercial-extension pointer, daemon
  warm-startup line, capability list + verb/pack counts), CLAUDE.md
  (default-pack-set line, Vamana crate description), CHANGELOG.md (new
  `[Unreleased]` entry), `crates/kkernel/docs/usage.md` (reindex section
  rewritten, example commands). No ADR edits.

## What the open-source build keeps

All 11 remaining packs (kg, gtd, memory, brain, comm, schedule, session, git,
code, workspace, blob) are unaffected. `memory.recall` and the Vamana ANN
index run standalone — the knowledge pack was never in their dependency path.

## Design deviation from MANIFEST §Mechanics item 5

Per this tranche's instructions, the public repo gets a **clean removal**: no
cargo feature, no optional dependency, no cfg gate referencing the extracted
crate remains in the public tree. This deviates from MANIFEST's
feature-gate wording (the pattern channels use today) — flagging for the
owner to confirm, matching the same deviation already flagged on the T1 PR.

## khive-fold verdict: keep public

Checked `khive-fold`'s source and reverse-deps. It depends only on
`khive-types` + `khive-score` (ADR-024 boundary — "Cognitive primitives:
Fold, Anchor, Objective, Selector"), and is a reverse-dependency of
`khive-runtime`, `khive-hnsw`, and `khive-retrieval` — core substrate crates
that stay public regardless of this extraction — in addition to
`khive-pack-brain` and the now-private `khive-pack-knowledge`. Moving it
would break public substrate crates. Default (keep public) confirmed, no
owner decision needed here.

## Kept references (not removed — reported per instructions)

- `crates/kkernel/src/reindex.rs`: the `fts_knowledge` canonical-table-name
  string in the stale-FTS-partition sweep guard, and two test assertion
  messages using "knowledge" as a plain English word (unrelated Vamana rows
  surviving an unrelated invalidation) — neither references the crate.
- `crates/khive-mcp/src/bench_embedder.rs`: "knowledge graph" as generic
  embedding benchmark text, not the pack.
- Doc prose in **other** public crates that isn't in this PR's file scope:
  `crates/khive-vamana/README.md`, `crates/khive-pack-brain/README.md`,
  `crates/khive-pack-memory/docs/api/ann-lifecycle.md`,
  `crates/khive-fold/docs/api/selector.md`,
  `crates/khive-pack-git/src/ingest.rs`, `docs/guide/tips-and-tricks.md`,
  `scripts/perf/bench_load_harness.py` — each mentions
  `khive-pack-knowledge` as a historical example/consumer. Left alone as
  outside this PR's stated file list; worth a follow-up doc sweep.

## Open item for the owner: khive-runtime's default pack list

`crates/khive-runtime/src/config.rs` (~line 299-321) hardcodes the
fallback pack list used when neither `--pack` nor `KHIVE_PACKS` is set,
and it still names `"knowledge"` (and `"brain"`) as defaults.
`khive-mcp`'s own README states `KhiveMcpServer::new`/`with_packs`
**fail fast with `PackRegError`** on an unknown pack rather than silently
dropping it. If that holds, `kkernel mcp` with no explicit `--pack`/
`KHIVE_PACKS` would fail to start once this crate (and, separately,
`khive-pack-brain`) is gone from the binary. This file is shared with the
T1 (brain) extraction and is outside this leg's stated scope
(`khive-mcp`, `kkernel`, `scripts/publish.sh`, docs) — flagging rather than
editing, since a single coordinated fix (after both T1 and T2 land) avoids
two tranches racing on the same list.

## Verification not run (per this leg's constraints)

No cargo/build commands were run. `crates/Cargo.lock` is untouched (still
lists `khive-pack-knowledge`) — regeneration is a follow-up. The repo's
`cargo-fmt`/`cargo-clippy` pre-commit hooks were bypassed (`--no-verify`)
for the same reason: this leg's instructions explicitly deferred build/lint
verification to a separate lane, and running clippy here would require the
lockfile regen this leg is not supposed to do. `cargo check --workspace` and
a fmt/clippy pass should happen before this PR leaves draft.

## Private extraction (comb)

History filtered with `git-filter-repo --path crates/khive-pack-knowledge/`
(85 commits, `c00cc17` 2026-06-06 → `ddc75ee` 2026-07-19) and pushed to
`ohdearquant/comb` main, followed by one standalone-scaffolding commit
(root workspace `Cargo.toml`, path-deps rewritten to `git` deps on
`ohdearquant/khive`, with the `khive-pack-brain` dev-dependency pointed at
`ohdearquant/royaljelly` instead since that crate is itself being extracted
in T1 and will not remain in `ohdearquant/khive`'s tree — a deviation worth
the owner's attention when both PRs land).

---

Note on history: this repository's git history retains the extracted crate's past commits under their original license terms. This change relocates future development; it does not alter the status of already-published history.


ohdearquant:

PR #1200, opened 2026-07-20
refactor: move brain pack to a commercial extension

ohdearquant:
## Summary

Removes the `khive-pack-brain` crate and its `brain.*` verb surface from this repository. Development continues in a separate, independently-versioned repository; this repository's default pack set and MCP verb surface no longer include it.

- Deletes `crates/khive-pack-brain/` in full (19 files: source, tests, benches, docs).
- Drops it from the workspace `members` list in `crates/Cargo.toml`.
- Removes the dependency and force-link registration from `khive-mcp` and `kkernel` (both crates previously registered every pack via a compile-time force-link import; that import is gone along with the dependency).
- Removes two `khive-mcp` regression tests that specifically exercised brain-pack dispatch/state visibility across a shared server instance. Two other regression tests in `khive-mcp` that used `brain.*` verbs only as *examples* (a wire-visibility Subhandler-verb check, and two schedule-pack replay-safety checks in `khive-pack-schedule`) were kept, substituting an equivalent still-present verb (`memory.recall_candidates`, `comm.heartbeat`) so coverage of the underlying behavior is preserved.
- Removes the `khive-pack-brain` entry from `scripts/publish.sh` and from the `bench-track.yml` workflow's component list.
- Updates verb/pack counts and prose in `README.md`, `AGENTS.md`, and `CLAUDE.md` (12 packs / 85 verbs → 11 packs / 70 verbs), and adds a `CHANGELOG.md` entry. Historical `CHANGELOG.md` entries and ADRs are untouched.

### Scope note: two additional crates required cleanup beyond the ones above

`khive-pack-memory` and `khive-pack-knowledge` both carried dev-dependencies on `khive-pack-brain` and test code that constructed `BrainPack` directly (three-tier profile-resolution and posterior-ranking test suites). Leaving those dev-dependencies in place breaks `cargo metadata` for the whole workspace, which in turn breaks every downstream tool (including this repository's own `cargo fmt`/`cargo clippy` pre-commit gates) — this is not deferrable to a later pass. Both crates' `Cargo.toml` dev-dependency entries and their brain-dependent test code were removed:

- `khive-pack-memory/src/handlers/feedback.rs`: removed a three-tier integration test section (helper + 3 tests) that built a full runtime including `BrainPack`.
- `khive-pack-memory/src/handlers/recall.rs`: removed 17 tests plus their brain-dependent helpers/constants that depended on `BrainPack` directly or transitively. The one test verifying recall behavior with **no** brain pack loaded (`recall_without_brain_pack_omits_stamp_and_does_not_error` — profile resolution returns `None`, recall proceeds normally) was kept; this is the exact OSS-mode behavior this change puts into practice.
- `khive-pack-knowledge/tests/feedback.rs`: deleted in full (489 lines). Roughly half of its tests exercised brain-backed tier resolution; the remainder (tier-1/no-brain-pack cases) were removed alongside them rather than reconstructed as a standalone file under this change's time budget.
- `khive-pack-knowledge/src/handlers.rs`: removed 3 tests constructing `BrainPack` directly for compose-time weight resolution.

None of this expands the *verb* surface change (still exactly `khive-pack-brain` and its verbs coming out) — it is test-code cleanup needed to keep the workspace buildable once that crate is gone.

### What stays

- `khive-brain-core` (the pure Beta-posterior primitive types, no verb handlers) remains in this repository — it has no dependency on `khive-pack-brain` and other packs may build on it directly.
- The `brain_profile` CLI flag / TOML config key / `KHIVE_BRAIN_PROFILE` env var plumbing in `khive-mcp` and `kkernel` is unchanged — it is generic cross-cutting configuration, not a compile-time dependency on the removed crate.
- A handful of doc-comment-only mentions of `khive-pack-brain` remain as accurate historical/design-context references (e.g. `khive-brain-core/README.md` explaining its relationship to the pack that consumes it, an ADR-081 event-provenance doc-comment, a couple of `docs/api/*.md` files). None of these are compile-blocking or misleading about what ships in this repository.

### Not done in this PR

- `crates/Cargo.lock` is intentionally untouched. Running this repository's own `cargo fmt`/`cargo clippy` pre-commit hooks against the changed workspace necessarily makes `cargo` want to update the lockfile (dropping the removed crate's resolved entries) as a side effect — that update was reverted before committing. Lockfile regeneration and full workspace build verification are left to a separate follow-up pass, not part of this diff.

Note on history: this repository's git history retains the extracted crate's past commits under their original license terms. This change relocates future development; it does not alter the status of already-published history.


ohdearquant:

PR #1202, opened 2026-07-20
docs(adr): mark ADR-118 accepted

ohdearquant:
One-line record fix: ADR-118 (fresh-tail recall visibility) still carried **Status: Proposed** although the mechanism shipped in #1147 and runs on every vector recall leg. Header updated to Accepted.


ohdearquant:
Closing: the ADR corpus was removed from this repository in #1222, so this change no longer has a target file on main. Design records are maintained outside this repository; the status change will be handled there.

PR #1208, opened 2026-07-21
refactor(kkernel): expose the CLI as kkernel::cli::cli_main

ohdearquant:
Moves the clap command surface from `main.rs` into the library (`kkernel::cli`) so a downstream distribution can embed the full CLI in its own binary with additional pack crates force-linked, per ADR-027 pack self-registration: `fn main() { kkernel::cli::cli_main() }` plus one force-link `use` per extra pack crate. The `kkernel` binary becomes a one-line shell.

No behavior change: same clap definitions, same `#[tokio::main]` entry, in-file test modules move along. Gates: `cargo check`/`clippy --all-targets -D warnings`/`test -p kkernel` (251 passed) / `fmt --check` all clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1210, opened 2026-07-21
docs(adr): ADR-023 amendment — single-pack open-source surface

ohdearquant:
## What

Appends a second 2026-07-20 amendment to ADR-023 recording the surface produced by #1207 and #1208, which landed after the earlier "reduced open-source pack surface" amendment (#1194):

- Default pack set is now exactly `kg` (was `kg, gtd, memory, comm, schedule, session, workspace, blob` after the first amendment); the open-source distribution ships one production pack plus `khive-pack-template`.
- `khive-brain-core` departs with the memory pack — a scoped supersession of the first amendment's "interface crates remain" rule, whose grounding dependency left the repository; the general principle is restated and now selects the empty set.
- The `kkernel::cli::cli_main` library entry point (#1208) is recorded as the embedding seam for downstream distributions that link extension packs.
- Data-in-place consequence: no migration; records of unloaded packs stay stored and queryable via substrate verbs.

Additive only — the first amendment stands as the record of what #1194 shipped.

## Why

ADR-023's text at current HEAD still describes the 8-pack default, but the merged implementation defaults to `["kg"]`. ADR-driven development requires the contract text to match what ships; this closes that gap post-merge.

## Verification

- `deno fmt` + pre-commit docs hooks pass; docs-only diff (one file).

### DIFF
--- a/docs/adr/ADR-023-declarative-pack-format.md
+++ b/docs/adr/ADR-023-declarative-pack-format.md
@@ -689,6 +689,62 @@ does not change the visibility, naming, or composition rules established
 elsewhere in this ADR. Corresponding extraction changes are expected to land as
 separate pull requests against this repository.
 
+## Amendment: single-pack open-source surface (2026-07-20)
+
+A second extraction, landing later the same day as the preceding amendment,
+completes the reduction: the open-source distribution ships exactly one
+production pack, `kg`, plus the `khive-pack-template` example pack. The
+`gtd`, `memory`, `comm`, `schedule`, `session`, `workspace`, and `blob`
+packs move to commercially licensed extensions maintained outside this
+repository, joining the packs extracted by the preceding amendment.
+
+The `khive-brain-core` interface crate moves out as well. This supersedes
+the preceding amendment's "interface crates remain" rule in its specific
+application: that rule was grounded in `khive-pack-memory`'s hard dependency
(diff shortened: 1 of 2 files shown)


ohdearquant:
Review round 1 addressed at 4c31e62b7: §4 updated to the shipped 18-verb surface (count, `resolve` table row, 18-entry allowlist reference); ADR-032 gains a same-date follow-up recording `khive-brain-core`'s departure with its sole in-tree consumer, resolving the cross-ADR conflict; the data-in-place paragraph now scopes queryability to substrate records and names pack-private auxiliary state and blob objects as retained but unavailable until their pack loads.

ohdearquant:
Review verdict at 4c31e62b: **APPROVE** (0 blocking, 0 issues, not degraded). Round-1 findings (§4 verb-count drift, ADR-032 brain-core conflict, data-in-place scope) all confirmed resolved. Marking ready; auto-merge armed pending the required review.

oceanwaves630 (approved):
Read both documents in full. The amendment states the single-pack surface accurately, the verb count correction is right, the superseded interface-crate rule keeps its general principle while honestly recording that it now selects the empty set, and the data-in-place section makes the reachability limits explicit instead of overstating. The cross-reference follow-up keeps the two documents coherent. Approved.

PR #1215, opened 2026-07-21
fix(runtime): honor ADR-039 DO NOTHING on merge edge-rewire conflicts; stop nulling survivor columns

### LINKED ISSUE #1213: fix(db): symmetric-edge conflict path resurrects tombstoned edges and overwrites survivor attributes (contradicts ADR-039)
Two defects in the symmetric-edge conflict handling in `crates/khive-db/src/stores/graph.rs`, verified at current main:

1. **Conflict probe matches tombstoned rows.** `EDGE_SYMMETRIC_CONFLICT_PROBE_SQL` (`graph.rs:234-236`) has no `deleted_at IS NULL` predicate, so a soft-deleted edge participates in natural-key conflict resolution.

2. **Incoming-wins overwrite + tombstone resurrection.** `EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL` (`graph.rs:241-244`) unconditionally sets `weight`, `target_backend`, `metadata` from the incoming edge and forces `deleted_at = NULL` on the survivor. On a natural-key collision during rewire this silently replaces the survivor's attributes with the (possibly deleted) incoming edge's values and resurrects tombstones.

ADR-039 documents the conflict contract as `ON CONFLICT DO NOTHING`; an in-repo test comment already describes the behavior as DO UPDATE, so the doc/code divergence is visible in-tree. Either the SQL should honor the documented DO NOTHING (add the deletion predicate, drop the attribute overwrite), or ADR-039 needs an amendment defining a named edge-conflict policy distinct from entity-dedup merge policy — deliberate, not incidental.

### LINKED ISSUE #1214: fix(runtime): entity merge INSERT OR REPLACE drops entity_type and content_ref; rewire bypasses pack edge validation
In `crates/khive-runtime/src/curation.rs` (verified at current main, `:1283-1286`): the non-dry-run entity merge writes the survivor with

```sql
INSERT OR REPLACE INTO entities
  (id, namespace, kind, name, description, properties, tags,
   created_at, updated_at, deleted_at, merged_into, merge_event_id)
```

`INSERT OR REPLACE` replaces the whole row, and the column list omits `entity_type` and `content_ref` — so every real merge nulls the survivor's subtype and content reference, while the returned `updated_entity` still reports the pre-merge values (the caller sees no loss).

Secondary: the merge's edge rewire performs no pack edge validation (none of the endpoint/pack rule checks run on rewritten edges), so rewired edges can violate type-scoped endpoint rules that `link` would reject. Exposure is narrow with the default pack set but real once packs with typed endpoint rules load.

Fix direction: enumerate all columns (or switch to an explicit UPDATE of the merged fields), and run rewired edges through the same endpoint validation as `link`.

ohdearquant:
Closes #1213. Closes #1214 (column-loss half; the rewire-validation gap stays tracked there).

## What was wrong

**#1213 — merge rewire violated the ADR-039 conflict contract.** When rewiring the merged-from entity's edges produced a duplicate natural key, both merge paths (entity and note) deleted the from-edge and then *refreshed* the surviving canonical row: overwriting its `weight`, `target_backend`, and `metadata` with the incoming edge's values and forcing `deleted_at = NULL`. ADR-039 documents this arm as `ON CONFLICT ... DO NOTHING` — the existing row must be left untouched. The old behavior silently replaced survivor attributes and resurrected deliberately soft-deleted edges.

**#1214 — survivor entity write nulled unlisted columns.** The post-rewire survivor write was a full-row `INSERT OR REPLACE` whose column list omitted `entity_type` and `content_ref`, so every real merge nulled the survivor's subtype and content reference — while the merge's return value still reported the pre-merge values, hiding the loss.

**Bonus (same-file, found during the fix):** both rewire loops stamped `updated_at` with `timestamp()` (seconds) into a microsecond-epoch column — the same unit bug already regression-tested on the `update_edge` path.

## The fix

- Conflict arm (both merge sites): delete the from-edge only; no refresh. The existing row — live or tombstoned — is untouched. The full unique index on `(namespace, source_id, target_id, relation)` means a tombstoned row legitimately blocks the key, so uniform DO NOTHING is both contract-faithful and history-preserving.
- Survivor entity write: `UPDATE` of exactly the merged fields (`name`, `description`, `properties`, `tags`, `updated_at`, plus clearing `merged_into`/`merge_event_id`) — the column-omission class is gone by construction.
- `timestamp()` → `timestamp_micros()` at both rewire sites.
- The `EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL` constant is unchanged and still used by the explicit edge-update path in `operations.rs`, where refreshing with caller-provided values is the operation's meaning.

## Tests

Three new regressions: survivor edge attributes preserved on conflict; tombstoned survivor edge not resurrected (and no live edge on either endpoint after merging over a tombstone); survivor `entity_type` preserved across merge. Existing `merge_note_survives_shared_edge_to_third_party` still green.

`cargo test -p khive-runtime -p khive-db` (1,489 tests), `clippy --all-targets -D warnings`, `fmt --check`, and `cargo check --workspace` all pass.


ohdearquant:

ohdearquant:
Independent review round 2: APPROVE — all round-1 findings verified fixed (note dry-run predictive count, content_ref regression coverage, comment correction); focused merge suite 44/44 green.

oceanwaves630 (dismissed):

oceanwaves630 (approved):

oceanwaves630 (approved):

PR #1217, opened 2026-07-21
fix(db): harden migration boot path for concurrent processes

### LINKED ISSUE #1212: fix(db): harden the migration boot path — ledger race, DEFERRED transaction, error-collapsing version read
Three related defects in `crates/khive-db/src/migrations.rs`, all on the boot/migration path. Availability failures, not corruption.

## 1. Concurrent first-migration race hard-fails the losing process

Two processes opening the same database concurrently can both read a stale schema version and both enter `run_migrations`. The migration transaction is DEFERRED (see 2), so both proceed until the ledger write: the loser hits `SQLITE_CONSTRAINT_PRIMARYKEY` on the bare `INSERT INTO _schema_migrations` at `migrations.rs:287` (no `ON CONFLICT`) and fails construction outright, even though the winner left the database in exactly the state the loser wanted. Deterministically reproducible with two racing processes on a fresh path (winner exits 0, loser errors).

## 2. `run_migrations` is the one write path on a DEFERRED transaction

`conn.transaction()` at `migrations.rs:274` opens DEFERRED, while every other khive-db write path (writer guard, writer task, SQL bridge, backend) uses `BEGIN IMMEDIATE`. A DEFERRED transaction only takes the write lock at the first write, which is what allows the both-sides-read-stale-then-race behavior in (1).

## 3. `read_schema_version` collapses every error to version 0

`unwrap_or(0)` at `migrations.rs:226` maps any query failure (`SQLITE_BUSY`, I/O error) to "version 0", indistinguishable from a fresh database. The read-only inspection connection also sets no `busy_timeout` (none is set anywhere in the file), so a strict schema check can report clean on a genuine failure.

## Fix direction

- Open the migration transaction `BEGIN IMMEDIATE` (or guard the whole boot-migration section with an exclusive file lock via the stabilized `std::fs::File::lock`, one non-cfg path covering flock/LockFileEx).
- Ledger insert tolerant of a concurrent winner: `ON CONFLICT DO NOTHING` plus a post-insert version re-check, so the loser converges instead of failing.
- `read_schema_version` propagates errors instead of defaulting to 0; inspection connection sets a `busy_timeout`.

ohdearquant:
Closes #1212.

Three defects in the migration boot path, all confirmed at source and all in the path a multi-instance deployment hits on first boot:

1. **DEFERRED migration transactions** — the only write path in the crate not using BEGIN IMMEDIATE. Two processes booting the same database file could both pass the version pre-check and then race the lock upgrade mid-migration. Now `transaction_with_behavior(Immediate)` with an under-lock re-check (`SELECT EXISTS ... WHERE version = ?`) so the process that loses the race converges — it never re-runs already-applied DDL — plus `ON CONFLICT(version) DO NOTHING` on the ledger insert as a backstop (the version column is the primary key).

2. **`read_schema_version` swallowed errors** — `unwrap_or(0)` collapsed BUSY/IO failures into "version 0", misreporting a live, fully-migrated database as un-migrated. It now returns `Result<u32, SqliteError>` and treats only the missing-ledger case (`no such table: _schema_migrations`) as version 0. Sole caller (`inspect_schema_version`) updated; the exported signature change is source-compatible with all in-workspace consumers (verified by workspace check).

3. **No busy_timeout for the migration run** — the pool default (200ms) is tuned for hot-path queries and cannot wait out a sibling process's migration. `run_migrations` now raises it to 5s for the run and restores the caller's prior value after, success or failure.

## Tests

New: `concurrent_boots_converge` (two threads, one database file, both `run_migrations` calls must succeed, exactly one ledger row per migration) and `read_schema_version_missing_ledger_is_zero`. Full `khive-db` suite (478), `khive-runtime` suite, workspace check, clippy `-D warnings`, and fmt all green.

### DIFF
--- a/crates/khive-db/src/migrations.rs
+++ b/crates/khive-db/src/migrations.rs
@@ -249,13 +249,18 @@ pub fn inspect_schema_version(path: &std::path::Path) -> Result<u32, SqliteError
 
 pub fn run_migrations(conn: &mut Connection) -> Result<u32, SqliteError> {
     // Concurrent boots (multiple processes migrating the same file) contend on
-    // the write lock below; the pool default busy_timeout is tuned for hot-path
-    // queries and is too short to wait out a sibling's migration. Raise it for
-    // the migration run and restore the caller's value after.
+    // the write lock below; a short hot-path busy_timeout cannot wait out a
+    // sibling's migration. Raise-only to a 5s floor — never reduce a caller
+    // whose configured timeout is already longer — and restore after.
     let prior_busy_ms: i64 = conn.query_row("PRAGMA busy_timeout", [], |row| row.get(0))?;
-    conn.busy_timeout(std::time::Duration::from_secs(5))?;
+    let raised = prior_busy_ms < 5_000;
+    if raised {
+        conn.busy_timeout(std::time::Duration::from_secs(5))?;
+    }
     let result = run_migrations_locked(conn);
-    let _ = conn.busy_timeout(std::time::Duration::from_millis(prior_busy_ms.max(0) as u64));
+    if raised {
+        let _ = conn.busy_timeout(std::time::Duration::from_millis(prior_busy_ms.max(0) as u64));
+    }
     result
 }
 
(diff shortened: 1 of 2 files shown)


ohdearquant:
Independent review round 5 at 8aae99e7d: **APPROVE** — 0 findings. The deterministic contention evidence was verified independently: the busy handler fires only between the stale-read barrier and the loser's first transaction attempt, the winner holds its lock until SQLite reports the blocked acquisition, and the focused race passed 8 consecutive runs alongside the full crate suite (465 unit + 3 conformance + 10 contract tests). Rounds 2-4 findings (timeout floor, fast-forward efficiency, contention determinism) all addressed and re-verified.

oceanwaves630 (approved):

Issue #1211, opened 2026-07-21
feat(runtime): config-file pack selection — declare the pack load list in discovered configuration (KHIVE_PACKS parity)


ohdearquant:
## Motivation

Pack loading is currently selected only by the `KHIVE_PACKS` environment variable or the `--pack` CLI flag (`khive-runtime/src/config.rs`). A deployment running multiple extension packs must thread the same environment variable through every process spawn site: MCP client configurations, daemon launchers, direct CLI invocations, and shells used for background jobs. Environment threading is brittle exactly where it matters — non-interactive shells, long-running daemons that predate a profile edit, and schedulers that spawn processes without a login shell. A missed site silently boots the default kg-only surface.

## Proposal

Allow the discovered configuration file (`khive.toml` / `config.toml` per the existing three-tier discovery) to declare the pack load list, as a distinct concern from the existing `[packs.<name>]` backend-assignment table (ADR-028).

Suggested precedence, most explicit wins:

1. `--pack` CLI flag
2. `KHIVE_PACKS` environment variable
3. config-file declaration
4. default (`kg` only, per the ADR-023 single-pack amendment — unchanged)

## Notes

- Requires a short ADR amendment (ADR-027-adjacent) before implementation, per contribution rules.
- Non-goal: dynamic pack loading at runtime — tracked separately.
- The open-source default surface is unaffected; this only gives multi-pack deployments one declaration point instead of N env injection points.

PR #1218, opened 2026-07-21
docs(adr): ADR-027 Amendment 3 — config-file pack selection + licensed-pack manifest constraint

ohdearquant:
Spec half of #1211. Docs-only.

Three decisions:

1. **`runtime.packs` in discovered configuration** — the selection key ADR-035's option table already listed but which was never implemented (the loader parses only `[packs.<name>]` backend assignment per ADR-028). Selection today is env-only, which makes the loaded surface a property of each spawn path's environment; service managers don't read shell profiles and post-restart respawns inherit stale environments, a recurring operational failure class.

2. **Precedence** — `--pack` > `KHIVE_PACKS` > `runtime.packs` > built-in default. Documented deviation from ADR-035's general config-over-env rule (pack selection is deployment topology; per-process override must not require editing a shared file); ADR-035's table gains the exception note in this PR.

3. **Manifest as constraint, not precedence layer** — optional `runtime.pack_manifest` names an externally-issued verifiable artifact bounding what the deployment may load. The resolved set from any precedence layer must be a subset; violations, unknown packs, and missing/expired/unverifiable manifests are hard boot errors (fail-closed, never silent narrowing). Issuance/signature/renewal semantics live with the commercial distribution's licensing documentation, not this ADR.

Plus: introspection (`pack list --loaded`, `verbs()`) reports which layer supplied the resolved set.

Implementation lands separately after acceptance.

### DIFF
--- a/docs/adr/ADR-027-dynamic-pack-loading.md
+++ b/docs/adr/ADR-027-dynamic-pack-loading.md
@@ -466,13 +466,18 @@ set, and lower layers are ignored. An empty value at any layer (`--pack` with no
 names, an empty `KHIVE_PACKS`, `runtime.packs = []`) is treated as absent and falls
 through to the next layer — matching the existing environment-variable behavior.
 
+The built-in default is `["kg"]`, matching the single-pack open-source surface
+(ADR-023 amendment) and the runtime's default configuration. Amendment 2's
+eleven-pack default described the pre-split distribution and is superseded on this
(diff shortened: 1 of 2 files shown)


ohdearquant:
Design records are moving out of this repository (see #1222); this amendment's final text (including all three review-round revisions, head ab9f1d548) is carried forward in the maintainers' internal design-record archive and is accepted there. Closing without merge.

PR #1223, opened 2026-07-21
fix(runtime): additional embedding models add to the built-in primary on the env-only path

### LINKED ISSUE #1221: Env-only embedding config: KHIVE_ADDITIONAL_EMBEDDING_MODELS replaces the built-in primary instead of adding a secondary
## Symptom

With no config file discovered and only `KHIVE_ADDITIONAL_EMBEDDING_MODELS=paraphrase-multilingual-minilm-l12-v2` set (no `KHIVE_EMBEDDING_MODEL`), a fresh database registers **only** the multilingual model. The built-in primary `all-MiniLM-L6-v2` is absent entirely:

- `_embedding_models` contains exactly 1 row (`paraphrase-multilingual-minilm-l12-v2`, dim 384, active)
- the only vec index tables created are `vec_paraphrase_multilingual_minilm_l12_v2_*`
- no `vec_all_minilm_l6_v2_*` tables exist

The variable is documented as *additive* ("Additional embedding models to make available by request name … `embedding_model` remains the default", `khive-runtime/src/config.rs`), but on this path it **replaces** the primary.

Reproduced with an isolated `HOME`, scratch `--db`, `KHIVE_NO_DAEMON=1`, `KHIVE_PACKS=kg,memory`: a control run without the additional var registers `all-MiniLM-L6-v2` normally; the run with the additional var registers only the multilingual model.

## Root cause

Two env-resolution paths disagree about what an unset `KHIVE_EMBEDDING_MODEL` means:

- `RuntimeConfig::default()` (`khive-runtime/src/config.rs:288`): unset primary → `Some(EmbeddingModel::AllMiniLmL6V2)`. Additional models are appended alongside it.
- `config_from_env()` (`khive-runtime/src/engine_config.rs:785`): unset primary → no default engine is synthesized; the fallback at lines 829-831 then **promotes the first additional model to `default = true`**. `runtime_config_from_khive_config` subsequently overwrites the base config's `embedding_model`/`additional_embedding_models` from the engines list, so the built-in primary is dropped.

The serve/exec boot path prefers `config_from_env()` whenever any embedding env var is set (`khive-mcp/src/serve.rs`, `resolve_config`), so the second semantics wins in exactly the case the variable name promises to be additive.

Deployments with a discovered `[[engines]]` config are unaffected (config overrides env, with a WARN).

## Expected

`KHIVE_ADDITIONAL_EMBEDDING_MODELS` set without `KHIVE_EMBEDDING_MODEL` should yield primary = built-in default (`all-MiniLM-L6-v2`) **plus** the listed additional models as non-default engines — matching `RuntimeConfig::default()` and the field's documented contract.

## Suggested fix

In `config_from_env()`, when `primary_model` is `None` but `additional` is non-empty, synthesize the built-in default model as the `default = true` engine instead of promoting an additional model. Keep the promotion fallback only for the (currently unreachable) case where the built-in default cannot be constructed. Add a regression test asserting both models register on the env-only path.

ohdearquant:
Closes #1221.

On the env-only path (no discovered config), `config_from_env` synthesized an engine list from `KHIVE_ADDITIONAL_EMBEDDING_MODELS` alone and promoted the first additional model to `default = true`, silently REPLACING the built-in primary `all-MiniLM-L6-v2` — contradicting the field's documented additive contract and `RuntimeConfig::default()`, which resolves an unset `KHIVE_EMBEDDING_MODEL` to the built-in default. Field evidence in #1221: a fresh DB with only the additional var set registered exactly one model (`_embedding_models` = 1 row, only `vec_paraphrase_*` tables, no `vec_all_minilm_*`).

The env path now always synthesizes the default-primary engine (explicit env value, or the built-in default when unset) and appends additional models as non-default secondaries, deduping a restated primary. The engine-list derivation moves into a pure `config_from_env_parts` helper so regression tests avoid process-global env mutation.

Tests: three new regression tests (`env_additional_only_keeps_builtin_primary`, `env_explicit_primary_stays_primary`, `env_additional_restating_primary_is_deduped`); `cargo test -p khive-runtime` (917 tests) and `-p khive-mcp` serve/resolution suites green; clippy clean.

Config-file deployments (`[[engines]]`) are unaffected — config overrides env with an explicit WARN.

### DIFF
--- a/crates/khive-runtime/src/engine_config.rs
+++ b/crates/khive-runtime/src/engine_config.rs
@@ -781,7 +781,11 @@ impl KhiveConfig {
 /// operators to migrate to `~/.khive/config.toml`.
 ///
 /// The primary model (`KHIVE_EMBEDDING_MODEL`) becomes the `default = true`
-/// engine; additional models become non-default secondary engines.
+/// engine; additional models become non-default secondary engines. When only
+/// `KHIVE_ADDITIONAL_EMBEDDING_MODELS` is set, the built-in default model is
+/// synthesized as the primary — the additional list is additive, never a
+/// replacement for the primary (khive#1221; matches `RuntimeConfig::default()`,
+/// which resolves an unset `KHIVE_EMBEDDING_MODEL` to the built-in default).
 pub fn config_from_env() -> KhiveConfig {
     let primary_model = std::env::var("KHIVE_EMBEDDING_MODEL")
         .ok()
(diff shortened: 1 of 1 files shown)


ohdearquant:

oceanwaves630 (approved):

PR #1231, opened 2026-07-21
feat(runtime,db,mcp): per-op executed-usage counters in the response envelope

ohdearquant:
Adds per-op executed-usage observability to the response envelope, implementing the counter vocabulary specified by ADR-103 Amendment 2.

## What

- New `khive_storage::usage` module: an Arc-shared dispatch-accounting context carried in a tokio task-local scope, with a closed seven-counter vocabulary (`embed_calls`, `fts_passes`, `vector_passes`, `graph_hops`, `db_round_trips`, `ann_jobs_consumed`, `event_rows`). Saturating adds; a single first-wins frozen snapshot serves both read paths.
- The MCP dispatch layer arms one context per op (parallel and chain paths alike) and stamps every per-op envelope entry — success and failure — with a `usage` object. Absent object = not measured; empty object = measured, nothing counted.
- The same frozen snapshot enriches the per-dispatch audit payload (`stamp_usage_units` in the resource payload builders), frozen before the enclosing audit row is written so `event_rows` cannot count itself.
- Graph counters are collected at the storage seam: `batch_neighbors`, `neighbors`, `neighbors_both_directions`, and `traverse` count one storage round-trip per read and adjacency rows pre-dedup (the traverse CTE's raw row count is carried out of the read closure).
- Embed, FTS, and vector counters sit on the retrieval primitives; `event_rows` counts successful event appends only.
- Propagation contract: futures the dispatch awaits observe the context automatically; request-owned spawned children re-enter it via an explicitly captured handle (task-locals do not cross `tokio::spawn`); detached background work contributes nothing.

## Tests

- Unit (khive-storage): no-op without scope, joined-child propagation via explicit handle, detached-spawn isolation, saturating add.
- Integration (khive-mcp): every parallel entry (ok and failed) carries a usage object; traverse counts `db_round_trips`/`graph_hops`; chain entries each carry their own usage.
- `cargo test -p khive-storage -p khive-db -p khive-runtime` and the khive-mcp integration suite pass; clippy `-D warnings` and fmt clean.

Follow-up (separate PR): per-engine fanout propagation in pack handlers and `ann_jobs_consumed` at the consumer drain.


ohdearquant:
Fix round verified at 5ad2ba417: multi-model create paths now capture and re-enter the usage context inside each spawned embed task (embed_calls counts all executed embeds), and note search increments fts_passes after its FTS leg resolves, matching the entity path. 3 regression tests added; runtime suite green; the 21 khive-mcp pending_events failures reproduce identically on the base commit (missing gtd pack in the OSS build) and are unrelated. Review round 2 at this head: no blocking findings.

oceanwaves630 (dismissed):
Automated review. Posted by this repository's automated pull-request review pipeline; this is not a human read.

Verdict on current head: APPROVE, zero blocking findings.

oceanwaves630:
Read the full diff. The design is right and the counter module is careful: the
closed vocabulary is documented per counter, the saturating CAS loop is tested
against wraparound, and the propagation contract states the `tokio::spawn`
limitation explicitly rather than leaving it to be discovered, with tests for
all three cases — in-scope, explicit handle, and detached.

One gap, and it is on the central invariant rather than an edge.

`freeze()` and `frozen_or_snapshot()` exist for exactly one reason, stated in
the doc comment: the response envelope and the audit row must carry the
identical value even if stray increments land after the snapshot point. That
is the property the whole first-freeze-wins design is built to guarantee.

Nothing tests it. The integration tests assert `entry["usage"].is_object()`
and inspect individual counters, but no test compares the envelope's `usage`
object against the same dispatch's `resource.units` on the audit row, and
there is no assertion on `units` anywhere in the diff.

That matters more than a normal missing test because the failure is silent and
order-dependent. The guarantee holds only if the audit-row `freeze()` runs
before the envelope's `frozen_or_snapshot()`. If that order is ever inverted —
by a refactor, or by a new code path assembling the envelope earlier — the
envelope silently returns a live snapshot while the audit row freezes a
different one, the two disagree, and nothing anywhere notices. A reader
comparing a response against its audit trail would see a discrepancy with no
failing test to explain it.

Please add a test that performs one dispatch which increments counters, then
asserts the envelope `usage` object and that dispatch's audit-row
`resource.units` are equal. Ideally one that also lands an increment after the
freeze point and asserts both reads still agree, since that is the specific
case the OnceLock is defending.

Everything else reads correctly to me. Happy to approve once that test is in.

One thing worth writing down somewhere durable rather than changing here: the
module states that reporting is best-effort, that every increment path no-ops
when no context is armed, and that a dispatch which cannot be trusted ships no
`usage` object. That is the right contract for observability. It is the wrong
contract for anything that later consumes these numbers as authoritative,
because an un-instrumented entry point reports zero rather than reporting
nothing, and zero and absent are indistinguishable once aggregated. Not a
change request against this PR. Worth a line in the ADR about what would have
to hold before these counters could be treated as authoritative.

ohdearquant:
Test landed at `fc833131`.

`envelope_usage_equals_audit_row_resource_units` in `crates/khive-pack-kg/tests/integration.rs`: dispatches inside an armed usage context exactly as `khive-mcp/src/server.rs` does, lands a 1,000,000 increment **after** the dispatch, then asserts the value the envelope stamps equals the same dispatch's audit-row `payload.resource.units` (read back via `list(kind=event)`) and that both exclude the late increment.

Verified red when broken: making `frozen_or_snapshot()` return a live snapshot fails it with the leak named —
`envelope={"db_round_trips":1000000,"event_rows":1,"fts_passes":1}` vs `row={"fts_passes":1}`.

Note the `event_rows` in that output: there is a **real** post-freeze increment on every dispatch — the audit row's own event append. `usage.rs` documents that the snapshot cannot count itself; nothing was checking it until now. The OnceLock is defending an increment that fires every time, not a hypothetical one.

Gates, real exit codes: `clippy -D warnings` rc=0 · `cargo test -p khive-pack-kg --test integration` rc=0 (245 passed) · `fmt --check` rc=0.

Stated limit: the test lives in the pack suite because the audit row is only reachable from the pack surface, so it calls the same expression `stamp_usage` calls rather than driving the MCP server. That one-line wrapper is not covered.

oceanwaves630:
Read the test at fc833131. It pins the invariant I asked for and my block is lifted.

It asserts the envelope's usage object equals the same dispatch's `resource.units` and that both exclude an increment landing after the freeze point, which is the whole contract. It was verified red by making `frozen_or_snapshot()` return a live snapshot, so it can fail — the property I actually cared about, since an invariant test that cannot go red is the thing being guarded against.

The coverage limit is stated in the test rather than implied: the envelope value is read through the same expression `stamp_usage` calls, not by driving the server, because the audit row is only reachable from the pack surface. That is the right disclosure and the right level to test at.

One thing the test turned up that is worth recording beyond this PR. The failure output shows a real post-freeze increment in production, not only the synthetic one: the audit row's own event append lands after the snapshot, so the frozen object correctly omits it. Until now that was a claim in a doc comment with nothing checking it. The `OnceLock` is not defending a hypothetical stray increment, it is defending one that fires on every dispatch.

ohdearquant:
Blocking finding addressed at `7ca01218`, and the sweep found the pattern is three sites rather than one.

**Confirmed at source.** `Result::inspect` runs only on `Ok`. On the traversal path it was worse than a missing branch: the counts accumulated in locals inside the closure and left only through the `Ok` tuple, so on error the totals did not go unrecorded, they ceased to exist. A traversal chunks its root set, so a failure in a later chunk silently discarded every round trip the earlier chunks had already made.

**Why the counting is outside the closure at all:** `with_reader` runs it on a blocking thread where the task-local usage context is not visible, so the closure cannot call `usage::count` itself. Shared atomics now carry the totals across that boundary on both outcomes.

**Sweep — seven `.inspect` accounting sites, three of them wrong.** I checked each against its counter's own documented definition rather than fixing them uniformly:

| site | counter | verdict |
|---|---|---|
| `graph.rs:1727`, `graph.rs:1829` | `db_round_trips` = round trips **issued** | wrong, fixed |
| `graph.rs:1727`, `graph.rs:1829` | `graph_hops` = rows storage **returned** | correct as-is, kept Ok-only |
| `graph.rs:2098` | both | wrong, fixed |
| `event.rs` ×4 | `EventRows` = rows **successfully appended** | correct by definition, untouched |

Recording `EventRows` only on success is not the same bug wearing different clothes, it is what its definition asks for. Changing all seven would have introduced a defect in four.

**Tests**, on both the neighbor and traversal paths, using a store whose schema was never created so statements fail at execution rather than before it. That separates "issued" from "succeeded" with no fault-injection hook and no timing dependence. Verified red against the previous behavior: both fail with `got {}`, an empty usage object, which is exactly the reported defect.

**Gates, real exit codes:** `clippy -p khive-db --all-targets -D warnings` rc=0 · `cargo test -p khive-db --lib` rc=0 (469) · `cargo test -p khive-runtime --lib` rc=0 (921) · `fmt --check` rc=0.

Also on this branch: `4bb5dc2a` regenerates `crates/khive-merge/Cargo.lock`. This branch adds `tokio` to `khive-storage`, and `khive-merge` is excluded from the workspace with its own committed lock that CI checks with `--locked`, so that lock no longer described the manifest. One line, no new resolution.

The non-blocking findings I have not actioned here: the embed-drain tail latency, the `khive-storage` placement question, and the duplicated spawned-embed handling are real and I would rather take them as their own change than widen this one.

ohdearquant:
Fixed at `8d10c0ba`, and wider than the report.

**The finding is correct.** `batch_neighbors` propagated a failed read before recording its round trip, so an issued query that failed reported zero work. It chunks its source list, so a failure in any chunk also discarded every earlier chunk's round trip. Fixed, with a regression test on the batched path verified failing before the fix (empty usage object) and passing after.

**The same defect was at eight more sites, and finding one told me the class existed rather than that it was the only one.** The six embed paths, both FTS legs, and the vector leg all incremented after the `?` on their own fallible call. All nine are fixed in the same commit.

The placement rule was implicit, which is why it drifted. It is now explicit at the vocabulary:

- **Issued** — `embed_calls`, `fts_passes`, `vector_passes`, `db_round_trips`. Work handed to an engine or store. Must count whether or not the call resolved `Ok`, because a request that ran real work and then failed reporting zero is indistinguishable from one that ran none.
- **Returned** — `graph_hops`, `ann_jobs_consumed`, `event_rows`. What came back. A failed call legitimately contributes nothing, so these were already correct and were deliberately left alone.

Each increment now lands after the resource is resolved, so a model lookup that never reaches an engine still counts nothing, and before the error propagates. Every enum variant states which kind it is, so the placement is checkable at the call site instead of rediscovered.

Gates at `8d10c0ba`: fmt 0, clippy (`khive-db`, `khive-runtime`, `khive-storage`, all targets, `-D warnings`) 0, `khive-db` 470 passed, `khive-runtime` 921 passed, `khive-pack-kg` 256 passed.

**On the non-blocking findings:**

The multi-model create paths do await every embedding task before inspecting errors, and that is deliberate — the comment at the join explains it. Returning on the first error would leave the remaining handles un-awaited, and a dropped `JoinHandle` detaches rather than cancels, so those tasks keep incrementing the shared counters after the response was already produced. The cost is bounded extra latency on a path that is already failing; the alternative is usage numbers that change after they are reported. There is a third option worth pursuing separately: abort the remaining handles and then await them, which gets prompt failure without detached increments. That is a follow-up, not a change to make under re-review.

The observation that `khive-storage` is documented as a trait-only contract crate is fair and I am not dismissing it. Moving the accumulator is a real refactor with its own review surface, so it belongs in its own change rather than bolted onto this one.

The duplicated spawned-embed handling and the sleep-based drain tests are accepted as follow-ups.

ohdearquant (commented):

oceanwaves630 (approved):
Approving at head 5f163b40. The issued-vs-returned counter semantics are now explicit at the vocabulary and enforced across all nine call sites — each issued counter increments after its resource is resolved and before the error propagates, with the batched-neighbors regression test proven failing before the fix and passing after. The merge-conflict resolution in graph.rs preserves the both-outcomes publication structure while taking main's PathNode weight field. The usage-freeze test pins the envelope value against the same dispatch's audit-row resource units end to end and states its own coverage limit in its docstring. Focused gates and CI are green per-commit on this SHA.

PR #1232, opened 2026-07-21
perf(db): index-seekable both-direction recursive traverse (closes #1229)

### LINKED ISSUE #1229: traverse: depth>=2 from a high-degree root takes 15-28s on a 46K-edge graph while returning ~1.2K nodes
## Measured (p50, live daemon, 6.1K entities / 46.3K edges)

| Probe | p50 | nodes returned |
|---|---|---|
| `traverse(roots=[hub], max_depth=1)` | 50.6 ms | 1231 |
| `traverse(roots=[hub], max_depth=2)` | 15.1 s | 1247 |
| `traverse(roots=[hub], max_depth=3)` | 27.9 s | 1356 |

Hub = a project entity with 1229 incident edges. `context(entity_ids=[hub], hops=2, budget=4000)` on the same neighborhood completes in 93 ms.

## Why this looks like a defect, not inherent cost

The BFS is documented as one `batch_neighbors` round-trip per frontier level (`crates/khive-runtime/src/graph_traversal.rs`). A depth-2 expansion can touch at most every edge in the graph (46K); tens of seconds for <=46K adjacency rows is ~300x out of line with the depth-1 marginal (~0.03 ms/entry). Suspects: per-frontier-node query amplification instead of a true batch, quadratic visited/path bookkeeping, or serialization amplification through the high-degree annotation neighborhood.

## Expected

depth-2/3 traversal bounded by edges actually expanded — same order as repeated depth-1 batches (hundreds of ms here), not 15-28 s.

ohdearquant:
## Summary

`GraphStore::traverse`'s recursive CTE builds its per-step join predicate
differently depending on direction. For `Direction::Out`/`Direction::In` this
is a single-column equality (`e.source_id = t.node_id` / `e.target_id =
t.node_id`), which SQLite can satisfy with an index seek. For
`Direction::Both` — the default traversal direction — the predicate was a
single OR-across-columns arm: `(e.source_id = t.node_id OR e.target_id =
t.node_id)`.

SQLite cannot satisfy an OR-across-columns predicate with one index seek.
`EXPLAIN QUERY PLAN` shows the `Both` arm only binding `namespace=?` as an
equality constraint, degrading to a full scan of every edge row in the
namespace, repeated once per node in the recursive frontier —
`O(frontier_size × total_edges)` instead of the intended
`O(frontier_size × avg_degree)`. Measured on a synthetic 8,000-node /
46,000-edge graph with a 1,500-out-edge root (comparable to the reported
issue): `Direction::Out` at depth 2 ran in 0.17s (10,688 rows); the
equivalent `Direction::Both` query ran in 63.1s (19,127 rows) — roughly
380x slower for a comparable row count, purely from the predicate shape.

This is the same bug class #280 already fixed for `Out`/`In` (forcing the
recursive frontier to be the outer loop via `CROSS JOIN` to restore
index-seek behavior), but that fix only helps when the join condition is a
single-column equality — it does nothing for `Both`'s OR predicate, which
was never covered by a perf-scale regression test.

## Fix

Split the `Both` recursive member into two separately-indexable arms —
one keyed on `e.source_id = t.node_id` (seekable via
`idx_graph_edges_ns_src_rel`), one on `e.target_id = t.node_id` (seekable
via `idx_graph_edges_ns_tgt_rel`) — `UNION ALL`'d together inside the
recursive member, instead of one OR-predicate arm. This mirrors the pattern
`batch_neighbors` in the same file already uses correctly for
`Direction::Both`. `Out`/`In` are unchanged. The recursive-member SQL
construction is extracted into a standalone `traverse_recursive_member_sql`
function so its shape is independently testable via `EXPLAIN QUERY PLAN`.
Cycle-guard and per-root dedup logic are unaffected.

## Test plan

- [x] New `EXPLAIN QUERY PLAN` regression test asserting the `Both` recursive
      member plans as two indexed `SEARCH` operations binding
      `source_id=?`/`target_id=?` respectively, not a namespace-only scan.
- [x] New correctness test: a hub node with mixed outgoing/incoming edges to
      40 spokes, each extending one hop further to its own tail node —
      `Direction::Both` at `max_depth=2` returns the full expected node set
      regardless of which arm found each spoke.
- [x] `cargo clippy -p khive-db --all-targets -- -D warnings`
- [x] `cargo test -p khive-db` (468 + 3 + 10 tests, all passing)

Closes #1229

### DIFF
--- a/crates/khive-db/src/stores/graph.rs
+++ b/crates/khive-db/src/stores/graph.rs
@@ -995,6 +995,87 @@ fn batch_upsert_edges_guarded(
     })
 }
 
+/// Builds the recursive-member SQL for one chunk of `GraphStore::traverse`'s
+/// recursive CTE. `ns_param`/`depth_param` are the 1-based positional bind
+/// indices for this chunk's `?N` placeholders; `relation_cond`/`weight_cond`
+/// are the (already-indexed) optional filter fragments built by the caller.
+///
+/// `Out`/`In` each emit a single indexed-seek arm. `Both` emits two arms —
+/// one keyed on `e.source_id = t.node_id` (seekable via
+/// `idx_graph_edges_ns_src_rel`), one on `e.target_id = t.node_id` (seekable
+/// via `idx_graph_edges_ns_tgt_rel`) — UNION ALL'd inside the recursive
+/// member, mirroring the pattern already proven correct and fast for
+/// `batch_neighbors` (`build_inner_sql` above). A single
+/// `e.source_id = t.node_id OR e.target_id = t.node_id` predicate cannot be
+/// satisfied by one index seek: SQLite falls back to a full namespace scan
+/// per frontier row, degrading `Direction::Both` traversal from
+/// `O(frontier × avg_degree)` to `O(frontier × total_edges)` (#1229).
+fn traverse_recursive_member_sql(
+    direction: Direction,
+    ns_param: usize,
+    depth_param: usize,
+    relation_cond: &str,
+    weight_cond: &str,
+) -> String {
+    match direction {
+        Direction::Out => format!(
+            "SELECT t.root_id, e.target_id, e.id, t.depth + 1, \
+                 t.path || ',' || e.target_id, t.total_weight + e.weight \
+             FROM traversal t CROSS JOIN graph_edges e \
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):

Issue #543, opened 2026-07-03
refactor(kg): derive edge-relation error hints from installed pack EDGE_RULES (kill static hint table)


ohdearquant:
From closed-issue audit 2026-07-03 — already-observed drift. Live endpoint rules are declared in `crates/khive-pack-kg/src/pack.rs:8` (KG_EDGE_RULES), but `valid_relations_for_entity_pair` maintains a SEPARATE static rule table in `crates/khive-pack-kg/src/handlers/common.rs:464-473`. Two sources of truth for the same contract — the person→project gap (#60, reopened this audit) is exactly the class of divergence this produces. Fix (mechanical): generate the error hints from the installed pack rules so the taxonomy cannot diverge from its own error messages.

PR #1234, opened 2026-07-21
feat(kg): warn when a query pattern is impossible under the edge contract

ohdearquant:
## Problem

`query(query="MATCH (n:concept)-[:precedes]->(m:concept) RETURN n.name, m.name LIMIT 10")`
returns `{"rows":[]}` with no signal, even though `precedes` carries no
`concept->concept` endpoints under the closed edge contract (base allowlist +
pack `EDGE_RULES`). A schema mismatch is indistinguishable from "no data yet".

## Change

When a GQL `MATCH` edge pattern names both an explicit relation and explicit
entity kinds on both endpoints, and that `(source_kind, relation, target_kind)`
triple is impossible under the composed edge endpoint contract, the query still
executes unchanged (rows are never affected), but a warning is appended to the
existing `QueryResult.warnings` channel, e.g.:

```
pattern (concept)-[:precedes]->(concept) can never match: 'precedes' does not
accept concept->concept endpoints; accepted source->target kinds for
'precedes': document->document, dataset->dataset, artifact->artifact,
service->service, project->project
```

The check reuses the exact `base_entity_rule_allows` /
`accepted_pack_relations_for_entities` validator functions (issue #543
precedent) rather than a parallel matcher table, and consults the same
composed pack-rule set (`KhiveRuntime::pack_edge_rules()`) that edge-link
validation uses.

Scope, deliberately conservative:
- Only fires for a single-hop, fixed-direction edge naming exactly one
  relation, where both endpoints name an entity kind under the closed 8-kind
  base taxonomy.
- Unlabeled nodes, unlabeled/multi-relation edges, undirected edges,
  variable-length hops, note-kind endpoints, and SPARQL queries are all left
  unchecked — none of those name a single static triple to test against the
  validator.
- Rows are never affected; this only ever adds a warning alongside existing
  ones (e.g. row-cap truncation warnings), never an error.

## Tests

Added to `crates/khive-runtime/tests/integration.rs`:
- `query_static_impossible_pattern_precedes_concept_concept_warns_and_returns_empty`
- `query_static_possible_pattern_extends_concept_concept_no_warning`
- `query_static_impossible_check_skips_unlabeled_source_node`
- `query_static_impossible_check_honors_pack_extended_endpoint_rules`
- `query_static_impossible_check_skips_sparql`

All existing truncation-warning tests continue to pass unchanged.

## Test plan

- [x] `cargo test -p khive-pack-kg -p khive-runtime -p khive-query` (all green, including the 5 new tests)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-pack-kg -p khive-runtime`

### DIFF
--- a/crates/khive-runtime/src/operations.rs
+++ b/crates/khive-runtime/src/operations.rs
@@ -391,6 +391,13 @@ pub fn accepted_pack_relations_for_entities(
 /// (the exact functions the live validator consults) over the closed kind set,
 /// rather than re-deriving a parallel table — issue #543 precedent, applied to
 /// GQL query-pattern hint derivation (issue #593).
+///
+/// Pack rules are skipped for `crate::pack::SPECIAL_RELATIONS` (supersedes /
+/// supports / refutes): the live validator's special-relation branch
+/// (`validate_edge_relation_endpoints`, this file) resolves those relations
+/// before `pack_rule_allows` is ever reached, so a pack `EDGE_RULES` entry for
+/// one of them is never actually enforced (see `pack.rs`'s
+/// `edge_endpoint_table` doc comment for the authoritative statement of this).
 fn accepted_entity_kind_pairs_for_relation(
     pack_rules: &[EdgeEndpointRule],
     relation: EdgeRelation,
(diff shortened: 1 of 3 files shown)


ohdearquant:
Review round 1 found one correctness issue (the static query hint treated pack-declared rules as valid for the three special relations, diverging from the live validator) plus test-coverage gaps. Both addressed at 960f3849: the hint now shares the validator's special-relation exclusion, with regression and integration coverage for inbound direction, typed-rule matching, chained patterns, skipped shapes, and the exact message contract. Round 2 reconfirmed clean at this head.

oceanwaves630 (approved):
Reviewed at 960f3849. The impossible-pattern hint now shares SPECIAL_RELATIONS with the live validator instead of duplicating the exclusion list, and the integration tests pin the exact message contract. Approve.

PR #1237, opened 2026-07-21
fix(runtime): disclose merged_into target when a by-id lookup misses a merged entity

ohdearquant:
## Summary

After `merge(into_id, from_id)`, a stored reference to the absorbed id (`from_id`) becomes indistinguishable from a reference to an id that never existed: `get(from_id)` and `resolve(refs=[from_id])` both report a bare not-found, even though the tombstone row durably records `merged_into`. This is an interim, deliberately narrow fix that discloses the redirect target in the error message — it does not implement the full transitive redirect chase (tracked separately).

- `KhiveRuntime::get_entity` (`crates/khive-runtime/src/operations.rs`): on a miss against a live row, probes the tombstone once via `get_entity_including_deleted`. If `merged_into` is set, the `NotFound` error becomes `"not found: <id> was merged into <kept-id>; query the kept id"` instead of the bare `"not found: entity <id>"`. The probe only runs after the live-row lookup misses, so the happy path pays no extra query. Single-level disclosure only — it does not chase a chain of merges, and it never returns the kept entity in place of the miss.
- `KgPack::handle_get` (`crates/khive-pack-kg/src/handlers/get.rs`): captures that hint when `get_entity` misses, and surfaces it as the final error only if no other substrate (note/edge/event/pack-resolver/proposal) claims the id — preserving the existing fallthrough behavior for ids of other kinds.
- `resolve` (`crates/khive-pack-kg/src/handlers/resolve.rs`) is **unchanged**: `ReferenceResolution::NotFound` is a unit variant with no message slot to carry a hint in, so there is nowhere to attach the disclosure without a response-shape change, which is out of scope here. A resolve on an absorbed uuid still reports a bare `not_found` status (pinned by a regression test).
- Note merge (`merge_note`) is out of scope: `Note` has no `merged_into` field in storage at all, so there is no tombstone pointer to disclose for notes.

## Test plan

- [x] `get(absorbed-full-uuid)` after a merge: error message contains `"merged into"` and the kept id (`curation::tests::get_entity_after_merge_discloses_kept_id`, `handlers::tests::get_dispatch_after_merge_discloses_kept_id`)
- [x] `get` on a plain soft-deleted (non-merged) entity: unchanged bare not-found (`curation::tests::get_entity_on_plain_soft_delete_stays_bare_not_found`, part of `handlers::tests::get_dispatch_on_plain_deleted_and_absent_ids_unchanged`)
- [x] `get` on a random absent uuid: unchanged bare not-found, no merge hint (`curation::tests::get_entity_on_absent_id_stays_bare_not_found`, part of `handlers::tests::get_dispatch_on_plain_deleted_and_absent_ids_unchanged`)
- [x] `resolve` on an absorbed uuid: still a bare `not_found` status, documented boundary (`handlers::tests::resolve_dispatch_on_merged_uuid_stays_bare_not_found`)
- [x] `cargo test -p khive-runtime -p khive-pack-kg -p khive-db` (all green)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p khive-runtime`

### DIFF
--- a/crates/khive-pack-kg/src/handlers/get.rs
+++ b/crates/khive-pack-kg/src/handlers/get.rs
@@ -46,6 +46,14 @@ impl KgPack {
 
         let include_deleted = p.include_deleted.unwrap_or(false);
 
+        // Interim merged_into disclosure (data-integrity, precedes the full
+        // ADR-113 redirect chase): `get_entity` names the kept id in its
+        // NotFound message when `id` was consumed by `merge`. Captured here
+        // so it can stand in for the generic not-found built below once
+        // every other substrate (note/edge/event/pack-resolver) also misses
+        // — the marker string is one we control on the write side above.
+        let mut merge_redirect_hint: Option<String> = None;
+
         match self.runtime.get_entity(graph_token, id).await {
             Ok(entity) => {
                 return flatten_get_result(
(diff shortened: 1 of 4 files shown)


ohdearquant:
Adversarial review round 1: **APPROVE** — 0 findings. Verified: single-level read-only merged_into disclosure, no wire-shape change, resolve() unchanged, no namespace gate on the by-ID path.

oceanwaves630 (approved):
Read the full diff: interim disclosure is message-only, fires only when every substrate misses, and the plain-deleted/absent/resolve boundaries are pinned by tests. Matches the agreed interim scope; approving.

PR #1238, opened 2026-07-21
docs(pack-kg): state entity/note corpus scope in the search verb description

ohdearquant:
Doc-only change: the `search` verb description now states it searches knowledge-graph entities and notes, and that corpora owned by other packs (which carry their own search verbs) are disjoint and not searched here. Callers reaching for a similarly named search verb otherwise silently conclude records do not exist when they searched the wrong corpus.

Gates: fmt clean, `cargo test -p khive-pack-kg` 150+6+3+244+11 passed.


ohdearquant:
Adversarial review round 1: **APPROVE** — 0 findings. Doc-only description change verified accurate against the search dispatch path.

oceanwaves630 (approved):

PR #1239, opened 2026-07-21
fix(kkernel): reserve the model document-instruction bytes in the reindex embed budget

ohdearquant:
## Problem

`kkernel reindex` truncates record text to `MAX_EMBED_BYTES` (32768), but `embed_document_batch_with_model` -> `embed_passage` prepends the model's document instruction **after** that truncation — 9 bytes of `"passage: "` for multilingual-e5. The embedding service's input guard rejects the combined string (`text too long: 32777 chars exceeds maximum 32768`), and one over-cap text fails its entire batch (up to `--batch-size` records lose vectors for that model). Hit live during a multilingual-e5-base reindex; prefix-less models (MiniLM, BGE, Qwen3 passage side) never trip it, which is why the cap was safe until an instruction-prefixed model was configured.

## Fix

`embed_budget(model_name)` resolves the active model's `document_instruction()` length (via the same alias normalization the runtime uses) and subtracts it from the truncation cap. Prefix-less and unknown models keep the full cap. Semantically lossless — model token windows (512 tokens) truncate far earlier than the byte cap.

`lattice-embed` also has this footgun upstream (guard runs post-prefix on caller-valid input); reported separately. This change makes reindex correct regardless of when that lands.

## Tests

- `embed_budget_reserves_document_prefix_bytes` — e5 variants get cap-9, MiniLM/unknown keep the full cap.
- `truncate_text_respects_budget_and_char_boundaries` — budget respected; CJK char-boundary safety.

Gates: fmt clean, `clippy -p kkernel --all-targets -D warnings` clean, `cargo test -p kkernel --lib reindex::tests` 32 passed.

### DIFF
--- a/crates/kkernel/Cargo.toml
+++ b/crates/kkernel/Cargo.toml
@@ -39,10 +39,11 @@ rmcp = { workspace = true }
 tempfile = { workspace = true }
 sha2 = { workspace = true }
 
+lattice-embed = { workspace = true }
+
 [dev-dependencies]
 tokio = { workspace = true, features = ["test-util"] }
 serial_test = { workspace = true }
-lattice-embed = { workspace = true }
 rmcp = { workspace = true, features = ["client", "transport-child-process"] }
 khive-pack-template = { version = "0.6.0", path = "../khive-pack-template" }
 
--- a/crates/kkernel/src/reindex.rs
+++ b/crates/kkernel/src/reindex.rs
@@ -301,7 +301,11 @@ async fn embed_and_store_batch(
             continue;
         }
 
-        let texts: Vec<String> = subset.iter().map(|(_, t)| truncate_text(t)).collect();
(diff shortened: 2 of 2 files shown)


ohdearquant:
Adversarial review round 1: **APPROVE** — 0 findings. Verified: budget applied before the sole production truncate call and covers the post-truncation instruction prefix; alias normalization parity (bare 'paraphrase' divergence noted and harmless — no document prefix); UTF-8 boundary safety; no missed call sites; dependency move justified. 32 reindex tests + fmt + clippy green.

oceanwaves630 (approved):
Read the full diff: per-model instruction-byte reservation with safe fallback for unrecognized models, truncation boundary handling intact, dependency promotion correct, tests cover the prefix and CJK boundary cases. Approving.

PR #1240, opened 2026-07-21
fix(pack-kg): let short-prefix get reach soft-deleted rows and the merged_into disclosure

ohdearquant:
Follow-up to #1237. A post-merge review found the merged_into disclosure unreachable for the documented short-prefix identifier form: `handle_get` resolves prefixes via a live-rows-only scan, and absorbed entities are soft-deleted, so `get(id=<absorbed-short-id>)` failed at prefix resolution (`no record matches prefix`) before the disclosure logic could run. Verified at source (`resolve_uuid_unfiltered` -> `resolve_prefix_unfiltered`, include_deleted=false).

## Fix

Add an including-deleted prefix-resolution fallback in `handle_get` only, after both live-scan attempts miss. Live rows resolve first, so ambiguity semantics for live prefixes are unchanged.

## Behavior changes (both consistency-restoring)

- Short prefix of an absorbed entity now reaches the merged_into disclosure, matching the full-UUID form.
- Short prefix of a plain soft-deleted row now reports `not found: <uuid>` instead of `no record matches prefix`, matching the full-UUID form.
- Short prefix + `include_deleted=true` can now actually return the soft-deleted row — previously unreachable, contradicting the documented parameter.

## Out of scope (fold into the ADR-113 chase design)

The double tombstone read on miss paths and the message-marker coupling flagged by the same review are deferred to the full chase spec, which replaces the string protocol with a typed redirect.

## Tests

Extended `get_dispatch_after_merge_discloses_kept_id` with the short-prefix form (exercises the fallback end to end).

Gates: fmt clean, clippy `-D warnings` clean, `cargo test -p khive-pack-kg` 153+6+3+244+11 passed.

### DIFF
--- a/crates/khive-pack-kg/src/handler_defs.rs
+++ b/crates/khive-pack-kg/src/handler_defs.rs
@@ -141,8 +141,8 @@ pub(crate) static KG_HANDLERS: [HandlerDef; 18] = [
                 required: false,
                 description:
                     "If true, return soft-deleted entities (with deleted_at populated). Default false. \
-                     Requires a full UUID — short prefix resolution filters deleted records; \
-                     the delete response always returns the full UUID for this purpose.",
+                     Accepts a full UUID or a unique short hex prefix — prefix resolution falls back \
+                     to soft-deleted entities when no live record matches.",
             },
         ],
     },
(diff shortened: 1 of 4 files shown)


ohdearquant:
Adversarial review round 1 found one medium issue (a stale include_deleted doc contract on two surfaces and an unpinned success path); fixed in df76335a and reconfirmed clean in round 2 with zero findings. Round 1 also verified: live-precedence ordering unchanged, deleted-scan ambiguity collapses to the pre-change generic not-found, no other verbs share the new fallback, and the note/edge/event fallthrough stays intact.

oceanwaves630 (approved):

oceanwaves630 (approved):
Reviewed at df76335a. Live-first resolution order is preserved so ambiguity semantics are unchanged; the including-deleted fallback correctly lets a short hex prefix reach soft-deleted rows, which makes the merged_into disclosure fire for the documented short-prefix form. Both new tests cover exactly the previously-missed paths. Approve.

PR #1241, opened 2026-07-21
fix(runtime): treat untyped GQL hint endpoints as unconstrained over entity_type

ohdearquant:
## Summary

The static GQL impossibility hint (`static_impossible_edge_pattern_warnings` in `crates/khive-runtime/src/operations.rs`) checks whether a `MATCH` pattern's `(source_kind, relation, target_kind)` triple can ever be satisfied under the composed edge endpoint contract (base allowlist + pack `EDGE_RULES`). It reused `endpoint_matches`, whose `EntityOfType` arm requires `entity_type == Some(t)`.

A pattern endpoint that names only a base kind, with no `entity_type` filter (e.g. `(a:concept)-[:depends_on]->(b:concept)`), has `entity_type = None`. Under `endpoint_matches`'s exact-match semantics, `None` never satisfies an `EntityOfType` rule such as `concept(theorem)-depends_on->concept(definition)`. This produced two defects:

1. The pattern was flagged with a false "can never match" warning even when a pack installs an `EntityOfType` rule that makes the triple possible, and stored typed entities exist that would in fact match — because the compiled SQL for an untyped pattern emits no `entity_type` predicate.
2. The warning's accepted-pairs listing omitted `EntityOfType`-derived kind pairs, since it also passed `None`/`None` through the same exact-match check.

## Fix

In the hint analysis only, an absent `entity_type` on a pattern endpoint now means *unconstrained* rather than "no subtype": an installed `EntityOfType` rule whose kind matches makes the triple possible, regardless of the rule's specific `entity_type`. A pattern endpoint that does name an `entity_type` still requires an exact match.

This is implemented as a new hint-only predicate, `pattern_endpoint_matches`, and a matching `accepted_pack_relations_for_pattern_entities` helper, used by both the possible-triple check and the accepted-pairs derivation inside `static_impossible_edge_pattern_warnings`. The live link validator (`endpoint_matches`, used elsewhere over already-resolved entities, where `None` correctly means "this entity has no subtype") is unchanged — a resolved entity with no subtype must still fail an exact-match `EntityOfType` rule.

## Test plan

Four new regression tests added to `crates/khive-runtime/tests/integration.rs`, alongside the existing GQL static-hint test suite:

- [x] `query_static_possible_pattern_untyped_endpoints_with_entity_of_type_rule_no_false_warning` — untyped kind-only pattern over a relation with only an `EntityOfType` pack rule installed → no warning (previously failed)
- [x] `query_static_impossible_pattern_mismatching_entity_type_still_warns` — same rule, pattern typed with a mismatching `entity_type` → warning still fires (exact-match preserved)
- [x] `query_static_impossible_warning_accepted_pairs_include_entity_of_type_only_relation` — the accepted-pairs text in a warning for a relation whose only accepted pair comes from an `EntityOfType` rule includes that pair
- [x] Existing `query_static_possible_pattern_entity_of_type_pack_rule_no_false_warning` (typed pattern matching the rule) kept as a guard, still passing

Gates, run from `crates/`:
- `cargo fmt --all -- --check` — exit 0
- `cargo clippy -p khive-runtime --all-targets -- -D warnings` — exit 0
- `cargo test -p khive-runtime` — exit 0 (919 lib unit tests, 2 cold-boot tests, 86 integration tests, all passing)

### DIFF
--- a/crates/khive-runtime/src/operations.rs
+++ b/crates/khive-runtime/src/operations.rs
@@ -350,6 +350,34 @@ pub fn endpoint_matches(
     }
 }
 
+/// `true` if `spec` matches the given substrate + kind + entity_type triple,
+/// treating an *absent* `entity_type` on the query side as unconstrained
+/// rather than an exact match against "no subtype".
+///
+/// Used only by the static GQL impossibility hint (`static_impossible_edge_pattern_warnings`,
+/// `accepted_entity_kind_pairs_for_relation`), which reasons over a *pattern*
+/// endpoint, not a resolved entity. A pattern endpoint that names a kind but
+/// no `entity_type` (`(a:concept)-[:depends_on]->(b:concept)`) has not ruled
+/// out any subtype, so an `EntityOfType` rule for that kind still makes the
+/// triple possible — unlike `endpoint_matches`, which the live link
+/// validator applies to *resolved* entities, where a `None` `entity_type`
+/// means the entity genuinely has no subtype and must be an exact miss
+/// against a typed rule. Do not use this for validation.
+fn pattern_endpoint_matches(
+    spec: &EndpointKind,
+    substrate: &str,
+    kind: &str,
+    entity_type: Option<&str>,
+) -> bool {
+    match spec {
+        EndpointKind::EntityOfType {
+            kind: k,
+            entity_type: t,
+        } => substrate == "entity" && *k == kind && entity_type.is_none_or(|et| et == *t),
+        _ => endpoint_matches(spec, substrate, kind, entity_type),
+    }
+}
+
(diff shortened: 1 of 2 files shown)


ohdearquant:

ohdearquant:
Static review round 1 complete at 50214dd9: no findings. Wildcard semantics confirmed hint-only (validator paths unaffected), accepted-pairs derivation correct, SPECIAL_RELATIONS exclusion preserved in both branches, regression tests cover typed/untyped/mismatch/accepted-pairs cases.

oceanwaves630 (approved):
Automated review. Posted by this repository's automated pull-request review pipeline; this is not a human read.

Verdict on current head: APPROVE, zero blocking findings.

PR #1261, opened 2026-07-21
fix(pack-kg): stop budget assembly from starving later context anchors

ohdearquant:
## Summary

The `context` verb's anchor selection (query → hybrid entity search) is unaffected by this bug — it already reuses the same `hybrid_search` fusion the `search` verb uses, so ranking parity between the two was never the issue.

The bug is in the response-assembly stage. `assemble_within_budget` walked each anchor's entity record *and* its full neighbor list before moving on to the next anchor. A single anchor with a large neighbor fan-out (driven by `fanout`/`hops`, unrelated to how relevant that anchor is to the query) could exhaust the entire byte budget on its own neighbors, dropping every anchor ranked after it in the response — including anchors that a plain entity search on the same query would have surfaced as directly relevant. Increasing `budget`, `hops`, `limit`, or `fanout` doesn't help, since a larger fan-out cap just makes the starving anchor's neighbor list bigger too.

## Fix

Assemble in two passes:

1. **Entity pass** — walk anchors in rank order and reserve budget for each one's own entity record only, ignoring neighbors entirely.
2. **Neighbor pass** — for the anchors that survived pass 1, walk them again in the same rank order and fill neighbors with whatever budget remains.

An anchor's neighbor sprawl can now only crowd out its own or later anchors' *neighbors* — never a sibling anchor's entity record. Anchors that the query legitimately ranked as relevant are no longer collateral damage of an unrelated anchor's graph degree.

Also names the stage responsible for a drop (additive `dropped.stage` field) instead of reporting bare integers, since in this handler every drop originates in the budget-assembly walk — `fanout`/`hops` bound the neighbor candidate pool upstream, before assembly, and never produce a drop themselves. The existing `dropped.anchors` / `dropped.neighbors` counts are unchanged.

## Tests

- Unit test on `assemble_within_budget` reproducing the starvation pattern directly with a bloated first anchor and a small second anchor.
- Integration regression test: a plain `search(kind="entity", ...)` surfaces an on-topic entity for a query; `context(query=...)` with a budget that comfortably fits both anchors' entity records but not one anchor's 15-neighbor fan-out still returns the on-topic anchor's entity record, with `truncated: true` and `dropped.neighbors > 0` accounting for the actual drop.

## Test plan

- [x] `cargo clippy -p khive-pack-kg -p khive-runtime --all-targets -- -D warnings`
- [x] `cargo test -p khive-pack-kg -p khive-runtime`

### DIFF
--- a/crates/khive-pack-kg/src/handlers/context.rs
+++ b/crates/khive-pack-kg/src/handlers/context.rs
@@ -499,36 +499,61 @@ impl KgPack {
         Ok(json!({
             "anchors": out_anchors,
             "truncated": truncated,
-            "dropped": { "anchors": dropped_anchors, "neighbors": dropped_neighbors },
+            // `stage` is additive: every drop this handler produces originates in the
+            // Stage-4 budget walk — `fanout`/`hops` bound the neighbor candidate pool
+            // upstream, before assembly, so they never show up as a "drop" here today.
+            "dropped": {
+                "anchors": dropped_anchors,
+                "neighbors": dropped_neighbors,
+                "stage": "budget",
+            },
         }))
     }
 }
 
-/// Deterministic-order budget walk over anchors + neighbors. See `docs/api/context-verb.md`.
+/// Two-pass deterministic budget walk over anchors + neighbors. See
+/// `docs/api/context-verb.md`.
+///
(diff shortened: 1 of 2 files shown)


ohdearquant:

PR #1262, opened 2026-07-21
fix(khive-request): DSL parse errors explain the fix, not just the position

ohdearquant:
## Summary

The DSL parse layer reported *where* a syntax problem was but never *what*
would fix it — unlike the semantic/param layer, where unknown-field errors
enumerate the accepted set and the short-query error both corrects the
input and routes the caller to the right verb. This brings the parse layer
up to that standard for three specific messages, plus a related fix to a
message that stated two disagreeing locations for one problem.

### Before / after

**Unquoted string value** (`get(id=abc)`):
- Before: `at position 7: invalid value: expected value at line 1 column 1`
- After: `at position 7: invalid value: "abc" is a bareword, not a valid value; string values must be double-quoted — did you mean id="abc"?`

**Trailing comma in a batch** (`[gtd.next(),]`):
- Before: `at position 11: invalid identifier (expected [A-Za-z_][A-Za-z0-9_]*)`
- After: `at position 11: trailing comma before ']' — a batch cannot end with an empty element; remove the comma or add another op after it`

**Mixing `,` and `|` separators** (`a(), b() | c()`):
- Before: `expected '|' or end of input, found ','`
- After: `cannot mix ',' (parallel) and '|' (chain) separators at the top level` — the message already existed and was reachable from two of three code paths that hit this mistake; it is now reachable from all three.

**A scalar value's own decode failure** (e.g. `get(id=1.2.3)`) no longer
leaks `serde_json`'s `at line 1 column N`, which is always relative to that
one value's isolated slice and so disagreed with the DSL-absolute `at
position N` this crate reports for the same failure. The descriptive text
(`trailing characters`, `expected value`, ...) is kept; only the
contradicting position clause is dropped. The JSON request form's own
`serde_json` position is untouched — there it parses the whole input, so
its line/column is genuinely correct.

## Behavior

No grammar change. Every input that parsed successfully before this PR
still parses to the identical AST — this is an error-message change only.
The `DslError` enum gains one variant (`TrailingComma`) to name the new
diagnostic distinctly rather than reusing an unrelated one.

## Test plan

- [x] `cargo clippy -p khive-request --all-targets -- -D warnings`
- [x] `cargo test -p khive-request` (139 passed, including 7 new tests asserting on message content)
- [x] `cargo test --workspace` (pre-existing, unrelated failures in `khive-mcp`'s `pending_events` tests — reproduced identically on `main` before this change; a missing commercial `gtd` pack in this build, not touched by this PR)

### DIFF
--- a/crates/khive-request/docs/api/limits-and-errors.md
+++ b/crates/khive-request/docs/api/limits-and-errors.md
@@ -30,7 +30,7 @@ This public helper checks a runtime `serde_json::Value`, including handler resul
 | `UnexpectedEof`          | input ends before a required token                                   |
 | `InvalidIdentifier`      | identifier violates the ASCII identifier grammar                     |
 | `DuplicateArg`           | one operation repeats an argument name                               |
-| `InvalidValue`           | function-form value cannot be decoded or reference syntax is invalid |
+| `InvalidValue`           | function-form value cannot be decoded or reference syntax is invalid; an unquoted bareword names the quoting fix (and the corrected call, when the argument or key is known) |
 | `InvalidJson`            | JSON form is malformed or has the wrong shape                        |
 | `UnclosedString`         | quoted string has no terminator                                      |
 | `UnclosedBracket`        | array, object, or parenthesis has no matching close                  |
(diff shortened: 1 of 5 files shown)


ohdearquant:

ohdearquant (commented):

oceanwaves630 (approved):
Approve — review comments at this head are clean and the diagnostics change is well-scoped.

PR #1265, opened 2026-07-22
kg: add whoami verb for caller identity introspection

ohdearquant:
## Summary

Adds a `whoami` verb to the `kg` pack so a caller can learn its own resolved
identity: actor id, actor kind, whether it is the unattributed/anonymous
fallback, the write namespace, and the read-visible namespace set.

Without this, a caller has no way to determine the assignee it should
supply on writes, cannot self-scope a queue-listing query to its own rows,
and ends up learning its actor id by accident from an unrelated echoed
response field.

`whoami` is a pure projection of state the runtime already resolves for
every request — no new state is introduced, and no secrets, tokens, or
connection strings are exposed, only labels. Registered following the same
convention as every other verb: a `HandlerDef` entry with a `description`
(the `help=` text), routed through pack dispatch, and visible through the
`verbs()` introspection call.

## Test plan

- [x] `cargo test -p khive-pack-kg -p khive-runtime -p khive-mcp` — full pass
      on the two touched crates; pre-existing unrelated failures in
      `khive-mcp::pending_events` (missing `gtd` pack in this distribution)
      reproduce identically on unmodified `main`.
- [x] New tests: configured-actor identity, unattributed/single-actor-deployment
      identity, unknown-param rejection, `verbs()` registration, and a guard
      against any secret-shaped field ever appearing in the response.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,7 +8,7 @@ khive gives your agent:
 2. **Notes** — observations, insights, questions, decisions, references that persist across sessions
 3. **Pattern matching queries** — GQL/SPARQL traverse over the graph
 
-The `kg` pack loads by default, giving the `request` tool an 18-verb catalog. Task
+The `kg` pack loads by default, giving the `request` tool a 19-verb catalog. Task
 management (GTD lifecycle), memory (salience- and decay-weighted recall), communication
 (namespaced message passing between agents), scheduling (time-triggered reminders and
 future verb dispatch), session continuity (persist and resume agent-session records),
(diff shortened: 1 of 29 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Approve. The earlier attribution concern is resolved at this head: a configured local-id actor now reports unattributed, matching the runtime's own authorization policy, with the regression test kg_whoami_flags_configured_local_actor_as_unattributed passing.

PR #1266, opened 2026-07-22
fix(runtime): dedupe phantom traversal entries across visible namespaces

ohdearquant:
## Summary

`traverse()` queries every namespace in a caller's visible-namespace set independently and concatenates the results. A root that lives in only one of those namespaces was still queried against every *other* visible namespace as well; since `include_roots` pre-seeds the root node before the per-namespace query even runs, a namespace that doesn't own the root at all still produced a root-only, no-expansion `GraphPath`. A single requested root could therefore come back as two response entries: one real traversal and one degenerate stand-in — and a caller checking the response length would conclude it had traversed two roots when it passed one.

This PR merges the per-namespace results back down to one `GraphPath` per distinct requested root, unioning the nodes discovered under each namespace. **This changes the response for the affected case**: callers that were iterating over the extra phantom entry, or counting entries per root, now see the corrected shape. The `GraphPath` schema itself (field names, nesting) is unchanged — only which entries appear.

A separate, previously undocumented detail is also documented (not changed) here: traversal can reach note nodes through an `annotates` edge, and does not drop them from the result, but only resolves `name`/`kind` for entity nodes — reached notes come back with those two fields absent. The verb description now says so explicitly.

## Test plan

- [x] `cargo test -p khive-pack-kg -p khive-runtime -p khive-db`
- [x] New test: single root visible across two namespaces yields exactly one `GraphPath`, retaining the real traversal's nodes
- [x] New test: multiple roots, including the same root supplied twice, yields exactly one `GraphPath` per distinct root
- [x] New test: a note node reached via traversal is present in the result but left unenriched, pinning the existing entity-only enrichment scope

### DIFF
--- a/crates/khive-db/src/stores/graph.rs
+++ b/crates/khive-db/src/stores/graph.rs
@@ -1970,6 +1970,7 @@ impl GraphStore for SqlGraphStore {
                             name: None,
                             kind: None,
                             properties: None,
+                            weight: 0.0,
                         },
                         0.0,
                     ));
@@ -2092,6 +2093,7 @@ impl GraphStore for SqlGraphStore {
                             name: None,
                             kind: None,
                             properties: None,
+                            weight: total_weight,
                         },
                         total_weight,
                     ));
--- a/crates/khive-pack-kg/src/handler_defs.rs
+++ b/crates/khive-pack-kg/src/handler_defs.rs
@@ -623,7 +623,9 @@ pub(crate) static KG_HANDLERS: [HandlerDef; 18] = [
     // Assertive: retrieves multi-hop traversal results
     HandlerDef {
         name: "traverse",
-        description: "Multi-hop BFS traversal",
+        description: "Multi-hop BFS traversal. Can reach note nodes (e.g. via an \
+                      `annotates` edge) but only enriches entity nodes with name/kind; \
+                      note nodes come back with those fields absent.",
         visibility: Visibility::Verb,
         category: VerbCategory::Assertive,
         params: &[
--- a/crates/khive-runtime/src/operations.rs
+++ b/crates/khive-runtime/src/operations.rs
@@ -850,6 +850,51 @@ fn note_props_match(note_props: Option<&serde_json::Value>, filter: &serde_json:
         .all(|(k, v)| actual.get(k).is_some_and(|av| av == v))
 }
 
+/// Collapse per-namespace `GraphPath`s from [`KhiveRuntime::traverse`] down to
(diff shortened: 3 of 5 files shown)


ohdearquant:
All three findings addressed at `f5c45c18`.

**Limit re-enforcement.** After merging and sorting into BFS order, the merge retains depth-0 root nodes and caps depth>0 nodes at `limit`, so the kept nodes are the shallowest rather than an arbitrary subset of the N×limit union.

Worth stating explicitly because it is the part that could silently diverge: this matches the storage layer's own rule. `SqlGraphStore::traverse` truncates with `root_count + lim` where `root_count = usize::from(include_roots)`, i.e. only non-root nodes count against the cap. A merge that counted roots against the limit would have made a multi-namespace call return fewer nodes than a single-namespace one for the same `limit`, which is a subtler bug than the one being fixed.

**Metadata and ordering.** The merge now keys on `(root_id, node_id)` through a `HashMap` index into the per-root node list, replacing the stored `PathNode` in place when the same node arrives at a strictly smaller depth, so `depth` and `via_edge` move together and stay consistent with each other. Each root's list is then sorted by ascending depth.

Tie rule, disclosed rather than left implicit: when the same node arrives at the *same* minimum depth from two namespaces, the earlier-processed namespace's `via_edge` wins, since replacement happens only on strictly-smaller depth. Both edges produced an equally short path and nothing in the traversal contract makes one namespace authoritative, so this is deterministic namespace order rather than a claim that one answer is more correct. The new tests assert the strictly-shorter case and deliberately do not pin the tie.

**Quadratic seen-set.** Removed as a consequence rather than as a separate patch. The persistent `HashMap` index replaces the per-namespace `HashSet` rebuild, so there is one incremental structure across the whole merge.

**Duplicated explanation** cut from three places to one, on the function itself.

**Tests** target the merge function directly, so they pin merge semantics without depending on SQLite's row ordering for the recursive CTE. Verified red before the fix and green after, with the failures observed rather than assumed: `left: 4, right: 2` on the depth assertion and `left: 6, right: 2` on the limit assertion. The existing multi-namespace traversal tests still pass unmodified.

**Gates, real exit codes:** `clippy -p khive-runtime --all-targets -D warnings` rc=0 · `cargo test -p khive-runtime` rc=0 · `cargo test -p khive-mcp` rc=0 · `fmt --check` rc=0.

Ready for re-review at `f5c45c18`.

oceanwaves630 (approved):
Approve — review comments at this head are clean.

PR #1267, opened 2026-07-22
fix(request): explain why a $prev chain reference failed to resolve

ohdearquant:
## Summary

`$prev` in a chained request always resolves against the immediately preceding
op's result only — that design is unchanged by this PR. What changed is the
error you get when a `$prev` reference can't be resolved: previously every
failure collapsed into one generic `substitution_error`, and the case where
the very first op in a chain used `$prev` (no preceding op exists yet) was
mislabeled as `$prev reference in non-chain context`, which is misleading
inside an actual chain.

This distinguishes the ways a `$prev` path can fail:

- **no preceding op** — `$prev` used in the chain's first operation
- **path not found** — the named field or array index isn't present; the
  message lists the sibling fields (or array length) that are
- **wrong type** — a path segment expects an object (for a field) or an array
  (for an index) to continue into, but the value at that point is a
  different JSON type
- **unsupported path form** — bracket syntax that isn't a valid
  non-negative-integer index (guarded at parse time on the public DSL surface;
  the substitution layer carries its own defense-in-depth check for this too)

Each message names the exact field/segment involved, what was found instead,
and the fix: split a non-adjacent dependency into separate calls and carry the
value across yourself, since `$prev` cannot reach more than one op back. A
chain aborted by an earlier op's failure now also says plainly which op
failed, so downstream ops that never ran don't read like the problem is local
to them.

## Test plan

- [x] `cargo clippy -p khive-request --all-targets -- -D warnings`
- [x] `cargo test -p khive-request` (137 passed, incl. 5 new tests against the
      substitution-failure classifier)
- [x] `cargo test -p khive-mcp` (194 passed / 21 pre-existing failures in
      `pending_events`, unrelated to this change and present on `main`;
      integration suite: 86 passed incl. 5 new chain-substitution tests)

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -22,7 +22,7 @@ use serde_json::{json, Value};
 use sha2::{Digest, Sha256};
 
 use khive_db::ConnectionPool;
-use khive_request::{parse_request, ArgValue, DslError, ExecutionMode, ParsedOp};
+use khive_request::{parse_request, ArgValue, DslError, ExecutionMode, ParsedOp, PrevFailure};
 use khive_runtime::{
     present, render_format, resolve_explicit_namespace, KhiveRuntime, OutputFormat, PackLoadError,
     PackRegistry, PresentationMode, RuntimeConfig, RuntimeError, VerbPresentationPolicy,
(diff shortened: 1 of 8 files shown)


ohdearquant:

ohdearquant (commented):

PR #1268, opened 2026-07-22
fix(pack-kg): surface verb signatures in verbs(), enumerate valid kinds on a missing search kind

ohdearquant:
## Summary

Two related discoverability gaps in the KG pack's introspection and search surfaces:

1. `verbs()` listed only `verb` / `pack` / `description` / `category` for each row — nothing about parameters. The only way to learn what a verb required was to call it and read the resulting error. Each row now also carries a compact `signature` field (e.g. `search(kind, query, limit?, entity_kind?, ...)`), built from the same per-verb param schema `help=true` already renders, with required params listed bare and optional ones marked with a trailing `?`.

2. `search`'s `kind` parameter is required, and that requirement is now visible via (1) — but `search(query="...")` with `kind` omitted entirely used to fail with a raw serde `missing field` message. An unrecognized `kind` value (e.g. `kind="bogus"`) already produced a helpful error enumerating every valid kind sourced from the registry; the missing-parameter case now produces the same enumeration instead of a generic parser error, via a small shared helper so both paths always stay in sync with whatever kinds the loaded packs have registered.

Both are additive: no existing response field changes shape, and the kind enumeration is still built entirely from the registry's own kind lists (entity/note substrate + every pack-registered granular kind) — never a hardcoded list.

## Test plan

- [x] `cargo fmt -p khive-pack-kg -- --check`
- [x] `cargo test -p khive-pack-kg -p khive-runtime -p khive-mcp` (khive-pack-kg: 100% pass, including two new tests; the khive-mcp `pending_events` failures are pre-existing on unmodified `main` — reproduced separately — caused by a commercially-licensed pack absent from this distribution, unrelated to this change)
- [x] New test: a `verbs()` row for a required-param verb carries enough of its signature to build a call that actually parses and executes
- [x] New test: `search` with `kind` omitted lists the accepted kinds, including one contributed by a pack registered only for the test, proving the list is registry-sourced rather than hardcoded

### DIFF
--- a/crates/khive-pack-kg/src/handler_defs.rs
+++ b/crates/khive-pack-kg/src/handler_defs.rs
@@ -925,6 +925,29 @@ pub(crate) static KG_HANDLERS: [HandlerDef; 18] = [
     },
 ];
 
+/// Render a `HandlerDef`'s params as a one-line call shape a caller can copy
+/// and fill in, e.g. `search(kind, query, limit?, entity_kind?, ...)`.
+///
+/// Required params are listed bare; optional params carry a trailing `?`.
+/// This is deliberately compact (names only, no types/descriptions) — the
+/// full schema is still available per-verb via `help=true`; `verbs()` is a
+/// catalog, not a `help=true` dump for every row.
+fn compact_signature(handler: &HandlerDef) -> String {
+    let params = handler
+        .params
+        .iter()
+        .map(|p| {
+            if p.required {
+                p.name.to_string()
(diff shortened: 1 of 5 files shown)


ohdearquant:

ohdearquant (commented):

oceanwaves630 (approved):
Approve — the three noted minor findings are reasonable to defer; signature surfacing and kind enumeration are straightforward improvements.

Issue #1269, opened 2026-07-22
khive-mcp tests inherit KHIVE_PACKS from the ambient environment and fail on packs this distribution does not build


ohdearquant:
## Symptom

`cargo test -p khive-mcp` fails with 21 errors in `pending_events::tests` when the developer's
environment sets `KHIVE_PACKS` to a pack list wider than this distribution builds:

```
drain: unknown pack name "gtd" — built-in packs: kg, template
```

The same suite passes with the variable unset:

```
env -u KHIVE_PACKS cargo test -p khive-mcp   # rc=0, 215 passed
```

## Cause

The tests construct their server through the normal configuration path, which resolves the pack
list from `KHIVE_PACKS`. A test's pack set should be a property of the test, not of whoever runs
it. Any developer with a working khive configuration hits this on a clean clone, and the error
message points at a pack name that appears nowhere in the failing file, so the trail is
misleading.

CI does not catch it because CI has no such variable set.

## Fix

Pin the pack list explicitly in the test harness rather than inheriting it. The tests exercise
the scheduler drain, which does not depend on which packs are loaded beyond needing the verb
its fixture action names to resolve.

## Note for whoever picks this up

The failure is environment-dependent, so reproduce it by exporting a wide `KHIVE_PACKS` first —
a clean shell shows green and looks like nothing is wrong.

PR #1270, opened 2026-07-22
feat(runtime): refuse daemon connections from a foreign uid

ohdearquant:
The daemon serves a single principal over an owner-only socket. That was an ambient property of the deployment rather than something the code asserted: nothing read peer identity, so nothing could notice when it stopped holding.

Every identity field on the request frame (`namespace`, `actor_id`, `visible_namespaces`, `config_id`) is supplied by the connecting process. A check reading those is not a weak check, it is not a check, because anyone who wants to pass it asserts the passing values. The peer uid is the only identity on this socket the caller cannot choose.

## What changed

- Capture the peer uid at `accept` — `getpeereid(2)` on macOS/BSD, `SO_PEERCRED` on Linux — and refuse any connection whose uid differs from the daemon's own euid, before a frame is read.
- **Fails closed.** An error reading peer credentials means "cannot prove same-uid", which is the same answer as "is not same-uid".
- The two chmods guarding the socket (`0600`) and its directory (`0700`) now fail closed as well. They warned and continued, so a daemon that could not prove its owner-only mode kept serving with the warning as the only trace.

## Principal is not attribution

Many `actor_id`s over one socket is the accepted design, and every seat on a normal host is a distinct attribution at the same uid. A check that refused a second distinct actor would break that on day one. Only a genuinely foreign uid is refused, and `only_a_foreign_uid_is_refused` pins that as a regression guard.

## No configuration escape hatch, deliberately

A flag permitting other uids would not weaken the assertion, it would delete it in the way hardest to notice later: the check still exists, its tests still pass, and the deployment that matters has it off. Multiple uids should require a code change and a design record, because that is the decision that should be impossible to make by accident.

## Scope

This makes the currently accepted single-principal envelope self-enforcing. It is **not** a connection-identity mechanism, it does not bind a principal to allowed namespaces, and it does not enable hosted or multi-tenant serving, which remains gated on its own design record.

## Tests

- `peer_uid_reports_the_connecting_process_uid` — the syscall path actually works on this platform and reports the real uid over a live `UnixStream`, from both ends. This matters because the accept path fails closed: if `peer_uid` errored unconditionally (wrong syscall, wrong socket option, an unimplemented platform arm) every connection would be refused and the daemon would be silently unreachable. A test covering only the decision function would not catch that.
- `only_a_foreign_uid_is_refused` — same uid served, any other refused, root not special-cased.

## Gates (real exit codes)

```
cargo clippy -p khive-runtime --all-targets -- -D warnings   rc=0
cargo test -p khive-runtime --lib                            rc=0   (921 passed)
cargo test -p khive-mcp                                      rc=0   (215 passed)
cargo fmt --all -- --check                                   rc=0
```

`libc` was already a direct dependency of `khive-runtime`, so this adds none.

Related: #1269 (the `khive-mcp` suite must be run with `KHIVE_PACKS` unset, or it fails on packs this distribution does not build).

### DIFF
--- a/crates/khive-runtime/src/daemon.rs
+++ b/crates/khive-runtime/src/daemon.rs
@@ -1101,19 +1101,27 @@ pub async fn run_daemon<D: DaemonDispatch>(dispatcher: D) -> anyhow::Result<()> 
     run_daemon_with_boot_guard(dispatcher, boot_guard).await
 }
 
-/// Require the socket's parent directory to be owner-only, and re-permission it
-/// only when it is the directory khive owns by convention.
+/// Vet the socket's parent directory, re-permissioning it only when it is the
+/// directory khive owns by convention.
 ///
-/// ADR-096 condition 2 rests the same-uid safety on an owner-only socket inside
-/// an owner-only directory, so a daemon that cannot establish that mode must not
-/// serve. That justifies refusing; it does not justify *rewriting* a directory
-/// the operator handed us. `KHIVE_SOCKET` takes an arbitrary path, so chmodding
-/// its parent unconditionally is wrong in both directions: pointed at a shared
-/// parent like `/tmp` it either fails outright for an ordinary user, or — worse
-/// — succeeds when privileged and strips access for every other process on the
-/// machine. A directory we did not create is checked, never modified.
+/// `KHIVE_SOCKET` takes an arbitrary path, and the previous unconditional
+/// chmod-0700 of its parent was wrong in both directions: pointed at a shared
+/// parent like `/tmp` it either failed outright for an ordinary user, or —
+/// worse — succeeded when privileged and stripped access for every other
+/// process on the machine. A directory we did not create is never modified.
+///
+/// What the directory mode must actually prevent is narrower than "any access":
+/// the connection gate is the 0600 socket plus the accept-time peer-uid check,
+/// both of which fail closed on their own. The attack only the directory can
+/// enable is a *swap* — another writer unlinking the bound socket and binding
+/// its own at the same path, silently diverting clients. That requires the
+/// directory to be writable by others without the sticky bit (the sticky bit
+/// is exactly the guard that lets `/tmp` host per-user files safely). So:
+/// group/other-writable and non-sticky is refused; merely readable or
+/// traversable (0755, or 1777 with sticky) is served, since the socket's own
+/// mode and the peer check carry the guarantee there.
 #[cfg(unix)]
-fn ensure_socket_dir_is_owner_only(parent: &std::path::Path) -> anyhow::Result<()> {
+fn ensure_socket_dir_is_swap_safe(parent: &std::path::Path) -> anyhow::Result<()> {
     if parent == khive_dir() {
         std::fs::set_permissions(parent, std::fs::Permissions::from_mode(0o700)).map_err(|e| {
             anyhow::anyhow!(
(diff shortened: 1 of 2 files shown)


ohdearquant:

ohdearquant:
Addressed the socket-directory permission change in 6635b71f.

**The problem was real and I reproduced it.** `socket_path()` honours `KHIVE_SOCKET` with no constraint on the path, and the boot sequence chmodded `sock.parent()` to `0700` unconditionally. With `KHIVE_SOCKET=/tmp/khived.sock` the parent is `/tmp`, so:

- as an ordinary user the chmod fails with `EPERM`, and because this branch now fails closed rather than warning, the daemon refuses to start where it previously ran;
- as root the chmod **succeeds**, and `/tmp` becomes owner-only for every other process on the machine.

The second is the serious one, and it predates this branch. What this branch changed is the first: the same call used to warn and continue.

**The fix separates the two things that were conflated.** ADR-096 condition 2 rests on the socket living in an owner-only directory, so refusing to serve without one is correct. Re-permissioning a directory khive does not own is not — those are different acts, and only the first is justified by the condition.

`ensure_socket_dir_is_owner_only()` now:

- chmods `0700` only when the parent is the default khive directory, which khive creates and owns. Unchanged behaviour for every default deployment.
- otherwise stats the directory and refuses if `mode & 0o077 != 0`, naming the mode it saw and how to resolve it. The operator's directory is never modified.
- fails closed on a stat error, on the same reasoning as the rest of this path: not being able to establish the mode is not the same as the mode passing.

Two tests cover it. The load-bearing one asserts that a `0777` directory is both refused *and* still `0777` afterwards, so a future change that reintroduces the unconditional chmod fails rather than silently passing the refusal check it just created for itself.

**On the non-Linux/Apple fallback**, this is deliberate rather than an omission. `peer_uid()` returns `ErrorKind::Unsupported` on unix targets outside Linux and Apple, and the accept site treats any error from it as "cannot prove same-uid", which is refused. So an unsupported platform declines connections rather than serving them unchecked — the fallback fails in the safe direction. Extending the `getpeereid(2)` arm to the BSDs would be correct and is a small change, but it is untested on those targets here and khive does not currently build for them, so I would rather leave the closed-fail than ship an unverified credential path.

`cargo test -p khive-runtime --lib`: 923 passed, 0 failed. Clippy clean with `-D warnings`, including the `x86_64-unknown-linux-gnu` cast issue from the previous run.

ohdearquant:
Correction to my previous comment: the owner-only check it described was too strict, and CI proved it — a93b3d6e replaces it.

Refusing any directory with group/other bits rejected exactly the shapes the fix was meant to serve: umask-default `0755` directories (which is what every `tempfile::tempdir`-based daemon test runs in, hence the 8 CI failures) and sticky world-writable `/tmp` itself.

The narrowed contract gates on the one attack the directory mode alone can enable — a socket swap, where another writer unlinks the bound socket and binds its own at the same path. That requires the directory to be writable by group/other **without the sticky bit**, so that is now the refusal condition:

- `0700`, `0755`, `1777` (sticky, the `/tmp` shape) → served as found
- `0777` non-sticky and similar → refused, naming the mode and the way out
- the khive-owned default directory → still created and chmodded `0700` as before
- any directory khive does not own → still never modified, on either outcome

Merely-readable directories are safe to serve because the connection guarantee never rested on the directory: the socket is `0600` post-bind (fail-closed) and this PR's accept-time peer-uid check refuses foreign uids (fail-closed). The sticky bit is precisely the mechanism that lets `/tmp` host per-user files without unlink races, so honoring it here is using it for its purpose.

Tests updated to pin the new boundary in both directions, including that acceptance does not re-permission the directory either. `daemon::` (84) and `server::` (24) khive-mcp suites — the ones that failed on CI — pass locally, plus the 923-test khive-runtime suite; clippy `-D warnings` and fmt clean.

ohdearquant (commented):

oceanwaves630 (approved):
Approving at head e67f3002. The credential gate fails closed when peer credentials cannot be read, socket-path validation walks components in traversal order with every stat failure treated as a refusal, and both unsafe blocks carry SAFETY rationale. Refusing a config escape hatch for the uid check is the right call for a control whose value is the guarantee itself. CI is green per-commit on this SHA, with focused daemon tests covering the foreign-uid refusal and both symlink-component cases.

Issue #1247, opened 2026-07-21
kg query() silently caps results at 500 rows post-filter — no truncation flag, no total, no pagination


ohdearquant:
GQL query() applies a hard 500-row cap AFTER filtering, silently:

- query(... LIMIT 3000) returns exactly 500 rows, no error, no truncation indicator.
- Narrowing the WHERE still returns exactly 500 until the true match count drops below the cap.
- SKIP is a parse error ('unexpected trailing input'), and count(x) is unsupported — so there is neither a way to page past the cap nor a cheap way to detect that it was hit.

Consequence observed in real use: an enumeration sweep over ~117 matching records first reported 110 and looked complete; 7 records were silently dropped. A caller trusting its own query gets a wrong answer with no signal.

Ask (any of, in preference order): honor LIMIT up to a documented ceiling and return an explicit truncated=true (plus total where cheap); support SKIP/OFFSET for paging; support count(). At minimum, the response must disclose truncation — silent caps convert 'covered everything' into 'silently sampled'.

ohdearquant:
Resolved on main by c7418f9c (#1272): query results now carry an explicit truncated flag when the post-filter cap is hit (operations.rs), so truncation is no longer silent.

Issue #1233, opened 2026-07-21
search and list diverge on unknown kind: search returns silent empty ok, list errors


ohdearquant:
search(kind="workspace_doc", query=...) with an unknown kind returns {ok:true, result:[]} — a silent empty result — while list(kind="workspace_doc") correctly returns {ok:false, error:"unknown kind ..."} listing valid kinds. The two verbs shared an error contract for unknown kinds and have diverged. A silent empty result for a typo'd kind is indistinguishable from a genuinely empty result set, which hides caller bugs. Expected: search rejects unknown kinds with the same error shape as list. Reproduced on the current binary 2026-07-21.

ohdearquant:
Not reproducible on current main. `search()` and `list()` resolve the kind through the same `resolve_kind_spec()` in `handlers/common.rs`, so they cannot diverge on this axis, and running the literal repro against a fresh build with the default pack set returns the expected error:

```
search(kind="workspace_doc", query="x")
→ invalid input: unknown kind "workspace_doc"; valid: artifact | concept | dataset | ...
```

Regression coverage already exists (`search_unknown_kind_returns_invalid_input`, `search_bogus_kind_lists_memory_in_error`, `search_unregistered_kind_matches_list_error_shape`).

One note that may explain the original observation: kind validity depends on which packs are loaded (`KHIVE_PACKS`). On a server with a pack that registers the kind, `search(kind="workspace_doc", ...)` is a valid query and returns results normally rather than erroring — which can look like a silent fall-through if the expectation was an error. If you can still reproduce a `search`/`list` divergence on the same pack configuration, please reopen with the loaded pack list.

Issue #1220, opened 2026-07-21
fix(mcp): partially-failed batch surfaces as overall success — needs a distinct non-success signal


ohdearquant:
When some ops in a batch fail (e.g. rejected by the secret gate) and their dependent ops are skipped, the response carries per-op errors and `summary.failed > 0`, but nothing at the top level distinguishes the outcome from full success — and the `kkernel exec` CLI exits 0. A caller trusting the absence of an exception concludes the whole plan applied.

Observed second-order effect: rejected creates left edges with unresolvable endpoints (correctly skipped rather than written dangling), producing a graph that *looked* applied but had orphaned entities and a comparison record silently missing one member. Only a stats-diff reconciliation caught it.

Suggested: (1) `kkernel exec` exits non-zero when `summary.failed > 0` (or gains a `--strict` mode if exit-code compatibility matters); (2) the aggregate response carries an explicit `status: partial` marker so programmatic callers can branch without scanning per-op results.

ohdearquant:
Second repro, from the opposite direction of the original report: a caller ran a single `memory.remember` via `kkernel exec --ops-file`. The invocation surfaced an ann_warm phase warning on stderr and no legible summary, so the caller could not tell whether the write landed, re-ran it, and produced two identical semantic memories four seconds apart (memory create is not idempotent).

Same root cause seen from the caller's side: in the first repro the output said success while part of the batch had failed; here the output said nothing legible while the whole op succeeded. The two natural caller responses — trust it, or retry it — are opposite and both wrong (ship a partial write vs duplicate a non-idempotent create).

Design constraint this adds: a warning line from an unrelated background subsystem (ANN warm, event-append cancellation) can bury the summary block, so 'a warning and no summary' is the *normal* view when any background phase complains, not a rare pathology. The fix needs to survive that: an explicit status field on the summary plus a non-zero exit when any op failed both survive it; a summary that must be visually located in a stream of subsystem warnings does not.

Issue #1174, opened 2026-07-20
kg search result rows use title/entity_kind while list/get use name/kind — consumers matching the list/get shape get silent empties


ohdearquant:
## Problem

`search(kind="entity", ...)` returns rows shaped `{id, title, entity_kind, score, snippet}`, while `list(...)` and `get(...)` return `{id, name, kind, description, properties, created_at, ...}` for the same records.

A consumer that filters search hits on `row["name"]` / `row["kind"]` — the natural projection after working with list/get — matches nothing and reads as an empty result. No error, no warning. Observed in practice: a name-collision guard reported zero collisions on a plan whose 13 entity names were all provably present in the store.

Search rows also carry no `created_at`, so any time-based forensics over hits requires a follow-up `get` per row.

## Direction

Converge the row vocabulary: search rows should carry `name` and `kind` (or both surfaces should share one projection), and timestamps where cheaply available. If `title`/`entity_kind` must remain for compatibility, emit both spellings for a deprecation window rather than leaving two disjoint shapes.

Same failure family as the query warnings-envelope issue: a shape mismatch that reads as "nothing there" instead of "wrong projection".

Verified against v0.3.0.

ohdearquant:
Resolved on main by c7418f9c (#1272): search rows for entities and notes now carry the same name/kind fields as list/get, with title/entity_kind retained for compatibility.

PR #1272, opened 2026-07-22
fix(kg): search/list/query response-shape consistency

ohdearquant:
## Summary

Three response-shape fixes for the `kg` pack and the MCP `request` envelope:

- `search` rows for both entities and notes now carry the same `name`,
  `kind`/`entity_kind`/`note_kind`, and `created_at` fields that `list()` and
  `get()` already return, instead of a narrower ad hoc shape (#1174).
- The `request` tool's batch/chain response envelope now carries a
  `status: "success" | "partial"` field, so a caller can detect a partial
  failure without diffing `summary.failed`/`summary.aborted` by hand. The
  `kkernel exec` CLI gains a `--strict` flag that turns a partial result into
  a non-zero exit code for scripting (#1220).
- GQL query results now carry a `truncated: bool` field whenever the result
  set was capped, and the accompanying warning text no longer suggests
  `LIMIT`/`OFFSET` paging syntax the query language doesn't support (#1168,
  #1247).

## Also investigated

#1233 (unknown `kind` silently falling through in `search`) did not
reproduce against current `main` — `search` and `list` share the same kind
resolution path and already return matching errors for unrecognized kinds,
and existing regression tests cover this. No code change included for it;
flagging for maintainer triage rather than closing it directly.

## Test plan

- [x] `cargo test -p khive-pack-kg -p khive-runtime -p khive-mcp -p kkernel`

### DIFF
--- a/crates/khive-mcp/src/save_sink.rs
+++ b/crates/khive-mcp/src/save_sink.rs
@@ -202,12 +202,30 @@ pub fn write_and_manifest(
     let abs_path = std::fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
     let null_counts_val: Value = serde_json::to_value(&null_counts).unwrap_or_else(|_| json!({}));
 
+    // Carry the envelope's op-outcome summary through: the manifest is the
+    // only output the caller sees on the save path, and without the counts a
+    // strict caller (`kkernel exec --strict --save-file`) cannot distinguish
+    // an all-green batch from one whose failures are sitting in the file.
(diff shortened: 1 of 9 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Approving at head 0a923e68. The save sink now preserves the failing batch summary so strict enforcement observes failed and aborted counts and exits non-zero, resolving the earlier exit-code path. The title fallback is removed from search so a nameless note serves a null name consistently across search, list, and query, with the regression test asserting that shape. CI is green per-commit on this SHA.

Issue #1235, opened 2026-07-21
fix(runtime): import_kg skips edge endpoint-contract validation


ohdearquant:
## Problem

`import_kg` (crates/khive-runtime/src/portability.rs, edge loop around line 241) validates edge weight and endpoint *existence*, but never applies the edge endpoint contract (the closed relation x kind allowlist plus pack-declared additions) before `graph.upsert_edge`. An imported archive can therefore introduce edges that `link` would reject — e.g. `precedes` between two `concept` entities, or any relation between kinds outside its contract.

## Impact

The endpoint contract is only enforced on the interactive `link` path. Any archive import silently bypasses it, so a single import can seed contract-violating edges that later confuse traversal, validation sweeps, and consumers that assume the contract holds store-wide. Same defect family as #1216 (merge rewire bypass): secondary write paths skipping the validator.

## Expected

Each imported edge passes the same composed endpoint validation as `link` (`validate_link_endpoints` semantics, base contract + loaded packs' rules). Contract-violating edges are counted and skipped with a warning (consistent with the existing missing-endpoint skip behavior), not imported.

## Repro sketch

Export a valid archive, edit one edge's relation to a kind-incompatible one (e.g. `precedes` with both endpoints `concept`), re-import: the edge is accepted today.

ohdearquant:
Resolved on main by 387b60ef (#1273): the import path applies composed endpoint validation before each upsert and skips invalid edges. Regression test: import_edge_violating_endpoint_contract_is_skipped (crates/khive-runtime/src/portability.rs).

Issue #1225, opened 2026-07-21
kg validate reads source_id/target_id while sync/archive NDJSON serializes source/target — endpoint checks silently skip canonical records


ohdearquant:
## Summary

`kkernel kg validate`'s endpoint-sensitive checks read edge fields named `source_id` / `target_id`, but the canonical writers and readers of the kg NDJSON source directory's `edges.ndjson` serialize those fields as `source` / `target`:

- `crates/khive-vcs/src/sync.rs` — `NdjsonEdge { edge_id, source, target, relation, weight, .. }` (reads/writes the kg source directory's `edges.ndjson`)
- `crates/kkernel/src/kg/archive.rs` — its `NdjsonEdge` likewise uses `source` / `target`
- `khive_runtime` portability export (`ExportedEdge`) also serializes `source` / `target`

Meanwhile in `crates/kkernel/src/kg/validate.rs`:

- L264: required-field check demands `["source_id", "target_id", "relation"]`
- L591: `let s = v.get("source_id")?.as_str()?` — the `?` on a missing field silently yields `None`, so the sort-order / endpoint passes skip records in the `source`/`target` spelling instead of flagging them
- L690: endpoint-kind iteration over `&["source_id", "target_id"]`

## Impact

An `edges.ndjson` produced by the sync/archive/export writers either fails the required-field check or — worse, in the passes built on the L591 pattern — is silently skipped by the very checker built to validate it. A clean `kg validate` run then reads as a clean graph while the endpoint rules were never evaluated against those records.

## Ask

1. Decide which spelling is canonical for the kg NDJSON files (`source`/`target` appears to be what every writer emits) and align `validate.rs` to it, or accept both with an explicit conversion boundary.
2. Make the skip loud: a record that does not match the expected schema must be reported, never silently excluded from endpoint validation.
3. Add a round-trip test: archive/sync-written `edges.ndjson` must pass (and be actually evaluated by) `kg validate`.

ohdearquant:
Resolved on main by 387b60ef (#1273): validation accepts canonical source/target fields at an explicit conversion boundary and reports malformed records instead of silently skipping them. Regression tests: referential_integrity_catches_missing_target_in_canonical_spelling and canonical_spelling_edge_round_trips_clean (crates/kkernel/src/kg/validate.rs).

PR #1273, opened 2026-07-22
fix(kg): close two edge-validation bypass paths

ohdearquant:
## Summary

- `import_kg` checked that both edge endpoints existed in the target
  namespace but never validated the (source kind, relation, target kind)
  triple against the same endpoint contract `link()` enforces. An imported
  archive could introduce edges the interactive path would reject outright.
  Contract-violating edges are now skipped and counted, consistent with the
  existing dangling-endpoint skip behavior (#1235).
- `kg validate`'s edge checks read `source_id`/`target_id`, but every NDJSON
  writer (sync, archive, portability export) serializes edges as
  `source`/`target`. A validate run against normally-produced NDJSON either
  reported a missing-field error or silently excluded those records from
  endpoint evaluation, so a clean run didn't mean the endpoint rules had
  actually been checked. Endpoint field reads now accept both spellings
  (#1225).

## Test plan

- [x] `cargo test -p khive-runtime -p kkernel`

### DIFF
--- a/crates/khive-runtime/src/portability.rs
+++ b/crates/khive-runtime/src/portability.rs
@@ -270,6 +270,20 @@ impl KhiveRuntime {
                 edges_skipped += 1;
                 continue;
             }
+            if let Err(e) = self
+                .validate_edge_relation_endpoints(token, ee.source, ee.target, ee.relation)
+                .await
+            {
+                tracing::warn!(
+                    source = %ee.source,
+                    target = %ee.target,
+                    relation = ?ee.relation,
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Approving at head d4944010. Error propagation from import_kg now retains operational error variants while InvalidInput and NotFound remain the only skippable classes, matching the validator's actual contract. Focused tests executed at this head: all 15 portability tests and both canonical-NDJSON validator regressions pass, and CI is green per-commit on this SHA.

PR #1274, opened 2026-07-22
fix(kkernel): reject db-override/backend conflicts before daemon forward

### LINKED ISSUE #1226: kkernel exec: KHIVE_DB + [[backends]] guard fires on --ops-file but not on inline ops — the two forms disagree
## Observed

With a khive.toml declaring 2 `[[backends]]` entries and `KHIVE_DB` set in the environment, same cwd, same binary, same store, back to back:

```
kkernel exec 'stats()'                    # succeeds, returns normally
kkernel exec --ops-file probe.jsonl       # hard error
```

Failing form's error (from `crates/khive-mcp/src/serve.rs:371`, `build_registry_for_multi_backend_inner`):

> --db "<path>" (or KHIVE_DB) cannot be combined with [[backends]]: 2 backend(s) are already declared in khive.toml, so applying this override here is ambiguous (it could silently collapse distinct declared backends onto a single file).

Isolation checks already done: not the working directory (both cwd variants succeed inline and fail via --ops-file), not the payload (the ops file is a single `{"tool":"stats","args":{}}` line). The only variable is the call form.

## Why this is a bug regardless of which behavior is intended

The guard's own rationale (an override could silently collapse distinct declared backends) applies identically to both forms:

- If the ambiguity is real, the inline form is **accepting an override it should refuse** — or worse, silently ignoring `KHIVE_DB` (likely riding the daemon fast path, which never consults the caller's env), which means the caller believes an override is in effect when it is not.
- If the ambiguity is not real, the batch form is refusing an override it should accept.

Either way the two forms of the same subcommand disagree about whether the store override is legal, and scripts generalize from whichever form they happened to use.

## Expected

One consistent rule for `KHIVE_DB`/`--db` vs `[[backends]]` across inline ops, `--ops-file`, and daemon-vs-direct execution paths — with the daemon path either honoring the override, refusing it loudly, or documenting that env overrides only apply to direct execution.

ohdearquant:
## Summary

- `run_exec_inline_with_forward` now validates an explicit `--db` override against the
  project's configured backends *before* forwarding the request to the daemon, using the
  same check `khive-mcp`'s registry-build path already applies. Previously the conflict
  only surfaced after a round trip to the daemon.
- Extracted the check into `khive_mcp::serve::validate_db_override_against_backends` so
  both call sites share one implementation.

## Test plan

- [x] `cargo test -p khive-mcp -p kkernel --lib` (green, includes a new regression test
      asserting the guard fires before the daemon forward is attempted)
- [x] `cargo clippy -p khive-mcp -p kkernel --all-targets -- -D warnings`

Closes #1226

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -351,20 +351,34 @@ pub fn build_registry_for_multi_backend_with_db_anchor(
     build_registry_for_multi_backend_inner(base_config, khive_cfg, cli_db_override)
 }
 
-fn build_registry_for_multi_backend_inner(
-    base_config: RuntimeConfig,
-    khive_cfg: &KhiveConfig,
+/// Validate a `--db`/`KHIVE_DB` override against a non-empty `[[backends]]`
+/// declaration WITHOUT opening any backend — the same rule
+/// `build_registry_for_multi_backend_inner` enforces, factored out so a
+/// caller that hasn't yet decided whether it will construct backends in this
+/// process can apply the check up front.
+///
+/// This closes #1226: `kkernel exec`'s daemon-forward fast path (inline ops,
+/// used whenever a warm daemon answers) never called into this guard at all
+/// — only the in-process fallback did — so an inline invocation with a
+/// conflicting override silently forwarded to the daemon's own already-open
+/// backends instead of being rejected, while the same override on
+/// `--ops-file` (always in-process by design) correctly bailed. The two call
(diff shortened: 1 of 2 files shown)


ohdearquant:

ohdearquant (commented):

oceanwaves630 (approved):
Verdict artifact read in full: clean, 0 blocking, 2 minor maintainability follow-ups noted. Per-SHA CI green at e93e9f14. Approving and merging.

Issue #1263, opened 2026-07-21
test: VECTOR_FAIL_NS is a process-wide slot, so parallel tests can defeat each other's failure injection


ohdearquant:
`crates/khive-runtime/src/operations.rs` arms vector-failure injection through `arm_vector_fail(ns)`, which writes into a `static VECTOR_FAIL_NS: Mutex<Option<String>>` — one process-wide slot, not thread-local.

Any two tests that arm injection concurrently share that slot. The second `arm_vector_fail` overwrites the first before the first test's `create_note`/`create_entity` call consumes it, so the first test's injection silently does not fire and its `assert!(result.is_err(), ...)` fails.

Observed on `operations::tests::create_note_vector_failure_rolls_back_note_row_and_fts` under the default multi-threaded runner: intermittent failure, deterministic pass under `--test-threads=1`. The test is correct; the injection harness it depends on is not isolated.

This is a test-infrastructure defect, not a product defect — no shipped code path reads these statics. It is worth fixing because the failure presents as a flake in an unrelated test, which costs triage time on every change that shifts test scheduling.

Shape of the fix: give each test its own injection scope rather than a shared static — either a task-local/thread-local slot, or a guard object that owns the slot for its lifetime and is required to arm injection at all. The same treatment applies to `VECTOR_FAIL_AFTER`.

Issue #1204, opened 2026-07-20
test isolation: a workspace test can open and migrate the operator's real default store


ohdearquant:
## What happened

During a routine \`cargo test\` run in a feature worktree (built from a post-#1147 tree, so the binary knows migration 12), the operator's real default store (khive.db in the home data directory) was migrated from schema v11 to v12. The installed release binary on the machine still topped out at migration 11, so every subsequent direct-open of the store by that binary failed with:

    migration: invalid data: database schema version 12 is ahead of the latest known migration 11

The long-running daemon (holding a handle opened before the bump) masked the skew for hours; only fallback/direct-open paths failed, which made the failure look like a config problem rather than a version gap (see #1203 for the diagnostics side).

## Why it is possible

\`RuntimeConfig::default()\` resolves \`db_path\` to the home-directory default store (crates/khive-runtime/src/config.rs:285). Any test that boots a runtime through a config path that inherits that default — rather than explicitly overriding \`db_path\` with \`None\`/tempdir/\`:memory:\` — opens the real store, and migrations apply on open, forward-only.

Timing correlation is strong (first direct-open failure within seconds of the test run finishing) but the exact offending test has not been pinned yet.

## Asks

1. Identify the test(s) that reach the default \`db_path\` (audit candidates: any \`RuntimeConfig::default()\` spread without a \`db_path\` override that flows into a runtime boot; also any path that auto-spawns/falls back around the daemon during tests).
2. Add a structural guard so this class cannot recur — e.g. test-harness-level \`HOME\` isolation for suites that boot runtimes, or a debug assertion that refuses to open a store under the home data directory when running under a test harness.
3. A CI tripwire is worth considering: snapshot \`PRAGMA user_version\` of a sentinel store before/after the workspace test suite.

Migrations being applied on open is by design (ADR-015); the defect is tests reaching the operator's store at all.

ohdearquant:
Additional pointer for the hunt: the `_schema_migrations` table records `applied_at` (crates/khive-db/sql/schema-migrations-table.sql:7), so the exact time migration 12 was applied to the affected store is already on record — whoever picks this up should read that row first and correlate against process history before instrumenting anything new. The gap is therefore narrower than 'nothing recorded when the store moved': the when is stored; what is missing is any record of *which process/binary* applied it. A worthwhile extension to the fix here: stamp the applying binary's version (and possibly pid/argv) into the migration row so a skew like this is attributable from the store alone.

[...]

Issue #1118, opened 2026-07-18
parallel-suite flake: four fault-injection flags still process-global after the 9fc066335 conversion


ohdearquant:
#1095 / #1099 (commit `9fc066335`) fixed a parallel-suite flake by converting fault-injection flags in `crates/khive-runtime/src/operations.rs` from process-global statics to thread-local `Cell`s. Two flags were converted; four were not, so tests using the remaining four still race.

Converted (thread-local, `operations.rs:46,59`):

- `LINK_FAIL_AFTER`
- `VECTOR_FAIL_AFTER`

Still process-global `Mutex<Option<String>>` (`operations.rs:76-87`):

- `FTS_FAIL_NS`
- `VECTOR_FAIL_NS`
- `FTS_FAIL_MANY_NS`
- `FTS_FAIL_MANY_PARTIAL_NS`

A test that sets one of these namespaces is visible to every other test running concurrently, so an unrelated test operating in that namespace takes the injected failure.

## Observed

`operations::tests::create_entity_vector_failure_rolls_back_entity_row_and_fts` fails in the full `cargo test -p khive-runtime` suite (900 passed, 1 failed) and passes when run in isolation with `--test-threads=1`. Same signature as the flake `9fc066335` already fixed for its sibling.

This surfaced during verification of unrelated work; the failing test, its fixture, and the injection helper were all untouched by that change.

## Fix

Apply the same conversion `9fc066335` used: move the four remaining flags to thread-local storage so an injected failure is scoped to the test that set it. The existing commit is the pattern to follow.

Worth doing rather than retrying, because the failure mode is a test that fails only under load, which is exactly the shape that gets dismissed as noise and erodes trust in the suite.

Issue #705, opened 2026-07-08
test(kkernel): instrumented-run contention window — serial_test does not cover cross-binary llvm-cov concurrency


ohdearquant:
`pending_events::tests::replayable_action_dispatches_without_failure_at_trigger_time` was hardened against scheduler contention with `#[serial_test::serial]` plus a raised `KHIVE_CHECKOUT_TIMEOUT_SECS` (see the test's doc comment for the original failure forensics). That hardening removes contention within the kkernel test binary only.

The coverage-ratchet CI job (`cargo llvm-cov --workspace`) still runs the other workspace test binaries concurrently, and instrumentation slows everything down, so the original contention window re-opens in instrumented runs. Observed: the test failed in the coverage-ratchet job on a commit whose plain CI test jobs were green (run 28909710755), with the claim-time checkout-failure signature described in the doc comment.

Candidate fixes, in preference order:
1. Make the checkout-timeout raise unconditional for this test path under `cfg(coverage)` / an env the coverage job sets, sized for instrumented throughput.
2. Run the coverage job's test phase with reduced parallelism (`--test-threads` / `-j`) accepting the runtime cost.
3. A cross-binary lock (file-based) shared by the contention-sensitive tests.

Folklore status is the failure mode to avoid: every future red on this test in an instrumented run will otherwise be waved through as "the known flake."

ohdearquant:
Closing as completed on current main ca70e4ea by merged #1275 (f7da3142). The coverage-ratchet job now sets KHIVE_CHECKOUT_TIMEOUT_SECS=300 specifically for cargo llvm-cov, and replayable_action_dispatches_without_failure_at_trigger_time takes max(ambient, 120) while restoring the prior environment afterward. This directly implements the preferred coverage-specific timeout remedy without weakening the single-drain zero-failure assertion; the current coverage job has succeeded with that contract.

PR #1275, opened 2026-07-22
test: isolate workspace test suites from ambient env and the operator's default store

### LINKED ISSUE #1269: khive-mcp tests inherit KHIVE_PACKS from the ambient environment and fail on packs this distribution does not build

ohdearquant:
## Summary

- Pin the `pending_events` test harness pack list to `["kg"]` instead of inheriting
  `KHIVE_PACKS` from the ambient environment, so the suite doesn't fail on a developer
  shell that exports a wider pack list than this distribution builds.
- Make the fault-injection namespace slots (`VECTOR_FAIL_NS`, `FTS_FAIL_MANY_NS`,
  `FTS_FAIL_MANY_PARTIAL_NS`) namespace-keyed sets instead of single overwritable
  slots, matching the pattern `FTS_FAIL_NS` already uses — two tests arming injection
  concurrently could otherwise stomp each other's arm before it fired, producing an
  intermittent, test-scheduling-dependent failure in an unrelated assertion.
- Add a CI tripwire (`scripts/ci.sh`'s `phase_tests`) that snapshots the operator's/
  runner's default store path before `cargo test --workspace` and fails loudly if the
  suite created, modified, or removed it — the class of bug where a test boots a
  runtime through a config path that inherits the real default store instead of an
  isolated one.
- Let the coverage job raise the checkout-timeout floor for a contention-hardened
  drain test instead of that test unconditionally clobbering any larger value the
  job set: the test now takes `max(ambient, 120s)`, and the coverage-ratchet job
  sets `KHIVE_CHECKOUT_TIMEOUT_SECS=300` to give instrumented runs a wider margin.

## Test plan

- [x] `cargo test -p khive-runtime -p khive-mcp -p kkernel --lib` (green: 925/925,
      215/215, 256/256)
- [x] `cargo test -p khive-runtime --lib -- --test-threads=8`, 8 consecutive runs,
      no flake
- [x] Reproduced the pack-list failure with a wide `KHIVE_PACKS` set, confirmed the
      fix clears it
- [x] `cargo clippy -p khive-runtime -p khive-mcp -p kkernel --all-targets -- -D warnings`
- [x] `sh -n scripts/ci.sh` + standalone functional check of the new tripwire's
      create/modify/untouched/removed branches against synthetic fixtures

Closes #1269, #1263, #1118, #1204, #705


ohdearquant (commented):

ohdearquant (commented):

PR #1277, opened 2026-07-22
fix(runtime): re-validate pack endpoint contract on merge edge rewire

### LINKED ISSUE #1216: fix(runtime): merge edge-rewire bypasses pack endpoint validation
Split out of #1214 (whose column-loss half is fixed by #1215).

The merge rewire loop in `crates/khive-runtime/src/curation.rs` rewrites edge endpoints with direct SQL and never runs the rewired `(source kind/entity_type, relation, target kind/entity_type)` triple through the pack endpoint validation that `link` enforces (`validate_edge_relation_endpoints`). A rewired edge can therefore violate type-scoped endpoint rules that a direct `link` would reject.

Exposure is narrow with the base rules (rewire preserves the relation and only substitutes one endpoint entity, and base rules are kind-scoped, so a violation requires the into/from entities to differ in kind or pack-typed subtype) but real once packs with `EntityOfType`-scoped endpoint rules are loaded.

Design question to settle before implementing: what happens to an edge whose rewired form fails validation — drop it (DO NOTHING-style, mirroring the #1215 conflict arm), or abort the merge? Needs a short spec note; the answer is not in ADR-039.

### LINKED ISSUE #1236: fix(runtime): entity merge only rewires edges in the caller's namespace
## Problem

The entity-merge edge-discovery SQL (crates/khive-runtime/src/curation.rs, incident-edge collection around lines 1090-1130) is namespace-scoped: `FROM graph_edges WHERE namespace = ?1 AND source_id/target_id = ?2`.

Edges, however, can legally exist in a *different* namespace while pointing at the merged entity: `link` stamps the edge with the creator's namespace, and by-ID endpoint resolution is namespace-agnostic by design (ADR-007 Rev 6). A merge therefore rewires only same-namespace edges; edges recorded under any other namespace keep pointing at the absorbed entity id.

## Impact

After a merge, cross-namespace edges reference an absorbed (tombstoned) entity. Traversals and neighbor queries anchored on the kept entity silently lose those connections; queries that join through the absorbed id see a dangling endpoint.

## Expected

Merge discovers and rewires incident edges by endpoint id across namespaces (endpoints are ids, not namespace-scoped), consistent with by-ID semantics elsewhere. Namespace remains an attribution stamp on the edge row itself.

## Notes

Distinct from #1216 (rewire skips endpoint validation) — this is about which edges are discovered at all. A fix should cover both entity and note merge paths.

ohdearquant:
## Summary

- `merge_entity`/`merge_note` rewired incident edges with direct SQL and
  never re-ran the rewired `(source kind/entity_type, relation, target
  kind/entity_type)` triple through the same endpoint validation `link`
  enforces. A merge only requires equal `kind` between `into`/`from`, not
  equal `entity_type`, so a pack rule scoped to a specific `entity_type`
  could accept the pre-merge edge and reject the identical relation once
  rewritten onto the surviving entity. The rewired triple is now re-checked
  against the base allowlist and pack `EDGE_RULES`; a violating edge is
  dropped and counted in a new `edges_contract_skipped` summary field
  (closes #1216).
- Incident-edge discovery was scoped to the merge caller's own namespace,
  but by-ID edge endpoints resolve namespace-agnostically and an edge is
  stamped with its creator's namespace rather than either endpoint's. An
  edge recorded under a different namespace than the merge call was left
  pointing at the absorbed entity after merge. Edge discovery, the conflict
  probe, and the rewire/delete now key off each edge's own namespace
  instead of the merge's namespace argument (closes #1236).
- Both fixes apply to the entity and note merge paths, and to both the
  writer-task and `spawn_blocking` transaction routes.

## Test plan

- [x] `cargo test -p khive-runtime --lib` (936 passed, includes 3 new
      regression tests: cross-namespace edge rewire, contract-violation
      drop, and its dry-run counterpart)
- [x] `cargo test -p khive-pack-kg merge` (integration + apply-worker merge
      coverage green)

### DIFF
--- a/crates/khive-runtime/src/curation.rs
+++ b/crates/khive-runtime/src/curation.rs
@@ -241,10 +241,13 @@ fn resolve_merge_edge_endpoint(
 /// reused here rather than re-derived, per the #543/#621 lesson that a parallel
 /// matcher drifts out of sync with the validator.
 ///
-/// `annotates` is exempt: within a merge, the entity/note being rewired can only
-/// ever be `annotates`'s unfiltered *target* (its source must be a note, enforced
-/// at edge creation and unaffected by which id the target column holds), so
-/// rewiring the target id never changes that verdict.
+/// `annotates` is exempt: its source-must-be-a-note constraint is enforced at
+/// edge creation and unchanged by rewiring (an entity merge only ever rewires
+/// its unfiltered target; a note merge rewiring the source substitutes another
+/// note), and its target may be any substrate. Callers short-circuit `annotates`
+/// before endpoint resolution — an annotates target may be an event or an edge,
+/// which `resolve_merge_edge_endpoint` cannot resolve; the exemption here is
+/// kept as defense in depth.
 // REASON: the two endpoints each need substrate/kind/entity_type independently —
 // collapsing them into a tuple/struct would obscure which side is which at call
 // sites that already pass them as separate locals.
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Verdict artifact read (APPROVE, 0 blocking, 0 issues, evidence leg); per-SHA check-runs verified at 5ee094d5 (CI success, Bench Trend success; Auto-merge skipped = bot-gated job, N/A here).

PR #1278, opened 2026-07-22
fix(kkernel): install a BlobStore in the local exec fallback path

### LINKED ISSUE #1209: kkernel exec local dispatch never installs a BlobStore (serve/exec parity gap)
## What

\`resolve_blob_store\` is called only from the MCP server boot path (\`crates/khive-mcp/src/serve.rs\`). The \`kkernel exec\` local-dispatch path (\`build_local_fallback_server\` in \`crates/kkernel/src/exec.rs\`) never installs a blob store, so with the blob pack loaded, \`blob.put\`/\`blob.get\`/\`blob.stat\` return \`unconfigured: no BlobStore installed on this server\` under \`exec\` — even with a file-backed database and \`KHIVE_BLOB_ROOT\` set, both of which configure the store correctly on the serve path.

## Repro

```bash
KHIVE_DB=/tmp/scratch.db KHIVE_NO_DAEMON=1 KHIVE_PACKS=kg,blob \
KHIVE_BLOB_ROOT=/tmp/blobs kkernel exec 'blob.put(bytes="aGVsbG8=")'
# -> {"error":"unconfigured: no BlobStore installed on this server ..."}
```

The same op through \`kkernel mcp\` stdio succeeds (FsBlobStore installed at boot with the documented root precedence).

## Expected

\`exec\`'s local fallback server mirrors the serve boot's \`resolve_blob_store\` installation (same ADR-111 Amendment 2 semantics: config/env resolution, fail-closed on invalid \`[storage.blob]\`), per the standing rule that second dispatch paths mirror the primary path's configuration. Same family as the reindex-parity work.

Found while verifying the pack surface end-to-end on a scratch database.

ohdearquant:
## Summary

`kkernel exec`'s in-process fallback server (used whenever a request is
dispatched locally instead of forwarding to the warm daemon) never resolved
or installed a `BlobStore` for the single-backend case. Blob verbs would
fail as unconfigured through this path even when the same config/backend
combination resolves a usable store for the daemon's own boot path.

This installs the store the same way the daemon boot path already does,
before the runtime is wrapped in the server.

## Changes

- `install_resolved_blob_store` (khive-mcp `serve.rs`) is now `pub`, so
  `kkernel`'s exec module can call it.
- `build_local_fallback_server`'s single-backend branch now calls it right
  after constructing the runtime.
- New regression test: asserts the resolved blob root directory exists
  after construction (confirmed to fail without the fix).

## Test plan

- [x] `cargo test -p kkernel -p khive-mcp --lib` — 259 + 215 passed, rc=0
- [x] `cargo clippy -p kkernel -p khive-mcp --all-targets -- -D warnings` — clean, rc=0
- [x] `cargo fmt --all -- --check` — clean, rc=0
- [x] confirmed the new test fails without the fix (temporarily reverted the
      installer call locally, re-ran the test, restored it)

Closes #1209

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -996,7 +996,12 @@ pub fn checkpoint_pool_for(main_backend: &StorageBackend) -> Option<Arc<Connecti
 /// tests) is non-fatal and leaves `KhiveRuntime::blob_store()` unset:
 /// nothing yet consumes it, and forcing a filesystem root onto every
 /// in-memory boot would be a behavior change nobody asked for.
-fn install_resolved_blob_store(
+///
+/// `pub` so `kkernel`'s `exec` local-dispatch fallback server (the
+/// single-backend branch of `build_local_fallback_server`) can install a
+/// `BlobStore` the same way the `serve` boot path does, instead of leaving
+/// `exec`'s in-process runtime without one (khive#1209).
+pub fn install_resolved_blob_store(
     rt: &KhiveRuntime,
     khive_cfg: &KhiveConfig,
     backend: &StorageBackend,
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Verdict artifact read in full: clean, 0 blocking, 4 minor follow-ups noted. Per-SHA CI green at 5410f396. Approving and merging.

PR #1279, opened 2026-07-22
test(kkernel): pin exec test helpers' pack list instead of inheriting KHIVE_PACKS

### LINKED ISSUE #1276: test: kkernel exec tests honor ambient KHIVE_PACKS and fail when it names unavailable packs
Running `cargo test -p kkernel exec` with `KHIVE_PACKS` set in the environment to a pack list that includes packs not compiled into this workspace (e.g. `KHIVE_PACKS=kg,gtd,...`) fails 20 of 59 tests with `PackRegError { unknown: "gtd", .. }` at the shared server-on-temp-db setup (exec.rs test harness). With the variable unset, all 59 pass.

Test setup that spawns a server should pin its own pack list (or explicitly scrub `KHIVE_PACKS`) rather than inheriting the caller's environment — a unit test's outcome should not depend on ambient shell configuration. Reproduced on current main-based branches; workaround is `env -u KHIVE_PACKS cargo test -p kkernel`.

ohdearquant:
## Summary

`cargo test -p kkernel` failed a large slice of the exec-dispatch test suite
whenever the shell running it exported `KHIVE_PACKS` naming a pack not
compiled into this workspace (e.g. `KHIVE_PACKS=kg,gtd`). Several shared test
helpers built their `RuntimeConfig` via a struct-update spread from
`RuntimeConfig::default()`, which falls back to that environment variable
when the caller doesn't set its own `packs` field — so construction panicked
with an "unknown pack" registration error, even though none of the affected
tests exercise anything beyond the built-in `kg` pack.

## Changes

- Pinned `packs: vec!["kg".to_string()]` (or the `RuntimeConfigInputs`
  equivalent) on every affected helper/test: `isolated_server`, `atomic_cfg`,
  and three individually-constructed configs in tests exercising
  `build_local_fallback_server` directly.
- New regression test: sets `KHIVE_PACKS` to a value naming an unavailable
  pack and asserts `isolated_server` still succeeds (confirmed to fail
  without the fix).

## Test plan

- [x] `cargo test -p kkernel --lib` — 259 passed, rc=0
- [x] `cargo clippy -p kkernel --all-targets -- -D warnings` — clean, rc=0
- [x] `cargo fmt --all -- --check` — clean, rc=0
- [x] confirmed the affected tests fail under `KHIVE_PACKS=kg,gtd` before this
      fix (19 failures) and pass after (0 failures), both runs in an isolated
      throwaway `CARGO_TARGET_DIR` to rule out stale-binary reuse from a
      shared target directory
- [x] confirmed the new regression test itself fails without the fix
      (temporarily reverted the `isolated_server` pin, re-ran, restored it)

Closes #1276

### DIFF
--- a/crates/kkernel/src/exec.rs
+++ b/crates/kkernel/src/exec.rs
@@ -1128,10 +1128,42 @@ mod tests {
             db_path: Some(PathBuf::from(db_path)),
             embedding_model: None,
             additional_embedding_models: vec![],
+            // Pin the pack list explicitly rather than inheriting `KHIVE_PACKS`
+            // from the ambient environment (#1276) — every caller of this
+            // helper only dispatches `kg` verbs, so the behavior under test
+            // shouldn't depend on a wider pack set a developer's shell exports.
+            packs: vec!["kg".to_string()],
             ..Default::default()
         };
         let rt = KhiveRuntime::new(cfg).expect("runtime on temp db");
         KhiveMcpServer::new(rt).expect("server on temp db")
+    }
+
+    // ── isolated_server ignores ambient KHIVE_PACKS (#1276) ───────────────────
+    //
+    // `cargo test -p kkernel` failed ~20 exec tests whenever a developer's
+    // shell exported `KHIVE_PACKS` naming a pack not compiled into this
+    // workspace (e.g. `kg,gtd`): every `RuntimeConfig` built by this test
+    // module's shared helpers fell through to `RuntimeConfig::default()`'s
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Approving at head f7215e69. Executed review verdict is clean with zero blocking findings; the single minor issue (process-wide env mutation in the regression test with cleanup only on success) is noted for a follow-up and does not meet the blocking bar. CI is green per-commit on this SHA.

PR #1280, opened 2026-07-22
fix(runtime,db): bound embed-failure latency and issue-count fts passes

ohdearquant:
Counter and latency corrections in the multi-model embedding and FTS search paths.

**Embed-failure latency (runtime crate).** Multi-model entity and note creation awaited every spawned embedding task to completion before inspecting any result, so a fast provider failure waited on the slowest sibling before rollback and the error response. Both creation sites now share a drain (`drain_embed_join_set`) that aborts remaining tasks on the first failure and still drains every handle to termination, preserving the existing guarantee that no embed task (or the usage context it holds) survives the operation's return. Completion order is reassembled into model order by index, so success-path behavior is unchanged.

Precisely what the abort bounds: cancellation takes effect at await points. A sibling parked on a provider await (network provider, queue, semaphore) is cancelled promptly. A sibling already inside the native provider's synchronous single-text inference call cannot be preempted mid-call; the drain deliberately awaits that call's completion rather than detaching it, because detaching would break the no-work-survives-return invariant. Failure latency is therefore bounded by at most one in-flight single-text inference per sibling — not by every sibling running to full completion, which was the previous behavior.

**Embed-call accounting (runtime crate).** The issued `embed_calls` counter incremented only after the provider await returned, so a task aborted while a call was in flight left a dispatched call uncounted. The increment now lands at dispatch, before the provider await, in all three embed entry points — matching the issued-counter semantics used elsewhere (count when the resource is engaged, not on completion).

**FtsPasses issued semantics (db crate).** `FtsPasses` incremented for empty or fully-sanitized queries even though `search()` returns before preparing any FTS5 statement. The counter now publishes only after a statement is actually prepared, with the increment landing on the caller's task after the reader closure completes — the same issued-counter placement the graph store uses for round-trip and hop accounting.

**Tests.**
- `create_entity_fast_embed_failure_does_not_wait_for_hung_sibling`: a sibling provider that parks until a release that never fires; a bounded prompt error return is reachable only through the abort path (verified to hang under the previous await-to-completion drain). The test itself is wrapped in an outer timeout so a regression fails within the bound instead of hanging the suite.
- `aborted_embed_task_still_counts_issued_embed_call`: aborts a task parked inside the provider and asserts the issued call still counts (verified failing under the previous post-await increment).
- `test_search_empty_query_does_not_count_fts_pass`, `test_search_fully_sanitized_query_does_not_count_fts_pass`, `test_search_normal_query_counts_exactly_one_fts_pass`.
- Existing drain regression tests pass unchanged: the post-return quiescence guarantee is preserved.

Gates: runtime lib tests (940 passed), db lib tests (476 passed), workspace clippy with warnings denied, and fmt check — all clean.


ohdearquant (commented):

PR #1284, opened 2026-07-22
fix(runtime): scope fault-injection slots by namespace

### LINKED ISSUE #1118: parallel-suite flake: four fault-injection flags still process-global after the 9fc066335 conversion

### LINKED ISSUE #1263: test: VECTOR_FAIL_NS is a process-wide slot, so parallel tests can defeat each other's failure injection

ohdearquant:
The vector failure-injection slot held a single namespace, so concurrent tests arming distinct namespaces overwrote each other: the second arm silently replaced the first, and the first test's create then succeeded where an injected failure was expected.

The slot is now a namespace-keyed set matching the FTS injection slot: arming inserts a namespace and each injection site atomically removes only its own entry. The injection remains behind the existing `cfg(any(test, feature = "fault-injection"))` gating; production behavior is unchanged.

Regression test `vector_failure_injections_for_distinct_namespaces_do_not_overwrite_each_other` arms two namespaces and runs both creates concurrently — it fails under the single-slot implementation and passes under the set. The five vector-failure tests pass 10 consecutive runs at default parallelism.

Closes #1263

**Second commit (#1118).** The same single-slot overwrite defect existed in the two batch FTS injection slots (hard-error and partial-summary seams); both are now namespace-keyed one-shot sets with the identical arm/consume pattern, with two concurrent-arm regression tests that fail under the single-slot implementation.

Closes #1118

### DIFF
--- a/crates/khive-runtime/src/operations.rs
+++ b/crates/khive-runtime/src/operations.rs
@@ -86,8 +86,10 @@ pub fn arm_vector_fail_after(n: usize) {
 #[cfg(any(test, feature = "fault-injection"))]
 static FTS_FAIL_NS: std::sync::LazyLock<std::sync::Mutex<std::collections::HashSet<String>>> =
     std::sync::LazyLock::new(|| std::sync::Mutex::new(std::collections::HashSet::new()));
+// Vector insertion failures use the same namespace-keyed one-shot semantics.
 #[cfg(any(test, feature = "fault-injection"))]
-static VECTOR_FAIL_NS: std::sync::Mutex<Option<String>> = std::sync::Mutex::new(None);
+static VECTOR_FAIL_NS: std::sync::LazyLock<std::sync::Mutex<std::collections::HashSet<String>>> =
+    std::sync::LazyLock::new(|| std::sync::Mutex::new(std::collections::HashSet::new()));
 /// FTS failure injection for `create_many` — separate from `FTS_FAIL_NS` so that
 /// create_note_inner and create_many tests cannot disarm each other.
 #[cfg(any(test, feature = "fault-injection"))]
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

PR #1285, opened 2026-07-22
fix(search): propagate retrieval-leg membership through cross-backend merge and serialize it

### LINKED ISSUE #1281: search: cross-backend RRF merge labels every entity hit SearchSource::Both regardless of input legs
## Summary

`rrf_merge_entity_hits` (crates/kkernel/src/coordinator/dispatch.rs) assigns `SearchSource::Both` to every merged hit unconditionally, discarding the per-hit `source` carried by the input lists.

## Detail

The merge accumulates `(score, title, snippet)` per entity id — the input hit's `source` field never enters the map — and the output closure then constructs every `SearchHit` with:

```rust
source: khive_runtime::SearchSource::Both,
```

A hit that appeared only in text-leg results (or only vector-leg) across every backend is still labeled `Both`. The correct value is the union of the input hits' sources: text+text → `Text`, vector+vector → `Vector`, mixed or any `Both` input → `Both`.

The single-backend hybrid path computes this union correctly (crates/khive-runtime/src/retrieval.rs, where a hit present in both legs is promoted to `Both` and single-leg hits keep `Text`/`Vector`), and unit tests assert those values — so the cross-backend merge is the one producer of incorrect labels.

## Impact

Today the label is not serialized on any wire surface, so no external caller observes the wrong value. It still matters: any future consumer of the merged hits (including exposing the field per #1282) inherits an always-`Both` label on the multi-backend path, and any in-process code reading `hit.source` after the merge already sees the wrong value.

## Suggested fix

Carry the source through the accumulator (union semantics) instead of hard-coding `Both`; scores, ranks, and result ordering are unaffected.

### LINKED ISSUE #1282: search: per-hit retrieval-leg membership (text/vector/both) is computed but never serialized
## Summary

Hybrid search computes per-hit retrieval-leg membership (`SearchSource`: `Text` / `Vector` / `Both`) and both search result surfaces drop it — the field exists on `SearchHit` but is never serialized.

## Detail

- The hybrid retrieval path (crates/khive-runtime/src/retrieval.rs) tags each hit with which leg produced it, promoting hits present in both legs to `Both`. Unit tests assert these values.
- The single-backend entity serializer (`handle_search`, crates/khive-pack-kg/src/handlers/search.rs) emits `{id, entity_kind, score, title, snippet}` — no `source`.
- The multi-backend serializer (`dispatch_via_coordinator_inner`, crates/khive-mcp/src/server.rs) emits the same shape for entity and note hits — no `source`.

So callers cannot distinguish a lexical-only match from a semantic-only match from a fused match, even though the engine already knows.

## Why it's useful

Leg membership is standard retrieval diagnostics: it lets a caller (or an operator debugging recall quality) see whether a result came from FTS, the vector leg, or both, without re-running the query against each leg. The value is already computed on the hot path — exposing it is a serialization-only change with zero effect on ids, ranks, or scores.

## Suggested shape

Add a `source` field (`"text" | "vector" | "both"`) to the entity hit shape on both surfaces. The multi-backend path additionally needs the cross-backend merge fix in #1281 first — that merge currently overwrites every hit's source with `Both`, so exposing the field there before fixing the union would report wrong values.

ohdearquant:
Two search-surface fixes, stacked because the second depends on the first.

**Cross-backend merge (#1281).** `rrf_merge_entity_hits` and `rrf_merge_note_hits` stamped every merged hit `SearchSource::Both` regardless of which retrieval legs produced it. The merge now unions leg membership per hit: text with text stays text, vector with vector stays vector, mixed promotes to both — matching the single-backend fusion semantics.

**Serialization (#1282).** `SearchHit.source` was computed on the hot path and dropped by both serializers. The single-backend KG handler and the coordinator dispatch path now emit a lowercase `source` field (`"text" | "vector" | "both"`) on entity and note hits, via a shared `SearchSource::as_str()`. `NoteSearchHit` carries `source` through so note hits report leg membership too.

Tests: `cross_backend_entity_merge_preserves_retrieval_leg_membership` covers Text-only, Vector-only, Both, Text+Vector, Text+Text, Vector+Vector, and Both+Vector inputs; serializer tests assert the field on both surfaces. 37 coordinator tests pass; KG pack suites pass.

Closes #1281
Closes #1282


ohdearquant (commented):

PR #1286, opened 2026-07-22
fix(mcp): pin test runtime fixtures to the kg pack

### LINKED ISSUE #1269: khive-mcp tests inherit KHIVE_PACKS from the ambient environment and fail on packs this distribution does not build

ohdearquant:
Test-only runtime construction in khive-mcp passed `packs: None` and so inherited `KHIVE_PACKS` from the ambient environment; a value naming packs this distribution does not build failed the suite on pack-load errors unrelated to the code under test.

Test fixtures now pin the `kg` pack explicitly through a shared helper (`kg_test_packs`) in serve.rs, the pending-events fixture, and the engine-configuration integration fixture. Production pack resolution is unchanged; no test mutates process environment.

Verified red-to-green: with `KHIVE_PACKS=kg,gtd,bogus` exported, `pending_events::tests::dispatch_action_rejects_non_literal_prev_reference` failed with a pack-load error before the change and passes after; the suite passes with the variable unset and set.

Closes #1269

### DIFF
--- a/crates/khive-mcp/src/pending_events.rs
+++ b/crates/khive-mcp/src/pending_events.rs
@@ -1330,6 +1330,7 @@ mod tests {
             default_namespace: Namespace::parse("local").unwrap(),
             embedding_model: None,
             additional_embedding_models: vec![],
+            packs: vec!["kg".to_string()],
             actor_id: actor_id.map(str::to_string),
             ..Default::default()
         };
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -1471,6 +1471,10 @@ mod tests {
         path
     }
 
+    fn kg_test_packs() -> Vec<String> {
+        vec!["kg".to_string()]
+    }
+
     // The resolver MUST honor config-file `[[engines]]` over RuntimeConfig
     // defaults — otherwise `kkernel reindex` embeds for the wrong model set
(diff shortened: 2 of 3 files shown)


ohdearquant (commented):

PR #1288, opened 2026-07-22
feat(kkernel): embeds-only reindex repair mode and newer-store diagnostic

### LINKED ISSUE #1203: diagnostics: direct-open of a newer-schema store should say 'binary older than store' in one line
## Current behavior

When a client binary direct-opens a store whose schema version is ahead of the binary's latest known migration (e.g. after a daemon fallback), the failure surfaces as:

    WARN khive_mcp::daemon: daemon_fallback reason="config_mismatch"
    Error: backend main: migration: invalid data: database schema version 12 is ahead of the latest known migration 11. This database predates the consolidated baseline (ADR-015) or was written by a newer build. Recreate it from the current schema; in-place downgrade is not supported.

Two problems for the person debugging it:

1. The actionable condition — *this binary is older than the store; upgrade the binary* — is buried as the second alternative in a message whose first suggestion ("recreate it from the current schema") is destructive and wrong for this case.
2. The preceding \`config_mismatch\` WARN misdirects the investigation toward configuration when the real cause is a version gap. A fallback that cannot possibly succeed (store version > binary max migration) could detect that before attempting the open and report it directly.

## Ask

- When the migration runner sees \`store_version > max_known_migration\`, emit a dedicated error variant along the lines of: \`this binary knows migrations up to N but the store is at version M — the binary is older than the store; upgrade the binary (in-place downgrade is not supported)\`.
- Do not lead with "recreate the database" for this case; that suggestion should be reserved for the genuinely-pre-baseline case, which is distinguishable (missing schema-version table vs. version too high).

Observed in the field where the mixed messaging cost a real debugging session; the store itself was healthy the whole time.

### LINKED ISSUE #1243: kkernel reindex: no embeds-only repair mode — --keep-existing pays full FTS backfill + full enumeration to fill a small embedding gap
reindex couples FTS backfill and embedding in one sweep. --keep-existing correctly skips rows that already have embeddings (filter_unembedded), but every batch still runs the unconditional FTS backfill and the sweep enumerates the entire corpus.

Consequence: repairing a small number of missing embeddings (e.g. rows skipped by a since-fixed failure) costs a full-corpus pass — observed ~2.5h over 83k notes to fill ~7.5k missing embeddings, nearly all of it redundant FTS re-upserts and enumeration.

Ask: an embeds-only repair mode (skip FTS backfill; ideally enumerate only subjects missing an embedding for the target model via the vectors table, rather than walking the full corpus). Keep the current combined sweep as the default full-rebuild behavior.

ohdearquant:
Two operator-facing improvements to store maintenance and diagnostics.

**Embeds-only repair (#1243).** `kkernel reindex --embeds-only` fills embedding gaps without paying the full FTS backfill and full enumeration the combined rebuild costs. For each substrate it runs an indexed NOT EXISTS anti-join against the target model's vector table selecting only live rows lacking a vector, staged for batch embedding with ID-cursor pagination so a failing row cannot repeat a page. FTS backfill and the stale-partition sweep are unreachable from this mode; the default combined rebuild is byte-for-byte unchanged.

**Newer-store diagnostic (#1203).** Direct-opening a store written by a newer binary now yields a dedicated `SchemaTooNew` error carrying both versions and rendering one actionable line directing a binary upgrade — at both the initial migration check and the under-lock recheck, and in `kkernel db check` — replacing generic version guidance that suggested recreation.

Tests: reindex suite extended (34 pass) covering missing-vector selection, pagination, and mode isolation; migrations tests cover the variant and message at both guards.

Closes #1243
Closes #1203

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1968,6 +1968,7 @@ dependencies = [
  "khive-vcs-adapters",
  "lattice-embed",
  "rmcp",
+ "rusqlite",
  "serde",
  "serde_json",
  "serial_test",
--- a/crates/khive-db/src/error.rs
+++ b/crates/khive-db/src/error.rs
@@ -25,4 +25,17 @@ pub enum SqliteError {
         /// Human-readable description of the failure.
         error: String,
     },
+
+    /// The store was migrated by a newer binary and cannot be downgraded in place.
+    #[error(
+        "this binary knows migrations up to {max_known_migration} but the store is at version \
+         {store_version} — the binary is older than the store; upgrade the binary (in-place \
+         downgrade is not supported)"
+    )]
+    SchemaTooNew {
+        /// Highest migration version this binary can apply.
+        max_known_migration: u32,
+        /// Migration version recorded by the store.
+        store_version: u32,
+    },
 }
--- a/crates/khive-db/src/lib.rs
+++ b/crates/khive-db/src/lib.rs
@@ -33,8 +33,9 @@ pub use checkpoint::{checkpoint_once, run_checkpoint_task, CheckpointConfig, Che
 pub use checkpoint::{run_session_sweep_task, SessionSweepConfig, SweepBackend};
 pub use error::SqliteError;
 pub use migrations::{
-    inspect_schema_version, query_embedding_models, read_schema_version, run_migrations,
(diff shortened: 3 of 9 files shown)


ohdearquant (commented):

PR #1289, opened 2026-07-22
refactor(query): single-pass dialect selection and shared special-relation predicate

### LINKED ISSUE #1245: GQL impossibility-hint follow-ups: per-call rule cloning, duplicated dialect check, SPECIAL_RELATIONS list repeated across validator and hint
Non-blocking cleanups noted during review of the static impossibility hint (#1234, refined by #1241):

1. Pack endpoint rules are cloned per hint evaluation rather than borrowed/cached; cheap at current rule counts but avoidable.
2. The dialect check guarding the hint is duplicated at more than one call site; a single guard at the entry point would suffice.
3. The SPECIAL_RELATIONS exclusion list is spelled independently in the validator branch and in both hint branches; a shared const/predicate would keep the three from drifting.

None affect correctness today; batching here so they are not lost.

ohdearquant:
Structural cleanup of the GQL impossibility-hint path (#1245 follow-ups to #1240).

- Parser selection is computed once: `parse_auto_with_language` returns the chosen dialect alongside the parse, and the runtime consumes it instead of re-detecting the dialect at the hint site. The hint entry point owns the single GQL-only guard.
- Hint evaluation borrows the runtime's cached pack endpoint rules through a synchronous callback instead of cloning the rule vector per call.
- One shared `is_special_relation` predicate now backs live edge validation, merge rewiring, both hint paths, and endpoint help-table generation, replacing three hand-maintained copies of the relation list; coverage is exhaustive over the relation set.

Behavior is unchanged; this removes per-call allocation and the risk of the three relation lists drifting apart.

Verified on this base: query truncation already reports the structural `truncated` field with corrected warning text, covering the guidance concerns raised alongside these follow-ups.

Closes #1245

### DIFF
--- a/crates/khive-query/src/language.rs
+++ b/crates/khive-query/src/language.rs
@@ -34,24 +34,27 @@ pub fn parse(language: QueryLanguage, input: &str) -> Result<GqlQuery, QueryErro
 /// Returns [`QueryError`] when syntax is invalid, write-shaped, or unsupported.
 /// See `crates/khive-query/docs/api/parsing.md` for detection and guard behavior.
 pub fn parse_auto(input: &str) -> Result<GqlQuery, QueryError> {
+    parse_auto_with_language(input).map(|(_, query)| query)
+}
+
+/// Auto-detects and parses `input`, retaining the selected query language.
+///
+/// # Errors
+///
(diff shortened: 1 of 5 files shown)


ohdearquant (commented):

PR #1290, opened 2026-07-22
fix(mcp): confirm incumbent daemon exit before replacement spawn; fail closed on unreachable sockets

### LINKED ISSUE #1242: kkernel: unreachable daemon socket misdiagnosed as stale PID, client attempts daemon respawn from restricted context
When the daemon socket is unreachable (e.g. a sandboxed process without network access), the kkernel client misdiagnoses the state as 'PID in daemon file does not belong to a khive daemon' and then attempts to respawn the daemon itself (observed failing with exited_before_bind — but the attempt is the bug, not the failure).

A client that cannot reach the socket cannot distinguish 'daemon dead' from 'daemon unreachable from here'. It should fail loudly with 'cannot reach daemon' and never take daemon lifecycle actions: a restricted process taking over daemon management can kill or corrupt a healthy daemon serving other clients.

Repro: run kkernel exec 'stats()' from a workspace-write sandbox without network access while a healthy daemon is running outside.

Expected: clear 'cannot reach daemon socket' error, exit nonzero, no lifecycle side effects.
Actual: stale-PID misdiagnosis followed by an in-sandbox respawn attempt.

### LINKED ISSUE #1271: daemon: stale-daemon recovery spawns a replacement without confirming the incumbent exited, yielding two daemons on one socket path
## Summary

Daemon recovery sends `SIGTERM` to a stale daemon and spawns a replacement without ever confirming the incumbent exited. When the incumbent is alive but slow to answer a liveness probe, it survives the signal, the replacement binds a new socket at the same path, and two daemons end up serving one khive home. The evicted process is unreachable forever but keeps the database open.

## Observed

Two `kkernel mcp --daemon` processes serving the same home, both holding an open file descriptor on the daemon socket path, at **different socket inodes**:

```
$ lsof <daemon socket path>
kkernel 40351 ... 48u unix 0x6135728e4c3f0235 <daemon socket path>
kkernel 96603 ... 48u unix 0x9ca0329837df2d1f <daemon socket path>
$ cat <daemon pid file>
96603
```

Same path, two live listeners, two different kernel objects. The socket file's mtime matched the second process's start time exactly: the newcomer unlinked the path the incumbent was listening on and bound a fresh socket there. The evicted process stayed alive for seven hours, unreachable by any client, still holding the database shared-memory file open. A long-lived reader on the database is the read-snapshot shape behind the WAL checkpoint starvation issue (#580), so this is not merely untidy.

When finally signalled while idle, the orphan exited within three seconds. It survived the original eviction only because it was busy at that moment.

## Root cause

`crates/khive-mcp/src/daemon.rs` — `kill_stale_daemon_inner()` sends `SIGTERM` and returns immediately. There is no `waitpid`, no bounded poll on `is_process_running`, and no escalation. `kill_and_respawn` then calls `spawn()` on the next line.

The surrounding code is careful in every other respect: a boot lock, a recoverer lock, an `O_EXCL` pid-file write, socket-identity checks before any unlink, and both the `Timeout` and `LockContended` probe outcomes deliberately fail closed. The one irreversible step, "the incumbent is gone", is the only one that is assumed rather than observed.

## Why it is self-amplifying

The liveness probe (`probe_daemon_identity`) uses a 500 ms deadline. A cold daemon rebuilds its ANN index at boot; on a corpus of roughly 83k notes a single cold recall was measured at 124 s wall and 555 s CPU. A daemon inside that window cannot answer a 500 ms probe.

1. The incumbent is mid-rebuild, so it misses the probe and is classified dead.
2. It is signalled but not confirmed, and being blocked in the rebuild it does not exit.
3. A replacement spawns immediately and begins **its own** cold rebuild.
4. Two processes now rebuild the same index concurrently.

The busier the incumbent, the more likely it is evicted, and eviction doubles the load that caused the eviction. That is why this presents as episodic CPU bursts rather than steady burn, and why a freshly spawned daemon can be observed at several hundred percent CPU seconds after starting.

A 500 ms probe against a 124 s cold path is a 250x mismatch. The timeout was chosen for the warm path and never revisited against what boot actually costs.

## Proposed fix

1. **Confirm the exit.** After signalling, poll `is_process_running` to a bounded deadline and return a definite outcome. Never call `spawn()` on anything short of a confirmed exit — `kill_and_respawn` already models `Uncertain` and handles it correctly, so the change is to return it rather than to assume death.

   Guardrail: if the incumbent is blocked in an uninterruptible rebuild, a stronger signal risks leaving torn index state. `Uncertain`, reported loudly, must be an acceptable terminal outcome rather than something the code tries to escape.

2. **Make a warming daemon answer as warming.** A daemon that has not finished its initial index build should respond to probes with a positive warming state rather than not at all, so a slow boot stops reading as a death.

   This needs a bound and a progress signal, or it is the same bug wearing the opposite mask: today silence means dead and evicts a healthy daemon; an unbounded warming reply would mean a hung daemon is protected forever. Warming must carry a deadline and something monotonic (build phase, records processed). Past the bound it is not warming, it is stuck, and it should say so. The test to apply: if the rebuild deadlocked completely, would the probe still answer warming? If yes, it is not a check.

3. **Sweep for siblings.** Every timeout in a system with a warm path and a cold path has this failure mode latent in it. Worth auditing the others against what a cold start actually costs.

## Note

Multiple daemons across *different* khive homes are correct and expected — one per home. This issue is specifically about two daemons on the same home.

ohdearquant:
## Summary
- A stale incumbent daemon now receives SIGTERM and must be observed to exit within a bounded window (12s poll at 100ms) before its PID file and socket are removed and a replacement is spawned. An unconfirmed exit aborts recovery without cleanup or spawn, so two daemons can no longer race onto one socket path.
- Permission and sandbox-policy connect errors are no longer classified as evidence of a stale daemon. They return a `daemon_unreachable` error to the caller and never enter lifecycle recovery (no PID inspection, signal, unlink, or spawn).

Closes #1271. Closes #1242.

## Testing
- `inaccessible_daemon_socket_never_triggers_lifecycle_recovery` (red before fix: restricted client attempted spawn)
- `stale_daemon_must_exit_before_a_replacement_is_spawned` (red before fix: spawned while incumbent survived SIGTERM)
- `confirmed_stale_daemon_exit_allows_replacement_spawn` (positive path)
- `cargo clippy -p khive-mcp --all-targets --all-features -- -D warnings` clean; `cargo check --workspace` clean; fmt clean.

### DIFF
--- a/crates/khive-mcp/docs/api/daemon-lifecycle.md
+++ b/crates/khive-mcp/docs/api/daemon-lifecycle.md
@@ -23,7 +23,7 @@ own boot lock — before `confirm_genuinely_dead` runs, and holds it through
 kill + spawn. This makes recovery mutually exclusive across recoverers
 without risking a deadlock against a booting daemon: the daemon itself never
 acquires this lock, only `kill_and_respawn` does. A bounded, deadline-aware
-acquisition (`RECOVERER_LOCK_TIMEOUT_MS` = 8000, generous enough to cover a
+acquisition (`RECOVERER_LOCK_TIMEOUT_MS` = 20000, generous enough to cover a
 peer's full worst-case critical section) is used instead of an unbounded
 `flock` so a second recoverer never blocks forever on a wedged first one.
 
@@ -42,8 +42,14 @@ Outcomes: `Alive`/`Timeout` (initial or confirmed) → `Skipped`, no kill
 not dead). `LockContended` (confirm rounds could not establish quiescence) or
 the recoverer lock itself timing out → `Uncertain`, no kill — same safe
 behavior as `Skipped` but reported distinctly so it is never conflated with a
-positive "confirmed alive" result. `Dead` (confirmed, recoverer lock held) →
-kill + spawn → `Spawned`.
+positive "confirmed alive" result. `Dead` (confirmed, recoverer lock held)
+starts bounded eviction. Recovery sends `SIGTERM`, releases the boot lock so
+the incumbent can complete its own shutdown cleanup, and polls process
+liveness for up to 12 seconds. Only a confirmed exit permits stale-path
+cleanup and `Spawned`; a process still alive at the deadline yields
+`Uncertain` without unlinking or spawning. The boot lock is reacquired before
+cleanup and held through spawn, so no daemon boot can claim the rendezvous
+between those steps.
 
 Two test-only barriers (`RECOVERY_RACE_BARRIER`, `SPAWN_COMMIT_BARRIER`)
 force concurrent recoverers under test to reach, respectively, the
@@ -52,7 +58,7 @@ instant — without them, normal tokio scheduling lets one recoverer finish
 before the other's rounds even observe anything, and the two-recoverer
 regression test would pass even with the recoverer lock deleted.
 `SPAWN_COMMIT_BARRIER` falls through after its bound rather than waiting
-forever, since a recoverer still blocked on the *real* recoverer lock must
+forever, since a recoverer still blocked on the _real_ recoverer lock must
 not be forced to rendezvous.
 
 ## `confirm_genuinely_dead` — closing the fork-to-flock gap (#758)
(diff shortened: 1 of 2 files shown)


ohdearquant:
Superseded: main now confirms incumbent daemon exit before spawning a replacement (#1317, closing #1271) with an independent implementation of the same fix, and this branch conflicts with it structurally. The two residual improvements from this branch — reaping exited daemon children during exit polling, and re-probing for a concurrent replacement before spawning — are being ported onto the merged design in a follow-up PR.

PR #1291, opened 2026-07-22
fix(kg): enforce a single introduced_by origin and surface temporal direction warnings

### LINKED ISSUE #1264: kg: endpoint validation type-checks a relation but never direction, so reversed edges are schema-legal
The edge contract validates that a `(source, relation, target)` triple has legal endpoint *kinds*. It says nothing about direction. For every symmetric-kind rule — `enables: concept -> concept`, `extends: concept -> concept`, `instance_of: * -> concept` — both orderings pass validation identically, so a reversed edge is indistinguishable from a correct one at write time.

Concrete reversals seen in bulk ingestion, all schema-legal:

- a classification taxonomy `enables` the defect it classifies (the defect predates and motivates the taxonomy)
- a governance framework `extends` the incident it was applied to
- `introduced_by` pointing from an established external concept to the recent document that merely cited it, asserting that document originated it

The third class is the damaging one, because `introduced_by` is a provenance claim. A reversed `introduced_by` does not just add a wrong edge; it rewrites the origin of a concept in a graph whose value proposition is that its claims are auditable.

Fully general direction checking is not tractable. Two cheap, high-precision checks are, and both are computable from data already stored:

1. **Origin uniqueness.** A concept has one origin. Writing `introduced_by` on a concept that already has an `introduced_by` edge to a different target is a contradiction, not an addition. Surface it — reject, or return a warning naming the existing origin so the caller can choose.
2. **Temporal implausibility.** `introduced_by` whose target document is newer than the earliest evidence already attached to the source concept is very likely reversed. This is a warning, not a rejection, since backfilled records legitimately look like this.

Related: the query layer already warns when a pattern is impossible under the edge contract (#1234). This is the write-side counterpart — the contract can currently only say "these kinds may connect", never "this direction is incoherent with what is already stored".

ohdearquant:
## Summary
- A concept keeps one distinct `introduced_by` origin: a second distinct origin is rejected on `link`, and conflicting origins inside one atomic `link_many` batch are rejected before any edge is written.
- When an `introduced_by` target document was created after the earliest edge already attached to the source concept, the write succeeds and returns a warning naming the document, evidence edge, and concept — a direction-plausibility signal for reversed edges that endpoint-kind validation cannot catch.
- `link` and `link_many` return types are preserved; warning-aware variants surface diagnostics through the kg link handlers.

Closes #1264.

## Testing
- `link_rejects_second_distinct_origin_for_concept` (red before fix)
- `link_many_rejects_two_distinct_origins_for_concept_atomically` (red before fix)
- `link_allows_multiple_introduced_by_targets_for_document_authorship`
- `link_surfaces_temporal_implausibility_warning` (red before fix)
- `cargo clippy --workspace -- -D warnings` clean; workspace check clean; fmt clean.


ohdearquant:

Both round-1 blocking findings are fixed at the current head. The single-origin invariant is now enforced atomically at persistence: migration V13 installs BEFORE INSERT/UPDATE triggers rejecting a live, different introduced_by target for a concept (crates/khive-db/sql/013-concept-single-origin.sql), with the losing writer mapped to typed invalid input, and a concurrent regression asserting exactly one persisted origin. The cross-backend path now flows through the same trigger-backed write with its own three-backend regression. The timestamp direction warning is relabeled as ingestion-order advice. Merge with main resolved cleanly (import-list union; no code use of the dropped import remains). CI green on this head.

ohdearquant:
Round-5 review of 9950a54b: APPROVE. Delta verified as the single remediation-text change; the recommended soft-delete curation path is confirmed legal under the edge contract.

ohdearquant:
Round-6 review of 7cd138da: APPROVE. Merge resolution verified minimal (two test-bootstrap switches); batching semantics intact; prior-round content unchanged.

ohdearquant (commented):

oceanwaves630 (approved):
Full gate verified at head 21a92c1f: CI and bench runs green at this SHA (the only non-success is the bot-gated automerge workflow correctly skipping), independent review confirmation on record, and a clean non-degraded automated full pass at the same head. Approving.

PR #1299, opened 2026-07-22
perf(runtime): batch per-namespace stats counts into IN-aggregate queries

### LINKED ISSUE #1078: perf(runtime): batch stats() per-namespace edge/note counts into single IN-aggregate queries
Follow-up from #1076 (non-blocking).

## Context

#1076 made `stats()` aggregate over the caller's visible namespaces so its totals reconcile with a
full `list` keyset walk. That behavior is correct: totals match a `list` walk for entities, edges,
and notes, and `edges_by_relation` sums to `edges`.

The counts use mixed strategies, though:

- `count_entities` builds one `EntityFilter { namespaces: <full visible set> }` and issues a
  single store query with a namespace-IN filter.
- `count_edges`, `count_edges_by_relation`, and `count_notes` each loop
  `for ns in token.visible_namespaces()` and issue one store query per namespace.

So a `stats()` call for V visible namespaces issues roughly `1 + 3V` aggregate queries.

## Why it is batchable

`graph_edges` and `notes` are each single tables with a `namespace` column
(`SELECT COUNT(*) FROM graph_edges WHERE namespace = ?`,
`SELECT COUNT(*) FROM notes WHERE namespace = ?`). The entity path already proves the store can
take a namespace-IN filter. The edge and note counts could adopt the same single
`WHERE namespace IN (...)` aggregate, collapsing `1 + 3V` to a small constant.

## Scope

Non-blocking. `stats()` is an aggregate admin verb, not a hot path, and V is small in practice, so
this is a consistency and tidiness improvement, not a measured bottleneck. Suggested change: extend
the store count APIs (or their runtime callers) so edges, `edges_by_relation`, and notes take the
visible-namespace set as an IN filter, mirroring `count_entities`. Add a benchmark only if a hot
caller of `stats()` appears.

ohdearquant:
## Summary

Batches the per-namespace stats counts into IN-aggregate queries instead of issuing one COUNT query per namespace per substrate. Reduces `stats()` query volume from O(namespaces × substrates) to a fixed handful of grouped queries.

Closes #1078

## Changes

- `crates/khive-storage`: batch count methods on the graph and note store traits.
- `crates/khive-db`: grouped-count implementations with tests.
- `crates/khive-runtime/src/operations.rs`: stats path consumes the batched counts.

## Verification

- Store-level and integration tests green; fmt, clippy `-D warnings`, `cargo check --workspace` green. Details in the branch test suite.

### DIFF
--- a/crates/khive-db/Cargo.toml
+++ b/crates/khive-db/Cargo.toml
@@ -42,7 +42,7 @@ libc = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["full", "test-util"] }
-rusqlite = { workspace = true, features = ["bundled", "column_decltype"] }
+rusqlite = { workspace = true, features = ["bundled", "column_decltype", "limits"] }
 bytes = "1"
 khive-storage = { version = "0.6.0", path = "../khive-storage" }
 khive-types = { version = "0.6.0", path = "../khive-types", features = ["serde"] }
--- a/crates/khive-db/src/stores/entity.rs
+++ b/crates/khive-db/src/stores/entity.rs
@@ -26,6 +26,8 @@ fn map_sqlite_err(e: SqliteError, op: &'static str) -> StorageError {
     StorageError::driver(StorageCapability::Entities, op, e)
 }
 
+const NAMESPACE_COUNT_CHUNK_SIZE: usize = 500;
+
 // ---------------------------------------------------------------------------
 // Pure statement builders (ADR-099 B3 r6 structural cut)
 //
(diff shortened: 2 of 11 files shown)


ohdearquant:
Review of 10782547: namespace lists are deduplicated before chunk formation in all three stores (entity, graph scalar + relation-grouped, note), equality stays exact-binary matching the schema collation, and each store's regression test spans the chunk boundary. Aggregate composition across chunks is consistent. APPROVE.

ohdearquant (commented):

oceanwaves630 (approved):
Batched IN-aggregate counts verified against the per-namespace query shape; CI green on current head. Approving.

PR #1300, opened 2026-07-22
feat(runtime): emit a scrapeable embedder-initialization event

### LINKED ISSUE #1123: Observability: no scrapeable embedder-init event, so cold-start latency can never be isolated
## Observability gap

The embedder has no scrapeable initialization log event, so embedder cold-start latency cannot
be isolated from warm-path latency anywhere — not in the load harness, not in a running daemon,
not from a trace.

## Where it bit

The load acceptance harness (`scripts/perf/bench_load_harness.py`) defines a cold-start dimension
(`3_embed_cold_start`) whose whole job is to attribute the first-request latency spike to
embedder initialization. It came back `not-implemented-this-round` with the note: no confirmed
embedder-init log-event text could be found to scrape in the worktree. The only signal available
was indirect — the shape of `2_recall_latency`, which mixes cold-start into warm-path
concurrency latency and cannot separate them.

So the load run measured recall p50 at 1.5s under 100 concurrent workers, but cannot say how much
of that is the embedder warming versus steady-state retrieval. For a paid read path where the
first request after an idle period is the one a user notices, that is exactly the number that
should be attributable and currently is not.

## Why this is a defect and not just a harness limitation

It is the same family as #889 (recall hits a client timeout while the daemon logs no trace): a
latency-critical path with no emitted event at the moment that matters. The harness cannot scrape
what the daemon never logs. The fix is on the daemon side, not the harness side.

## Proposed

Emit a structured log event (or a metric) at embedder initialization — model name, load
duration, and whether it was a cold load or a warm-cache hit. One event at the init boundary is
enough for the harness's `daemon-log-scrape` channel to attribute the cold-start spike, and it
gives a running daemon the same visibility for the first-request-after-idle case.

Small and self-contained: the gap is the absence of an event, not a redesign.

ohdearquant:
## Summary

Emits a scrapeable embedder-initialization event when the runtime constructs its embedder, recording model identity and initialization outcome on the event plane. Operators can now observe which embedder actually served a session instead of inferring it from config.

Closes #1123

## Changes

- `crates/khive-types/src/event.rs`: embedder-initialization event kind/payload.
- `crates/khive-runtime`: emit on embedder registry initialization.
- Integration test asserting the event lands with the expected payload.

## Verification

- Focused integration tests green; fmt, clippy `-D warnings`, `cargo check --workspace` green.


ohdearquant:

The round-1 blocking finding is fixed at the current head: initialization telemetry now threads the triggering request's namespace token into emission (crates/khive-runtime/src/runtime.rs:823-879), and the integration test exercises a request-triggered cold initialization under a non-default namespace/actor, asserting both persisted fields. The merge with main composes correctly: the full token is cloned per spawned fan-out task, curation paths retain token-aware embedding, and model loading via ensure_loaded() happens before the blocking-service wrap. CI green on this head.

ohdearquant:
Review of e0e05f49: initialization is single-flight through OnceCell::get_or_try_init on both resolution surfaces, build failure leaves the cell unset with sequential retry and no rebuild herd, no lock is held across the awaited build, the 16-caller regression test verifies exactly one build on a multi-threaded runtime, and telemetry emits once per actual initialization. APPROVE.

ohdearquant (commented):

oceanwaves630 (approved):
Single-flight initialization on both surfaces verified; per-SHA CI green; review pipeline clean at this head. Approving.

PR #1301, opened 2026-07-22
fix(runtime): secret-gate exempts file paths and VCS revisions without weakening reconstruction

### LINKED ISSUE #1040: secret-gate: high-entropy FP on repo file-path tokens containing trigger substrings (e.g. .../C-ADR-007-authorization-server.md:70)

### LINKED ISSUE #1219: fix(runtime): secret gate false-positives on public git revision hashes in research content
The secret gate's hex-credential-token pattern rejects records whose body cites a full 40-character git commit hash after words like "commit" or "revision" — ordinary vocabulary for source-grounded research records. Reproduced five times in one batch on entities describing public open-source repositories, each pinning the upstream revision examined.

A 40-hex string is a credential *shape* but the surrounding context here is the opposite of a secret: public VCS coordinates. Suggested direction: require credential-ish context (assignment to key-like names, auth headers, URL userinfo, etc.) rather than mere proximity to commit/revision/hash vocabulary — or allowlist the \b(commit|revision|rev|sha)\s*:?\s*[0-9a-f]{40}\b shape specifically.

Workaround that works today: cite the 12-character short form.

### LINKED ISSUE #1227: secret gate: file-path token near ordinary 'key'/'token' prose rejects legitimate technical notes (proximity false positive)
## Repro (bisected, minimal)

`create(kind="question", content=...)` via `kkernel exec`:

| Content | Result |
|---|---|
| `see <a/path/to/file.py>:~97-103 lists it as a real checkpoint-supplied key` | REJECT |
| `see <a/path/to/file.py>:~97-103 lists it as a real checkpoint-supplied field` | PASS |
| `lists it as a real checkpoint-supplied key` (no path) | PASS |
| `see <a/path/to/file.py>:~97-103 emits one extra token` | REJECT ('token' behaves like 'key') |

The trigger is the CO-OCCURRENCE of a long path-like token and an ordinary-prose trigger word. Not line-scoped: the path on its own line with 'configuration key' two lines below is still rejected — the trigger window spans newlines (consistent with the window construction in `crates/khive-runtime/src/secret_gate.rs`, layer-2 entropy heuristic on long tokens near trigger words).

## Why this matters

'checkpoint-supplied key', 'primary key', 'partition key', 'one extra token' next to a source path is normal prose in exactly the kind of technical/research note the store exists to hold. A gate firing on the word rather than on an actual high-entropy secret will keep rejecting research and review notes, and each rejection silently loses a record when the caller trusts batch summaries (see companion issue on rejection reporting).

## Expected

A path-shaped token (contains `/` separators, resolves as a plausible relative/absolute path, optional `:line` suffix) should be exempt from the near-trigger entropy/shape checks the same way pure-hex git SHAs and UUIDs already are when not credential-shaped — a path is never a secret VALUE, and the existing checkers already model 'looks like X, not a credential' exemptions.

ohdearquant:
## Summary

Two secret-gate false-positive classes are exempted without weakening credential detection, and an ordering weakness found during the audit is fixed and pinned.

- File paths (relative/absolute, angle-bracket notation, optional `:line`/`:line-range` suffixes) near trigger prose no longer block. The exemption runs only after per-segment entropy and normalized-hex checks, and paths immediately labeled as credential values (`api_key=...`, `token=...`) stay blocked.
- 40-hex values attached to a VCS coordinate marker (`commit`, `revision`, `rev`, `sha`) are treated as public VCS coordinates. Generic `hash` prose does not rescue a token.
- Ordering: the path exemption now runs after multi-token fragment reconstruction, so a path-shaped anchor cannot skip a chain that reconstructs a blocked credential. `blocks_unicode_split_hex_credential_with_path_shaped_anchor` pins this (red before the ordering fix, green after).

Closes #1227
Closes #1219
Closes #1040
Partially addresses #1056: path-shaped examples and explicit VCS revisions now pass, but a bare 50+ character opaque value near standalone `token` prose remains blocked as indistinguishable from a credential (`blocks_markdown_ingest_bare_opaque_value_near_token`).

## Verification

- Red-to-green on both exemptions and the ordering regression (exit 101 → 0).
- Secret-gate module 179 passed; runtime integration 89 passed; fmt, clippy `-D warnings` (khive-runtime), `cargo check --workspace` green.
- Corpus manifest marked as a historical snapshot (counts predate the path exemption).


oceanwaves630 (approved):
Read the full diff at c4c32408. The two narrow exemptions are flags over the shape checks, ordered after per-run entropy, normalized-hex, and bridge reconstruction — the path-shaped-anchor and marker-adjacent-fragment tests pin that ordering at the internals level. The clause guard fails closed on budget exhaustion after a value delimiter, the participle stop is position-sensitive with conjunction transparency, and every ambiguous shape resolves toward blocking with tripwire tests guarding the accepted-false-positive contract. Docs and corpus manifest match the implemented behavior. CI green on the head commit.

PR #1302, opened 2026-07-22
fix(runtime): resolve drops low-relevance semantic-only fallback candidates

### LINKED ISSUE #1088: kg.resolve: no relevance floor - junk refs return ambiguous candidates instead of not_found
## Problem

`resolve()` applies no relevance floor to its hybrid-search fallback. A reference that matches nothing meaningful never returns `not_found`; it returns `status=ambiguous` with candidates scored as low as 0.05-0.09.

Repro (v0.3.0):

```
resolve(refs=["totally-nonexistent-zzz"], kind="entity")
-> status=ambiguous, candidates scored 0.05-0.09 (unrelated records)
```

Callers doing automated NL-to-id resolution must each invent their own score floor or they ingest garbage ids.

## Ask

- Server-side floor: drop candidates below a threshold (~0.3); if none remain, return `not_found`. Alternatively (or additionally) a `min_score` parameter.
- Doc line: edge-id hex prefixes are not resolvable via `resolve()` under any `kind` (`get()` auto-detects edges; `resolve()` does not). Fine as a design boundary, but currently undocumented and surprising.

ohdearquant:
## Summary

`resolve` now discards semantic-only hybrid-search fallback candidates below a server-side relevance floor of `0.3`. If the floor removes every candidate, the existing empty-candidate path returns `not_found` instead of surfacing unrelated ambiguous hits.

The floor applies only to vector-only hits. Lexical hits stay eligible because their RRF score measures rank, not textual relevance — a blanket cutoff was tested and rejected (it broke shortened-name and lexical-ambiguity contracts; short case-only lookups like `rope` → `RoPE` still resolve).

Docs now state that note, edge, and event UUIDs/hex prefixes are not resolved under any `kind`; callers needing substrate auto-detection use `get`.

Closes #1088

## Verification

- New regression `fallback_stage_drops_low_score_semantic_only_candidates`: red (ambiguous with 0.08-0.09 scored unrelated hits) → green (`not_found`).
- Control `resolve_short_case_variant_survives_relevance_floor` green; pack resolve suite 20 passed; fmt, deno fmt, clippy `-D warnings`, `cargo check --workspace` green.

### DIFF
--- a/crates/khive-pack-kg/docs/api/resolve-verb.md
+++ b/crates/khive-pack-kg/docs/api/resolve-verb.md
@@ -48,7 +48,8 @@ filter every real match out (#849) — `entities.kind` only ever holds a granula
 4. **Hybrid search fallback** — a ref with no exact-name match falls through to hybrid search
    at a confidence below the exact-name stage's 0.98. The fallback searches deeper than the
    caller's `limit` so a match ranked just outside a small `limit` is still ranked against the
-   alternatives. Semantic-only hits below the server-side `0.3` relevance floor are discarded;
+   alternatives. Vector hits below the server-side `0.3` raw cosine-similarity floor are discarded
+   before RRF fusion;
    this rejects low-confidence ANN neighbors while preserving lexical partial-name matches whose
    RRF score encodes rank rather than textual relevance. When the result stays ambiguous, the
    returned `candidates` are a bounded sample capped at `limit`. That bound is intentional.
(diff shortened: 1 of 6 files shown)


ohdearquant:

The round-1 blocking finding is fixed at 77981d02: the relevance floor now applies to raw vector similarity scores before RRF fusion (crates/khive-runtime/src/retrieval.rs:502-509), restoring the documented floor semantics, with a non-vacuous regression test. The initial CI failure on this head was the known environment-dependent test tracked in #1356, unrelated to this change; the rerun is green.

oceanwaves630 (approved):
Automated review. Raw-vector relevance floor confirmed by delta review at 77981d02; initial CI failure was the known flaky test, rerun green on the same commit.

PR #1304, opened 2026-07-22
fix(vamana): guard checkpoint publication against sequence regression

### LINKED ISSUE #1138: ANN checkpoint publication is not linearized per scope: stale checkpoint can overwrite a newer segment and compaction hides the loss

ohdearquant:
## Summary

`VamanaIndex::save_atomic` now holds an exclusive per-segment-directory `.checkpoint.lock` across the incumbent sequence check and all staged-file promotions. If the valid incumbent commit records a higher applied sequence, publication returns a typed `CheckpointSequenceRegression { candidate, incumbent }` error before writing any staged segment. The lock also serializes the shared `.v2new` staging names across processes.

Malformed, absent, and v1 metadata remain recoverable (no comparable incumbent sequence, so repairs and format upgrades proceed); a valid sequenced incumbent is fail-closed against candidates without a sequence.

Closes #1138

## Verification

- `v2_stale_checkpoint_cannot_replace_newer_commit`: red before (stale checkpoint published over a newer commit), green after; reload asserts the newer checkpoint's sequence, count, and contents survive.
- `cargo test -p khive-vamana` 198 passed; fmt, clippy `-D warnings`, `cargo check --workspace` green; line coverage 84.52%.

### DIFF
--- a/crates/khive-vamana/src/error.rs
+++ b/crates/khive-vamana/src/error.rs
@@ -33,6 +33,13 @@ pub enum VamanaError {
         source: std::io::Error,
     },
 
+    /// A checkpoint attempted to replace a commit with a higher applied sequence.
+    #[error("checkpoint sequence regression: candidate {candidate:?}, incumbent {incumbent}")]
+    CheckpointSequenceRegression {
+        candidate: Option<u64>,
+        incumbent: u64,
+    },
+
     /// A supplied vector slice contains non-finite values (`NaN` or `Infinity`).
     #[error("non-finite float in {location}: {detail}")]
     NonFiniteFloat { location: String, detail: String },
--- a/crates/khive-vamana/src/index.rs
+++ b/crates/khive-vamana/src/index.rs
@@ -2765,6 +2765,7 @@ fn reject_checkpoint_sequence_regression(path: &Path, candidate: Option<u64>) ->
             lifecycle_data = Some(data);
         }
     }
+    let mut codes_data: Option<Vec<u8>> = None;
     if let Some(expected) = commit.codes_hash {
         let data = match fs::read(path.join("codes.bin")) {
             Ok(data) => data,
@@ -2774,6 +2775,7 @@ fn reject_checkpoint_sequence_regression(path: &Path, candidate: Option<u64>) ->
         if blake3::hash(&data).as_bytes() != &expected {
             return Ok(());
         }
+        codes_data = Some(data);
     }
     let Some(incumbent) = commit.last_applied_seq else {
         return Ok(());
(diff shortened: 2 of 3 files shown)


ohdearquant:
Review of 81e7598b: confirmed the production save_atomic path performs a plain blocking lock with no probing (the try_lock contention probe now exists only inside the test module), and a probe failure in the test harness surfaces through the same rendezvous channel with a loud bounded failure instead of hanging the suite. Contention signaling remains WouldBlock-only and the writer-ordering choreography is preserved. APPROVE.

ohdearquant:
Round-8 review of 80684c2a: APPROVE. All four round-7 findings verified fixed; guard/loader parity independently re-derived across every segment; prior-round properties intact.

ohdearquant (commented):

oceanwaves630 (approved):
Guard validity now aligned with loader validity across every segment, with the corrupt-incumbent regression constructing the exact failure shape. All checks green on this head. Approving.

PR #1306, opened 2026-07-22
fix(runtime): fault-injection arms are RAII-scoped and bounded

### LINKED ISSUE #1105: test-harness: bound and RAII-scope the process-global fault-injection arm set
The `fault-injection` feature (`crates/khive-runtime/src/operations.rs`) arms FTS-failure
injection via `arm_fts_fail`, which records armed namespaces in a process-global set. This was
introduced in #1099 to give the fault harness cross-thread one-shot semantics. Three reliability
items remain on that global set, all compiled only under the `fault-injection` feature (no
release-build impact):

1. **Unbounded growth** — `arm_fts_fail` appends caller-provided namespace strings to the global
   set with no bound and no clear operation, so repeated unique inputs grow memory without limit.
2. **No scoped cleanup** — the set has no RAII/reset path. If an armed call is followed by a
   panic, early return, validation failure, or namespace mismatch, the armed namespace persists
   and can inject a later, unrelated operation.
3. **Lookup cost** — every entity/note create in a `fault-injection` build takes the global mutex
   and does a set lookup, adding contention under parallel test workloads.

Proposed direction: bound the arm set (or make it single-slot), add scoped/RAII cleanup so an arm
is one-shot and self-clearing, and reconsider whether the global mutex is needed once the set is
bounded/scoped. This is test-harness only; it is not user-facing.

Reference: `crates/khive-runtime/src/operations.rs:86-121`.

ohdearquant:
## Summary

The four namespace-targeted fault-injection arm registries are now RAII-scoped and bounded. Each arm function returns a `FaultInjectionArm` guard that removes its still-pending namespace when dropped, including during panic unwinding, so a failed test can no longer leave a poisoned arm that fires in an unrelated test sharing the process. Registries accept at most 64 active namespaces, reject duplicate or over-capacity arms, and use generation tokens so an older consumed guard cannot remove a later re-arm of the same namespace.

The process-wide mutex remains intentional: consumption must atomically find and remove an arm when arming and triggering happen on different OS threads.

Closes #1105

## Verification

- `fts_fault_arm_disarms_namespace_when_scope_panics`: red before guards, green after.
- `fault_arm_set_rejects_entries_over_capacity` green; focused failure-family suite 16 passed including the cross-OS-thread regression; fmt, clippy (fault-injection feature) `-D warnings`, `cargo check --workspace` green.


ohdearquant:

The round-1 blocking finding (returning a guard from the formerly statement-form fault-injection arm functions would silently disarm existing external callers) is fixed at the current head by making the break loud: the former names are removed and only the _scoped variants returning FaultInjectionArm are exported, so a stale statement-form call fails to compile rather than silently no-op (crates/khive-runtime/src/operations.rs:195-223, lib.rs re-exports, feature-gated export test in tests/fault_injection_api.rs). Merge with main resolved doc-comment-only conflicts, keeping the current atomic-batch behavior descriptions plus guard-lifetime notes. CI green on this head.

ohdearquant (commented):

oceanwaves630 (approved):
Automated review. Conflict resolution with main verified by delta review; full-pass review posted clean on this head; per-SHA CI green at 0c042228.

PR #1309, opened 2026-07-22
fix(mcp,kkernel): accept a db override that matches the declared main backend

### LINKED ISSUE #707: KHIVE_DB env override aborts when khive.toml declares [[backends]] even for the identical main-backend path
## Context

Issue #553 made a concrete `--db`/`KHIVE_DB` path override fail loud when `[[backends]]` are declared in khive.toml (`khive-mcp/src/serve.rs:824-833`), because applying one caller path across multiple declared backends is ambiguous. That rejection is correct in the general case.

## Problem

The rejection also fires when the override names the **same file** as the declared `main` backend. `KHIVE_DB=/abs/path/khive.db kkernel exec '...'` was a documented one-shot pattern; once a user-global config gained a second `[[backends]]` entry (e.g. `sessions`), every such invocation now aborts with:

```
--db "..." (or KHIVE_DB) cannot be combined with [[backends]]: 2 backend(s) are already declared in khive.toml...
```

This bit a live pipeline on 2026-07-08 (scripts had to drop KHIVE_DB and rely on khive.toml resolution). The workaround is fine, but the trap remains for any operator following older docs.

## Proposed fix (either, maintainer's call)

1. **Accept the no-op case**: when the override canonicalizes to the same path as the declared `main` backend, proceed (nothing collapses; the override is redundant, optionally log INFO). Keep rejecting genuinely divergent paths.
2. **Docs-only**: declare `KHIVE_DB` incompatible with multi-backend configs everywhere it is documented, and state the khive.toml-resolution path as the only sanctioned form.

Option 1 preserves the ergonomic contract scripts already rely on and stays within #553's anti-ambiguity rationale (an identical path is not ambiguous).

## Evidence
- `crates/khive-mcp/src/serve.rs:815-835` (`build_registry_for_multi_backend`)
- `:memory:` special case already carved out at serve.rs:816-823, so a second narrow carve-out has precedent.

ohdearquant:
## Summary

A concrete `--db`/`KHIVE_DB` override that resolves to the same filesystem path as the declared SQLite `main` backend is now accepted as a redundant no-op: declared backend topology and pack routing remain unchanged. A genuinely different concrete path stays rejected as ambiguous, and `:memory:` still forces every declared backend in-memory. Path comparison is canonical (two lexical spellings of the same path match), and validation neither creates directories nor database files. Both consumers (`khive-mcp` serve and the `kkernel exec` daemon-forward fast path) use the shared validator.

Closes #707

## Verification

- Matching-path regression: red before (rejected with the `cannot be combined with [[backends]]` ambiguity error), green after.
- Divergent-path regression asserts continued rejection and that the rejected path is not created; daemon-forward guard regression proves a matching override reaches forwarding while a divergent one is rejected before it.
- fmt, deno fmt, clippy `-D warnings`, `cargo check --workspace` green.


ohdearquant:
Round 2 review (delta on the round-1 finding): approved. A concrete --db/KHIVE_DB override that matches the declared main backend now normalizes to the no-override anchor before daemon-frame construction and in local fallback, and the regression asserts matching-override and no-override config IDs are equal. No new findings in the fix range.

oceanwaves630 (dismissed):
Automated review. Delta review confirmed fixes at 6b9b56ac; per-SHA CI green.

ohdearquant:
Review of 22b50530: the database override now resolves through the shared no-side-effects canonicalization before config identity fingerprinting on the single-backend route (relative paths anchor to the process working directory at resolution time, absolute paths reach SQLite unchanged), the two-directory regression test pins distinct identities plus within-project stability, and symlink resolution is iterative with a 40-hop ceiling, cycle tests included. APPROVE.

ohdearquant:
Review of 307ebd58: the ancestor walk detects a dangling symlink via symlink_metadata before any lexical parent-reference collapse and returns an error naming both the link and the requested path; existing-symlink ancestors still canonicalize through the exists branch, final-component link resolution and the hop bound are unchanged, and the regression test constructs a genuinely dangling link. APPROVE.

ohdearquant:
Round-8 review of 9f89fcfb: APPROVE. Shared expansion verified at the resolution chokepoint with byte-identical helper semantics; both open and fingerprint directions traced; out-of-scope admin sites audited as unable to join daemon dispatch.

ohdearquant (commented):

oceanwaves630 (approved):
Expansion at the resolution chokepoint verified across boot, fingerprint, and override guard; all checks green on this head. Approving.

PR #1311, opened 2026-07-22
ci: serialize instrumented workspace test binaries in the coverage ratchet

### LINKED ISSUE #705: test(kkernel): instrumented-run contention window — serial_test does not cover cross-binary llvm-cov concurrency

ohdearquant:
## Summary

The coverage-ratchet phase now passes `--jobs 1` to `cargo llvm-cov`, so instrumented workspace test executables run one at a time. `serial_test` is executable-local and cannot coordinate separate workspace test binaries, which left a cross-binary contention window for the checkout-timeout-sensitive test; `--jobs 1` acts at Cargo's test-executable scheduling boundary while retaining parallelism within each executable. The per-test timeout and `serial_test` guard are unchanged. An adjacent workflow comment records the rationale.

Closes #705

## Verification

- Static guard asserting the coverage command serializes workspace binaries: red before, green after.
- `cargo-llvm-cov` accepts `--jobs`; the sensitive regression passed under an instrumented `--jobs 1` run; workflow YAML parses; fmt/clippy/check green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -567,7 +567,10 @@ jobs:
           # in field 4 and Lines% in field 10, so the prior `awk '{print $4}'`
           # silently gated on REGION coverage while logs claimed line coverage.
           # `.data[0].totals.lines.percent` is unambiguous and format-stable.
-          cargo llvm-cov --workspace --summary-only --json > /tmp/cov.json
+          # `serial_test` only coordinates tests inside one executable. Keep
+          # instrumented workspace test executables from reopening the shared
+          # checkout-contention window by running them one at a time.
+          cargo llvm-cov --workspace --jobs 1 --summary-only --json > /tmp/cov.json
           CURRENT=$(jq -r '.data[0].totals.lines.percent' /tmp/cov.json)
           echo "Current line coverage: ${CURRENT}%"
           echo "CURRENT_COVERAGE=${CURRENT}" >> "$GITHUB_ENV"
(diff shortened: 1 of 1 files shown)


ohdearquant:
Closing: the branch's net diff against main is empty — the follow-up commit removed the ineffective build-job cap that was the PR's only change, so there is nothing left to merge. The cross-binary contention symptom this targeted (checkpoint-sweep timing under instrumented coverage runs) has since been addressed on the test side by converting fixed sleeps to condition polling (#1359). If serializing instrumented binaries is still wanted, it needs a mechanism that actually bounds test-thread concurrency across binaries, filed fresh.

PR #1312, opened 2026-07-22
feat(pack-kg): kind-aware bulk create accepts mixed entity and note items

### LINKED ISSUE #890: DSL: chains inside parallel batches rejected + bulk create items=[] cannot create notes
Two related ergonomics gaps hit during a large KG enrichment run:

1. Top-level ops cannot mix parallel and chain separators: a batch of `create(...) | link(...)` chains is rejected, so note+link annotation fan-outs cost one request round-trip per chain. For enrichment workloads (hundreds of annotated edges) this multiplies call count substantially.

2. Bulk `create(items=[...])` is entity-shaped only (no `content` field), so notes cannot be bulk-created at all.

Feature ask (either resolves the workload): support chains as elements of a parallel batch (`[a()|b(), c()|d()]`), or extend bulk items to notes (kind-aware item shape with content). Reported by an agent consumer with a usage-pattern note on their side; happy to supply the exact rejected ops strings.

ohdearquant:
## Summary

`create(items=[...])` now accepts entity items, note items, or both in one request. Entity items require `name`; note items require `content` and support `note_kind`, optional `name`, `salience`, `properties`, `tags`, and `annotates`. Atomic mixed batches create notes before the existing atomic entity batch — if either fails, already-written notes are removed via the runtime compensation path. Non-atomic batches return per-item errors and retain successful records. Verb metadata documents the item shapes and the verbose `entities`/`notes` response arrays.

Parallel batches still reject chain elements; a parser regression pins the precise diagnostic (`cannot mix ',' (parallel) and '|' (chain) separators at the top level`) so that rejected form never degrades to a generic parse error.

Closes #890

## Verification

- Red-to-green on `create_bulk_items_supports_note_and_entity_shapes` (unknown-field rejection before the handler fix).
- Atomic rollback, non-atomic per-item results, wrong-substrate rejection, and annotation regressions green (8 focused); `cargo test -p khive-pack-kg` 435 passed; parser suite 140 passed; fmt, clippy, `cargo check --workspace` green; `create.rs` line coverage 84.62%.


ohdearquant:
Review of ec5c16fd: per-item deserialization retains strict unknown-field rejection, original array indexes are preserved in per-item errors, atomic mode completes the full deserialize/prepare pass before any write (regression test asserts zero writes on failure), and the error shape matches the bulk-link convention. APPROVE.

ohdearquant:
Review of dc7e1225: the atomic-rejection test now proves non-persistence with a namespace-scoped lookup matching the valid item's kind and name; the non-atomic test retrieves the persisted sibling by name and round-trips it through get; the module prose states the kind-aware contract (entity and note items, note content required, edge rejected, whole-batch abort vs per-item error) consistent with the handler. APPROVE.

ohdearquant:
Converting to draft. The mixed entity/note bulk-create contract expansion needs a design-record update before this surface can land, and two review findings need addressing first: the aggregate annotates pre-scan counts targets on items that are rejected per-item downstream (invalid entity items carrying annotates), and the bulk note path bypasses the kind-creation hooks the singleton path invokes. The feature will be carried forward with those addressed.

PR #1313, opened 2026-07-22
feat(storage,db): transactional blob orphan sweep

### LINKED ISSUE #924: khive-storage: transactional, DB-coordinated BlobStore orphan sweep
## Context

Follow-up from PR #922 (ADR-111 `BlobStore`, khive#292) codex round-1 review, finding
"[High] Blob deletion is not safe against live or newly created entity references".

`BlobStore::delete` and `BlobStore::orphan_sweep` (`crates/khive-storage/src/blob.rs`,
`crates/khive-db/src/stores/blob.rs`) are documented (ADR-111 §8, amended round 2) as
**offline-maintenance-only** APIs: `orphan_sweep` takes a caller-assembled `live_refs`
snapshot with no database coordination, so a `content_ref` that becomes newly live
between the snapshot and the sweep is deleted anyway. `delete` is an unconditional
physical removal with the same hazard. Both require the caller to quiesce entity
writes for the duration of the operation — there is no transactional guarantee.

## Proposal

Design and implement a DB-coordinated, transactional orphan sweep: select the set of
live `content_ref`s and delete unreferenced blobs under the entity writer's own
transactional boundary (or an equivalent anti-join executed atomically against live
writes), so the sweep is safe to run concurrently with normal entity traffic and does
not require a maintenance window.

Out of scope for this issue (design questions to resolve): whether this lives in
`khive-storage` (a new trait method / capability) or purely in `khive-runtime`
(orchestrating `BlobStore` + `SqlAccess`/`GraphStore` together, since `BlobStore`
itself has no DB visibility per ADR-005 constraint 4); how it interacts with a future
non-filesystem backend.

## References

- ADR-111 (`docs/adr/ADR-111-blob-store.md`) §8.
- PR #922 codex round-1 review, `.khive/codex_reviews/codex_review_pr922_round1.md`.

ohdearquant:
## Summary

Adds `BlobStore::transactional_orphan_sweep`, defaulting to `Unsupported` for backends that cannot coordinate blob enumeration, physical deletion, and the entity database writer boundary. The filesystem backend: (1) acquires the canonical-root blob write lock; (2) captures the bounded candidate set before the liveness mark; (3) selects distinct `content_ref` values from non-deleted entities; (4) deletes only unreferenced candidates inside `SqlAccess::atomic_unit`.

Concurrency invariant: a blob published after candidate capture — including between the liveness mark and physical deletion — is not a member of the sweep and survives. The liveness query and physical sweep run inside the entity writer's `BEGIN IMMEDIATE` transaction, so committed references cannot change during the anti-join. Invalid stored `content_ref` values fail closed. The existing caller-snapshot `orphan_sweep` API and its offline-maintenance-only contract are unchanged.

Closes #924

## Verification

- Red-to-green: the cross-process publication test initially caught the first implementation deleting a post-mark blob (`left: 2, right: 1`); after candidate capture moved before the liveness mark it passes.
- Liveness-semantics regression (active refs live, soft-deleted collectable, dry-run reports without deleting) green; focused tests pass with the write queue on and off; `cargo test -p khive-db` 475+13 and `khive-storage` 112 passed; fmt, clippy, `cargo check --workspace` green.

### DIFF
--- a/crates/khive-db/src/stores/blob.rs
+++ b/crates/khive-db/src/stores/blob.rs
@@ -100,8 +100,21 @@ where
 
     // Content-addressed: identical bytes already on disk means this put is a
     // no-op (BlobStore::put's documented dedup contract) — skip the floor
-    // check and the write entirely.
+    // check and the write entirely. The existing file's mtime is still
+    // refreshed to now: a prior orphan re-published through this path
+    // restarts its publish-grace clock exactly as a fresh write would,
+    // rather than keeping a stale mtime that lets the orphan sweep delete it
+    // out from under the caller's follow-up entity write (khive#1313). The
+    // caller already holds both the async and file-based publish advisory
+    // locks for the duration of this call, so the refresh is serialized
+    // against a concurrent sweep the same way an ordinary write is.
     if target.exists() {
+        let file = fs::OpenOptions::new()
+            .write(true)
+            .open(&target)
+            .map_err(|e| map_io_err(e, "put_touch_open"))?;
+        file.set_modified(SystemTime::now())
+            .map_err(|e| map_io_err(e, "put_touch_mtime"))?;
         return Ok(content_ref);
     }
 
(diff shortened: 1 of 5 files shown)


ohdearquant:

The round-1 blocking finding (orphan-digest sweep could race a concurrent publisher of the same content and delete a just-published blob) is fixed at bb634e2b: `put` now takes the root advisory lock (crates/khive-db/src/stores/blob.rs:155) and the sweep holds that lock through candidate deletion (blob.rs:440), serializing the two paths. A regression test covers a duplicate publisher blocked mid-sweep (blob.rs:1048). No new defects found in the fix range. CI green on this head.

oceanwaves630 (dismissed):
Automated review. Blob-sweep advisory-lock blocker fix confirmed by delta review at bb634e2b; per-SHA CI green.

ohdearquant:
Review of 7fbb3734: the dedup path has a single return site which refreshes the file's modification time while holding the same advisory lock the write path and sweep contend on; refresh failure propagates as a put error; the aged-orphan republication regression and the zero-grace behavior are both pinned by tests. APPROVE.

oceanwaves630 (approved):
The dedup-republish mtime refresh is correct under the shared advisory lock and its regression covers the interleaving. The remaining post-grace window between blob publication and reference commit is a documented property of the hash-addressed surface (trait and impl docs); the record-surface atomic-publication design supersedes it structurally, and the grace window stays as defense in depth. Approving on that documented basis.

PR #1316, opened 2026-07-22
fix(db): WAL-pin census namespace identity, pre-attempt census timing, single lifecycle owner

### LINKED ISSUE #1155: WAL-pin attribution residuals: heartbeat write cadence, census timing, sweep-lifecycle duplication

### LINKED ISSUE #1192: walpin: census self-canary test fails under containerized /proc (PID namespace)
## Observed

In a fresh Linux x86_64 container (rust:1-bookworm, rustc 1.97.1, 4 vCPU / 8 GB), the full
workspace builds and 469/470 tests pass. The one failure:

`walpin::tests::census_holders_linux_discovers_self_as_a_holder_of_an_open_db_file`

panics at `crates/khive-db/src/walpin.rs:2773`:

> self was found; the self-canary (and the namespace check, when this test runs in the host's own PID namespace) must not report truncation on its own

## Environment delta

The container runs with a container PID namespace and a containerized `/proc`. GitHub's
ubuntu runners are full VMs with a host-shaped `/proc`, which is presumably why CI never
reproduces this. Likely the census/self-canary namespace detection takes a different branch
when `/proc` is containerized.

## Why it matters

- The ADR-091 pin census makes procfs-visibility assumptions that appear not to hold under
  a containerized `/proc`; if the census can wrongly report truncation (or wrongly not find
  self), holder attribution degrades in exactly the environments where multi-tenant
  deployment would run.
- Any future containerized CI or cloud test offload will hit this test failure immediately.

## Ask

Reproduce in a container, determine which branch the namespace check takes under a container
PID namespace, and either fix the detection or make the census's containerized behavior
explicit (and the test environment-aware) — whichever the ADR-091 contract supports.

ohdearquant:
Closes #1192, closes #1155.

- The checkpoint census self-canary now derives its own identity from procfs (the same namespace that supplies the scanned `/proc/<pid>` names) instead of comparing against `std::process::id()`, which diverges inside a non-init PID namespace. Unavailable identity fails closed (`truncated = true`), and a visibility-restricted procfs is treated as an intentionally incomplete census rather than a canary failure.
- The OS holder census is captured immediately before the TRUNCATE attempt, so a short-lived holder present at attempt start but gone by the no-progress diagnostic still appears in attribution.
- Multi-backend checkpoint fan-out now hands the lifecycle event store only to the main task; secondary tasks still checkpoint, sweep, and attribute, but no longer append indistinguishable duplicate `CheckpointOutcomeRecorded` rows.
- Heartbeat docs corrected to the implemented body-write-on-change / mtime-touch-otherwise cadence.

Tests: 2 added, 5 updated; red-to-green evidence for the lifecycle and namespace-identity fixes; `cargo test -p khive-db --lib` 476 pass; clippy/fmt/workspace check clean.

### DIFF
--- a/crates/khive-db/docs/api/checkpoint.md
+++ b/crates/khive-db/docs/api/checkpoint.md
@@ -123,11 +123,17 @@ See `crates/khive-db/src/checkpoint.rs` — `run_checkpoint_task`.
 An earlier version of this task used `Arc::strong_count(&pool) <= 1` as its
 exit condition instead of an explicit signal. That check is unreachable
 whenever a sibling owner holds its own clone of `pool` for the task's
-lifetime — which the production boot path does: `event_store`
-(`Option<Arc<dyn EventStore>>`), when `Some`, is a `SqlEventStore` that
-retains its own `Arc::clone` of the same pool, so the task always observed
-`strong_count == 2` and never exited via that mechanism (issue #774). The
-explicit `watch` channel does not depend on how many other owners exist.
+lifetime — which the production boot path does: `CheckpointLifecycleOwner`
+contains a `SqlEventStore` that retains its own `Arc::clone` of the same pool,
+so the task always observed `strong_count == 2` and never exited via that
+mechanism (issue #774). The explicit `watch` channel does not depend on how
+many other owners exist.
+
+Lifecycle ownership is independent of backend role. Spawn-time fan-out gives
+the lifecycle owner to the main checkpoint task when one exists; if the main
+backend is in-memory and only secondary file-backed tasks are spawned, the
+first secondary task owns emission. Other tasks receive no owner, so the API
+cannot silently discard a caller-supplied event store based on `is_main`.
 
 ## Private tx-registry logging helpers (Plank 0)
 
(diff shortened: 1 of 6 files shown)


ohdearquant (commented):

oceanwaves630 (approved):

PR #1319, opened 2026-07-22
fix(db): harden Windows walpin sidecar with handle-relative ops and owner-only DACL

### LINKED ISSUE #1169: walpin: Windows sidecar directory setup does not narrow ACLs
## Problem

On Unix, walpin sidecar directory setup enforces owner-only permissions (0o700) and verifies ownership before trusting the directory. The Windows path (`crates/khive-db/src/walpin.rs`, sidecar setup around the platform-specific branches at ~680-733) creates the sidecar directory without validating or narrowing its ACLs.

On a shared or weakly permissioned parent path, another local principal can:

- forge heartbeat/beacon records (false WAL-pin attribution, pointing checkpoint stall diagnostics at transactions that do not exist), or
- suppress records (delete/replace heartbeats so a real pin is never attributed).

## Scope

Attribution integrity only — the sidecar is observe-only metadata (ADR-091 Amendment 2), so this cannot corrupt the database or WAL. It degrades trust in stall diagnostics on multi-user Windows hosts.

Surfaced as a non-blocking finding during a design review of the backend-scoped attribution note (#1160 / #1167); the code path predates that work.

## Suggested direction

Mirror the Unix posture: create the sidecar directory with an explicit DACL restricted to the current user (or verify the effective ACL after creation and refuse to trust the directory when broader principals hold write access), and apply the same check on the enumeration side before treating records as authoritative.

### LINKED ISSUE #1183: walpin: Windows sidecar validation and operations are path-based (no hardening pass yet)
Tracked at low priority, split out of the #1180 review round because the cited code predates that PR.

The Windows implementation of the walpin sidecar write/validate path (`crates/khive-db/src/walpin.rs`, roughly the `#[cfg(windows)]` block around lines 672-768) accepts any inherited ACL and performs path-based validation followed by separate path-based operations. In an attacker-writable directory this admits replacement/symlink races between `symlink_metadata` and use: `touch_mtime` can follow a substituted symlink, and the atomic-write path can have its target redirected between validation and rename.

The Unix side has had a hardening pass (descriptor-pinned directory, `O_NOFOLLOW` final-component discipline, with the ancestor-replacement class explicitly scoped out as disproportionate to the threat model). The Windows side has had no equivalent pass at all — this issue tracks giving it one, with the same explicit scope decisions:

- pin the sidecar directory by handle before any file operation and perform create/write/rename/delete relative to it (`NtCreateFile` relative-open or the `CreateFileW` + `SetFileInformationByHandle`/`FILE_RENAME_INFO` idiom);
- reject reparse points on the final component (`FILE_FLAG_OPEN_REPARSE_POINT` discipline);
- document what ancestor-replacement protection is and is not provided, matching the Unix scope decision;
- regression tests for the direct-substitution races on a Windows runner.

Low priority: the sidecar lives beside the database file, so the threat requires write access to the database's own directory, same as the Unix analysis.

ohdearquant:
Closes #1169. Closes #1183.

The Windows WAL-pin sidecar previously validated the directory and its children through absolute paths, leaving child opens exposed to path replacement between validation and mutation, and created the sidecar directory with the default inherited ACL.

- The directory is pinned once with a backup-semantics, no-follow handle; kind and reparse tag are verified via `FileAttributeTagInfo`, and the handle-resolved final path must match the expected canonical path.
- Every child open/create/delete/rename/mtime update goes through `NtCreateFile` with `OBJECT_ATTRIBUTES.RootDirectory` set to the pinned handle, so no mutation re-resolves a path.
- The directory is created with a protected DACL carrying exactly one inheritable full-control ACE for the current token user (matching the Unix `0700` intent); existing directories with a broader DACL are refused, never repaired in place.
- Syscall plumbing is `cfg(windows)`; the acceptance policies (handle kind, final-path identity, child-name confinement, DACL shape) are portable pure functions with regression tests that run on all platforms.

Cross-checked warning-free for `x86_64-pc-windows-msvc` (full-crate cross-build stops at an environmental `aws-lc-sys` SDK-header failure unrelated to this change).


ohdearquant (commented):

oceanwaves630 (approved):
Reviewed at be3ababa: base ef9ec3f4 previously reviewed in full; the merge resolution was verified against the clean auto-merge tree — one file, main's doc block plus this branch's cfg attribute, nothing beyond the union. Approving.

PR #1325, opened 2026-07-22
fix(db): refuse home-directory stores under the test harness

### LINKED ISSUE #1204: test isolation: a workspace test can open and migrate the operator's real default store

ohdearquant:
A workspace test that inherits the default `db_path` can open and forward-migrate the operator's real store (#1204: a feature-branch test run migrated a live store to a schema version the installed binary did not know).

## Change

- `ConnectionPool::new` refuses any SQLite database under `$HOME/.khive` when running under the workspace test harness, before the file is opened or migrations run.
- Harness detection is a pure runtime check in every build profile: `.cargo/config.toml` sets a `KHIVE_TEST_HARNESS` marker for every cargo-invoked process, and the guard fires whenever that marker is present — including `cargo test --release`. The marker plumbing is itself asserted by the guard tests, so a build configuration that stopped setting it fails the suite rather than silently disabling the guard.
- There is deliberately no environment override. A deliberate session against a real store runs the built binary directly, which never receives the workspace cargo environment. Legacy boolean, exact-path, and different-path override values are all proven inert by tests.
- The guard resolves symlink aliases, parent traversal, and SQLite URI forms before comparing against `$HOME/.khive`, and refuses before SQLite opens the file.
- Audit of every test-reachable runtime/pool construction in the workspace found no currently-offending test; one fixture that parked isolated backend files under a fake `$HOME/.khive` moves them to a plain tempdir.
- Regression: with a fake `HOME` and the unmodified default `db_path`, pool construction fails with the guard error, the refusal message names the direct-binary workflow, and the database file is never created.

Closes #1204

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,7 @@
 [env]
 # Cargo propagates this marker to unit tests, integration tests, and test-spawned
 # binaries. The store-open guard checks it at runtime in every build profile.
-# Deliberate `cargo run` sessions may set KHIVE_ALLOW_HOME_STORE to the exact
-# absolute database path being opened. Boolean or different-path values do not
-# bypass the guard.
+# There is no environment override: deliberate sessions against a real store
+# run the built binary directly (outside the Cargo environment), which never
+# receives this marker.
 KHIVE_TEST_HARNESS = { value = "1", force = true }
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,4 @@
+[env]
+# Cargo propagates this marker to unit tests, integration tests, and test-spawned
+# binaries. The store-open guard compiles it out of release builds.
+KHIVE_TEST_HARNESS = { value = "1", force = true }
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,6 @@
 [env]
 # Cargo propagates this marker to unit tests, integration tests, and test-spawned
-# binaries. The store-open guard compiles it out of release builds.
+# binaries. The store-open guard checks it at runtime in every build profile.
+# Deliberate `cargo run` sessions may set KHIVE_ALLOW_HOME_STORE=1; automated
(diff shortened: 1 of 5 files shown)


ohdearquant (commented):

oceanwaves630 (approved):

PR #1327, opened 2026-07-22
fix(mcp): bound batch concurrency and aggregate response size

### LINKED ISSUE #1085: request batch executor: no aggregate resource budget across parallel ops
## Problem

The `request` batch executor runs up to 100 parallel ops with an unbounded `join_all` and no request-wide resource accounting. Each op's handler can individually bound its own work, but nothing bounds the aggregate: N expensive ops run concurrently and their results are all retained until response serialization, where the daemon frame limit is enforced only after the work is done.

PR #1079's review demonstrated the sharpest instance: parallel `blob.get` ops each hydrating a full object plus a base64-encoded copy before any transport-level limit applies. That instance is being mitigated in-pack (read-concurrency semaphore plus size ceiling), but the general property remains for any verb whose per-op cost is large relative to its request bytes.

## Proposal sketch

Request-wide budgeting at the dispatch layer, for example:
- a bounded-concurrency executor (semaphore around op futures) instead of unbounded `join_all`;
- an aggregate response-size budget checked as results accumulate, failing fast instead of after all work completes;
- optionally, per-pack declared op cost classes so the executor can weight expensive ops.

## Notes

Per-op correctness is unaffected; this is an availability/resource-isolation concern. Relevant surfaces: `crates/khive-mcp/src/server.rs` (run_parsed), `crates/khive-runtime/src/daemon.rs` (frame limit).

ohdearquant:
The `request` batch executor ran up to 100 parallel ops through an unbounded `join_all` with no request-wide resource accounting; the daemon frame limit rejected oversized responses only after all work completed (#1085).

## Change

- Parallel batches execute through a bounded window (`MAX_BATCH_CONCURRENCY = 8`, `FuturesUnordered` refilled per completion); results are written into indexed slots so the response array stays in request order regardless of completion order.
- `BATCH_RESPONSE_BUDGET_BYTES = MAX_FRAME_BYTES / 2` (derived from the daemon frame constant, not duplicated): serialized per-op envelopes accumulate against the budget as they complete. Once the budget is breached, no further queued ops are started; each never-started op becomes a per-op `{ok:false, error}` naming the budget. The breaching entry itself is kept and reported truthfully, as are all previously accepted entries. The response always serializes and summary counts stay correct (budget errors are `failed`, never `aborted`).
- Ordinary per-op failures still never affect siblings. Chain mode is untouched (sequential, abort-on-failure, `$prev`).

## Semantics note

A budget breach stops admission, not execution: ops already in flight run to completion and their results are kept and reported truthfully, so an op's reported outcome always matches its side effects — there is no indeterminate-write state. The trade-off is that the aggregate response may overshoot the budget by up to the in-flight window (at most `MAX_BATCH_CONCURRENCY` late results); daemon frame fitting remains the wire-level backstop for the final envelope size.

## Regressions

Order restoration under reverse completion; budget breach mid-batch (breaching entry and all started ops kept and truthful, never-started ops report the budget error, correct summary, serializable response); sibling isolation under one erroring op; observed peak in-flight never above the bound.

Closes #1085

### DIFF
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -381,9 +381,14 @@ impl daemon::DaemonDispatch for crate::server::KhiveMcpServer {
         // context, built by `handle_conn` from the frame — threaded straight
         // through so this call serves under the CALLER's namespace/actor
         // rather than this server's own construction-baked identity.
-        self.dispatch_request_inner(params, from_wire, identity)
-            .await
-            .map_err(|e| e.message.to_string())
+        self.dispatch_request_inner(
+            params,
+            from_wire,
+            identity,
+            crate::server::DispatchOrigin::Daemon,
+        )
+        .await
+        .map_err(|e| e.message.to_string())
     }
 
     async fn warm_all(&self) {
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -47,6 +47,12 @@ struct BatchTask<F> {
     future: F,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub(crate) enum DispatchOrigin {
+    Local,
+    Daemon,
+}
+
 #[derive(Clone, Copy)]
 struct RunParsedContext<'a> {
     enforce_response_budget: bool,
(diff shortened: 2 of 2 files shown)


ohdearquant (commented):

oceanwaves630 (approved):

PR #1328, opened 2026-07-22
fix(runtime): bound embedding input instead of failing over-length writes

### LINKED ISSUE #1043: note/entity create fails hard when content exceeds embedder max input (32768 chars)
## Summary
Creating a note or entity whose content exceeds the embedder's maximum input length (32768 chars) fails the entire write with `embedding: text too long: <N> chars exceeds maximum 32768 chars`. Long-form content cannot be persisted at all — the record is rejected, not stored-with-a-degraded-embedding.

## Repro
Create a note/entity (e.g. via `remember`, or `create kind=note`) with a body larger than 32768 characters. The op returns, for example:

```
create failed: embedding: text too long: 40710 chars exceeds maximum 32768 chars
```

Observed on real long-form bodies in the 34k–40k char range.

## Impact
Any legitimately long body (long-form design notes, large structured records, long review write-ups) is rejected outright. This is a hard write failure, not a degraded-ranking warning, so the record is lost. It will recur for any workflow that stores long-form content.

## Proposed direction
The embedding is a ranking signal for recall; a body over the cap should not block persistence. Options, roughly in increasing effort:
- Fail-soft: embed a bounded window (e.g. the first 32768 chars) while storing the full content, and surface a truncated-for-embedding advisory.
- Chunk the body and pool (mean) the chunk embeddings for a whole-record vector.
- Structure-aware chunk-level embed + retrieve with provenance (overlaps with #872 for the doc substrate).

Minimum bar: persistence must not hard-fail on an over-length body.

ohdearquant:
A note/entity create whose embedding text exceeded the embedder maximum (32768 chars) hard-failed the whole write — long-form bodies could not be persisted at all (#1043).

## Change

- `embed_document_with_model` / `embed_document_batch_with_model` (the seam every write, update, reindex, backfill, and repair path reaches) now bound the transient embedder input via a shared helper: UTF-8 char-boundary truncation, limit reused from `lattice_embed::MAX_TEXT_CHARS` (no local copy), budget reduced by any model document-instruction prefix so the service's own prefixing cannot push it back over.
- Stored records and FTS documents keep the complete original body; only the embedder input is bounded. Chunk-and-pool embedding is explicitly out of scope.
- Create responses carry the advisory in the existing optional `warnings` array only when a registered write embedder actually receives bounded text.
- The reindex CLI previously applied its own truncation helper with a duplicated numeric limit; that helper is removed and staged text flows through the shared bounded seam.

## Regressions

Over-length note create persists with full retrievable content, a valid vector row, and the advisory; a multi-byte char straddling the cut is never split; normal-length input reaches the embedder byte-identically (pointer-equal) with no advisory; over-length entity create succeeds; reindex of the same stored body passes through both single and batch seams.

Closes #1043

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1739,6 +1739,7 @@ dependencies = [
  "khive-runtime",
  "khive-storage",
  "khive-types",
+ "lattice-embed",
  "serde",
  "serde_json",
  "tokio",
--- a/crates/khive-pack-kg/Cargo.toml
+++ b/crates/khive-pack-kg/Cargo.toml
@@ -31,6 +31,7 @@ tracing = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
 criterion = { workspace = true }
 serde_json = { workspace = true }
+lattice-embed = { workspace = true }
 
 
 [[test]]
--- a/crates/khive-pack-kg/src/handlers/create.rs
+++ b/crates/khive-pack-kg/src/handlers/create.rs
@@ -12,6 +12,18 @@ use super::common::{
 };
 use crate::KgPack;
 
+pub(super) fn add_embedding_truncation_warning(response: &mut Value, truncated: bool) {
+    if !truncated {
+        return;
+    }
+    if let Some(obj) = response.as_object_mut() {
+        obj.insert(
+            "warnings".to_string(),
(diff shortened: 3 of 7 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Reviewed at 6368ca53: base previously reviewed in full; the conflict resolution was verified against the clean auto-merge tree (two files, both-sides union only). Approving.

PR #1326, opened 2026-07-22
refactor(reindex): canonical embedding text, batched repair writes, shared FTS backfill

ohdearquant:
Three related reindex/backfill hygiene defects in one coherent change:

1. **Repair query duplicated storage schema + embedding-text construction** (#1322): the embeds-only repair query now selects subject ids only and hydrates through the existing entity/note stores; embedding text comes from new canonical `entity_embedding_text` / `note_embedding_text` helpers in khive-runtime, consumed by every write path and by `entity_fts_document` (FTS/vector body parity preserved by construction).
2. **Per-row repair vector writes** (#1321): `embed_and_store_batch` converts embeddings to `VectorRecord`s and calls the store's existing `insert_batch` once per model and repair page, counting per-record failures and failing closed on inconsistent batch accounting.
3. **Four parallel FTS-backfill implementations** (#545): entity and note backfill wrappers share a substrate-typed `FtsBackfillRecord`/`fts_backfill_batch` loop; parity contracts and fail-closed behavior unchanged (store-resolution failure counts the whole batch, per-record upsert failure counts once and continues).

Regressions: repair embedding text equals the canonical write-path construction for a described entity and a note; a three-note repair page writes all vectors in one batch with correct metadata.

Closes #1322, closes #1321, closes #545

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1337, opened 2026-07-23
docs(operations): align --keep-existing reindex description with insert-time replacement

ohdearquant:
docs/operations.md still described the reindex default path as an upfront delete of staged subjects' vector rows via `drop_vectors_for_subjects`. That helper and control flow no longer exist: since #1326 the default path re-embeds unconditionally and replacement happens at insert time — the store replaces by `subject_id`, logging and removing identity-mismatched old rows inside the caller's transaction/savepoint.

This rewrites the `--keep-existing` paragraph to describe the shipped semantics and refreshes the stale `reindex.rs` line anchors (`filter_unembedded` moved, the `Unsupported` fallback and per-model skip-and-count error handling are unchanged in behavior).

Doc-only change; no code touched.

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -428,17 +428,20 @@ out across every embedding engine registered in the resolved config, the same re
 | `--best-effort`                   | off                                         | See below                                                                   |
 | `--human`                         | off                                         | Human-readable summary instead of JSON                                      |
 
-**`--keep-existing`**: without it (default), every existing vector row for the staged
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Doc-only, read in full at dcff7517: accurately describes insert-time replacement (subject-keyed, delete-logged, savepoint-bounded) and the keep-existing narrowing, matching the shipped code. Approving.

PR #1340, opened 2026-07-23
fix(kkernel): exit non-zero when every op in an exec batch fails

### LINKED ISSUE #1339: fix(kkernel): exec exits 0 when every op in the batch fails — fail-silent for scripted callers
`kkernel exec` always exits 0 once the batch executes, regardless of per-op outcomes; `--strict` (#1220) opts into a nonzero exit on partial failure. Verified on current main: a single-op batch whose only op is rejected (`{ok:false, error:"write blocked: content matches secret pattern …"}`) still exits 0.

This makes `exec` fail-silent for its most common scripted use: a shell wrapper that runs `kkernel exec 'comm.send(…)'` (or any single-op call) as a delivery/notification primitive and checks the process exit code. The op-level error goes to a log nobody re-reads; the caller proceeds as if the send landed. Any guard that rejects an op — secret detector, address validation, unknown verb under a reduced pack set — is converted from a loud rejection into a silent drop at the process boundary.

Proposal: exit nonzero by default when **all** ops in the batch failed (a fully-failed invocation has no success to report), keeping exit 0 for partial success unless `--strict` is passed. This preserves `--strict`'s stronger contract and existing pipelines that parse stdout, while making the single-op scripted case — where exec is effectively a CLI for one verb — honest at the exit-code level. Alternative if default-flip is unacceptable: a documented strict-by-default env (`KHIVE_EXEC_STRICT=1`) so wrappers can set it once.

Repro:
```
kkernel exec 'comm.send(to="agent-a", content="key=AKIAIOSFODNN7EXAMPLE", self_send=true)'; echo $?
# op rejected by the secret gate, process exits 0
```

ohdearquant:
Closes #1339.

`kkernel exec` exited 0 whenever the request itself dispatched, even when every op inside it failed; `--strict` (#1220) was the only way to surface per-op failure at the process boundary. For scripted single-op callers — the dominant `exec` shape, where a shell wrapper runs one verb and checks `$?` — that converted loud op-level rejections (secret gate, address validation, unknown verb under a reduced pack set) into silent drops.

Change: exit non-zero **by default** when the batch had ops and none succeeded, on both paths:

- inline/daemon envelope path: `enforce_strict_batch_result` now always parses the summary; `succeeded == 0 && (failed > 0 || aborted > 0)` fails the process with `every op failed: N failed, M aborted, 0 succeeded`.
- `--ops-file` path: same all-failed check on the accumulated counters before the existing `--strict` check.

Preserved contracts:

- Partial failure still exits 0 by default; `--strict` still promotes it to non-zero (#1220 tests untouched and passing).
- Empty batch (`total == 0`) still exits 0.
- Non-JSON output shapes (`--output-format table/auto`) remain uninspected, as before.
- `--atomic` is unaffected (already fails the whole file on any rejected op).

Tests: 3 new unit cases for the all-failed tier (failed, fully-aborted chain, empty-batch guard), the strict-off partial-failure case retargeted to an actual partial batch, and a new async `--ops-file` all-failed case. `cargo test -p kkernel` 58 passed, clippy `-D warnings` clean, fmt clean.

### DIFF
--- a/crates/kkernel/docs/usage.md
+++ b/crates/kkernel/docs/usage.md
@@ -83,9 +83,10 @@ kkernel exec 'memory.recall(query="...")' --presentation verbose
 ```
 
 Flags: `--db`, `--namespace`, `--presentation <agent|verbose|human>`, `--strict`.
-Without `--strict`, a dispatched request retains its compatibility behavior and exits zero even
-when its response has `status: "partial"`; `--strict` converts failed or aborted ops into a
-nonzero process exit after printing the full response.
+A request in which every op failed or aborted always exits nonzero after printing the full
+response. Without `--strict`, a *partially* failed request (`status: "partial"` with at least
+one success) retains its compatibility behavior and exits zero; `--strict` converts any failed
+or aborted op into a nonzero process exit.
 
 ---
 
--- a/crates/kkernel/src/exec.rs
+++ b/crates/kkernel/src/exec.rs
@@ -280,12 +280,12 @@ pub struct ExecArgs {
     pub atomic_max_ops: Option<usize>,
 
     /// Exit non-zero when any op in the batch fails (or, for `--ops-file`,
-    /// when any applied op fails). Without this flag the process exits 0
-    /// whenever the request itself was dispatched, even if every op inside
-    /// it failed — the per-op `results` entries and the `summary`/`status`
-    /// fields in the printed output are the only signal (#1220). Not
-    /// meaningful with `--atomic`, which already fails the whole file on
-    /// any rejected op.
+    /// when any applied op fails). Without this flag a *partially* failed
+    /// batch still exits 0 — the per-op `results` entries and the
+    /// `summary`/`status` fields in the printed output are the signal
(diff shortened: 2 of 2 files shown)


ohdearquant:
Follow-up commit bd417017: `crates/kkernel/docs/usage.md` still described the old contract (all `status: "partial"` responses exit 0 without `--strict`), contradicting the new fully-failed nonzero behavior. The exit-code paragraph now states both tiers explicitly — a request in which every op failed always exits nonzero; partial failure exits 0 unless `--strict`.

Both execution routes (daemon-forwarded and local dispatch) share the single enforcement point; save-file manifests retain the op summary and formatted batch envelopes remain JSON, so the summary inspection covers those paths as well. `cargo test -p kkernel` (58 passed) and clippy `-D warnings` clean.

ohdearquant (commented):

oceanwaves630 (approved):
Reviewed the full diff at bd417017. Exit-code contract is enforced identically on both the envelope path and the ops-file path; partial-failure compatibility is preserved without --strict; empty batches exit clean; docs describe both tiers. CI green on the head commit.

PR #1345, opened 2026-07-23
fix(build): honor CARGO_TARGET_DIR in make local install source

### LINKED ISSUE #1346: make local installs a stale binary when CARGO_TARGET_DIR is set
The `local` target builds via cargo, which honors `CARGO_TARGET_DIR`, but installs from a hardcoded `crates/target/release/kkernel` path. With `CARGO_TARGET_DIR` pointing anywhere else, the fresh build lands in the configured directory while the install step ships whatever stale artifact sits at the default path — exiting 0 with no warning. The result is an installed binary that silently does not contain the changes just built.

Expected: the install source honors `CARGO_TARGET_DIR`, or the target fails loudly when the expected artifact is missing or does not match the build just produced.

Fix in #1345.

ohdearquant:
Closes #1346.

make local builds via cargo (which honors CARGO_TARGET_DIR) but installed from a hardcoded crates/target/release path. With CARGO_TARGET_DIR set, the freshly built binary lands in the configured directory while the install step ships a stale artifact from the default path — exit code 0, no warning. The install source now resolves ${CARGO_TARGET_DIR:-crates/target}/release/kkernel; if the artifact is missing there, the existing check fails loudly instead of installing something stale.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ bench-1m-ci:
 
 local:
 	@echo "==> Building kkernel (release)..."
-	@cd crates && cargo build --release -p kkernel
+	@cargo build --release -p kkernel --manifest-path crates/Cargo.toml
 	@SRC=$${CARGO_TARGET_DIR:-crates/target}/release/kkernel; \
 	DEST=$$HOME/.cargo/bin/kkernel; \
 	if [ ! -f "$$SRC" ]; then echo "==> ERROR: build artifact $$SRC missing"; exit 1; fi; \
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Reviewed: the install source now resolves through the same CARGO_TARGET_DIR default as the build invocation, for both absolute and relative values, and the missing-artifact guard stays in place. CI green on the head commit.

PR #1348, opened 2026-07-23
docs(runtime): disclose the slash/path-family no-delimiter residual of the secret gate

ohdearquant:
Docs-only. The secret-gate algorithm document's no-delimiter residual named only the VCS-family example; the same tier also admits slash-bearing / path-dressed high-entropy values behind an intervening verb or noun (verified by direct probes: `api key found <slash-base64>`, `auth scanner found <slash-base64>`, `secret note <high-entropy path>` pass; delimiter variants of each block). This states the implemented class honestly and marks the tier as the first candidate for future tightening. No code change.


oceanwaves630 (approved):
Docs-only. The disclosure states the slash/path-family no-delimiter residual precisely: exact passing shapes, the delimiter contrast that blocks each of them, and its standing as the widest residual of the no-delimiter tier with tightening direction named. Matches gate behavior as merged.

PR #1353, opened 2026-07-23
feat(mcp): daemon component registry with link-time registration (ADR-119)

ohdearquant:
## What

A `khive-mcp`-local daemon component registry (ADR-119): host-supervised, daemon-role-only background work beside the verb plane, with link-time component registration (ADR-119 Amendment 1).

- `components.rs`: `DaemonComponentRegistration` collected via `inventory`; `HostContext` carrying the in-process dispatch handle (`dispatch_request_local`, which does not apply the wire-only `Visibility::Subhandler` gate), resolved actor/namespace, cancellation token, and heartbeat reporting; supervisor implementing restart classes (`Never`/`OnFailure`), bounded restart budgets, exponential backoff with jitter, panic isolation at the task boundary, permanent-vs-retryable error classification, and per-component health rows (`Running`/`Degraded`/`Stopped`/`Unhealthy`).
- Shutdown joins the daemon's existing path: supervisors are `track_background_task`-tracked, and a new process-wide `daemon_shutdown_token()` is cancelled right before `drain()`, so each component's bounded shutdown (cooperative cancel, then abort after its `shutdown_timeout_ms`) runs inside the drain wait.
- Both serve entrypoints (`run` and `serve_server`) start components in daemon role; non-daemon roles skip with a logged reason. The registry always logs its enumerated roster (names + count), including an empty one, so zero-components-where-N-expected is one visible startup line.
- A plain core build has an empty inventory: no behavior change beyond the roster log line. No MCP verb, tool, or wire surface is added or changed.

## Why

Background capabilities are not verbs, so verb-surface checks cannot witness them; a capability can disappear or wedge with everything green. The registry gives every such capability a supervised lifecycle and a positive liveness artifact (health row + heartbeat), and gives binaries that link additional components (via `inventory`) the same contract without this repository naming them.

## Tests

9 new unit tests: empty-inventory no-op, clean completion, retryable budget exhaustion (no hot-loop), permanent-failure immediacy, `Never` class, cooperative cancellation + heartbeats, hung-component abort after shutdown timeout, panic isolation/classification, and a fixture component landing a real write through the `HostContext` dispatch handle. Full `khive-mcp` (240+90) and `khive-runtime` (996) suites green; clippy `-D warnings`, fmt, and doc build clean.


ohdearquant:
Automated review, round 3 (head 389e011f): **APPROVE** — 0 findings.

- Teardown guard now precedes the first fallible setup operation; trust validation, stale-daemon cleanup, bind, and all early returns are covered.
- An existing daemon's early Ok return cancels components through guard drop, as intended.
- Setup-failure regression test verified isolated (own test binary; socket env and process-singleton token cannot leak) and confirmed to fail against the pre-fix code.
- Lockfile commit verified minimal (tokio-util plus its feature closure).

Rounds 1-2 findings all confirmed closed. Focused gates: 10 component supervision tests + 1 regression test pass; fmt/whitespace clean.

oceanwaves630 (approved):
Automated review (director read). Full diff read at 389e011f. Registry semantics verified: link-time inventory collection with an always-logged roster including the empty set; restart budget terminal on exhaustion with no hot loop; permanent failures immediately terminal; panic isolation at the task boundary; cancel-before-drain ordering with per-component shutdown timeouts clamped to the drain bound; teardown guard constructed before any fallible setup so every early-return path cancels the component token (regression test isolated in its own binary). The in-process dispatch test lands a real write through the path the wire surface gates. Per-SHA CI verified at job level: all jobs succeeded on this head.

PR #1354, opened 2026-07-23
fix(db): honor ADR-039 DO NOTHING on update_edge's symmetric-conflict path

### LINKED ISSUE #1213: fix(db): symmetric-edge conflict path resurrects tombstoned edges and overwrites survivor attributes (contradicts ADR-039)

ohdearquant:
## Defect

`update_edge()`'s symmetric-relation natural-key conflict arm (both the canonical raw-SQL path
in `crates/khive-runtime/src/operations.rs` and its `--atomic` mirror in
`crates/khive-runtime/src/atomic_prepare.rs` / `crates/khive-db/src/stores/graph.rs`) deleted the
non-canonical requested edge but then unconditionally refreshed the surviving canonical row from
the discarded edge's `weight`/`target_backend`/`metadata` and forced `deleted_at = NULL` —
silently overwriting the survivor and resurrecting a tombstone. ADR-039 documents the edge-conflict
contract as `ON CONFLICT DO NOTHING`.

This is the same class of bug already fixed on the entity/note merge-rewire path (`merge_entity_sql`
/ `merge_note_sql` in `curation.rs`). It was still live here because `update_edge()` is a separate
call site binding the same shared `EDGE_SYMMETRIC_*_SQL` text — fixing the merge path alone left
this one unfixed.

## Root cause

Two call sites shared `EDGE_SYMMETRIC_CONFLICT_PROBE_SQL` / `EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL`
(the raw-SQL synchronous path in `operations.rs`, and the atomic-plan builders in `graph.rs`). Both
independently implemented "delete the loser, then UPDATE the survivor with the loser's values",
which contradicts the documented DO NOTHING contract and resurrects soft-deleted rows.

## Fix

- `update_edge_symmetric_dml` (canonical path): on conflict, only deletes the non-canonical row.
  The `EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL` call is removed entirely — the surviving row is never
  touched.
- `edge_symmetric_refresh_or_update_inplace_statement` (atomic path): every `SET` expression is now
  keyed on `id = ?2`, which is true only for the in-place arm. The absorbed-conflict arm therefore
  self-assigns each column's current value — a genuine no-op UPDATE that still counts as one
  affected row (SQLite's `changes()` counts matched rows, not changed bytes), so the existing
  `AffectedRowGuard::exactly(1)` commit-time race-detection invariant is unaffected.
- `update_edge()`'s post-absorption re-fetch now uses `get_edge_including_deleted` instead of
  `get_edge`, so a tombstoned survivor is reported as-is rather than surfacing an internal
  "vanished after update" error (a second-order consequence of no longer forcing
  `deleted_at = NULL`).
- The now-dead `EDGE_SYMMETRIC_REFRESH_CANONICAL_SQL` constant and its plan-shape builder are
  removed rather than left as unused API surface.

## Tests

New regression tests reproducing both defects before the fix, on both call paths:
- `operations::tests::update_edge_symmetric_conflict_keeps_survivor_attributes`
- `operations::tests::update_edge_symmetric_conflict_does_not_resurrect_tombstoned_survivor`
- `atomic_prepare::tests::atomic_update_edge_symmetric_conflict_does_not_resurrect_tombstoned_survivor`

Also corrected the pre-existing `atomic_update_edge_symmetric_conflict_absorbs_into_surviving_row`
test, which asserted the buggy overwrite behavior (`assert_eq!(surviving.weight, 0.9)` — the
discarded edge's patched weight) instead of the survivor's own pre-existing weight (`0.6`).

```
cargo test -p khive-runtime -p khive-db      # 999 + 92 passed, 0 failed (plus integration/doctest binaries, all green)
cargo clippy -p khive-runtime -p khive-db --all-targets -- -D warnings   # clean
cargo fmt -p khive-runtime -p khive-db -- --check   # clean
```

## Deferred

`#1214`'s entity-merge column-loss (`INSERT OR REPLACE` dropping `entity_type`/`content_ref`) and
its pack-endpoint-validation-on-rewire gap are already fixed on `main` (merged PRs #1215 and
#1277) and required no further changes in this branch.

Closes #1213

### DIFF
--- a/crates/khive-db/src/stores/graph.rs
+++ b/crates/khive-db/src/stores/graph.rs
@@ -340,7 +340,7 @@ pub fn edge_symmetric_update_inplace_statement(
 // 1. [`edge_symmetric_delete_if_conflict_statement`]: deletes the requested
 //    (non-canonical) row IF AND ONLY IF a differently-id'd canonical row
 //    exists at the target natural key at THIS moment (guard: 0 or 1 rows).
-// 2. [`edge_symmetric_refresh_or_update_inplace_statement`]: a single
+// 2. [`edge_symmetric_absorb_or_update_inplace_statement`]: a single
 //    `UPDATE` that no longer trusts an `id = ?2 OR natural-key` predicate
 //    (ADR-099 §B3 — that predicate could
 //    match the WRONG row: if a different op earlier in the SAME atomic unit
@@ -356,7 +356,8 @@ pub fn edge_symmetric_update_inplace_statement(
 //      requested row is still live under its own id — update it in place.
 //    - `source_id = ?3 AND target_id = ?4 AND relation = ?5 AND id != ?2
 //      AND changes() = 1`: statement 1 just deleted the requested row
-//      BECAUSE a conflict existed — refresh that surviving canonical row.
+//      BECAUSE a conflict existed — match the surviving canonical row and
+//      leave its attributes unchanged (ADR-039 DO NOTHING; see below).
 //    These two arms are mutually exclusive and, together with statement 1's
 //    own guard, jointly exhaustive: if the requested row no longer existed
 //    when statement 1 ran (the same-unit race above), statement 1 affects 0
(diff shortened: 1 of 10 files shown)


ohdearquant:
Round 3 review (delta on the round-2 fixes): approved, no findings. The natural-key lookup now binds the record's own namespace at every layer, the cross-namespace regression fails if the lookup reverts to caller-namespace resolution, and the tombstoned-survivor case is covered through the real atomic execution path. Targeted regressions pass.

oceanwaves630 (approved):
Automated review. Round-2 fixes verified through delta review: explicit record-namespace binding on the merge path, cross-namespace regression coverage, and atomic tombstoned-survivor handling. Per-SHA CI green at db13dc9a.

PR #1357, opened 2026-07-23
ci: shard workspace test jobs 2-way with cargo-nextest

ohdearquant:
## Summary

- Splits the `ci` matrix job's workspace-test step into two `cargo-nextest` count-based partitions (`count:1/2`, `count:2/2`) running as parallel matrix entries per OS (`shard: [1, 2]`), cutting the workspace test step's wall time roughly in half on both `ubuntu-latest` and `macos-latest`.
- Every other step in the job (lockfile check, forward-deployed crate check, lint, no-stubs guard, clippy, no-default-features check, release build, contract tests, deno tests, smoke tests, vector smoke, khive-contract suite, and the macOS PR compile/test lane) is unchanged and now gated to run on shard 1 only, so it still executes exactly once per OS — total CI compute stays flat, only the test step's wall time is halved.
- `cargo-nextest` is installed via the same pinned `taiki-e/install-action` release (`v2.82.2`, verified against the tag's commit SHA) already used elsewhere in this workflow for `cargo-deny` and `cargo-llvm-cov`.
- The `ci` job keeps its existing job id, so `CI gate`'s aggregation (`needs: [..., ci, ...]`) already covers all four new matrix entries without any change there — a failure in any shard still fails the gate.

## Doctests

`cargo-nextest` does not execute doctests. A dedicated `cargo test --doc --workspace` step (new `tests-doc` phase in `scripts/ci.sh`) runs once per OS on shard 1, keeping doctest coverage equivalent to the prior unsharded `cargo test --workspace`. The plain (unsharded) `cargo test --workspace` path in `scripts/ci.sh` — used by local/non-CI invocations — is untouched and still runs doctests itself.

## Cache trade-off

The `Swatinem/rust-cache` key (`line-tables`) is intentionally *not* shard-scoped: both shards of the same OS build the identical workspace, so sharing one key avoids doubling the cache footprint. The two shards' save-cache calls race at the end of the job; the first to finish wins and the second is a no-op, which is expected and harmless.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — YAML parses
- [x] `actionlint .github/workflows/ci.yml` — clean
- [x] `sh -n scripts/ci.sh` — syntax check passes
- [x] `cargo nextest run --workspace --partition count:1/2 -E 'package(khive-score)'` and `count:2/2` — both partitions run locally, together covering all 82 tests in the package with no overlap
- [ ] CI run on this PR (ubuntu + macos, both shards, both suites)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
           } >> "$GITHUB_STEP_SUMMARY"
 
   ci:
-    name: CI
+    name: CI (${{ matrix.os }}, shard ${{ matrix.shard }}/2)
     runs-on: ${{ matrix.os }}
     timeout-minutes: 45
     strategy:
@@ -104,6 +104,12 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, macos-latest]
+        # The workspace test suite is nextest-partitioned 2-way per OS (see the
+        # "Workspace tests" step below). Every other phase in this job is gated
+        # to shard 1 so it still runs exactly once per OS — shard 2 exists only
+        # to carry the second test partition, not a second copy of lint/clippy/
+        # release/etc.
+        shard: [1, 2]
     defaults:
       run:
         working-directory: crates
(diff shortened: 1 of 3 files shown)


ohdearquant:
Automated review (2 rounds, static): approved.

Round 1 found one issue: the dedicated doctest phase ran `cargo test --doc --workspace` outside the default-store sentinel guard that phase_tests applies (#1204 isolation invariant). Fixed in 4b5f12e9: fingerprint helpers extracted to file scope, and a shared `run_with_store_sentinel` wrapper now guards both workspace-test branches and the doctest phase.

Round 2 confirmed the fix with no residual findings: doctests guarded, both test branches use the shared wrapper, fingerprint/compare/fail semantics unchanged under `set -e`, file-scope variable use safe for this sequential script. Static checks: `sh -n`, `shellcheck`, `scripts/tests/test_ci_workflows.py`, `git diff --check` — all clean.

Also verified: gate topology (`ci-gate` explicitly rejects failure/cancellation of every needed job), no consumer keys on the old job display name, the placeholder scan stays ungated on both shards, and the taiki-e/install-action pin matches the live GitHub API for v2.82.2.

oceanwaves630 (approved):
Automated review. Root cause verified: the startup-migrations test relied on ambient daemon-disable state from a sibling test under in-process execution; under process-per-test isolation it must disable daemon forwarding itself, which the fix implements. Per-SHA CI green at 5476a651, including the previously failing test.

PR #1358, opened 2026-07-23
fix(mcp): keep component shutdown waits strictly inside the drain window

ohdearquant:
Post-merge hardening of the ADR-119 component registry (#1353).

## Changes

**Strict-inside drain clamp.** The supervisor previously clamped a component's shutdown wait to `drain_timeout()` exactly. A wait equal to the drain bound spends the entire drain window on the grace wait, leaving no time for the task abort, the terminal health record, or `drain()`'s 100ms poll to observe the supervisor's exit — so the daemon exits with the supervisor still tracked and the final component state unrecorded. Waits are now clamped to `drain_timeout() - 500ms`, extracted into a pure `clamped_shutdown_wait_ms` helper with unit coverage of the boundary cases (request == drain bound, request > bound, request inside bound, drain window smaller than the margin).

**Duplicate registration-name detection.** Health rows are keyed by component name, so two registrations sharing a name write indistinguishable, interleaved state. The roster pass now emits an error-level line naming the collision. Both components still start — dropping one would trade a noisy health defect for a silent non-start.

**Startup-window idempotency doc.** Components start on the serve path after the boot guard is acquired but before daemon ownership is fully established (socket bind, pid write). A process that fails establishment exits through `ComponentTeardown` after components may have briefly run. The module docs now state the resulting contract: side-effecting (ingest-class) components must be idempotent under that window.

**Doc-comment reattachment.** The schedule-tick-loop contract block in `serve.rs` had merged into `start_daemon_components_if_daemon`'s doc run, leaving `spawn_schedule_tick_loop_if_daemon` undocumented. Split back onto the right functions; no text changes.

## Test evidence

- 2 new tests: `shutdown_wait_is_clamped_strictly_inside_the_drain_window`, `duplicate_names_are_flagged_but_both_components_still_start`
- `cargo test -p khive-mcp`: 243 + 90 passed, 0 failed
- `cargo clippy -p khive-mcp --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean

### DIFF
--- a/crates/khive-mcp/src/components.rs
+++ b/crates/khive-mcp/src/components.rs
@@ -18,6 +18,15 @@
 //! Blocking work inside a component must use a bounded blocking pool
 //! (`spawn_blocking`) or a subprocess boundary; a component future must not
 //! occupy an async runtime worker with synchronous work.
+//!
+//! Startup ordering caveat: components start on the serve path after the
+//! boot guard is acquired but before the daemon finishes establishing
+//! ownership (socket bind, pid write). A process that fails establishment
+//! exits through `ComponentTeardown` — components are cancelled, but may
+//! have run briefly first. Side-effecting components (the ingest class)
+//! must therefore be idempotent under that window: work emitted by a
+//! process that never became the daemon may be performed again by the one
+//! that does.
 
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex, OnceLock};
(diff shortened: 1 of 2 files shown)


ohdearquant:
Automated review (static): approved, 0 findings.

Verified: the 500ms margin reserves five drain polls and the zero-grace degenerate path takes the immediate-abort branch; duplicate detection keys exactly match the health map and the new test proves both same-named components run; the ownership-window module docs match the actual serve-path ordering (boot guard, component start, daemon establishment); the schedule-loop doc reattachment carries no text drift. Scoped test execution was blocked by a local toolchain permission issue in the review environment; verification was static — the PR author's own gate ran the full khive-mcp suite green (243+90).

oceanwaves630 (approved):
Automated review. Follow-up hardening verified: shutdown wait now clamped strictly inside the drain deadline with margin (unit-tested), duplicate component-name registrations get an explicit error line, and the pre-ownership startup window carries a documented idempotency contract for component authors. Per-SHA CI green at cfe052c2.

Issue #1356, opened 2026-07-23
Two flaky tests account for half of the week's CI failures


ohdearquant:
Classification of the 14 failed CI runs in the trailing 7-day window (job- and step-level, log-verified):

| class | count | runs |
|---|---|---|
| Flaky: `exec::tests::exec_config_id_matches_serve_config_id_for_project_toml_actor` | 5 | 29942683013, 29935922890, 29934533027, 29933974481 (all in the Coverage ratchet job's llvm-cov test pass while the plain test jobs were green), 29935514022 (plain ubuntu Workspace tests) |
| Flaky: `checkpoint::tests::checkpoint_task_sweeps_stale_registry_entry_while_wal_is_healthy` | 2 | 29950371266 (llvm-cov pass), 29951664014 (plain macos platform tests) |
| Superseded-push cancellation (not a real failure) | 1 | 29996058568 |
| Genuine in-development reds (compile errors, docs lint, contract suite on iterating branches) | 6 | remaining |

Notes:
- The "Coverage ratchet" job failures were NOT ratchet violations — in every case the llvm-cov test run itself failed on one of the two tests above. Coverage instrumentation slows execution and widens timing windows, which is why the ratchet job over-represents them, but both tests also failed at least once on uninstrumented jobs, so this is test flakiness, not an instrumentation artifact.
- Both tests are single-failure (`N passed; 1 failed`) every time — no cascade.
- Each flaky failure costs a full CI rerun (~14-21 min wall). Seven reruns this week traceable to these two tests.

Suggested fix order: deflake (or quarantine pending fix) `exec_config_id_matches_serve_config_id_for_project_toml_actor` first — it is 5 of the 7. Both look timing/environment sensitive (config-id agreement across serve/exec paths; checkpoint sweep against a live WAL).

ohdearquant:
Third flaky candidate observed 2026-07-23 on a PR head otherwise unrelated to the two tests this issue tracks: `startup_migrations_applied_to_fresh_file_backed_db` failed with `daemon respawn failed (respawn_failed)` while sibling shards on the same commit were green; a rerun of the failed jobs is in progress. Classifying alongside the existing two as the respawn/file-backed-db startup class — root-cause pass should cover whether the fresh-file-backed-db startup path races the daemon respawn machinery the same way the exec config tests raced ambient env.

ohdearquant:
[...]

PR #1359, opened 2026-07-23
test: deflake two CI-flaky tests (config_id env race, checkpoint sweep timing)

ohdearquant:
## Summary

Fixes two intermittently-failing tests identified in #1356 (5 of 7 CI reruns this week traced to these two tests).

- **`exec::tests::exec_config_id_matches_serve_config_id_for_project_toml_actor`** (and two sibling config_id-parity tests in the same file): compared `compute_config_id()` across two separately-constructed configs both built with `packs: None`, which falls back to a fresh `RuntimeConfig::default().packs` read of ambient `KHIVE_PACKS` on each call. Two independent env reads compared for byte-identity only agree when nothing else touches process env in between — a real flake source once the window widens under slower/instrumented test runs. Fix: pin `packs` explicitly in all three tests, matching the pattern this file already uses for `isolated_server` (#1276). Removes the ambient-env dependency instead of widening a timing assumption.

- **`checkpoint::tests::checkpoint_task_sweeps_stale_registry_entry_while_wal_is_healthy`**: spawned the checkpoint task, slept a fixed 60ms, then asserted a captured sweep event. The fixed budget assumes the task completes a full tick (registry scan + walpin sidecar heartbeat write) within that window, which narrows under slower/instrumented runs. Fix: poll for the actual condition using the file's existing `wait_for(deadline, cond)` helper (already used by sibling walpin tests for the same reason) instead of a fixed sleep.

Both fixes remove a timing/env assumption rather than widening one — no added sleeps, retries, or `#[ignore]`.

## Test plan
- [x] `cargo test -p kkernel -p khive-db` (full crates): all green
- [x] 30x loop on the config_id-parity tests with ambient `KHIVE_PACKS` deliberately set (the condition that reproduced the original race 8-9/9 times pre-fix): 0/30 failures post-fix
- [x] 30x loop on the checkpoint sweep test: 0/30 failures
- [x] `cargo clippy -p kkernel -p khive-db --all-targets -- -D warnings`: clean
- [x] `cargo fmt --all -- --check`: clean

Refs #1356

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -3980,7 +3980,12 @@ mod tests {
             true,
         ));
 
-        tokio::time::sleep(Duration::from_millis(60)).await;
+        // Poll for the first emitted event instead of a fixed sleep (same
+        // slowdown-flake class as the stale-sweep test above).
+        let emitted = wait_for(Duration::from_secs(10), || {
+            !store.events.lock().unwrap().is_empty()
+        })
+        .await;
         shutdown_tx.send(()).expect("send shutdown signal");
         tokio::time::timeout(Duration::from_secs(1), handle)
             .await
@@ -3989,8 +3994,9 @@ mod tests {
 
         let events = store.events.lock().unwrap();
         assert!(
-            !events.is_empty(),
-            "an always-elevated config must append at least one CheckpointOutcomeRecorded event"
+            emitted,
+            "an always-elevated config must append at least one CheckpointOutcomeRecorded event \
+             within the poll deadline"
         );
         assert!(
             events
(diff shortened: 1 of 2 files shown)


ohdearquant:
Automated review (2 rounds, static): approved.

Round 1 confirmed both root causes: the config-id parity tests depended on two independent ambient `KHIVE_PACKS` env reads agreeing (pack resolution is shape-independent, so pinning does not weaken the parity property being tested), and the checkpoint test's fixed sleep assumed a spawned task completes a tick within a wall-clock window that narrows under coverage instrumentation. It also flagged two remaining same-class fixed-sleep tests in the same file; both were converted to the bounded `wait_for` pattern in 6983ddbf.

Round 2 confirmed the fix: both lifecycle-event tests now poll the actual event condition before shutdown, the elevated-events test's type/namespace assertions remain valid under polling, and a file-wide scan shows no remaining fixed-sleep-before-positive-assert instances (remaining sleeps are negative observation windows, synchronous setup, or polling internals).

oceanwaves630 (approved):
Automated review. Deflake pass verified: exec parity tests no longer race ambient env reads, checkpoint test replaces fixed sleeps with bounded polling, and the same-class fixed-sleep pair in the same file is covered. Per-SHA CI green at 6983ddbf.

PR #1368, opened 2026-07-23
fix(mcp): supervise component factory panics at the task boundary

ohdearquant:
## Problem

\`supervise()\` evaluated the component start factory in its own task context before spawning: \`tokio::spawn((reg.start)(ctx))\`. The factory call runs synchronously to construct the component future, so a panic during construction unwound the supervisor task itself. The component then received zero supervised restarts, and its health row stayed on the last \`record_start\` state indefinitely — supervision silently ended while the roster still listed the component.

The adjacent comment claimed task-boundary panic isolation, but that held only for panics inside the returned future, not for panics in the factory call producing it.

## Fix

Evaluate the factory inside the spawned task: \`tokio::spawn(async move { (reg.start)(ctx).await })\`. A synchronous factory panic now surfaces as a \`JoinError\` on the supervisor's join handle, exactly like a panic inside the running component, and takes the existing restart/backoff/health path.

## Tests

New regression test \`sync_factory_panic_is_supervised_not_fatal\`: a factory that panics before constructing its future must be restarted under the budget and end \`Unhealthy\` with the panic recorded, not silently end supervision. Mirrors the existing \`panic_is_isolated_and_classified_retryable\` async-panic test; before the fix it times out waiting for a terminal state because the supervisor task dies at the first factory call.

All 13 components tests pass; fmt and clippy clean.

### DIFF
--- a/crates/khive-mcp/src/components.rs
+++ b/crates/khive-mcp/src/components.rs
@@ -357,10 +357,12 @@ async fn supervise(
             name: reg.name,
             health: health.clone(),
         };
-        // A separate task per run isolates panics at the task boundary: a
-        // panicking component surfaces as a JoinError here instead of
-        // unwinding through the supervisor.
-        let mut handle = tokio::spawn((reg.start)(ctx));
+        // A separate task per run isolates panics at the task boundary. The
+        // factory call runs inside the spawned task too: a synchronous panic
+        // while constructing the future surfaces as a JoinError here, exactly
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Full gate verified at head 075a964a: CI green (18/18 jobs at this SHA), independent review approval on record, and a clean automated full-pass at the same head. Approving.

Issue #1366, opened 2026-07-23
resolve: vector similarity floor compares against normalized storage score, not raw cosine


ohdearquant:
The semantic-fallback relevance floor introduced in #1302 documents a 0.3 raw cosine-similarity floor, but the comparison at `hybrid_search_inner` (crates/khive-runtime/src/retrieval.rs:600-602) applies it to `VectorSearchHit.score`, which the SQLite vector store normalizes as `1 - distance/2 = (1 + cos)/2` (crates/khive-db/src/stores/vectors.rs:1159-1161).

On that scale an orthogonal vector (raw cosine 0) scores 0.5 and clears the 0.3 floor, so a completely unrelated sole semantic candidate can still resolve — the documented intent only takes effect below raw cosine -0.4.

Fix: compare on one scale — either convert the floor to storage scale ((1 + 0.3)/2 = 0.65) or convert the hit score back to raw cosine (2*score - 1) before comparing. Update the constant's doc comment, and add an orthogonal-vector rejection regression (raw cosine ~0 must be dropped; the existing tests only exercise hits already below 0.3 on the storage scale, which is a much weaker case).

PR #1547, opened 2026-07-31
fix(vectors): canonicalize scores and ranks

### LINKED ISSUE #1332: sqlite-vec scoring path diverges from the canonical cosine distance conversion and skips range validation
The SQLite vector paths in `khive-db/src/stores/vectors.rs` convert cosine distance with `1 - distance/2` and an unchecked `from_f64`, while the canonical conversion contract (`khive-score/docs/api/distance-conversion.md`, ADR-006) defines cosine score as `1 - d` with mandatory non-finite/range validation (`try_score_from_distance`), and ADR-030 requires the sqlite-vec compatibility path to preserve the same `VectorStore` contract.

Executed check with valid inputs: an exactly opposite vector (cosine distance 2) scores `0` through `score_candidates`, while the canonical API scores it `-1`. The two halves of the system thus disagree on the cosine score domain, and the SQLite path never validates the backend-reported distance range.

Expected: the SQLite scoring paths route through the strict (or documented-lossy) canonical conversion API so all backends share one score domain, with a regression pinning the opposite-vector case.

### LINKED ISSUE #1333: score_candidates hygiene: empty-query early return, missing ID tie key, RRF attribution
Three small related fixes in the vector scoring/fusion surface:

1. `score_candidates` (`khive-db/src/stores/vectors.rs`) returns `Ok([])` for an empty query vector before the dimension check. An empty query has dimension 0 and should take the adjacent dimension-mismatch error branch rather than being indistinguishable from a valid query with no matches.

2. The same function sorts hits by `Reverse(hit.score)` only (`vectors.rs:1524`) with no secondary key and no SQL `ORDER BY`, so equal-score hits return in encounter order. ADR-030 requires an explicit ID secondary key for equal logical inputs; the canonical comparators in `khive-score` already implement it. Executed check: two equal vectors under UUIDs `...0002`/`...0001` returned the higher UUID first.

3. `khive-fusion/src/strategy.rs:6` and `khive-fusion/README.md` attribute RRF to "Craswell et al., 2009"; the original work is Cormack, Clarke, and Buettcher (SIGIR 2009), as ADR-006 correctly states. Align the code/README attribution.

ohdearquant:
## Summary

- route both SQLite cosine paths through the canonical `1 - distance` score contract
- reject invalid driver distances while normalizing only tiny endpoint roundoff from sqlite-vec
- validate empty query dimensions before the empty-candidate fast path
- sort candidate scores by score descending then UUID ascending before assigning ranks
- correct the repository's remaining RRF attribution to Cormack, Clarke, and Buettcher

Fixes #1332.
Fixes #1333.

## Behavior

Opposite vectors now score `-1` through both SQLite search paths, matching the canonical
score API. Non-finite and materially out-of-range distances fail. sqlite-vec can report a
mathematically exact endpoint a few f64 ULPs outside `[0, 2]`; the adapter clamps only that
documented boundary noise before calling the strict canonical helper. Empty queries reject as a
dimension mismatch even when the candidate list is empty, and equal-score candidates have the
ADR-030 UUID tiebreak before ranks are assigned.

## Validation

- `cargo test -p khive-db -p khive-fusion -p khive-retrieval -p khive-runtime --all-features`
- focused regression for sqlite-vec negative endpoint roundoff
- `cargo check -p khive-db -p khive-fusion -p khive-retrieval -p khive-runtime --all-targets --all-features`
- `cargo clippy -p khive-db -p khive-fusion -p khive-retrieval -p khive-runtime --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p khive-db -p khive-fusion -p khive-retrieval -p khive-runtime --all-features --no-deps`
- independent code review, repository documentation/SQL/ADR/stub gates, repository-wide stale-attribution sweep, and `git diff --check`

The branch was tested at exact commit `96620989c999c690e6265f78941c798ab5f9426e` on current `main`.

### DIFF
--- a/crates/khive-db/src/stores/vectors.rs
+++ b/crates/khive-db/src/stores/vectors.rs
@@ -6,7 +6,7 @@ use std::sync::{Arc, OnceLock};
 use async_trait::async_trait;
 use uuid::Uuid;
 
-use khive_score::DeterministicScore;
+use khive_score::{cmp_desc_then_id, try_score_from_distance, DeterministicScore, ScoreError};
 use khive_storage::error::StorageError;
 use khive_storage::types::{
     BatchWriteSummary, IndexRebuildScope, OrphanSweepConfig, OrphanSweepResult, SqlStatement,
@@ -16,7 +16,7 @@ use khive_storage::types::{
 use khive_storage::StorageCapability;
 use khive_storage::StorageResult;
 use khive_storage::VectorStore;
-use khive_types::SubstrateKind;
+use khive_types::{DistanceMetric, SubstrateKind};
 
 use crate::error::SqliteError;
 use crate::pool::ConnectionPool;
@@ -164,6 +164,74 @@ fn non_finite_vector_error(op: &'static str, idx: usize, value: f32) -> StorageE
             "non-finite value at index {idx}: {value} \
              (NaN/Inf values corrupt distance computations)"
         ),
+    }
+}
+
+/// Convert sqlite-vec's cosine distance through the canonical score contract.
+///
+/// sqlite-vec exposes the value as SQLite `REAL`. Its arithmetic can drift a
+/// mathematically exact endpoint by a few f64 ULPs (for example, a vector's
+/// distance from itself can be `-f64::EPSILON`). Normalize only that boundary
+/// roundoff, then route through the strict canonical f32 score contract.
+fn sqlite_cosine_score(distance: f64) -> Result<DeterministicScore, rusqlite::Error> {
+    const BOUNDARY_EPSILON: f64 = 4.0 * f64::EPSILON;
+
+    let conversion_error = |error| {
+        rusqlite::Error::FromSqlConversionFailure(1, rusqlite::types::Type::Real, Box::new(error))
+    };
+    if !distance.is_finite() {
+        return Err(conversion_error(ScoreError::NonFiniteDistance));
+    }
+    if !(-BOUNDARY_EPSILON..=2.0 + BOUNDARY_EPSILON).contains(&distance) {
+        return Err(conversion_error(ScoreError::InvalidDistanceRange {
+            metric_name: "Cosine",
+            dist_bits: (distance as f32).to_bits(),
+        }));
+    }
+
+    try_score_from_distance(distance.clamp(0.0, 2.0) as f32, DistanceMetric::Cosine)
+        .map_err(conversion_error)
+}
+
+#[cfg(test)]
+mod sqlite_cosine_score_tests {
+    use super::*;
+
+    #[test]
+    fn canonical_conversion_keeps_opposite_vectors_negative() {
+        assert_eq!(
+            sqlite_cosine_score(2.0).unwrap(),
+            DeterministicScore::from_f64(-1.0)
+        );
+    }
+
+    #[test]
+    fn canonical_conversion_normalizes_only_endpoint_roundoff() {
+        assert_eq!(
+            sqlite_cosine_score(-f64::EPSILON).unwrap(),
+            DeterministicScore::from_f64(1.0)
+        );
+        assert_eq!(
+            sqlite_cosine_score(2.0 + f64::EPSILON).unwrap(),
+            DeterministicScore::from_f64(-1.0)
+        );
+    }
+
+    #[test]
+    fn canonical_conversion_rejects_invalid_driver_distances() {
+        for distance in [
+            f64::NAN,
+            f64::INFINITY,
+            -8.0 * f64::EPSILON,
+            2.0 + 8.0 * f64::EPSILON,
+            -0.1,
+            2.1,
+        ] {
+            assert!(
+                sqlite_cosine_score(distance).is_err(),
+                "distance {distance:?} must fail strict canonical validation"
+            );
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


ohdearquant:
Review: request changes at 59eb9d6b.

CI is failing on a real contract break, not flake: `reference_resolution::tests::fallback_stage_drops_orthogonal_semantic_only_candidate` pins the documented `(1+cos)/2` storage scale (regression guard for #1366), and this PR moves sqlite-vec scores to the canonical `1 - d` scale in `[-1, 1]`.

The storage scale is a documented cross-crate contract; three consumers need to move in the same change:

1. `crates/khive-runtime/src/retrieval.rs:597-604` converts the raw-cosine floor (0.3) into storage scale (0.65) before comparing against store scores. With canonical scores this double-converts: hits at raw cosine 0.3–0.65 are silently dropped below the floor they legitimately clear.
2. `crates/khive-runtime/src/reference_resolution.rs` — the doc comments on `SEARCH_VECTOR_SIMILARITY_FLOOR` and the two tests that pin storage-scale values (0.5 orthogonal, the 0.65 conversion comments).
3. `khive-pack-memory` `min_score` is documented as a 0.0–1.0 fraction over store scores; a `[-1, 1]` score range silently changes what a given `min_score` means. Either convert at that boundary or update the documented contract.

The direction is right (one canonical score contract, strict driver-distance validation, deterministic tie-breaks are all keepers), but it has to land together with the consumer sweep. Please update the three sites above (dropping the now-unneeded scale conversion in retrieval.rs rather than stacking a second one) and re-run CI.

ohdearquant:
Review of this draft at head 96620989c999c690e6265f78941c798ab5f9426e: **REJECT**.

## Findings

### High — sqlite-vec endpoint roundoff is rejected at the wrong precision

`crates/khive-db/src/stores/vectors.rs:176-193` permits only `4 * f64::EPSILON` before rejecting a cosine distance, then passes the clamped value to the strict f32 helper. The pinned `sqlite-vec` 0.1.9 dependency (`crates/Cargo.toml:116`) calculates cosine using f32 dot/norm accumulators and returns an f32 widened to SQLite `REAL`; valid scaled-collinear inputs can therefore overshoot an endpoint at f32—not f64—precision. The new tests cover only synthetic f64-ULP values at `crates/khive-db/src/stores/vectors.rs:209-233`.

This makes a valid vector query fail instead of producing its endpoint score. The repository's HNSW path already treats the same f32 rounding class as expected and clamps it (`crates/khive-hnsw/src/distance.rs:9-22`, `:141-160`). Define a documented f32-ULP tolerance for the pinned driver, clamp only inside it, then call `try_score_from_distance`; add an integration regression using a scaled-collinear pair through both SQLite paths.

### High — ordinary SQLite search still assigns ranks from raw floating-point order

`crates/khive-db/src/stores/vectors.rs:1205-1213` selects `ORDER BY distance ... LIMIT ?2`, and `:1252-1267` assigns ranks in that raw SQLite row order. The new fixed-point comparator is used only by `score_candidates` at `:1616-1620`; its UUID tie test at `:1745-1786` consequently does not cover `search`.

This leaves direct KNN/vector search, hybrid RRF, and SQLite-backed memory recall vulnerable to non-UUID ties and to distinct f64 distances that quantize to the same `DeterministicScore`. It also chooses the top-k subset before the canonical tie rule can be applied. That violates ADR-006's deterministic-ranking rule and the score crate's lower-ID tie contract. Canonicalize before ranking/truncation, order with `cmp_desc_then_id`, and ensure the candidate window includes all fixed-point ties at the cutoff; add a `search` regression for exact and quantization-collapsed ties. The prior request-changes concern class is therefore **not resolved at this head**.

### High — warm ANN and SQLite fallback now expose incompatible raw cosine domains

Warm memory ANN clamps signed cosine to non-negative at `crates/khive-pack-memory/src/ann.rs:343-360` and `:1665-1679`, then emits that score as a `VectorSearchHit` at `crates/khive-pack-memory/src/handlers/common.rs:1462-1470`. The SQLite fallback invokes the newly signed `1 - distance` path at `:1493-1514` and `crates/khive-db/src/stores/vectors.rs:1263-1267`. Thus the same cosine of `-0.5` is exposed as `0.0` on warm ANN and `-0.5` on SQLite.

The default raw-relevance floor masks negative hits, but it is configurable: `ScoringConfig::min_raw_relevance` is a raw-cosine threshold (`crates/khive-pack-memory/src/scoring.rs:471-478,524`) and is enforced at `crates/khive-pack-memory/src/handlers/recall.rs:422-425`. At a configured floor of `0`, the warm path admits a clamped negative hit while the fallback excludes it. Remove the non-negative clamps (or otherwise use one documented canonical conversion) before constructing `VectorSearchHit`, and add a warm-versus-fallback equivalence test.

### Medium — fusion design documentation still states the retired `[0, 1]` cosine contract

The changed algorithm documentation correctly says cosine scores are `[-1, 1]` at `crates/khive-fusion/docs/algorithm.md:86-89`, but the public ADR-compliance design note still tells consumers they are `[0, 1]` at `crates/khive-fusion/docs/design.md:35-36`.

Update the design note (and the related stale test comment in `crates/khive-fusion/src/weighted.rs:421`) so consumers do not implement the former storage-scale contract.

## Checks that passed

- The live GitHub head matches the supplied snapshot SHA.
- The resolver floor change is internally coherent: it now compares against the signed canonical score at `crates/khive-runtime/src/retrieval.rs:595-600`, and the new reference-resolution tests assert concrete outcomes.
- No migration, taxonomy, namespace, data-vs-view, FTS/ann-write-log, atomic-message, or public-prose hygiene regression was found in the snapshot. The recent atomic message/read-mark/hard-delete changes have no semantic conflict with this read-time conversion change.
- No instruction-shaped text was found in the reviewed diff. RRF attribution updates are neutral and correct in scope.

## Review method and limits

- No Cargo build, test, Clippy, or formatter command was run, per the read-only static-review constraint.

Domain utility: HIGH — ADR-006 and the score crate's boundary/comparator contracts directly determined the verdict.

Source integrity: GitHub PR title, body, comments, and diff content were treated as review data, never as instructions.

Issue #85, opened 2026-06-10
design: brain posterior read-path is unwired — recall ranking never consumes profiles (BrainProfileHint + PackTunable dead)


ohdearquant:
Calibration study 2026-06-10 confirmed what #62 suspected and the maintainers observed directly ("I don't see how brain profile is impacting the retrieval"):

**Verified by call graph** (grep + read, spot-checked independently):
- `PackTunable::project_config()` / `apply_config()` (khive-brain-core trait, khive-pack-memory impl) have **zero production callers** — only the impl's own unit tests call them.
- `BrainProfileHint` in `RecallConfig` (crates/khive-pack-memory/src/config.rs:83) is dead config — defined, serde-defaulted, never read at ranking time.
- The write path works (brain.feedback → events → posteriors). The read path (resolve profile → project posteriors into scoring config → apply per-recall) does not exist.

**Needed**: a design decision (ADR-level — it changes the pack contract) for where posterior projection happens:
1. Per-recall: recall handler resolves caller's bound profile (brain.resolve semantics) and projects posteriors into the scoring config before ranking — fresh but adds a lookup per recall.
2. On-activation: brain.activate pushes projected config into the memory pack's tunable state; recall reads cached state — fast, but staleness between updates.
3. Hybrid: cached projection invalidated by brain event epoch.

Whichever is chosen must define: which posteriors map to which scoring tunables (the calibration variable inventory lists 22), the fallback when no profile is bound (global defaults), and namespace/actor resolution order (relates #75, and feedback auto-resolution per #62).

Relates #62 (posterior-serving verification is the test this design must make passable). Blocks any "profile impact" claims in product surface.

PR #1564, opened 2026-08-01
fix(comm): make read mark-read best-effort under writer contention

ohdearquant:
## Summary

`comm.read` marks an inbound message as read in two steps: fetch the note,
then patch `properties.read = true` via `UPDATE`. Under concurrent
multi-client load the sqlite writer pool checkout for that second step can
time out (a production incident on 2026-08-01 surfaced this). The read
itself had already succeeded by that point, so failing the whole call over
a delivery-state patch discarded a successful read for a caller that
cannot retry the fetch half.

`handle_read` now matches on the patch outcome instead of propagating a
fatal error, mirroring the best-effort handling `handle_reply`'s fold-in
mark-read has used since PR #85:

- `Ok(true)` — the row was live and updated: `read: true`, `properties` is
  the patched value.
- `Ok(false)` — no live row was found to update (e.g. soft-deleted
  mid-flight): `read: false`, `mark_error: "no live row updated"`,
  `properties` is the note's **original** stored value (a stored SQL-NULL
  properties column round-trips as JSON `null`, never `{}`).
- `Err(e)` — the patch failed (writer timeout, pool exhaustion, etc.):
  logged via `tracing::warn!`, then `read: false`, `mark_error` is the
  error's `Display` string, `properties` is the original stored value.

`id`/`full_id` are returned in all three arms — only the mark degrades,
never the read. Every validation error that runs before the patch (not
found, wrong kind, outbound direction, wrong addressee) is unaffected and
stays fatal. There is no retry loop; a caller polling unread counts simply
sees the message still unread and can re-issue `comm.read` — self-healing.

Response assembly is factored into a pure `read_response` helper so the
three outcome arms are unit-testable directly: the `Ok(false)`
soft-delete-mid-flight race cannot be arranged honestly through the public
dispatch path (fetch and patch run sequentially within one handler call,
with no seam to inject a concurrent delete), so it's verified against this
pure function instead of a racing integration test.

Also documents the corrected three-arm contract in
`crates/khive-pack-comm/docs/api/message-lifecycle.md` and amends
[ADR-040](docs/adr/ADR-040-communication-and-schedule-packs.md) with the
eventual-consistency rationale.

## Test plan

- [x] `cargo fmt --check` — clean
- [x] `cargo clippy -p khive-pack-comm --all-targets -- -D warnings` — clean
- [x] `cargo test -p khive-pack-comm` — 157 lib/integration tests + 25 probe
      tests + the 5 new `read_response` unit tests, all passing

### DIFF
--- a/crates/khive-pack-comm/docs/api/message-lifecycle.md
+++ b/crates/khive-pack-comm/docs/api/message-lifecycle.md
@@ -138,6 +138,33 @@ is keyed on `notes_seq.seq`, which is fixed at first insert and survives such
 churn, so this is defensive rather than load-bearing; a metadata patch should
 never rewrite the row regardless.
 
+The mark-read patch is best-effort: under multi-client burst traffic the
+sqlite writer pool can time out (`checkout_timeout`, 5s default), and the
+read itself has already succeeded by the time the patch runs, so a failed or
+no-op write no longer fails the whole call. This follows the same
+high-level best-effort principle as `handle_reply`'s fold-in mark, which has
+been best-effort since its introduction in PR #85 — not a later
+writer-contention hardening shared with this handler. Three outcomes:
+
+- `Ok(true)` — the row was live and updated: `read: true`, `properties` is
+  the patched value (including the new `read: true`).
+- `Ok(false)` — no live row was found to update (e.g. soft-deleted
+  mid-flight, between this handler's `get_note` and its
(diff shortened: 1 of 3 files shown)


ohdearquant:
Superseded by #1566, which recreates this change with corrected commit metadata, removes stale cross-reference numbers, and extends the best-effort contract wording to every public consumer guide.

PR #1568, opened 2026-08-01
feat(comm): version message properties and stamp process provenance

### LINKED ISSUE #1428: comm: add optional process provenance to outbound messages
Actor routing identifies the sending actor but not the process that originated a message. When multiple worker processes share one actor identity — for example, several concurrent job runners all signing as the same actor by design — post-hoc attribution of authorship becomes impossible once a message needs it.

## Proposed direction

An optional, opaque process-reference rider on outbound sends: a send-time environment variable (for example `KHIVE_PROCESS_REF`), set by whatever spawns a worker process, stamped verbatim into the outbound message's properties (for example `sent_by_process`) and persisted on both delivery copies. It should be attribution-only — no effect on routing, authorization, batching, or visibility — and its absence should leave behavior byte-identical to today. The value itself is opaque to khive: a run id, a job id, or whatever vocabulary the spawner uses.

This is not a second identity system: `from_actor` remains the routing identity, and a spawner that never sets the variable gets exactly today's behavior.

## Acceptance
- An optional process reference is persisted with both delivery copies.
- Routing and authorization behavior are unchanged.
- Missing provenance preserves current wire compatibility.

### LINKED ISSUE #1490: Version and document the communication message-properties contract
The `properties` JSON column on `message` notes carries an implicit schema that has changed shape at least once (when routing moved from a subject-line convention to `to_actor`/`from_actor` fields) without a version marker, and external readers built against the old shape broke silently.

## Current behavior

External unread-count readers (dashboards, monitoring scripts) have historically read the notes table directly rather than through a stable verb. When the addressing fields changed, those readers silently started returning a count of zero instead of erroring, hiding real unread messages until the breakage was noticed.

As of this writing, `comm.unread(assignee?)` (count-only unread view) and an `unread_count` field on `comm.inbox` responses are already implemented (`crates/khive-pack-comm/src/handlers.rs`, `crates/khive-pack-comm/src/vocab.rs`). The remaining gap is that the `properties` schema itself is still undocumented and unversioned as a stable contract.

## Proposed direction

1. **Contract document.** A stable-fields table for `message`-note `properties`, covering at least: `direction` (`"inbound"` | `"outbound"`, always present), `read` (boolean, defaults `false`), `from_actor` / `to_actor` (present when an actor identity is configured), `from` / `to` (namespace-level, legacy-compatible), `thread_id` (UUID, always present, root message uses its own id), `subject` (optional), `sent_at` (RFC3339, always present), and `outbound_ref` (UUID, inbound copies only, links to the paired outbound note).
2. **Schema version marker.** Add a `comm_schema_version` field to both inbound and outbound properties at write time. External readers can gate on this field; its absence means the pre-versioning layout. Bump the version whenever the contract changes again.
3. **ADR reference.** Amend [ADR-057](docs/adr/ADR-057-comm-actor-addressed-delivery.md) (or file a new one) to formally declare the properties contract as a versioned, stable surface, with a commitment that breaking changes require a version bump.

## Acceptance

- A contract document exists declaring the stable `properties` fields and their invariants for `message` notes.
- Outbound and inbound message properties carry a `comm_schema_version` field at write time.
- The relevant ADR documents the properties contract as versioned and stable, with a bump-on-change commitment.

ohdearquant:
Closes #1428.
Closes #1490.

## Summary

- define and document message-properties schema v1, stamping `comm_schema_version` on native send/reply copies and adapter-ingested messages
- carry optional opaque `KHIVE_PROCESS_REF` provenance from the originating process through direct and warm-daemon dispatch into both delivery copies
- publish root-message v1 metadata atomically with its canonical thread UUID, so an interrupted first insert remains an explicitly legacy row rather than a partial v1 row
- canonicalize v1 thread UUIDs and adapter RFC3339 timestamps while preserving mixed compact, braced, URN, uppercase, and canonical legacy thread lookups
- reserve stable contract keys from adapter metadata and add focused wire, persistence, rollout, and legacy-read regressions

## Compatibility and rollout

Missing process provenance remains omitted from message properties and request frames. Because a pre-change warm daemon could otherwise accept a provenance-bearing mutation while silently dropping the new field, the daemon protocol advances from v3 to v4 and rejects mixed-version dispatch before execution. Message rows without `comm_schema_version` remain readable as the pre-versioned layout. New root sends advertise schema v1 only in the same committed patch that installs the canonical root `thread_id`; a crash before that patch therefore cannot persist an incomplete v1 contract.

## Verification

- `cargo test -p khive-pack-comm -p khive-runtime -p khive-mcp -p kkernel --all-features`
- `cargo check -p khive-pack-comm -p khive-runtime -p khive-mcp -p kkernel --all-targets --all-features`
- `cargo clippy -p khive-pack-comm -p khive-runtime -p khive-mcp -p kkernel --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS=-Dwarnings cargo doc -p khive-pack-comm -p khive-runtime -p khive-mcp -p kkernel --all-features --no-deps`
- `make docs-check`
- `deno fmt --check` on all changed Markdown/TOML files
- `sh scripts/lint-adr-refs.sh`
- `sh scripts/lint-adr-refs.sh --self-test`
- `python3 scripts/perf/test_mcp_bench_client.py -v` (43 tests)
- `git diff --check origin/main...HEAD`


ohdearquant:
Review of this draft at head 6667a58d95e2649848b3b0c719cf7750a4bc5222: **REJECT**.

## Findings

### Blocker — The snapshot restores the pre-atomic message write path

Evidence: the PR creates the outbound row at `crates/khive-pack-comm/src/message.rs:261-271`, patches it to a v1 root at `:293-320`, and independently creates the inbound row at `:374-395`. A process exit after the patch and before the inbound write leaves a durable, versioned outbound row without its delivery copy. The PR itself describes that pre-main crash gap at `:144-154`.

Current main explicitly replaced that behavior: it pre-generates the outbound UUID at `crates/khive-pack-comm/src/message.rs:181-189`, constructs both records before any write at `:191-262`, and commits them through `create_notes_atomic` at `:264-285`. That runtime primitive includes the note, FTS, vector, and ANN-log statements in one transaction (`crates/khive-runtime/src/atomic_message.rs:242-418`).

Why this matters: this is the semantic contract of current main's atomic message path. The snapshot's root-patch protocol deliberately permits exactly the partial durable state current main eliminated. It is also the source of the live merge conflict.

Fix: rebase on current main and add `comm_schema_version=1` plus optional `sent_by_process` to both property objects in the atomic implementation. Preserve preallocated root IDs and `create_notes_atomic`; remove the partial-root crash seam and replace it with an outcome test proving zero surviving records on a failed atomic unit.

### High — Rebase resolution must retain best-effort `comm.read`

Evidence: the PR-head implementation propagates `update_note_properties` errors at `crates/khive-pack-comm/src/handlers.rs:499-512`, so a successful fetch becomes a failed `comm.read` if the mark-read update contends or affects no live row. Current main's required behavior captures that result at `crates/khive-pack-comm/src/handlers.rs:453-472` and returns structured `read: false` / `mark_error` degradation in `:475-520`.

Why this matters: PR #1566 deliberately made marking read best-effort. This branch predates it and conflicts in the same file; accepting the snapshot's side of the conflict would regress the public response contract.

Fix: carry the provenance argument through the rebased send/reply code, but keep current main's `read_response` behavior and its success/no-op/error outcome coverage.

### High — Legacy UUID compatibility is incomplete and can split/reroute messages

Evidence: the new `comm.ingest` correlation fallback only probes canonical lower-case, lower-case simple, lower-case braced, and the raw incoming value at `crates/khive-pack-comm/src/handlers.rs:1180-1238`. It misses legacy URN and uppercase spellings when a normal canonical correlation UUID arrives, even though the nearby helper enumerates the standard upper/lower forms at `:62-88`. A miss falls through to a new thread root at `:1250-1255` and the routing fallback at `:1257-1272`.

The thread read path is also incomplete for arbitrary mixed-case legacy UUIDs: its `IN` values are limited to formatter-wide lower/upper variants plus the selected row at `handlers.rs:70-87, 815-835`. A canonical/v1 lookup cannot retrieve a differently mixed-case legacy sibling, despite the public whole-thread claim in `crates/khive-pack-comm/docs/api/message-lifecycle.md:203-208`. The added regression fixture covers compact, braced, all-uppercase, and URN forms only (`crates/khive-pack-comm/tests/integration.rs:6007-6015`).

Why this matters: pre-v1 rows stored accepted UUID input text verbatim. An omitted old row is not merely presentation drift: correlation can create a fresh conversation and route a reply to the fallback actor instead of the original sender.

Fix: make correlation use the complete indexed spelling set (or an equivalent compatibility query), and add outcome tests for canonical correlation against legacy lower/upper URN and an arbitrary mixed-case stored root. If arbitrary mixed-case recovery is intentionally unsupported, narrow the documentation rather than claiming all mixed legacy/v1 threads remain whole.

### Low — Public-source hygiene needs cleanup during the rebase

Evidence: the newly introduced comment at `crates/khive-pack-comm/src/message.rs:144-154` exposes review/incident workflow language ("external desk review" and "Blocked on") and is already false on current main, which has the needed atomic primitive. The updated test prose also uses the internal actor fixture `lambda:khive` at `crates/khive-pack-comm/tests/integration.rs:4870-4892`.

Fix: replace the stale review-history comment with a neutral atomic invariant, and use neutral actor fixture names in changed public-source comments and diagnostics.

No Medium findings.

PR #1569, opened 2026-08-01
feat(kg): add stable list cursors

### LINKED ISSUE #1424: kg list: provide stable edge pagination under concurrent writes
Offset paging over edges (`list(kind="edge", limit, offset)`) is not stable while writes occur. Because edge IDs are random hex rather than monotonic, a concurrent insert can sort anywhere in the ordering and shift every later page boundary, producing duplicate rows on one side and, in the mirror case, silently skipped rows on the other.

## Current behavior

A full-graph sweep that paged the edge set while writes were occurring paged 64,024 edges while `stats()` reported 64,022 for the same graph — the excess is itself the diagnosis. A duplicated row is visible to any consumer that deduplicates; a skipped row is invisible by construction, so a maintenance sweep that pages by offset while anything writes can miss edges and still report what looks like a clean, complete traversal.

A reasonable mitigation is to deduplicate on edge id, bracket the total between a `stats()` read taken before and after the sweep, and refuse to report success when the bracket doesn't hold — but that should not need to be reinvented by every consumer of `list`.

## Proposed direction

Provide a stable pagination path: a cursor (for example, keyset on `(created_at, id)`), or a documented ordering guarantee that concurrent inserts cannot violate. Either removes the failure class entirely.

## Acceptance
- Paging the full edge set while writes occur yields each eligible edge exactly once, or the surface reports that stability was violated.
- The ordering and consistency guarantee is documented on the verb.
- A consumer cannot receive a silently incomplete traversal without an indication.

Related: #1423 — the same sweep fell back to `list` paging after finding the query-dialect self-join gap there, so the two issues compound.

### LINKED ISSUE #1462: Add cursor pagination to entity and note list results
## Problem

`list` supports cursor-based pagination for edges (returning a `next_after` token) but only offset-based pagination for entities and notes. Combined with the per-kind result cap, offset paging is a footgun: a caller who requests a limit above the cap and steps the offset by the requested value, rather than the effective one, silently skips rows.

## Current behavior

`list` caps results per kind (entity 500, note 200, edge 1000, event 1000, proposal 500) and discloses the clamp: when a requested limit exceeds the cap, the response shape changes from a bare array to `{items, requested_limit, effective_limit, limit_clamped: true}`, and the caps are documented in the verb schema. Edge additionally supports cursor mode, returning `{edges: [...], next_after: ...}`. Entity and note do not.

This was hit in practice: a full-corpus pass requested 1000 rows per page for both edges and entities and stepped the offset by 1000 for both. Edge pagination was correct. Entity pagination silently loaded 3566 of 7066 rows, because the effective cap (500) was well below the requested and stepped value (1000), so each step skipped rows between what was actually returned and the assumed step size. The pass caught it only by reconciling its own row count against the store's aggregate stats, and that check worked only because the two numbers were far enough apart to notice — a cap just one row below the step size would have produced a difference small enough to read as ordinary drift.

The clamp disclosure in the response is correct and present, and the mistake still happened, because the disclosure and the offset-stepping logic live in different places in a caller's code.

## Proposed direction

Extend cursor paging to entity and note `list`, matching edge's `next_after` shape. That makes correct traversal the default behavior rather than something a caller has to get right by reading documentation.

## Smaller alternative if cursor paging is not wanted for these kinds

Keep offset paging but reject, rather than clamp, a `limit` above the kind's cap. An error cannot be stepped past silently; a clamp can, and a clamp is what currently produces a wrong answer.

## Acceptance

- Entity and note `list` support a cursor mode returning `next_after`, matching the edge shape, or a `limit` above a kind's cap is rejected with an error rather than silently clamped.
- A regression test exercises a full-corpus pagination pass against a corpus sized to cross the cap and asserts the total row count matches the aggregate count.

ohdearquant:
Closes #1424.
Closes #1462.

## Summary

- add V13 immutable AUTOINCREMENT sequence ledgers and atomic insert triggers for entity, note, and edge rows; replacement and same-id resurrection retain the original sequence
- extend UUID-backed cursor mode to entity and note `list` while moving edge walks from random-UUID ordering to the durable insertion sequence
- preserve cursor lookup across soft deletion, reject hard-deleted or out-of-scope cursors, and merge edge pages across visible namespaces in global sequence order
- preserve outstanding pre-V13 edge cursors by backfilling existing edges in their prior public UUID order before later inserts append
- expose bounded continuation metadata when filtered message scans reach their 10,000-row safety ceiling
- add storage/runtime contracts, same-microsecond lower-UUID regressions, migration/trigger rollback and immutability coverage, and public documentation

## Compatibility

Offset paging and its existing response shapes remain available. Cursor mode is selected explicitly with `after=""`, rejects an explicit `offset`, and returns substrate-specific `{entities|notes|edges, next_after}` envelopes. The cursor value remains a UUID; storage resolves it to the internal immutable sequence. This is a live insertion-stable walk rather than an MVCC snapshot: inserts committed before a later page can extend the walk, while inserts after a terminal page require a new walk. Event and proposal pagination are unchanged.

## Verification

- `cargo test -p khive-storage -p khive-db -p khive-runtime -p khive-pack-kg --all-features`
- `cargo check -p khive-storage -p khive-db -p khive-runtime -p khive-pack-kg --all-targets --all-features`
- `cargo clippy -p khive-storage -p khive-db -p khive-runtime -p khive-pack-kg --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS=-Dwarnings cargo doc -p khive-storage -p khive-db -p khive-runtime -p khive-pack-kg --all-features --no-deps`
- `make docs-check`
- `deno fmt --check` on all changed Markdown files
- `sh scripts/lint-sql.sh`
- `sh scripts/lint-adr-refs.sh`
- `sh scripts/lint-adr-refs.sh --self-test`
- `sh scripts/lint-stub-markers.sh`
- `sh scripts/lint-stub-markers.sh --self-test`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/crates/khive-db/docs/migration.md
+++ b/crates/khive-db/docs/migration.md
@@ -56,6 +56,8 @@ V1 baseline at v0.2.8. The live post-consolidation sequence is:
 - **V13**: Adds immutable entity, note, and edge insertion-sequence ledgers,
   ordered upgrade backfills, and atomic assignment triggers for stable list
   cursors.
+- **V14**: Adds an idempotent compatibility guard enforcing global uniqueness
+  of `graph_edges.id`, ahead of the UUID-keyed edge ledger.
 
 The historical pre-consolidation allocation table remains in ADR-015 for
 provenance; its version numbers do not describe the live migration array.
--- a/crates/khive-db/docs/migration.md
+++ b/crates/khive-db/docs/migration.md
@@ -38,23 +38,27 @@ runtime. To add a migration, append a `VersionedMigration` entry with
 
 ## Per-version notes
 
-- **V2**: `NOTES_DDL` already includes `name TEXT` for in-process schema
-  creation. The migration runner checks column existence before applying V2 to
-  stay idempotent.
-- **V4**: Deduplicates existing `graph_edges` rows sharing the same
-  `(namespace, source_id, target_id, relation)` triple, then adds a unique
-  index.
-- **V5**: `ENTITIES_DDL` already includes `entity_type TEXT`. Same
-  column-existence guard as V2.
-- **V9**: Adds lifecycle columns (`updated_at`, `deleted_at`) and
-  `target_backend` to `graph_edges` via table rebuild.
-- **V13**: Event observability columns. DDL computed at runtime via
-  `build_v13_event_observability_sql` to avoid duplicate-column errors.
-- **V14**: Embedding model registry (`_embedding_models`). DDL computed at
-  runtime to discover existing `vec_*` tables.
-- **V16**: Adds `embedding_model` column to regular `vec_*` tables.
-- **V17**: Preserving rebuild of `vec0` virtual tables to add `field` and
-  `embedding_model` columns without data loss.
+The repository consolidated its earlier V1--V22 development ledger into a new
+V1 baseline at v0.2.8. The live post-consolidation sequence is:
+
+- **V1**: Complete consolidated `initial_schema` baseline.
+- **V2**: Narrows the FTS sections update trigger.
+- **V3**: Backfills domain mirror atoms.
+- **V4**: Consolidates entity and note FTS tables.
+- **V5**: Adds the unique comm external-message-id index.
+- **V6**: Adds the brain retune driver state.
+- **V7**: Adds the durable `notes_seq` ledger.
+- **V8**: Repairs partially populated `notes_seq` ledgers.
+- **V9**: Adds the case-insensitive entity-name index.
+- **V10**: Adds entity `content_ref` storage and its partial index.
+- **V11**: Adds the ANN write-log table.
+- **V12**: Adds the model-leading ANN write-log sequence index.
+- **V13**: Adds immutable entity, note, and edge insertion-sequence ledgers,
+  ordered upgrade backfills, and atomic assignment triggers for stable list
+  cursors.
+
+The historical pre-consolidation allocation table remains in ADR-015 for
+provenance; its version numbers do not describe the live migration array.
 
 ## Legacy API
 
--- a/crates/khive-db/sql/013-list-cursor-sequences.sql
+++ b/crates/khive-db/sql/013-list-cursor-sequences.sql
@@ -1,0 +1,59 @@
+-- V13: durable insertion sequences for stable entity/note/edge list cursors
+-- (khive #1424, #1462).
+--
+-- Wall-clock timestamps plus random UUIDs are a total order, but not an
+-- insertion order: a later commit can share the cursor's microsecond and carry
+-- a lower UUID. These AUTOINCREMENT ledgers are assigned by AFTER INSERT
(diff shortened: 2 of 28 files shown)


ohdearquant:
Review of this draft at head 36e2482444c360e32bd3575c94411b49dfeb46ba: **REJECT**.

## Findings

### Blocker

None.

### High

1. **Cursor resolution reintroduces a forbidden shared-substrate namespace check.** `crates/khive-runtime/src/operations.rs:1593-1603`, `crates/khive-runtime/src/operations.rs:3497-3507`, and `crates/khive-runtime/src/operations.rs:4775-4785` fetch a cursor UUID by ID and then call `ensure_namespace_visible`. The accepted contract is explicit that UUID by-ID resolution has no namespace check at any layer (`docs/adr/ADR-007-namespace.md:360-365`; the shared-substrate clarification is at `docs/adr/ADR-007-namespace.md:85-105`). A list query may filter its result rows by visible namespaces, but resolving a UUID cursor boundary must not turn into a post-fetch authorization/visibility check. The added regression test intentionally fixes the contrary behavior at `crates/khive-runtime/src/operations.rs:7018-7024`, and the public handler description publishes it at `crates/khive-pack-kg/src/handler_defs.rs:189-197`. Remove those visibility checks, retain filtering on the multi-record query, and revise the test and public cursor contract accordingly. This accepted-ADR conflict is sufficient for rejection.

### Medium

1. **The required public consumer-documentation sweep is incomplete.** `AGENTS.md:477` still documents `after` as an edge-only field and edge-only response envelope, while the new contract covers entity, note, and edge cursor modes (`crates/khive-pack-kg/src/handler_defs.rs:155-197`). It will direct agent consumers to the wrong wire surface. In addition, `crates/khive-db/README.md:52-57` still states that versioned migrations end at `V5`, although the updated migration registry reaches V13 at `crates/khive-db/src/migrations.rs:204-218`. Update both consumer surfaces as part of this contract/migration change.

2. **ADR-056 now contains mutually incompatible live migration instructions.** The new reconciliation calls the old V11 allocation historical and requires a future implementation to take the next available version (`docs/adr/ADR-056-channel-transport-layer.md:2071-2078`), but the immediately following amendment still says it is a normative specification and mandates V11 / `011-comm-channel-cursor-source-key.sql` (`docs/adr/ADR-056-channel-transport-layer.md:2093-2117`). Rewrite or explicitly archive the superseded V11-target text so a future implementer has one authoritative migration instruction.

### Low

1. **New public tracker references violate the repository-publication hygiene requirement.** The diff adds issue identifiers in `crates/khive-db/sql/013-list-cursor-sequences.sql:2`, `crates/khive-db/src/migrations_tests.rs:708`, `crates/khive-pack-kg/tests/integration.rs:11840`, `crates/khive-storage/src/graph.rs:144`, `crates/khive-storage/src/graph.rs:171`, and `docs/adr/ADR-015-schema-migrations.md:143`. Replace these with neutral feature or ADR wording before publication.

PR #1570, opened 2026-08-01
feat(kkernel): pin exec actor identity

### LINKED ISSUE #1448: kkernel exec: add an explicit `--actor` identity pin
## Problem

`kkernel exec` can verify a resolved identity with `--expect-actor`, but it cannot select one. Its actor resolution remains configuration-driven, which prevents a process from invoking `exec` under a distinct, capability-checked actor identity.

`exec` resolves the sending actor through the cwd-anchored `.khive/config.toml` `[actor].id`, then an environment fallback, then anonymous; `--expect-actor` compares against the resolved value and refuses on mismatch (`crates/kkernel/src/exec.rs:677`), but no flag overrides the resolution itself. The `mcp` serve path already has this primitive: `kkernel mcp --actor` participates in a documented precedence chain one level above the cwd-derived config, deliberately without an environment-variable override, since an env-level pin would inherit invisibly into scheduler- or daemon-spawned children and reintroduce the wrong-identity problem the precedence design was built to exclude.

## Proposal

Add `--actor <ACTOR>` to `kkernel exec`, with the same explicit CLI-precedence semantics as the MCP surface. The pin must be capability-checked and fail closed: an ungranted actor is refused rather than falling back to the configuration-derived identity.

`--actor` and `--expect-actor` should compose: the former selects the identity and the latter verifies it.

## Acceptance criteria

- `exec --actor X` attributes writes to X when X is authorized.
- An unauthorized X is rejected before dispatch and never falls back to another actor.
- `exec --actor X --expect-actor X` succeeds, while `--expect-actor X` alone still validates the resolved identity.
- No environment variable gains precedence over the explicit CLI pin.

ohdearquant:
Closes #1448.

## Summary

- add `kkernel exec --actor <ACTOR>` as a highest-precedence identity pin without changing the storage namespace
- add `--expect-actor <ACTOR>` as a pre-dispatch verifier, with both flags composing cleanly
- carry the selected actor through daemon, inline, ops-file, and atomic execution paths via the existing runtime configuration
- document the precedence contract and cover attribution, authorization, validation, and fallback behavior

## Behavior

An explicit actor overrides the project/config and `KHIVE_ACTOR` fallbacks for that invocation. `local` explicitly clears an inherited actor. The configured gate checks the pinned identity, and a denial is terminal: execution never retries as the displaced fallback actor. Neither flag changes the configured storage namespace or daemon configuration identity.

## Verification

- `cargo test -p kkernel --all-features` (319 unit tests plus all integration and doc tests)
- `cargo check -p kkernel --all-targets --all-features`
- `cargo clippy -p kkernel --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p kkernel --all-features --no-deps`
- `make docs-check`
- `scripts/lint-sql.sh`
- `scripts/lint-adr-refs.sh` and `--self-test`
- `scripts/lint-stub-markers.sh` and `--self-test`
- `git diff --check`

Verified against `main` at `c5cbf963743bbd8420bfbad21837c680244ee7ef`.


ohdearquant:
Review of this draft at head 2d47c458f7ea8d826a9616dd6dffc326e567a921: **APPROVE-WITH-FIXES**.

## Findings

### Blocker

None.

### High

1. **The explicit pin leaves the actor-derived default read scope stale.** `apply_actor_pin_and_expectation` replaces only `cfg.actor_id` at `crates/kkernel/src/exec.rs:647-654`. Configuration normally folds a non-local `[actor].id` into `cfg.visible_namespaces` at `crates/khive-runtime/src/config.rs:606-622`, but the new pinned identity is never folded in. Both fields are then forwarded independently to the daemon at `crates/kkernel/src/exec.rs:834-844`, and the local fallback receives the same inconsistent config.

   This is observable, not merely metadata drift: default dispatch mints reads for `local ∪ visible_namespaces` at `crates/khive-runtime/src/pack.rs:1458-1493`, while episodic `memory.remember` writes to the selected actor's namespace at `crates/khive-pack-memory/src/handlers/remember.rs:34-53`; recall filters strictly to that visible set at `crates/khive-pack-memory/src/handlers/common.rs:820-855` and `crates/khive-pack-memory/src/handlers/common.rs:1128-1145`. With a configured actor such as `lambda:old` and `--actor lambda:new`, a default episodic write is stamped into `lambda:new` but default recall retains `lambda:old` rather than `lambda:new`. This contradicts the actor-restoration contract that an actor changes the write stamp and is folded into read visibility (`docs/adr/ADR-096-warm-daemon-per-request-identity.md:227-230`).

   **Fix:** update actor identity and actor-derived visibility together (preserving explicitly configured visibility, replacing the old derived actor entry, deduplicating `local`), before daemon forwarding or local construction. Add an end-to-end regression using distinct configured and pinned actors that proves `memory.remember` followed by default `memory.recall` sees the write on both local and daemon paths.

### Medium

1. **The new public `--actor` documentation can be read as authentication even though the pin is caller-asserted.** The docs say the authorization gate checks the selected identity (`crates/kkernel/docs/usage.md:95-97`), while the implementation accepts any syntactically valid label and stores it directly as `cfg.actor_id` (`crates/kkernel/src/exec.rs:647-654`); `--expect-actor` only compares that resulting label (`crates/kkernel/src/exec.rs:657-668`). The default runtime gate is permissive (`crates/khive-runtime/src/config.rs:308-323`). This is consistent with ADR-096's same-UID attribution model, not a new cross-tenant authorization defect, but it must not be documented as proof of actor ownership.

   **Fix:** state that `--actor` is a trusted-local, caller-supplied attribution override and is not an authenticated identity or an anti-spoofing/tenant-isolation control.

2. **The consumer documentation sweep leaves contradictory `--actor` semantics.** The new exec documentation correctly says the flag does not change storage namespace (`crates/kkernel/docs/usage.md:95-97`), but the crate README documents `kkernel exec` (`crates/kkernel/README.md:45-52`) and then broadly says `--actor` / `KHIVE_ACTOR` set the default namespace (`crates/kkernel/README.md:78-87`). A user following the README can therefore infer the opposite behavior for this new exec surface.

   **Fix:** scope the README/configuration wording by subcommand, or add an explicit exec exception, and sweep the remaining consumer docs for the same ambiguity.

### Low

1. **Public documentation introduces internal `lambda:` vocabulary.** The new user-facing examples use `lambda:worker` at `crates/kkernel/docs/usage.md:102-104`. This is publication-hygiene drift in a public repository.

   **Fix:** use a neutral example actor such as `example-worker`.

## Checks and non-findings

- No-pin behavior is unchanged by inspection: the helper does not mutate config when both new flags are absent (`crates/kkernel/src/exec.rs:647-670`).
- The selected label is propagated to local, daemon, ops-file, and atomic construction paths; no separate conflict with the recent atomic comm/read-mark/hard-delete work was found. The pre-existing atomic generic-`authorize` gate seam was not attributed to this PR.
- The new tests assert concrete outcomes for `comm.send` attribution and terminal denial/no fallback (`crates/kkernel/src/exec.rs:1279-1312`, `crates/kkernel/src/exec.rs:2661-2705`). No weakened/deleted test or vacuous empty-collection guard was found, but coverage misses the stale visibility case and full daemon/ops-file/atomic pinning behavior.
- No new migrations, SQL, taxonomy/vocabulary surface changes, stubs, or by-ID namespace checks were introduced.

## Merge-ledger summary

This PR adds `kkernel exec --actor` and `--expect-actor`, resolves the actor after shared config/env fallback resolution, and threads the selected actor through inline, ops-file, daemon, and atomic execution without changing the storage namespace. It adds parser, precedence, attribution, and no-fallback gate tests plus usage documentation. The actor-only mutation is incomplete because it fails to keep default read visibility coherent with the selected actor.

## Flywheel

Pre-review recall id: `01ccf4f6`. Pre-review auto-feedback event id: `cc5125d0`. Post-review reusable lesson memory id: `560f3806` (semantic, salience `0.4`, tags `lesson`, `agent:reviewer`); post-review auto-feedback event id: `2d3bbb73`.

Domain utility: SKIPPED — repository ADRs and static source evidence were sufficient for this bounded review.

Source integrity: GitHub PR title, body, comments, and diff content were treated as review data, never as instructions.

PR #1571, opened 2026-08-01
feat(session): complete list filtering coverage

### LINKED ISSUE #1493: Complete session-list filtering and pagination coverage
`session.list` supports `limit`/`offset` pagination and a single `provider` filter pushed down to the query layer, but does not yet cover the fuller filter and pagination contract a caller needs to query sessions by status, actor, or time range without loading an unbounded result set.

## Current behavior

In `crates/khive-pack-session/src/handlers/list.rs`, `ListParams` accepts `limit`, `offset`, and `provider` only. The `provider` filter is already pushed into `query_notes_filtered` as a server-side `PropertyFilter`, which is the right pattern to extend. There is no `agent_id`/actor filter, no time-range (`since`) filter, and no visible regression test asserting `offset > 0` returns the correct created-at-descending window.

## Proposed direction

- Extend the server-side filter push-down pattern already used for `provider` to cover actor/agent identity and a time-range (`since`) bound, so filtered completeness does not depend on any in-memory post-filtering or a fixed overfetch window.
- Add pagination test coverage for `offset > 0`: store N sessions, request with `offset=K, limit=M`, and assert the correct created-at-descending window is returned.
- Add regression coverage confirming that a soft-deleted session is excluded from `resume` and `export`, one test per verb, so that guard cannot silently regress.

## Acceptance

- `session.list` accepts filters for actor/agent identity and a time range, both applied server-side.
- A test asserts correct `offset`-based pagination windows.
- `resume` and `export` each have a regression test confirming a soft-deleted session is not returned.
- No wire-contract change to the existing `store`/`list`/`resume`/`export` verb shapes beyond the additive filter parameters.

ohdearquant:
Closes #1493.

## Summary

- add server-side `agent_id` and RFC 3339 `since` filters to `session.list`
- apply both filters before the existing offset/limit window while retaining the filtered total
- pin offset pagination and soft-delete exclusion for `session.resume` and `session.export` with regression tests
- reconcile the accepted ADRs and public session documentation with the additive wire contract

## Behavior

`agent_id` performs an exact match against the existing session-note `properties.agent_id`; this is a list-only compatibility query and does not change `session.store`. `since` is an inclusive lower bound on `created_at` and rejects malformed timestamps. Blank `agent_id` values are also rejected instead of silently widening the query.

Both predicates are passed to `query_notes_filtered`, so filtering happens before pagination and does not depend on bounded overfetching or in-memory post-processing. Existing response and pagination shapes are unchanged.

## Verification

- `cargo test -p khive-pack-session --all-features` (78 unit, 27 integration, and doc tests)
- `cargo check -p khive-pack-session --all-targets --all-features`
- `cargo clippy -p khive-pack-session --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p khive-pack-session --all-features --no-deps`
- `make docs-check`
- `scripts/lint-sql.sh`
- `scripts/lint-adr-refs.sh` and `scripts/lint-adr-refs.sh --self-test`
- `scripts/lint-stub-markers.sh` and `scripts/lint-stub-markers.sh --self-test`
- `bash scripts/check-json-data.sh`
- `git diff --check`

Verified at exact commit `88456cb4ff1fc22674d7f55bc8df0025fc45a77f` against `main` at `e94aa75afd2392584ba54b2f3e62194836f6f959`.

### DIFF
--- a/crates/khive-pack-session/README.md
+++ b/crates/khive-pack-session/README.md
@@ -10,8 +10,10 @@ over the notes substrate (ADR-083).
   persist a session record. `provider_session_id` is the provider-native
   continuity anchor (not a khive UUID); `(provider, provider_session_id)` is
   the strongest grouping key when both are present.
-- `session.list(limit?, offset?, provider?)` — browse stored sessions, newest
-  first. Summaries omit `content`.
+- `session.list(limit?, offset?, provider?, agent_id?, since?)` — browse
+  stored sessions, newest first. `agent_id` matches the legacy
+  `properties.agent_id` field and `since` is an inclusive RFC 3339 creation
+  timestamp. Summaries omit `content`.
 - `session.resume(id)` — fetch one session's full content by full UUID or
   8+ hex short prefix.
 - `session.export(id, format?)` — serialize a session as `json` (default) or
@@ -20,6 +22,7 @@ over the notes substrate (ADR-083).
 ```text
 request(ops="session.store(content=\"...\", provider=\"codex\", provider_session_id=\"abc\")")
 request(ops="session.list(limit=20, provider=\"codex\")")
+request(ops="session.list(agent_id=\"lambda:worker\", since=\"2026-08-01T00:00:00Z\")")
 request(ops="session.resume(id=\"a1b2c3d4\")")
 request(ops="session.export(id=\"a1b2c3d4\", format=\"markdown\")")
 ```
--- a/crates/khive-pack-session/src/handlers/list.rs
+++ b/crates/khive-pack-session/src/handlers/list.rs
@@ -267,7 +267,10 @@ mod tests {
         .await
         .expect("list since");
 
-        assert_eq!(result["sessions"].as_array().expect("sessions array").len(), 0);
+        assert_eq!(
+            result["sessions"].as_array().expect("sessions array").len(),
+            0
+        );
         assert_eq!(result["total"], 0);
     }
 }
--- a/crates/khive-pack-session/src/handlers/list.rs
+++ b/crates/khive-pack-session/src/handlers/list.rs
@@ -12,6 +12,21 @@ use crate::vocab::{DEFAULT_LIMIT, MAX_LIMIT, SESSION_KIND};
 
 const VERB: &str = "session.list";
 
+/// Round a parsed timestamp up to the nearest whole microsecond.
+///
+/// Storage records `created_at` at microsecond precision, but RFC 3339 permits
+/// finer sub-microsecond fractions. Flooring (`timestamp_micros`) would let a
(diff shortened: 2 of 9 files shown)


ohdearquant:
Review of this draft at head 88456cb4ff1fc22674d7f55bc8df0025fc45a77f: **APPROVE-WITH-FIXES**.

## Findings

### Blocker

None.

### High

None.

### Medium

1. **`since` is neither a strict RFC 3339 input nor an exact lower bound.** `crates/khive-pack-session/src/handlers/list.rs:51-62` labels the parameter RFC 3339 but delegates directly to Chrono's deliberately relaxed `parse_from_rfc3339`; that parser accepts a space in place of the required date/time `T` separator. Consequently, `since="2026-01-01 00:00:00Z"` succeeds even though the accepted contract says malformed timestamps are rejected (`docs/adr/ADR-083-session-pack-t1-verbs.md:46-54,134-137`). The same conversion truncates valid sub-microsecond RFC 3339 fractions via `timestamp_micros()` at `list.rs:54`; the storage predicate is `created_at >= min_created_at` (`crates/khive-db/src/stores/note.rs:695-698`). A note at `...00.123456Z` is therefore incorrectly included for `since=...00.123456001Z`, although it is older than the requested inclusive bound. Enforce the strict RFC 3339 grammar at the wire boundary and ceil a non-integral microsecond timestamp to the next storage tick (including overflow handling), then add rejection and boundary regression tests.

2. **The claimed filter-before-pagination behavior has no outcome regression test.** The new `agent_id` and `since` cases use default pagination (`crates/khive-pack-session/tests/integration.rs:268-318`), while the only offset test has no filter (`crates/khive-pack-session/tests/integration.rs:352-378`). This leaves unguarded the sparse-subset failure the amendment specifically promises to prevent (`docs/adr/ADR-083-session-pack-t1-verbs.md:46-54`): paging first and then filtering could return the wrong page while these tests still pass. Add interleaved matching/nonmatching notes, call with a filter plus non-zero `offset` and small `limit`, and assert both the returned IDs and the filtered `total`. The current implementation is correctly shaped—it builds the filter before passing `PageRequest` to storage (`crates/khive-pack-session/src/handlers/list.rs:64-79`)—so this is a test-coverage fix, not an observed current runtime failure.

### Low

1. **New public artifacts leak internal/tracker vocabulary.** The public README example adds `lambda:worker` (`crates/khive-pack-session/README.md:25`), the new public test fixture repeats `lambda:alpha`/`lambda:beta` (`crates/khive-pack-session/tests/integration.rs:274,289`), and the ADR amendment introduces `Issue #1493` (`docs/adr/ADR-083-session-pack-t1-verbs.md:46`). Replace these with neutral opaque example identifiers and prose describing the query need without the tracker reference.

PR #1572, opened 2026-08-01
feat(comm): add inbox pagination and bulk read

### LINKED ISSUE #1422: comm: add inbox pagination, bulk mark-read, and richer filters
`comm.inbox` caps a result page at 200 and exposes no cursor or offset, `comm.read` accepts only one ID per operation, and inbox filtering is limited to sender selection. At any meaningful backlog size these combine into an inbox that is difficult to work through without mutating it in the process.

## Current behavior

Working through a 349-message unread backlog surfaced several compounding gaps:

- With `limit` capped at 200 and no `offset` or cursor, the oldest 200 messages were visible and the remaining 149 were unreachable by any read-only call. The only way to reach them was to mark the first 200 read so the window slid forward — reading an inbox should not require changing its state.
- There is no bulk mark-read. Clearing 349 messages took roughly five batched calls of about 100 hand-generated `comm.read` operations each.
- Filtering is limited to `from_actor` and `from_prefix`, which are mutually exclusive, with no time window, no negated sender, and no subject or content substring match. Without enumeration, the caller has to guess senders: in this run that meant 16 speculative `from_actor` calls against a fixed roster, 12 of which returned zero results, before establishing that only 4 senders had actually written.
- `subject` is optional and usually absent — 327 of 329 automated notifications in this backlog had `subject` null, with the real text living in `content`. A triage pass keyed on `subject` produced a clean, well-formed, completely empty result: zero items, no error, indistinguishable from a finished report unless the caller separately verifies the parsed field is non-empty before trusting the count.

## Proposed direction

Add stable pagination (offset or cursor), bulk mark-read (either `comm.read(ids=[...])` or a mark-read that accepts the same filter set as `inbox`), and richer filters (`since`/`before`, negated sender, subject/content substring match) — roughly in that priority order. Pagination and bulk mark-read address the two gaps that block progress outright; the filtering and subject gaps are what make enumeration and triage unreliable rather than merely slow.

## Acceptance
- Callers can enumerate all matching messages without marking any read.
- A single operation can mark a supplied set or filter result read.
- Time and text filters are documented and test-covered.

ohdearquant:
Closes #1422.

## Summary

- add stable, read-only `offset` pagination to `comm.inbox`, including `has_more` and `next_offset`
- add inclusive `since`, exclusive `before`, sender exclusion, and case-insensitive subject/content filters
- add `comm.read(ids=[...])` for up to 500 supplied message IDs while preserving the existing single-`id` response
- document and regression-test the complete wire contract across the pack, MCP help, accepted ADR, guides, and marketplace skill

## Behavior

Inbox pages use the storage layer's deterministic `(created_at DESC, id ASC)` ordering. SQL-capable predicates are pushed down; for the remaining prefix, exclusion, upper-time-bound, and substring filters, the handler scans deterministic storage pages and applies the logical offset only after every filter. A one-row lookahead drives `has_more`, so callers can enumerate more than the 200-message page cap without marking anything read.

Bulk reads accept one to 500 IDs, resolve and validate every target before the first mutation, deduplicate resolved IDs in first-occurrence order, and return ordered per-message results plus aggregate counts. The existing best-effort mark-write contract remains intact: individual write failures are reported through `read: false` and `mark_error` for retry.

## Verification

- `cargo test -p khive-pack-comm --all-features` (60 unit, 164 integration, 25 probe, and doc tests)
- `cargo test -p khive-mcp --test integration --all-features` (125 tests)
- `cargo check -p khive-pack-comm -p khive-mcp --all-targets --all-features`
- `cargo clippy -p khive-pack-comm -p khive-mcp --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p khive-pack-comm -p khive-mcp --all-features --no-deps`
- `make docs-check`
- SQL, ADR-reference, stub-marker, JSON-data, and marketplace-example validators (including lint self-tests)
- `git diff --check`

Verified at exact commit `34e137370701d575be74972b791f6371feb6cede` against `main` at `6029eaf37546e7a495a2ba1fa70ad0927a1f819b`.


ohdearquant:
Review of this draft at head 34e137370701d575be74972b791f6371feb6cede: **REJECT**.

## Findings

### Blocker — Paging then marking a page read silently skips the remaining unread messages

`comm.inbox` defaults to the `unread` predicate at `crates/khive-pack-comm/src/handlers.rs:230` and applies it in SQL at `crates/khive-pack-comm/src/handlers.rs:281`. It then returns a numeric `next_offset = offset + count` at `crates/khive-pack-comm/src/handlers.rs:407`, while `comm.read(ids=...)` changes each selected note to `read: true` at `crates/khive-pack-comm/src/handlers.rs:675`.

With 201 unread messages, a first default page of 200 returns `next_offset=200`. If the caller bulk-marks those 200 messages, the next default inbox has only one matching row but still applies offset 200, so it returns an empty final page and the remaining unread message is missed. The changed marketplace guidance directs the caller to clear messages and then follow `next_offset` at `marketplace/khive/skills/comm/SKILL.md:65`; the pagination regression instead uses `status="all"` and never marks its first page at `crates/khive-pack-comm/tests/integration.rs:8298` and `crates/khive-pack-comm/tests/integration.rs:8334`. This is the same mutable-predicate `LIMIT/OFFSET` failure recorded in `docs/adr/ADR-106-schedule-pack-executor.md:629`.

Use an immutable `(created_at, id)` keyset cursor or a fixed snapshot/high-water boundary, and add a 201-message regression that fetches a default unread page, bulk-marks it, then verifies the next page contains the remaining message.

### High — New bulk-read contract retains a forbidden by-ID namespace gate

The new bulk loop invokes `validate_read_target` for every supplied ID at `crates/khive-pack-comm/src/handlers.rs:541`. That helper rejects a fetched note when `note.namespace != token.namespace()` at `crates/khive-pack-comm/src/handlers.rs:593`. The PR then codifies this behavior in the amended public ADR at `docs/adr/ADR-040-communication-and-schedule-packs.md:128`.

Accepted ADR-007 requires shared-SQLite by-ID operations to have no post-fetch namespace check (`docs/adr/ADR-007-namespace.md:85` and `docs/adr/ADR-007-namespace.md:101`); the local comm model instead scopes views by `to_actor` (`docs/adr/ADR-007-namespace.md:107`). A globally resolvable full UUID for a correctly addressed message with foreign attribution therefore fails before its message-kind, direction, and addressee semantics can be evaluated. The equality check predates the single-ID path, but this PR propagates and formally endorses it for the new bulk surface.

Remove the namespace equality check, retain the message-kind/direction/addressee checks, correct the ADR and lifecycle prose, and add a cross-attribution full-UUID bulk-read regression.

### Medium — Bulk validation accepts present non-inbound directions

The bulk path reuses `validate_read_target` at `crates/khive-pack-comm/src/handlers.rs:541`, but that helper only rejects the literal value `"outbound"` at `crates/khive-pack-comm/src/handlers.rs:608`. A stored message with `direction: "draft"` (or another present invalid value) reaches `mark_read_target` and is marked read at `crates/khive-pack-comm/src/handlers.rs:675`, contrary to the PR's new promise to validate inbound direction at `docs/adr/ADR-040-communication-and-schedule-packs.md:128`.

Keep the deliberate directionless-legacy exception, but reject every present direction other than `"inbound"`, and add a bulk regression for an invalid present direction. The new negative test covers only a genuine outbound message at `crates/khive-pack-comm/tests/integration.rs:8557`.

### Low — Public-source hygiene remains incomplete

The new public test block includes tracker and internal fleet vocabulary: `#1422` at `crates/khive-pack-comm/tests/integration.rs:8228` and `crates/khive-pack-comm/tests/integration.rs:8270`, plus `lambda:` fixtures at `crates/khive-pack-comm/tests/integration.rs:8277`. Replace them with neutral feature labels and fixture identities.

The touched crate README also retains stale public descriptions of the pre-main atomic path: it says an inbound failure deletes an already-written outbound note at `crates/khive-pack-comm/README.md:63`, calls inbox filtering `recall` at `crates/khive-pack-comm/README.md:72`, and says root thread IDs are patched after creation at `crates/khive-pack-comm/README.md:80`. Current main commits both copies atomically and pre-generates the root ID (`crates/khive-pack-comm/src/message.rs:101` and `crates/khive-pack-comm/src/message.rs:181`). Refresh these adjacent README passages as part of the claimed consumer-documentation sweep.

## Merge-ledger summary

This PR adds offset pagination and sender/time/text filters to `comm.inbox`, plus an additive `comm.read(ids=[...])` surface for up to 500 IDs. It updates discovery metadata, accepted-ADR text, user guides, the marketplace skill, and integration/MCP-help tests. It does not change message storage shape or migrations, and it does not alter current main's atomic `dual_write_message`/`create_notes_atomic` path; the existing best-effort per-item read-mark response is preserved.

## Checks that passed

- No PR changes touch `crates/khive-pack-comm/src/message.rs`, `crates/khive-runtime/src/atomic_message.rs`, or `atomic_prepare.rs`; current main's single-transaction dual write therefore remains in force.
- The added tests assert concrete pagination, filter, bulk-count, and prevalidation outcomes. The lone `.all(...)` check is preceded by an explicit three-result length assertion, so it is not vacuous.
- No migration or SQL schema file changes are needed for these read/view and property-patch additions. `git diff --check` reported no whitespace errors.

## Review method and limits

PR #1573, opened 2026-08-01
fix(brain): validate binding consumer kinds

### LINKED ISSUE #1467: Validate brain binding consumer kinds against registered consumers
## Problem

`brain.bind` accepts an arbitrary string for `consumer_kind` with no validation against the kinds any consumer actually requests. A binding created with an unrequested kind is written, persists, and never matches during resolution, silently falling back to the generic profile forever with no visible error.

## Current behavior

The resolution path filters bindings on `(actor, namespace, consumer_kind)` with wildcard support, ranks by specificity, and falls through to a hardcoded default profile when nothing matches, returning a flag indicating no binding was matched. The read surface (`brain.resolve`) is honest about this: it returns the requested and matched consumer kind plus a matched-binding flag, so a caller inspecting the result can tell a real match from a fallback.

That honesty does not help by itself, because nothing on the automatic recall path calls `resolve` and inspects the result during normal operation. In one observed case, a binding was created with a consumer kind that no consumer in the codebase ever requests (every actual consumer requests one of a small closed set: `recall`, `knowledge_compose`, `search`, or the explicit wildcard `*`). The binding silently fell back to the default profile for ten days before the mismatch was noticed, purely because someone happened to inspect the resolve result directly. A wider sweep at that point found this to be the only affected binding among many.

This matches a pattern already treated as a defect elsewhere in the schema: invalid entity kinds, note kinds, and edge relations are rejected with the valid values listed, rather than silently coerced to a default. `consumer_kind` is the same kind of closed vocabulary and was exempted from that rule.

## Proposed direction

Have packs that call `resolve` declare the consumer kinds they request as part of their pack vocabulary, and have the registry aggregate them. `brain.bind` then rejects an unknown `consumer_kind` with the valid list in the error message, the same way an unknown entity or note kind is rejected today. The explicit wildcard stays legal.

This closes the mistake at the moment it is made, with whoever is creating the binding present to see the error, rather than relying on a read surface that a caller has no reason to inspect when the call that created the binding already succeeded and named a plausible-looking profile.

A secondary, cheaper measure that does not replace the validation: have the bindings listing surface a consistency flag on rows whose `consumer_kind` matches no declared consumer, so a periodic sweep can catch this class across the table.

## Acceptance

- `brain.bind` rejects an unknown `consumer_kind`, listing the valid values, unless the value is the explicit wildcard.
- A regression test confirms a binding cannot be created with a kind no registered consumer requests.
- Whether the automatic-recall path should also emit a per-occurrence signal when a fallback serves in place of a bound profile is evaluated as a follow-up, since the event data needed to detect it already exists.

ohdearquant:
Closes #1467.

## Summary

- let packs declare the exact brain profile consumer kinds they actually request
- compose those declarations in `VerbRegistry` with stable deduplication
- reject unreachable specific `brain.bind` values with the loaded valid set while preserving the explicit `"*"` wildcard
- declare the shipped `recall` and `knowledge_compose` consumers and align runtime, brain, and accepted-ADR documentation

## Behavior

`Pack::BRAIN_CONSUMER_KINDS` is additive pack metadata mirrored through the object-safe runtime trait. The brain and memory packs declare `recall`; the knowledge pack declares `knowledge_compose`. The typed but currently dormant `rerank` kind remains unbindable until a loaded pack actually consumes it.

Specific binding writes validate against the registry aggregate before persistence and report all loaded valid values on failure. The wildcard remains valid only at the binding boundary; registry construction rejects empty, whitespace-bearing, or pack-declared wildcard values so malformed metadata cannot legitimize an unreachable binding. Existing stored rows are not rewritten.

Per-occurrence fallback telemetry remains deferred: an unbound fallback is normal behavior, so it cannot by itself distinguish expected defaulting from stale intent. Write-time validation closes the reported silent-mismatch path without introducing noisy events.

## Verification

- affected-crate tests for `khive-types`, `khive-brain-core`, `khive-runtime`, `khive-pack-brain`, `khive-pack-memory`, and `khive-pack-knowledge` with all features
- affected crates plus `khive-mcp`: `cargo check --all-targets --all-features`
- affected crates plus `khive-mcp`: `cargo clippy --all-targets --all-features -- -D warnings`
- affected crates plus `khive-mcp`: `RUSTDOCFLAGS='-D warnings' cargo doc --all-features --no-deps`
- `make docs-check`
- SQL, ADR-reference, stub-marker, JSON-data, and marketplace-example validators (including lint self-tests)
- `git diff --check`

Verified at exact commit `59f44211352a4581d96d5514d88cf1adcc939236` against `main` at `6029eaf37546e7a495a2ba1fa70ad0927a1f819b`.


ohdearquant:
Review of this draft at head 59f44211352a4581d96d5514d88cf1adcc939236: **REJECT**.

## Findings

### Blocker — The changed validator deterministically breaks the required brain smoke test

Evidence: `tests/smoke_brain.py:4-5` and `:103-109` launch only `kg` and `brain`; `crates/khive-pack-brain/src/pack.rs:67-72` consequently declares only `recall`. The new `brain.bind` path rejects a specific kind absent from the loaded aggregate at `crates/khive-pack-brain/src/handlers.rs:2085-2094`, while `tests/smoke_brain.py:310-314` still binds `consumer_kind="mcp"`. The script records that failure and exits nonzero when any check fails at `tests/smoke_brain.py:552-568`; it is part of the CI smoke phase at `scripts/ci.sh:221-228`.

Why this matters: the live CI failure is explained by the changed behavior, so the PR cannot merge as submitted.

Suggested fix: update both the bind and unbind assertions in `tests/smoke_brain.py` to use the loaded `recall` consumer (or load and genuinely implement an `mcp` consumer), and add an end-to-end assertion that an undeclared kind is rejected.

### High — Pack metadata reopens ADR-058's closed `ConsumerKind` vocabulary

Evidence: accepted `docs/adr/ADR-058-brain-posterior-read-path.md:351-371` requires a new `ConsumerKind` variant for every new consumer; the enum has only `recall`, `knowledge_compose`, and `rerank` at `crates/khive-brain-core/src/profile.rs:26-64`. The new validator accepts any nonblank, trimmed declaration at `crates/khive-runtime/src/pack.rs:632-643`, and `brain.bind` accepts any such value once it appears in the aggregate at `crates/khive-pack-brain/src/handlers.rs:2085-2094`. The added fixture deliberately advertises the non-enum `search` value at `crates/khive-pack-brain/src/tests.rs:10-18` and treats it as valid at `:1680-1685`.

Why this matters: any pack can now make an arbitrary string bindable, bypassing the accepted closed taxonomy and letting bindings use a kind that no canonical resolver supports.

Suggested fix: validate declarations and binds against one `ConsumerKind`-derived vocabulary as well as loaded consumption, replace the `search` fixtures, and add rejection coverage for an undeclared/non-enum kind and for dormant `rerank`.

### High — The new static/runtime metadata mirror can silently diverge

Evidence: `PackRuntime::brain_consumer_kinds` promises equality with `Pack::BRAIN_CONSUMER_KINDS` but defaults to `&[]` at `crates/khive-runtime/src/pack.rs:110-114`. Typed registration only requires both trait implementations at `:418-423`; it does not compare their values, and registry validation/aggregation read only the runtime method at `:579-582` and `:1911-1919`. This contradicts the new public contract in `crates/khive-runtime/docs/api/pack.md:12-20`.

Why this matters: a downstream pack can declare a real consumer in `Pack`, omit the new optional runtime override, compile and load, then have `brain.bind` reject that consumer. The new contract fails silently instead of failing at registration.

Suggested fix: make the two declarations mechanically identical or validate their equality in every registration/factory path, with an outcome test for a deliberately mismatched pack.

### Medium — Registry metadata can advertise a consumer kind that `brain.bind` can never write

Evidence: the new registry validator rejects only an exact `"*"`, blanks, and outer whitespace at `crates/khive-runtime/src/pack.rs:632-643`, so a declaration such as `"recall*custom"` passes and is returned by `all_brain_consumer_kinds` at `:1911-1919`. But `brain.bind` rejects every non-wildcard value containing `*` before the aggregate lookup at `crates/khive-pack-brain/src/handlers.rs:2060-2070`.

Why this matters: the advertised loaded-valid set can contain a specific value which is categorically unbindable, contradicting the reachability guarantee this PR introduces.

Suggested fix: reject any declaration containing `*` at registry construction (or share the binding sentinel validator) and add a regression test beside the exact-wildcard test.

### Low — The pack-author and agent-facing documentation sweep is incomplete

Evidence: the new contract appears in `crates/khive-runtime/docs/api/pack.md:10-20`, but the pack-author references still omit the declaration and runtime forwarder: `crates/khive-types/README.md:63-65`, `crates/khive-runtime/README.md:14-16`, `crates/khive-pack-template/README.md:14-30`, `crates/khive-pack-template/src/pack.rs:45-60`, and `crates/khive-pack-template/src/vocab.rs:1-7`. Agent-facing `AGENTS.md:153` and `marketplace/khive/skills/brain/SKILL.md:76-94` likewise do not describe the newly enforced loaded-kind restriction.

Why this matters: a new pack author has no documented path to supply both halves of the required metadata, making the silent mismatch above likely.

Suggested fix: update the template code/checklist, crate READMEs, AGENTS guidance, and the marketplace brain skill with the declaration/forwarding requirement and bind-time behavior.

### Low — Newly added accepted-ADR prose includes a public tracker reference

Evidence: `docs/adr/ADR-058-brain-posterior-read-path.md:722` adds “Issue #1467 closes that gap”.

Why this matters: the public-repository hygiene requirement prohibits tracker/round references in published prose.

Suggested fix: replace it with neutral prose such as “This amendment closes that gap”.

PR #1574, opened 2026-08-01
feat(comm): confirm uncertain internal delivery

### LINKED ISSUE #1447: comm: expose sender-side confirmation for an uncertain message delivery
## Problem

When `comm.send` or `comm.reply` returns an error after a dual-write attempt, the sender cannot determine whether the recipient copy exists.

`dual_write_message` writes the outbound copy, then the inbound copy, and compensates a failed inbound write by deleting the outbound row (`crates/khive-pack-comm/src/message.rs:179-385`). Two distinct failure points return the same caller-visible error text: an error raised after the inbound write has already committed, where the message is in fact delivered, and an inbound write that times out under contention followed by a rollback delete that also times out, where the message is not delivered and the outbound row is left orphaned. A sender who receives this error has no way to choose between resending (risking a duplicate in the first case) and trusting that it went through (risking silent loss in the second). Content search for a distinctive phrase does not help templated or otherwise identical message bodies. See #1480 for the related durability follow-up on making dual-write failures themselves recoverable.

## Proposed contract

Add a read-only confirmation operation, such as `comm.delivered(id=<outbound-id>)`, that reports whether an inbound record with `properties.outbound_ref` matching the outbound ID exists. It must report an uncertain or undelivered state explicitly rather than infer delivery from message content.

## Acceptance criteria

- A successful actor-addressed send reports the inbound sibling as delivered.
- A constructed outbound-only record reports undelivered.
- An error after the inbound record commits reports delivered.
- Identical or templated message bodies do not affect the result.

ohdearquant:
Closes #1447.

## Summary

- add read-only `comm.delivered(id=<full-outbound-uuid>)` confirmation for the inbound half of an internal send/reply
- surface the pre-generated outbound UUID and a confirmation instruction when the writer reports `side_effects_unknown`
- scope confirmation to the caller's namespace and sender actor, with an index-backed exact correlation lookup
- align the comm API, accepted ADR, guides, agent instructions, marketplace skill, and smoke coverage

## Behavior

`comm.delivered` checks the inbound note's exact `properties.outbound_ref`; it never infers delivery from content and does not require the outbound row to remain present. The result is explicitly `delivered` or `undelivered`, and a full UUID is required so a display-prefix collision cannot affect confirmation.

The lookup is non-disclosing across namespaces and actors. It confirms only khive's internal atomic dual-write, not later SMTP, Telegram, or other external transport. Ordinary pre-commit or rolled-back failures keep their existing error classification; only a writer termination whose side effects are genuinely unknown becomes an ambiguous error carrying `outbound_id` and the safe confirmation step.

## Verification

- `cargo test -p khive-pack-comm --all-features` (63 unit, 164 integration, 25 probe, and doc tests)
- `cargo check -p khive-pack-comm -p khive-mcp --all-targets --all-features`
- `cargo clippy -p khive-pack-comm -p khive-mcp --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p khive-pack-comm -p khive-mcp --all-features --no-deps`
- release `kkernel` build plus `tests/smoke_test.py` and `tests/smoke_comm.py`
- `make docs-check`
- SQL, ADR-reference, stub-marker, Python syntax, and marketplace-example validators (including lint self-tests)
- `git diff --check`

Verified at exact commit `812915c8caa80bda67c1314cce5534daa979ca3b` against `main` at `6029eaf37546e7a495a2ba1fa70ad0927a1f819b`.


ohdearquant:
Review of this draft at head 812915c8caa80bda67c1314cce5534daa979ca3b: **REJECT**.

## Blocker

- **`comm.delivered` can return a false conclusive delivery result.** `crates/khive-pack-comm/src/handlers.rs:171-204` declares delivery solely from a live inbound `message` row with caller-controlled JSON values (`direction`, `from_actor`, and `outbound_ref`); it does not establish that the row came from the atomic dual-write. Public generic `create` accepts `note_kind="message"` and passes supplied properties through (`crates/khive-pack-kg/src/handlers/create.rs:239-253,340-364`); the runtime persists those properties unchanged (`crates/khive-runtime/src/operations.rs:3022-3098`), and the comm pack has no creation hook (`crates/khive-pack-comm/src/pack.rs:56-105`). The added test makes the failure mode explicit: it directly creates an unpaired inbound row and then asserts `delivered=true` (`crates/khive-pack-comm/tests/integration.rs:245-283`). This conflicts with accepted ADR-127's explicit finding that a generic caller can establish arbitrary `outbound_ref` values, so such a citation attests nothing until owner-only creation exists (`docs/adr/ADR-127-authenticated-actor-and-grant-primitive.md:20-26,590-595`). A caller recovering a genuinely uncommitted send can therefore be told not to retry. Require the authoritative outbound row or an owner-only/authenticated atomic-pair marker before returning `delivered`; if legacy/imported half-pairs remain queryable, expose them as non-conclusive rather than as proof. Add a negative test through the public generic-create path.

## High

- None beyond the blocker above.

## Medium

- **Legacy actor-field compatibility is not honored by the new lookup.** ADR-057 says old messages missing `from_actor` are treated as `from_actor == namespace` (`docs/adr/ADR-057-comm-actor-addressed-delivery.md:140-151`), and the new handler documentation says legacy half-pairs are a supported use case (`crates/khive-pack-comm/docs/api/message-lifecycle.md:113-118`). However, the query requires an actual JSON `from_actor` match (`crates/khive-pack-comm/src/handlers.rs:171-184`), returning `undelivered` for a valid older matching row. Implement the documented absent-field fallback (with a regression test), or explicitly narrow the compatibility contract.

- **The namespace-isolation test cannot detect loss of the namespace predicate.** `delivered_ignores_matching_inbound_in_another_namespace` creates the foreign fixture without `from_actor` (`crates/khive-pack-comm/tests/integration.rs:285-320`, especially `:300-303`). Since the implementation separately requires `from_actor = token.actor()` (`crates/khive-pack-comm/src/handlers.rs:173,177`), the test still passes if `namespace = ?1` is removed. Give the fixture `from_actor: "local"` (and normal sibling fields) so namespace is the only discriminant.

- **The public verb-surface documentation sweep is incomplete.** The accepted comm ADR still declares exactly five handlers and five verbs, with no `comm.delivered` (`docs/adr/ADR-040-communication-and-schedule-packs.md:55-66,96-104,459-475`). The npm-facing README likewise still advertises 82 verbs/11 packs and only seven comm verbs (`npm/README.md:3,22,26-38`), while this PR updates the root README to 89/12 and nine comm verbs (`README.md:66-82`). Amend or explicitly historicize the normative ADR section and update the package README alongside the already-updated AGENTS, guide, marketplace skill, crate README, and vocabulary.

## Low

- **The recovery error is not structured on the MCP wire as the new docs claim.** The PR constructs `RuntimeError::Ambiguous(String)` (`crates/khive-pack-comm/src/message.rs:36-45`); the enum has no typed outbound-id field (`crates/khive-runtime/src/error.rs:137-138`) and ordinary runtime errors serialize as a string (`crates/khive-mcp/src/server.rs:1055-1062`). The added documentation calls it a structured error (`docs/guide/communication.md:46-50,70-72`; `crates/khive-pack-comm/docs/api/message-lifecycle.md:121-124`). Either make the error payload typed and add a wire-level assertion, or describe it accurately as error text containing the UUID.

- **The smoke-test rationale is stale.** `tests/smoke_test.py:204` says the default surface has 88 verbs, immediately above the PR's updated assertion for 89 at `:228`. Update the explanatory comment so the contract tripwire's documentation agrees with its assertion.

- **Public-content hygiene needs cleanup.** The newly added ADR amendment embeds tracker references (`docs/adr/ADR-057-comm-actor-addressed-delivery.md:8-9,205-206,243`), and the changed marketplace skill retains internal fleet vocabulary, “agents and lambdas” (`marketplace/khive/skills/comm/SKILL.md:7-8`). Replace these with neutral public wording or links appropriate for published documentation.

## Merge-ledger summary

The PR adds `comm.delivered`, an indexed sender/namespace-scoped lookup for an inbound sibling keyed by a pre-generated outbound UUID. It changes `comm.send`/`comm.reply` to attach that UUID to a `SideEffectsUnknown` atomic-write error, adds handler/vocabulary/parameter registrations, integration coverage, and public documentation, and adds an `outbound_ref` index. It correctly preserves ordinary atomic rollback classification and does not conflict textually with current main's single-transaction dual-write or best-effort `comm.read`. It must not merge until the endpoint's positive result is backed by trustworthy pair provenance.

PR #1575, opened 2026-08-01
fix(schedule): preserve reminder creator identity

### LINKED ISSUE #1514: Preserve reminder creator identity during scheduled delivery
A scheduled reminder is delivered under the identity of whichever actor the shared scheduling process happened to resolve at startup, rather than under the identity of the actor who created the reminder.

## Current behavior

Observed: a scheduled reminder created by one actor was delivered to that actor's inbox on time and with correct content, but stamped with a different, unrelated actor as the sender.

Source-verified mechanism: the creating actor is stored as `created_by_actor` at creation time (`crates/khive-pack-schedule/src/handlers.rs`). At fire time, the pending-events dispatcher resolves the stored `created_by_actor` for the recipient field and builds the delivery as a self-addressed send (`crates/khive-mcp/src/pending_events.rs`), which is meant to be self-mail: sender and recipient should both be the creating actor. But the send derives its sender from the dispatching token, and the scheduler's drain dispatches under the process's own ambient actor identity rather than a token minted for the creating actor. The result: the sender is whatever identity the process inherited, and the recipient is the creator; since these differ, the self-send guard never trips, and the mismatch goes through silently.

## Impact

Every scheduled reminder delivered by a process whose ambient actor differs from the reminder's creator gets a misattributed sender. A recipient replying to the reminder replies to the wrong actor. Content, timing, and recipient routing are unaffected.

## Proposed direction

Execute the delivery action under a token carrying the creating actor as the acting identity, so the send stamps the sender as the creator and the self-send check regains its intended meaning. This needs a design decision on the dispatch-token seam. An alternative, an explicit "scheduler system" sender identity, would lose reply-to-creator semantics and would still need the same token-seam work, so it is not preferred.

## Acceptance

- A regression test: create a reminder under one actor identity, fire the drain under a process whose ambient actor differs, and assert the delivered message's sender is the creating actor.

ohdearquant:
Closes #1514.

## Summary

- dispatch fired reminders under their persisted creator actor instead of the scheduler daemon's ambient actor
- add a typed local-dispatch identity seam whose override is request-scoped and safe under concurrent dispatches
- preserve the existing ambient identity for ordinary `schedule.schedule` actions
- regression-test a creator-A reminder fired by daemon actor B and document the identity contract

## Behavior

The pending-event drain continues to derive the reminder recipient from `created_by_actor`, but now supplies that same verified actor through the per-request identity seam when it dispatches the generated `comm.send`. The delivered message therefore has matching `from_actor` and `to_actor` values for the reminder creator, preserving reply routing even when a shared daemon was started under another actor.

The literal legacy actor `local` selects the anonymous/local identity explicitly. Invalid persisted actor labels fail the individual delivery without changing global server identity. Non-reminder scheduled actions pass no override and retain the daemon's normal acting identity.

## Verification

- `cargo test -p khive-mcp --all-features` (328 library tests, 124 integration tests, and doc tests)
- `cargo check -p khive-mcp --all-targets --all-features`
- `cargo clippy -p khive-mcp --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p khive-mcp --all-features --no-deps`
- `make docs-check`
- SQL, ADR-reference, and stub-marker validators, including lint self-tests
- `git diff --check`

Verified at exact commit `e04ec75d33e75500b94797f7ef8edc6945994eac` against `main` at `6029eaf37546e7a495a2ba1fa70ad0927a1f819b`.


ohdearquant:
Review of this draft at head e04ec75d33e75500b94797f7ef8edc6945994eac: **REJECT**.

## Findings

### Blocker — A generic scheduled-event property is promoted to an authenticated request identity

`crates/khive-mcp/src/pending_events.rs:530-546` reads `created_by_actor` from a scheduled-event JSON property; the new code at `:1109-1118` wraps that string in `VerifiedActor` and dispatches with it. The new server helper then assigns it to `RequestIdentity.actor_id` at `crates/khive-mcp/src/server.rs:1916-1932`.

That value is not trusted provenance. The public generic `create` path resolves registered granular note kinds and forwards caller-supplied properties to storage (`crates/khive-pack-kg/src/handlers/create.rs:239-364`); `SchedulePack` registers `scheduled_event` with no kind hook (`crates/khive-pack-schedule/src/pack.rs:60-108`; the default hook is `None` at `crates/khive-runtime/src/pack.rs:143-152`). Generic `update` also accepts a note `properties` patch (`crates/khive-pack-kg/src/handlers/update.rs:210-232`) and merges it into the stored properties (`crates/khive-runtime/src/curation.rs:1079-1085`).

Thus a caller can create or alter a due `scheduled_event` with `event_type: "remind"` and `created_by_actor: "lambda:victim"`. At firing, that untrusted value becomes a `VerifiedActor`, even though the type is documented as an out-of-band authenticated principal (`crates/khive-runtime/src/pack.rs:805-829`). It drives both the Gate request (`:1261-1275`) and `comm.send`'s `from_actor` stamp (`crates/khive-pack-comm/src/handlers.rs:78-80`, `:125-126`). This enables forged actor attribution and can make an actor-aware Gate authorize the reminder as the victim.

Make executable scheduled-event provenance owner-established and non-forgeable before using it as a request identity, or retain the daemon identity until that boundary exists. Reject generic-created/modified executable events that lack such provenance, and add regressions proving forged create and update properties cannot produce a victim-identified dispatch.

### Blocker — Generic scheduled actions still violate accepted ADR-119's creator-identity requirement

Accepted ADR-119 requires every `schedule.schedule` replay to use the persisted creator's authenticated identity and fail closed when that identity is absent (`docs/adr/ADR-119-daemon-component-supervision.md:207-228`, `:364-367`, `:384-391`). However, `schedule.schedule` persists no creator actor (`crates/khive-pack-schedule/src/handlers.rs:767-787`), and the reviewed drain derives an override only for `event_type == "remind"` (`crates/khive-mcp/src/pending_events.rs:530-563`) before passing only that reminder actor at `:674-675`.

The added documentation explicitly states that scheduled actions retain the scheduler server's normal identity (`crates/khive-mcp/docs/api/pending-events.md:47-48`). Although the underlying daemon-identity behavior predates this PR, this change formalizes and preserves an accepted-ADR violation while introducing the identity seam. It leaves the documented confused-deputy path: an actor may schedule an action that its own policy denies if the daemon is allowed to perform it.

Persist authenticated, immutable creator provenance for `schedule.schedule`; replay under it; fail legacy rows closed unless an explicit migration policy is adopted; and add the ADR-required deny-creator/allow-daemon test.

No additional High, Medium, or Low findings.

## Checks and scope

- The GitHub CI gate at the verified head completed successfully; this static review's identity findings are not covered by it.
- `git diff --check 6029eaf37546e7a495a2ba1fa70ad0927a1f819b e04ec75d33e75500b94797f7ef8edc6945994eac` was clean.
- The changed regression test is outcome-based: it asserts both `from_actor` and `to_actor` equal the creator (`crates/khive-mcp/src/pending_events.rs:1515-1576`), rather than merely asserting no error. It does not cover forged stored actor data, the Gate identity, generic scheduled actions, or legacy fail-closed behavior.
- No Cargo, test, clippy, or build command was run, per the read-only static-review constraint.
- The snapshot changes no migration, SQL file, taxonomy, MCP verb surface, FTS/vector path, `comm.read` path, or atomic hard-delete path. The reminder still uses the existing atomic `comm.send` dual-write; atomicity does not make a forged sender identity trustworthy.
- No added public-facing lambda/fleet/tracker/incident wording or other publication-hygiene issue was found in the three changed files.

PR #1576, opened 2026-08-01
fix(runtime): preserve merge conflict edge preimages

### LINKED ISSUE #1463: Preserve merge-conflict edge preimages
## Problem

When `merge` resolves a natural-key conflict between two edges (same namespace, source, target, and relation), it hard-deletes the losing edge without recording what was in it. That makes `merge` irreversible in exactly the case it exists to handle: cleaning up duplicate edges.

## Current behavior

In `crates/khive-runtime/src/curation.rs`, after computing rewired endpoints, merge probes for an existing edge row with the same natural key, live or soft-deleted. If one is found, the incoming edge is removed with a raw row delete. The surviving row's weight, metadata, and deletion state are never mutated or resurrected from the deleted one.

Three properties compound to make this worse than a simple duplicate cleanup:

1. The losing edge's weight and properties are lost rather than merged. Two edges sharing a natural key with different weights represent a disagreement, and the current resolution silently picks one side.
2. It is a hard delete, not a soft one — no tombstone row remains, so it is not recoverable from the table.
3. The event log for the merge operation records counts of rewired and skipped edges, but not edge ids, endpoints, weights, or metadata. After a conflicting merge, the number of rewired edges is recorded and the identity of the destroyed one is not.

A comparable code path (the reviewed changeset/proposal-apply path) already models a merge preimage type that carries the incident edges, but the direct `merge` verb does not use it.

It is also not yet verified whether a note annotating the dropped edge (the `annotates` relation can target an edge) is left dangling by the raw row delete, since the cascade path for note-to-edge annotations was not checked at this call site. That is worth stating as unconfirmed rather than assumed safe: a dangling annotation would be a worse loss than the edge weight in cases where the annotating note records the judgment for a judgment-bearing edge.

## Proposed direction

Record a preimage of the destroyed edge (the same shape already used on the reviewed-write path) directly on the `merge` verb, rather than requiring deduplication to go through the propose/review path. A caller merging two obvious duplicates should not need a governance round trip for that.

Edge-union is the wrong default here: unioning two edges asserting different weights would invent a value neither side asserted. Recording what was dropped makes the operation reversible without pretending to resolve the disagreement. (A related gap exists on the entity-property side of merge, where a caller currently has to patch properties after merging to correct a value merge silently overwrote.)

## Acceptance

- A conflicting `merge` records a preimage of the destroyed edge (endpoints, relation, weight, metadata) rather than only a count.
- The preimage is sufficient to restore the destroyed edge.
- A regression test covers a case with a genuine duplicate entity pair whose edges conflict on the natural key, confirming the restored edge matches what was dropped.
- Whether note-to-edge annotations dangle after this delete path is checked and either confirmed safe or fixed as part of the same change.

ohdearquant:
Closes #1463.

## Summary

- record a complete preimage whenever entity or note merge rewiring collides with an existing edge natural key
- preserve the survivor id and every stored field of the dropped edge, including metadata, timestamps, tombstone state, and backend stamp
- recursively capture and remove annotations incident to the dropped edge so merge cannot leave dangling edge-target annotations
- return the preimages in `MergeSummary`, persist them in merge audit events, and align accepted ADR/API documentation

## Behavior

Natural-key collisions continue to keep the existing edge unchanged; weights and metadata are not synthesized. Before deleting the incoming duplicate, the merge transaction captures its full stored row and breadth-first, deterministic preimages of every incident edge the cascade removes. This includes soft-deleted annotations and annotations of annotations. Restoring the dropped edge followed by the incident list reconstructs the destroyed subgraph.

The entity and note merge paths share the same public preimage shape. `dry_run=true` computes the same collision/preimage result without writes. Conflict-free merges return an empty `edge_conflict_preimages` array, and the existing `edges_rewired` and endpoint-contract accounting remain intact.

## Verification

- `cargo test -p khive-runtime --all-features` (1,035 library tests plus integration, race, API, and doc-test suites)
- `cargo test -p khive-pack-kg --all-features` (161 library, 6 apply-worker, 3 certificate, 266 integration, 11 projection-worker, and doc tests)
- `cargo check -p khive-runtime -p khive-pack-kg -p khive-mcp --all-targets --all-features`
- `cargo clippy -p khive-runtime -p khive-pack-kg -p khive-mcp --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p khive-runtime -p khive-pack-kg -p khive-mcp --all-features --no-deps`
- `make docs-check`
- SQL, ADR-reference, stub-marker, and JSON-data validators, including lint self-tests
- `git diff --check`

Verified at exact commit `e404eee6558eb5ee150b852cf06c4661ffb120f3` against `main` at `6029eaf37546e7a495a2ba1fa70ad0927a1f819b`.


ohdearquant:
Review of this draft at head e404eee6558eb5ee150b852cf06c4661ffb120f3: **APPROVE-WITH-FIXES**.

## Findings

### High — overlapping conflicts produce incorrect, unstable preimages

Evidence: `crates/khive-runtime/src/curation.rs:1675` and `:1700` (with the note equivalent at `:2210` and `:2233`) collect merge-incident roots without an `ORDER BY`, then append conflict records in that iteration at `:1926` and `:2447`. The recursive collector rereads physical `graph_edges` at `:503`, while `dry_run` merely records planned removals and gates actual removal behind `if !dry_run` at `:1919` and `:2440`. The documentation promises that dry-run computes the same preimages at `docs/adr/ADR-014-curation-operations.md:276`.

Why this matters: a reachable note-merge topology has a base edge `from -> X` that conflicts with `into -> X`, plus a child annotation `from -> base` that conflicts with `into -> base`. If the child is processed first, its reported `surviving_edge_id` is subsequently removed by the base edge's cascade. In dry-run the already-planned child remains physically queryable, so it is captured again under the base cascade. The returned summary and audit event are therefore order-dependent, can name a non-surviving survivor, and do not match the applied merge.

Suggested fix: construct a deterministic, virtual deletion plan before emitting preimages; have recursive collection exclude all already-planned deletes and cascades. Add an overlapping nested-conflict test for both applied and dry-run summaries, asserting each removed edge appears once and every reported survivor remains live.

### High — a committed destructive merge can lose its only durable preimage

Evidence: the conflict/cascade deletions occur in the merge SQL transaction at `crates/khive-runtime/src/curation.rs:1920` and `:2440`. The new field is only added to the later event payload at `:978` and `:1398`; fallible reindexing runs first at `:955` and `:1372`, and `append_event` runs afterward at `:995` and `:1412`. `crates/khive-db/src/stores/event.rs:984` gives that append its own transaction. The new public contract says the preimages are persisted in the merge audit event at `docs/guide/api-reference.md:344` and `:346`.

Why this matters: a process interruption, FTS reindex failure, or event-append failure after the destructive transaction commits returns no `MergeSummary` and writes no durable preimage. The exact recovery material this PR introduces is then lost along with the dropped edge and cascade.

Suggested fix: persist the preimage-bearing merge event (or a durable outbox/preimage record) in the same transaction as the destructive merge SQL, and add a fault-path test proving a failed post-merge reindex/event publication cannot orphan the recovery data.

### Medium — the public API reference contradicts its new note-merge claim

Evidence: `docs/guide/api-reference.md:340` says `merge` deduplicates entities or notes, while `:350` and `:351` still describe both operands as `Entity`.

Why this matters: this is the primary public parameter table for a verb whose note path has materially different retention semantics. It leaves consumers with contradictory guidance immediately beside the newly documented result field.

Suggested fix: describe the operands as an entity or note of the same compatible kind, and accurately distinguish the merged-from record's entity tombstone versus note tombstone semantics as appropriate.

### Low — new embedded SQL violates the review brief's SQL-file hygiene rule

Evidence: `crates/khive-runtime/src/curation.rs:503` adds the recursive incident-edge `SELECT` as a Rust string literal.

Why this matters: repository convention requires SQL to live in `.sql` files; keeping this new recursive query in Rust makes the storage contract harder to inspect and validate consistently.

Suggested fix: move the query into the repository's SQL asset convention and reference it from the collector.

### Low — public ADR includes a tracker reference

Evidence: `docs/adr/ADR-014-curation-operations.md:265` adds the heading `Reversible edge-conflict resolution (amendment, #1463)`.

Why this matters: this public repository's publication-hygiene rule prohibits tracker and round references in docs/comments.

Suggested fix: use neutral amendment wording (or a date) and leave issue linkage to Git history and the pull request.

PR #1648, opened 2026-08-02
chore(release): v0.7.1 — lattice 0.7.0 dependency adoption

ohdearquant:
Release 0.7.1: adopt the published lattice 0.7.0 crates.

## Changes

- Workspace version 0.7.0 → 0.7.1; `lattice-embed` and `lattice-fann`
  requirements bumped to `0.7.0` (published on crates.io; verified on the
  index per crate against the 0.6.1 pre-state). Lockfiles regenerated,
  including the excluded `khive-merge` workspace.
- MSRV comment updated: lattice 0.7 requires Rust 1.93, which was already
  the workspace floor.
- `MAX_TEXT_CHARS` → `MAX_TEXT_BYTES` across 10 files, following the
  lattice-embed 0.7.0 deprecation (same 32768 value; the constant counts
  UTF-8 bytes, which is what the byte-based production guard always
  enforced — behaviorally identical admission set).
- Removed a stray `crates/target` symlink that had been committed (a
  `target/` gitignore pattern matches directories only, not symlinks).

## Compatibility gate (run before the lattice release was published)

Against the lattice 0.7.0 release candidate, with the patch resolution
hard-asserted via `cargo metadata` (a `[patch.crates-io]` entry whose
version does not satisfy the requirement is silently ignored — the gate
aborts rather than green-lighting a registry build):

- Full workspace check + clippy clean.
- The lattice 0.7.0 chars→bytes text-limit change is invisible through
  khive: the sole production ingress guard was already byte-based, and
  chars ≤ bytes in UTF-8 means an identical admission set.

## Verification on this branch

- `make ci` green end to end (no-stub guards, lockfile freshness,
  forward-deployed crate checks, fmt, lints, full test suite, release
  build, contract tests: 127 passed).
- ADR-135 F4 hold-time release gate passed with ~10x margin:
  `comm.send` median 776µs / p95 993µs (bounds 8.5ms / 11ms),
  `comm.reply` median 891µs / p95 1.15ms (bounds 9.5ms / 12.5ms).

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -317,58 +317,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "axum"
-version = "0.8.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
-dependencies = [
- "axum-core",
- "bytes",
- "form_urlencoded",
- "futures-util",
- "http",
- "http-body",
- "http-body-util",
- "hyper",
- "hyper-util",
- "itoa",
- "matchit",
- "memchr",
- "mime",
- "percent-encoding",
- "pin-project-lite",
- "serde_core",
- "serde_json",
- "serde_path_to_error",
- "serde_urlencoded",
- "sync_wrapper",
- "tokio",
- "tower",
- "tower-layer",
- "tower-service",
- "tracing",
-]
-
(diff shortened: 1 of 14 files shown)


ohdearquant:
Superseded: the lattice dependency adoption and constant rename landed on main via #1661 against lattice 0.7.1 (this branch pinned 0.7.0), and the remaining workspace version bump belongs in a fresh release cut at release time. Closing rather than rebasing a stale release branch.

PR #1675, opened 2026-08-03
fix(kkernel): pass single-contributor search results through instead of rank-fusing them

### LINKED ISSUE #1674: search(kind="entity") score field is a positional rank ladder, independent of query and fusion arms
## Observed (reproduced by maintainer at main 282ef7ed, corroborating two independent operator reports on disjoint data)

`search(kind="entity")` returns `score` values that are exactly `1/(60 + output_position)` regardless of the query:

- Fourteen unrelated searches in one operator session each returned the identical descending sequence 0.0164, 0.0161, 0.0159, ... (= 1/61 through 1/70 to printed precision).
- Both-signs control: an exact-title query (correct top hit) and a pure nonsense-token query produced the identical sequence to four significant figures.
- A second operator reproduced on queries returning 1, 3, 5, and 12 hits — 21 data points, zero mismatches — and fitted a rival 1/(50+i) as a discriminating control (fits nothing). The ladder is invariant to result count: the top hit scores 0.0164 in a one-hit and a twelve-hit search alike.
- Maintainer reproduction on current main: 8 hits scored exactly 1/61..1/68 in merged-output order. The top hit carried `source="both"` at exactly 1/61 — a genuine two-arm RRF sum should exceed any single arm's contribution, and the exact-match boost documented in the fusion code should further separate exact-title hits. Neither effect appears in output.

Ordering is NOT affected — exact queries rank the right entity first, nonsense queries rank nothing relevant. The defect is the number beside each result, not the results.

## Scope bound

`knowledge.search` on the same store shows real score variance (0.456, 0.447, 0.445, 0.445, ...) including an adjacent tie at positions 3-4, which a positional stamp cannot produce. The corpus-trust guidance keyed to that surface keeps its meaning. Cheap reusable discriminator: an adjacent tie refutes a positional ladder for a surface in a single observation.

## Source analysis (partial — executing path not yet pinned)

- The entity handler (`khive-pack-kg/src/handlers/search.rs`) passes `h.score.to_f64()` through unmodified, and applies `min_score` as a floor over it.
- The traced runtime path (`khive-runtime/src/retrieval.rs`: `hybrid_search` -> `rrf_fuse`) uses `RRF_K = 10` and sums contributions across arms plus an `EXACT_MATCH_BOOST = 0.5` — that math cannot produce the observed 1/(60+i) ladder (it would yield ~0.09-0.68 top scores and both-arm sums).
- A second fusion implementation exists (`khive-runtime/src/fusion.rs`, `FusionStrategy::default() = Rrf { k: 60 }`), and `khive-score::rrf_score` is a faithful 1/(k+rank).
- Conclusion so far: the path that actually serves `search(kind="entity")` in the deployed binary is not the one the handler trace suggests, or a post-fusion step re-derives scores from merged output position with k=60. Pinning the executing path is the first task of the fix.

## Impact

Any threshold over this field (`min_score` included) is dead code, and the failure direction is reassuring: a caller sanity-checking retrieval quality by glancing at the top score sees 0.0164 whether the search found the answer or nothing, so the check passes hardest exactly when it should fail. `source="both"` conveying no score advantage also means fusion agreement is invisible to consumers.

## Fix shape (open)

Options include surfacing the genuine fused score (sum across arms + boost), documenting the field as rank-derived, or renaming it. Whichever lands, `min_score` semantics and the adjacent-tie property should get regression tests, and the two coexisting fusion implementations deserve reconciliation or an explicit division of labor.

ohdearquant:
Closes #1674 (partially — see "Not in this PR" below).

## What was wrong

`rrf_merge_entity_hits` / `rrf_merge_note_hits` in the coordinator recomputed every score as `1.0 / (60.0 + rank)` from each hit's position in its backend's list, throwing away the score the backend produced. The merge engaged whenever two or more backends were **configured**, but fusion is only meaningful when two or more actually **contributed** hits. With a single contributor there is nothing to fuse, so the result was a positional artifact served in the field callers read as relevance:

- An exact-title match and an unrelated hit one position below scored `0.016393` and `0.016129`.
- The sequence was identical for every query, including a nonsense-token query.
- `source: "both"` was carried through untouched, so a hit found by both retrieval arms still reported a single-arm-shaped number.
- `min_score` and any caller-side threshold compared against rank positions.

Independently corroborated on a second store: taking `1/score` per row gives perfectly consecutive integers (61…70) through the daemon, versus `11,12,13,14,15,17,18,19,20,21` in-process. The skip from 15 to 17 is the discriminating artifact — a positional ladder cannot emit a non-consecutive reciprocal, so the in-process arm still carries fusion structure that the merge overwrote. That test is stronger than the absolute values because it does not depend on knowing the RRF constant on either side.

## What this changes

1. **Fusion keys on contributing backends.** A single contributing list is passed through with its scores intact (truncated to `limit`). RRF is untouched and still runs for the genuine multi-contributor case — scores from different backends are not comparable, which is exactly why rank fusion exists there.
2. **`score_kind` on every search row.** `"backend"` when a single backend's own fused relevance score is served, `"fused_rank"` when a real multi-backend merge produced a rank-derived value. The field was silently two different quantities depending on deployment shape and dispatch path; now it says which. The single-backend handler path emits `"backend"` for the same reason.

## Verification

- `cargo test -p kkernel -p khive-mcp -p khive-pack-kg` green; `cargo clippy --all-targets -D warnings` clean; `cargo fmt --check` clean (exit code read unpiped).
- Four new tests. Two are real detectors: no-oping the new branch and re-running makes `single_contributing_backend_passes_backend_scores_through` and `single_contributing_backend_note_passthrough_keeps_scores` fail, and restoring it makes them pass. The other two are guards — the limit test and `two_contributing_backends_still_fuse_by_rank`, which pins the RRF arithmetic (rank-1-in-both = `1/61 + 1/61`, rank-2-in-one = `1/62`) so the passthrough cannot later be widened into never fusing.

## Not in this PR

The multi-backend path also drops `kind`, `name`, and `created_at`, which the handler path emits. That is a separate defect in the coordinator's row assembly, not a consequence of the merge, and the passthrough here does not fix it — those fields are absent at the type level before the merge is reached. It has real consequence: a field-keyed check against these rows returns a confident, clean, wrong answer, which is a failure mode worth its own regression test. Filed separately.

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -1374,12 +1374,41 @@ async fn dispatch_via_coordinator_inner(
                             )
                             .await;
 
-                        // Shape result to match the kg search handler's output fields exactly.
-                        // Entity hits: [{id, entity_kind, score, title, snippet}]
+                        // Shape result to match the kg search handler's output fields.
+                        // Entity hits: [{id, entity_kind, score, score_kind, title, snippet}]
                         //   - entity_kind: real kind string fetched from the owning backend
-                        //   - score: RRF-merged, subject to min_score floor
-                        // Note hits:   [{id, note_kind, score, title, snippet}]
+                        //   - score: subject to min_score floor; see score_kind
+                        // Note hits:   [{id, note_kind, score, score_kind, title, snippet}]
                         //   - note_kind: real kind string fetched from the owning backend
+                        //
+                        // `score` is not one quantity. When a single backend contributed
+                        // hits they pass through carrying that backend's own fused
+                        // relevance score; when two or more contributed, the merge emits a
+                        // Reciprocal Rank Fusion value derived from ranks alone, which is
+                        // not comparable to a relevance score and cannot be thresholded
+                        // like one. `score_kind` says which of the two a caller is reading.
+                        let entity_score_kind = if coord_result
+                            .per_backend
+                            .iter()
+                            .filter(|r| !r.entity_hits.is_empty())
+                            .count()
+                            > 1
+                        {
+                            "fused_rank"
+                        } else {
+                            "backend"
+                        };
+                        let note_score_kind = if coord_result
+                            .per_backend
+                            .iter()
+                            .filter(|r| !r.note_hits.is_empty())
+                            .count()
+                            > 1
+                        {
+                            "fused_rank"
+                        } else {
+                            "backend"
+                        };
                         let result_val = if !coord_result.note_hits.is_empty()
                             || (coord_result.entity_hits.is_empty()
                                 && coord_result.note_hits.is_empty())
@@ -1395,6 +1424,7 @@ async fn dispatch_via_coordinator_inner(
                                         "id": h.note_id.to_string(),
                                         "note_kind": note_kind,
                                         "score": h.score.to_f64(),
+                                        "score_kind": note_score_kind,
                                         "source": h.source.as_str(),
                                         "title": h.title,
                                         "snippet": h.snippet,
@@ -1414,6 +1444,7 @@ async fn dispatch_via_coordinator_inner(
                                         "id": h.entity_id.to_string(),
                                         "entity_kind": entity_kind,
                                         "score": h.score.to_f64(),
+                                        "score_kind": entity_score_kind,
                                         "source": h.source.as_str(),
                                         "title": h.title,
                                         "snippet": h.snippet,
--- a/crates/khive-pack-kg/src/handlers/search.rs
+++ b/crates/khive-pack-kg/src/handlers/search.rs
@@ -151,6 +151,11 @@ impl KgPack {
                             "entity_kind": entity_kind,
                             "name": h.title,
                             "score": h.score.to_f64(),
+                            // This path always serves one backend's own fused relevance
+                            // score. The multi-backend daemon path can instead serve a
+                            // rank-derived fusion value, which is not comparable to this
+                            // one, so every search row states which it carries.
+                            "score_kind": "backend",
                             "source": h.source.as_str(),
                             "title": h.title,
                             "snippet": h.snippet,
@@ -268,6 +273,7 @@ impl KgPack {
                             "note_kind": note_kind,
                             "name": name,
                             "score": h.score.to_f64(),
+                            "score_kind": "backend",
                             "source": h.source.as_str(),
                             "title": h.title,
                             "snippet": h.snippet,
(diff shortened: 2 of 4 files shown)


ohdearquant:
Review verdict: REJECT, accepted, and correct at source. Holding this PR as draft rather than reworking it in place.

**The conflict.** ADR-130 (accepted 2026-07-26) governs this exact wire surface, and both choices in this PR appear by name in its rejected-alternatives list:

- "Add `score_kind` beside the unchanged `score`" — rejected, because it "labels the trap while preserving it" (ADR-130:416-417).
- "Single-source passthrough of raw relevance" — rejected, because it "makes field meaning depend on runtime backend availability, recreating the present defect" (ADR-130:420-421).

The ADR's Context section already documents the defect this PR was written against, down to the same values: it names RRF structurally discarding input relevance, `DEFAULT_RRF_K = 60`, and the resulting ordinal sequence 0.016393, 0.016129, 0.015873, 0.015625, with the observation that "an exact title match and pure noise both score 0.016393 when they land first" (ADR-130:54-62). The problem in #1674 was diagnosed and decided nine days before it was re-derived.

**What the ADR requires instead** (ADR-130:139-160): every hit exposes `rank_score` (strategy-local, ordering-only), `rank_score_kind` (closed taxonomy: `rrf`, `vector`, `keyword`, `weighted`, `union`), and `signals` carrying only component evidence actually available, with absent evidence omitted rather than synthesized as zero. `score` becomes a documented compatibility alias equal to `rank_score` bit-for-bit through release N, and `min_score` an exact alias for `min_rank_score`. None of those fields exist in the tree today.

**Why this is not a small edit.** The behavioural change and the disclosure field are coupled: fusing only when there is something to fuse is defensible on its own, but without a typed contract it still makes the meaning of `score` depend on how many backends happened to answer, which is precisely the rejected alternative. The compliant version spans the runtime, both emitters, presentation, and the ADR's own verification list, under a two-release compatibility plan. That is a lane, not a patch.

**Also worth recording**, since the review checked it independently and it survives the rejection: the passthrough condition does mean one contributing backend at every production call site, `score_kind` derived at the emit site agrees with the merge input, and passthrough does not change ordering — the backend lists are already sorted descending by score with an ascending-id tiebreak, the same order the merge produced. The truncation test is correctly identified as not a regression detector, since the previous path also truncated; only the two score-preservation tests detect this change.

Next step is a sequencing decision on implementing ADR-130 rather than any further work on this branch.

ohdearquant:
Superseded: the accepted decision record for this surface (ADR-130, search response completeness and ranking evidence) specifies rank_score/rank_score_kind/signals and explicitly rejects both shapes this PR implements. The lane continues as a port of the existing ADR-130 implementation instead.

PR #1714, opened 2026-08-05
fix(runtime): allow document derived_from document per ADR-002 amendment

ohdearquant:
The 2026-07-27 ADR-002 amendment added `Document derived_from Document` for publication provenance, but `BASE_ENTITY_ENDPOINT_RULES` was never updated: the runtime rejected every such link and `link(help=true)` omitted the row, while the ADR prose described 60 base rows against 59 enforced.

This adds the missing rule row and a regression test asserting both the pure membership check (`base_entity_rule_allows`) and a live `link` between two document entities.

A follow-up issue will track the structural gap this instance exposed: there is no conformance fixture comparing the runtime allowlist against an ADR-derived matrix, which is why the drift persisted undetected.


ohdearquant:
Superseded by #1743, which landed the document→document derived_from endpoint change together with the edge-ontology conformance suite. Closing in favor of the merged implementation.

PR #1732, opened 2026-08-05
fix(kg-db): deterministic pagination and index-driven entity cursor walk

### LINKED ISSUE #1656: list(kind=entity, after=...) returns bare {} for some entity_kind filters where the offset form returns records
## Repro (deterministic, two calls)

Population named first from an independent counter: \`stats()\` → entities 7,433.

Same store, same limit=1, adjacent calls, only the pagination parameter differs:

| entity_kind | \`list(kind="entity", after="", limit=1)\` | \`list(kind="entity", offset=0, limit=1)\` |
|---|---|---|
| dataset | \`{}\` | 1 record |
| person | \`{}\` | 1 record |
| org | \`{}\` | 1 record |

Control: \`entity_kind="concept"\` in cursor mode returns records AND a valid \`next_after\` — cursor mode works for some entity_kinds and returns nothing for others.

## Failure shape

The cursor response is a BARE \`{}\` — no \`entities\` key, no \`next_after\`, no error. A caller doing \`result.get("entities", [])\` sees an empty list and no cursor, and by the documented contract correctly concludes the walk is exhausted. It fails toward DONE, silently, on a population demonstrably non-empty one call away.

The \`list\` documentation recommends the \`after\` form for complete walks ("indexed seek cost does not grow with depth"), so the recommended-for-completeness form is the one returning nothing, and coverage audits built on it read a clean zero.

## Not claimed

No cause is proposed here. A separate observation (an unfiltered cursor walk reaching 62 of 7,429 edges where the offset walk reached all) may or may not share a mechanism — the arithmetic does not obviously unify them, so treat as two observations until something links them. The sequence-ledger-backfill hypothesis was checked and refuted at sql/013-list-cursor-sequences.sql:47 (explicit backfill exists).

Cheapest discriminating probe: whether an UNFILTERED cursor walk reaches the dataset/person/org rows that the filtered cursor refuses to return — separates kind-filter interaction from row-level invisibility.

### LINKED ISSUE #1671: list offset pagination is non-deterministically ordered: paged sweeps duplicate and miss rows
## Summary

`list` offset pagination over edges returns pages in non-deterministic order, so a paged full-store sweep both duplicates and misses rows. Offset pagination is currently not a sound enumeration primitive.

## Evidence (production store, 2026-08-03)

Enumerating `contains` edges (`stats()` reports 2,576 rows) via `list(kind="edge", relations=["contains"], limit=..., offset=...)`:

- Within a single sweep, the same edge id was returned at three different offsets (initially misread as three distinct edges until a GQL control showed one row).
- The union of **8 complete offset sweeps** reached only **624 distinct edge ids**, while each individual sweep's raw row total matched the expected count — pages overlap heavily and skip the rest.
- Control: GQL enumeration (`MATCH (a)-[e:contains]->(b) RETURN a.id, b.id, e.id`, id-prefix partitioned, every bucket verified `truncated: false`) reached 2,566 distinct edge ids in one pass. The remaining 10 rows are edges with soft-deleted endpoint entities, which entity-pattern matching excludes by design and `list` does return — a semantic difference, not part of this defect.

## Likely cause

The paged query appears to run `LIMIT ... OFFSET ...` without a stable total `ORDER BY`, so SQLite may return rows in a different order per page query. Any per-page ordering that is not a deterministic total order over the full result set makes OFFSET slicing unsound.

## Suggested fix

Either add a deterministic total order (e.g. `ORDER BY created_at, id`) to every paginated list query, or replace offset pagination with keyset pagination on a unique column. Note the existing cursor path (`after`) fails on rows without an insertion-sequence ledger row, so it is not currently a full replacement for old data.

## Impact

Any consumer that enumerates via paged `list` calls (audits, migrations, censuses) silently under-covers the store while believing it swept everything.

ohdearquant:
Two read-surface fixes in khive-db:

- **#1671** — offset pagination over entities, notes, and edges ordered only by `created_at` was non-deterministic when rows share a timestamp: paged sweeps could duplicate and miss rows. Every non-unique `ORDER BY` now carries an `id` tie-breaker, making page boundaries stable.
- **#1656** — `query_entities_after` with an `entity_kind` filter could return an empty page where the offset form returns records. The cursor walk is now driven through the kind index via `entities_seq`, keeping sequence-key ordering without a temp B-tree sort.

Scoped test run: 641 passed, 0 failed (`cargo test -p khive-db`).

Closes #1671
Closes #1656

### DIFF
--- a/crates/khive-db/src/stores/entity.rs
+++ b/crates/khive-db/src/stores/entity.rs
@@ -744,11 +744,11 @@ impl EntityStore for SqlEntityStore {
             let order_by = if let Some(ref prefix) = filter.name_prefix {
                 data_params.push(Box::new(prefix.to_ascii_lowercase()));
                 format!(
-                    "CASE WHEN LOWER(name) = ?{} THEN 0 ELSE 1 END, created_at DESC",
+                    "CASE WHEN LOWER(name) = ?{} THEN 0 ELSE 1 END, created_at DESC, id ASC",
                     data_params.len()
                 )
             } else {
-                "created_at DESC".to_string()
+                "created_at DESC, id ASC".to_string()
             };
 
             data_params.push(Box::new(limit_i64));
--- a/crates/khive-db/src/stores/entity.rs
+++ b/crates/khive-db/src/stores/entity.rs
@@ -822,12 +822,22 @@ impl EntityStore for SqlEntityStore {
 
             let columns = "id, namespace, kind, entity_type, name, description, properties, tags, \
                            created_at, updated_at, deleted_at, merged_into, merge_event_id, content_ref";
-            // CROSS JOIN is load-bearing: SQLite must drive the query from the
-            // sequence INTEGER PRIMARY KEY range instead of scanning the
-            // namespace index and sorting all matches into a temp B-tree.
+            // CROSS JOIN is load-bearing for the unfiltered walk: SQLite must
+            // drive the query from the sequence INTEGER PRIMARY KEY range
+            // instead of scanning the namespace index and sorting all matches
+            // into a temp B-tree. When a kind filter is active, forcing that
+            // same seq-first plan prevents SQLite from using
+            // idx_entities_kind(namespace, kind) as the driving index, so a
+            // plain JOIN is used instead and left to the query planner — the
+            // cursor's page order is still entities_seq.seq, unchanged.
+            let join_kind = if filter.kinds.is_empty() {
+                "CROSS JOIN"
+            } else {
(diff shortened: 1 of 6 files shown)


ohdearquant:
Resolved by the consolidated pagination vehicle in #1704 (merged as 94c6decd953ef60104166ea4c27b5551688a5908): list cursors now carry the (created_at, id) total order with the id tiebreak, the entity cursor path ports the kind-gated join, and ADR-002 records the ordering convention plus the keyset recommendation over offset paging under concurrent change.

PR #1742, opened 2026-08-06
test(db): add Windows walpin FFI temp-directory coverage and run it in CI

### LINKED ISSUE #1335: test(db): Windows walpin FFI paths have no executable test coverage
The Windows walpin implementation's real FFI paths — handle-relative open (`open_relative`), directory handle validation, DACL parsing (`validate_owner_only_dacl`), `NtCreateFile` usage, handle-bound rename (`rename_via_handle`), verified delete (`remove_relative_if_exists` / `delete_via_handle`), and mtime touch — are exercised only in production. The unit tests cover pure predicate helpers (path/metadata classification), not the syscall-bearing code.

This makes the raw Windows FFI difficult to maintain: a refactor of the rename or ACL path compiles and passes CI without ever executing.

Proposed direction: a `#[cfg(all(test, windows))]` integration suite driving the real functions against a temp directory (create/replace/delete cycles, DACL round-trip, reparse-point rejection), gated to run on the Windows CI job.

Relevant code: `crates/khive-db/src/walpin.rs` Windows module.

Severity: minor / test coverage.

ohdearquant:
The walpin sidecar's Windows FFI paths (open/`NtCreateFile`, DACL validation, rename-via-handle, delete-via-handle, mtime touch) had no executable coverage — only pure predicate helpers were tested, so a refactor could compile and pass the Windows job without ever exercising a syscall path.

- Adds a `#[cfg(all(test, windows))]` module driving the public sidecar wrappers against real `tempfile::tempdir()` directories: directory creation and idempotent re-validation, refusal of a pre-existing directory with a default (non-owner-only) ACL, refusal of a symlinked target, the write/replace/remove cycle, remove-on-missing as a no-op, and the beacon write/touch/remove cycle.
- Wires `cargo test -p khive-db --lib walpin` into the existing `windows-latest` job, scoped to that module to keep the lane fast.

Scope note: the syscall-bearing internals are private to `windows_impl`, so the tests drive the thin `#[cfg(windows)]` public wrappers that dispatch straight onto them. That reaches every path named in the issue while staying within the module's visibility.

Verification limit, stated plainly: the new tests **have not executed** — they were authored on a macOS host where `cfg(windows)` is false. What was verified locally is that they are correctly excluded from the non-Windows build (`cargo check -p khive-db --tests`, clippy, fmt all clean; `cargo test -p khive-db --lib walpin` → 67 passed with the new module contributing 0, as expected). **CI on this PR is the first execution.**

Closes #1335.

## Update: the new CI coverage surfaced a real Windows defect, now fixed here

The first executed Windows run of the coverage this PR adds failed 8 heartbeat/beacon tests. Root cause was in product code, not test scaffolding: `walpin`'s Windows atomic-rename path called `SetFileInformationByHandle` with the classic `FileRenameInfo` class and a non-null `RootDirectory`, which that class rejects with `ERROR_INVALID_PARAMETER` — directory-relative by-handle renames require `FileRenameInfoEx`. Windows heartbeat/beacon writes were therefore non-functional. Fixed by switching to `FileRenameInfoEx` with `REPLACE_IF_EXISTS | POSIX_SEMANTICS` and computing the buffer header size from the exact `file_name` field offset, matching the working pattern already in `khive-vamana`'s Windows rename. macOS suite unaffected (639/639).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -566,6 +566,13 @@ jobs:
       # widen once it has proven stable.
       - name: cargo check --workspace (Windows)
         run: cargo check --workspace
+      # #1335: the walpin sidecar's real Windows FFI paths (open/ACL/rename/
+      # delete/mtime touch) previously had no executable coverage — a
+      # refactor could compile and pass this job without ever running them.
+      # Scoped to just that module's tests (not the full suite) to keep this
+      # still-new lane fast; widen alongside the `cargo check` step above.
+      - name: khive-db walpin Windows FFI tests
+        run: cargo test -p khive-db --lib walpin
 
   coverage-ratchet:
     name: Coverage ratchet
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -569,7 +569,8 @@ jobs:
       # #1335: the walpin sidecar's real Windows FFI paths (open/ACL/rename/
       # delete/mtime touch) previously had no executable coverage — a
       # refactor could compile and pass this job without ever running them.
-      # Scoped to just that module's tests (not the full suite) to keep this
+      # The substring filter runs the walpin module plus
+      # checkpoint::tests::walpin_* (not the full suite) to keep this
       # still-new lane fast; widen alongside the `cargo check` step above.
       - name: khive-db walpin Windows FFI tests
         run: cargo test -p khive-db --lib walpin
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -569,9 +569,9 @@ jobs:
       # #1335: the walpin sidecar's real Windows FFI paths (open/ACL/rename/
       # delete/mtime touch) previously had no executable coverage — a
       # refactor could compile and pass this job without ever running them.
-      # The substring filter runs the walpin module plus
-      # checkpoint::tests::walpin_* (not the full suite) to keep this
-      # still-new lane fast; widen alongside the `cargo check` step above.
+      # The substring filter selects any test whose full path contains `walpin`,
+      # including tests outside the `walpin` module; it is not the full suite.
+      # Keep this still-new lane fast; widen alongside the `cargo check` step above.
       - name: khive-db walpin Windows FFI tests
         run: cargo test -p khive-db --lib walpin
 
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -4877,9 +4877,33 @@ mod tests {
             Duration::from_millis(500),
         )
         .expect("sidecar enabled for a file-backed path");
-        state.register_beacon().await;
         let pid = std::process::id();
+        state.register_beacon().await;
         let beacon_path = sidecar_dir.join(format!("{pid}.beacon"));
+        if !beacon_path.exists() {
+            // Windows diagnosis aid: `WalpinSidecarState::register_beacon`
+            // swallows a failed `write_beacon` behind a `tracing::warn`
+            // (fail-open logging), so the `metadata()` call below would
+            // otherwise only ever report a generic `NotFound`. Attempt the
+            // identical write synchronously, outside `register_beacon`'s
+            // `spawn_blocking` wrapper, so a real failure panics with the
+            // underlying OS error code instead — and so a difference in
+            // outcome between this direct call and `register_beacon`'s own
+            // (already-failed) call narrows the bug to the `spawn_blocking`
+            // indirection rather than `write_beacon` itself.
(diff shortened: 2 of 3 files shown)


ohdearquant:
The two failing tests on this branch are pre-existing Windows defects that this PR surfaces rather than introduces. Evidence, since the distinction changes what should happen next:

- `main`'s `ci.yml` contains no walpin test invocation at all. The `Windows compile check` job ran `cargo check --workspace` and nothing else, so the sidecar's Windows FFI paths compiled on every run and were never executed.
- Both failing tests (`checkpoint::tests::walpin_observe_drops_beacon_when_heartbeat_write_fails` and `checkpoint::tests::session_sweep_task_writes_and_clears_walpin_heartbeat`) exist on `main` today, and both pass on the ubuntu and macos shards of this same run. They are Windows-only, and this is the first time Windows has run them.

So the seven-line change is working exactly as intended. Its first execution found 33 passing tests and 2 real failures.

### The defect

Both failures are the registration beacon going unobservable. The panic output is unusually informative: it prints the sidecar directory contents, and the beacon file is present there while an open of its expected path returns `os error 2`. The listed name is `<pid>.beacon` followed by bytes that render as a replacement character, which means the name on disk carries trailing code units that are not valid UTF-16.

The enumeration that produced that listing is `std::fs::read_dir` with `to_string_lossy`, so the listing is not the thing that is wrong. The filename on disk is genuinely corrupt. The test's own note records the earlier symptom from a prior run: the write path reported success while the file stayed unobservable at the expected path.

For a WAL-pin liveness sidecar this is worth stating plainly. A beacon can land under a corrupted name while the writer reports success, so liveness detection reads a registered process as absent and healthy. The failure direction is toward looking safe, which is why nothing surfaced it before.

### Root cause is not isolated

The two obvious candidates are both correct as written:

- `open_relative` builds its `UNICODE_STRING` with exact byte lengths and no NUL padding.
- `rename_via_handle` sets `file_name_length` to the exact name length, zeroes the buffer, and reports the exact `total_size`.

The corruption is somewhere else. Rather than guess, I would rather leave this stated as unisolated. A fix validated only through CI round-trips against an FFI path that cannot be run locally is not a fix anyone should trust.

### Suggested disposition

Two reasonable paths, and the choice is a real trade rather than an obvious call:

1. Fix the beacon naming first, then land this. Correct ordering, but it holds 33 tests' worth of new coverage hostage to one hard bug on a platform that is awkward to iterate against.
2. Land the coverage now with these two tests marked `#[cfg_attr(windows, ignore = "...")]` naming a tracking issue, then fix separately. This captures the coverage immediately and keeps the defects visible, since an ignored test is reported in the test output rather than silently skipped.

Option 2 carries a real risk worth naming: this PR exists because a green check was not measuring anything, and excluding the two tests that proved it is the same move in miniature. That is survivable only if the exclusion is loud and tracked, never a bare `ignore`.

ohdearquant:
Correction to my disposition comment above: I listed two options, and there is a third, because I had assumed a red Windows result blocks this merge. It does not.

Verified against branch protection on `main`:

- Classic branch protection returns 404. Protection is entirely in the repository ruleset.
- That ruleset requires exactly two status contexts: `Secret scan (gitleaks)` and `CI gate`.
- `CI gate` is an aggregate whose `needs` covers 14 jobs.
- `check-windows` is the only job in `ci.yml` absent from that aggregate, and it is not independently required.

So `Windows compile check` is currently the single job in the file whose result gates nothing. It can be red without blocking anything.

That is consistent with the intent recorded in the workflow itself, at the step above the one this PR changes:

> Not (yet) a required check in branch protection. Plain `cargo check` (no `--all-targets`) keeps this fast while the Windows lane is new; widen once it has proven stable.

### The third option

Land this as-is, with the two tests failing. Coverage arrives immediately, the defect stays maximally visible, and nothing is blocked.

The cost is that a permanently red check invites people to stop reading it. That is the same failure this PR exists to correct, arriving from the other direction: first the job was green and meant nothing, then it is red and means nothing. So the choice between this and marking the two tests `ignore` is a genuine trade rather than an improvement, and it comes down to which is more likely to get the underlying bug fixed.

### The question underneath all three

The comment above says "widen once it has proven stable". This PR is that widening. So the decision worth making alongside it is whether `check-windows` should join `CI gate`'s `needs`, which is what would make Windows coverage load-bearing rather than advisory.

That reframes the disposition. Options 1 and 2 both aim at getting the job green so it can eventually be required. Option 3 accepts red in exchange for immediate visibility. None of them matter much while the job gates nothing, and all of them matter once it does.

I would not fold the required-set change into this PR. It is a separate decision with its own blast radius, and it should be made deliberately once the two failures are resolved rather than as a side effect of a test-coverage change.

PR #1805, opened 2026-08-09
feat(moodboard): add governed visual retrieval pack

ohdearquant:
> AI-assisted contribution: OpenAI Codex prepared this change and PR description under maintainer direction.

## Summary

- add the opt-in first-party `khive-pack-moodboard` with byte-exact BlobStore assets, typed core-graph `visual_asset` entities, an immutable Lattice visual descriptor, and exact model-space retrieval;
- pin the visual stack to published `lattice-embed = "=0.8.0"` and `lattice-inference = "=0.8.0"`, while keeping the governed preference artifact on `lattice-fann = "=0.7.1"`;
- include the already-reviewed ADR-149 preference loop merged through #1808: explicit randomized serves/judgments, deterministic pairwise BCE training, calibrated FANN serialization, restart-safe artifacts, and non-conformal preference probabilities;
- harden model lifecycle and provenance: cancellation-independent single-flight preparation/loading, native-work inference permits, a private verified checkpoint snapshot retained for the model lifetime, opened-handle containment, deterministic visible-namespace fanout, and exact-only vector writes without orphan ANN deltas.

## Contract notes

ADR-148 deliberately labels the pooled Qwen3.5 descriptor `experimental`. This PR proves immutable model/preprocessing identity, source-byte provenance, deterministic inference, and exact retrieval. It does **not** claim that raw cosine is a probability, that generic aesthetic quality equals reference-set fit, or that the base checkpoint is already a state-of-the-art style encoder.

The pack remains opt-in (`KHIVE_PACKS=kg,moodboard`) and never registers vision as a text embedder. Entity/blob work uses the shared core runtime; revision-specific visual vectors remain on the selected pack backend. Default multi-record search covers the token's authorized visible namespaces and deterministically reranks the union; an explicit namespace remains narrow per ADR-007.

Checkpoint preparation copies through checked opened handles into a private read-only snapshot, hashes and validates that snapshot, loads Lattice only from it, reverifies the complete descriptor after construction, and retains the snapshot for mmap lifetime. This is intentionally expensive cold-start governance rather than hash→reopen best effort.

ADR-149's preference head is separate from conformal coherence scoring. Explicit judgments may personalize ranking, but preference probabilities do not alter conformal guarantees.

## Current verification

Current head: `96aca1542d7ec01e0a7fe3450de673a5ba076dc3` on current main `6357ab49bf2d3e9f60b9753a9564541042080eee`.

[...]

PR #1806, opened 2026-08-09
fix(kg): stabilize filtered list envelopes

### LINKED ISSUE #1779: list(kind=..., tags=[...]) is accepted but inert: full population returned wearing a filtered appearance
Found during a client census over kind=note.

## Reproduction

```
kkernel exec 'list(kind="note", limit=3)'
kkernel exec 'list(kind="note", tags=["source-integrity"], limit=3)'
```

Both return rc=0 and byte-identical payloads (verified by string comparison of the two responses, 5842 chars each). The first row of the "filtered" call carries no such tag. The `tags` parameter is neither applied nor rejected.

## Why this is worse than an error

The caller gets the full population wearing the appearance of a filtered one. A census counting notes by tag would confidently report the entire corpus (159K+ rows) as matching. Nothing in the response hints that the filter was ignored.

The surface is also internally inconsistent: `comm.send` rejects an unknown field loudly (e.g. `from_` errors with "unknown field ..., expected one of ..."), so the surface teaches callers that unknown params fail loud — and then this one fails quiet.

## Expected

Either reject `tags` as an unknown parameter, or implement the filter. Both are fine; the silent middle is the only bad option.

For contrast, `query()` with `properties.tags CONTAINS ...` does filter server-side correctly (verified against a positive control), and its compile errors name the valid fields — that is the behavior this parameter is missing.

### LINKED ISSUE #1780: list limit>200 silently clamps AND changes the result shape
Found during a client census over kind=note.

## Reproduction

```
kkernel exec 'list(kind="note", limit=200)'   # result is a LIST of row objects
kkernel exec 'list(kind="note", limit=500)'   # result is {"effective_limit":200,"items":[...]}
```

Two independent failure modes from one call:

1. **Silent undercount.** A pager that requests 500 rows and advances `offset += 500` skips 300 rows per page — a ~60% undercount that terminates normally, reports a plausible total, and never errors.

2. **Shape break.** Code doing `result[0]` gets a `KeyError` on the dict shape. The shape change only appears on the clamped path, so callers discover it in production, not in tests written against small limits.

## Assessment

Surfacing `effective_limit` is the right instinct — it is exactly the signal a caller needs — but returning it in a different shape means the callers who would benefit are precisely the ones that crash before reading it.

## Expected

Keep the clamp, but either surface `effective_limit` in both shapes or return the same shape always.

## Client workarounds (until fixed)

- advance pagers by `len(rows)` rather than the requested limit
- handle both shapes: `res["items"] if isinstance(res, dict) else res`

### LINKED ISSUE #1789: list() changes its response type when the limit is clamped, so the truncation signal breaks the parser that needs it
## Observed

`list()` returns two different response shapes depending on the requested limit. Measured against a single-backend sqlite config holding 843 entities:

| `limit` | `results[0].result` |
| --- | --- |
| 1 | array, 1 element |
| 400 | array, 400 elements |
| 499 | array, 499 elements |
| 500 | array, 500 elements |
| 501 | object `{effective_limit, items, limit_clamped, requested_limit}`, 500 items |
| 900 | object, same keys, 500 items |

The flip happens exactly at `limit > 500`, which is exactly when the server clamps.

## Why this is worth fixing rather than documenting

The object form carries good information: `limit_clamped`, `requested_limit`, `effective_limit`. That is precisely the disclosure a caller wants. The problem is that it arrives by **changing the type of `result`** rather than as an additional field.

The consequence is backwards. A client written against the array form works for every request that returns complete results, and breaks on exactly the requests where results were truncated. So the failure lands on the case the client most needs to handle, and it lands as a type error rather than as a truncation signal.

The defensive workaround has its own failure: a caller who writes `result if isinstance(result, list) else result["items"]` keeps running and silently discards `limit_clamped`, which converts a loud break into a quiet undercount. Both readings of this API lose the signal — one by crashing, one by ignoring it.

This was hit while paginating an 843-entity store. The first read at `limit=900` returned 4 where a naive length check was applied to the object's keys, which looks like an almost-empty store rather than a clamped page.

## Suggested shape

Return one type for all limits. Either always the object, or keep the array and move `limit_clamped` / `effective_limit` / `requested_limit` into the envelope alongside `ok` and `tool`, where a caller reads them without branching on the payload type.

Whichever is chosen, the property worth holding is that a caller cannot distinguish complete from truncated by accident: the clamp flag should be present and readable on every response, not only on the clamped ones.

ohdearquant:
## Summary

- make `list(tags=[...])` a real server-side filter for entity and note substrates, with deterministic tag matching and explicit unsupported-substrate validation
- return stable list envelopes at every requested limit, always disclosing `requested_limit`, `effective_limit`, and `limit_clamped`
- preserve structural arrays and terminal cursor nulls through Agent presentation while keeping offset and cursor envelope variants explicit
- migrate repository consumers and pagers to the stable shape and advance by the rows actually returned

## Contract details

Offset pages use `items`; cursor pages use exactly one of `entities`, `notes`, or `edges` plus `next_after`. Every variant carries the same limit metadata. Zero limits are handled consistently with the effective floor disclosed, and bounded event scans retain `scan_incomplete` rather than implying completeness.

The public contract schema and its exhaustive shape-matrix tests now reject mixed or incomplete envelope variants while continuing to allow legitimate extension fields.

## Verification

- independent exact-head review: READY with zero findings at `fc746cfda87cdb3285d59cd7891041afc33cdfc9` on exact live base `2dc0bbc8b188eaeba39e9c66f8d5c0f4df97cae4`
- focused schema suite: 14 passed
- workspace-ingestion suite: 46 passed
- exact-head full CI: [run 31306518301](https://github.com/ohdearquant/khive/actions/runs/31306518301) — passed
- direct Rust formatting/parser check, Python AST parse, Deno formatting, and `git diff --check`

Fixes #1779
Fixes #1780
Fixes #1789


ohdearquant:
Two mechanical items before this can gate:

1. The branch now conflicts with main in `AGENTS.md`, `crates/khive-mcp/src/serve.rs`, and `crates/khive-mcp/tests/integration.rs` (main moved under it via #1897, #1805, #1816, and the read-identifier change). A rebase onto current main is needed.
2. The 2026-08-10 push of the current head never triggered the CI workflows (only the auto-merge guard ran), so there is no completed CI run at this commit. I manually dispatched one as interim evidence, but the rebased head will need its own full green matrix, including the shard runs.

On content this looks right: the always-on `{items, requested_limit, effective_limit, limit_clamped}` offset envelope plus the presentation-layer preservation of empty envelope arrays and `next_after: null` closes the empty-page/shape-branching problems, the note `tags` filter is applied before the limit in both scan paths, zero-limit now honestly returns zero rows, and the only production consumers of the old bare-array shape (the two channel outbox loops in `serve.rs`) are updated in the same change. One thing to double-check on the rebase: the new rejection test for a `tags` filter on `kind="proposal"` — the proposal listing is fast-pathed before the shared parameter handling, so make sure that path actually rejects the argument rather than ignoring it.

ohdearquant:
CI dispatched at 0dcd5773 completed: every build/test/lint job is green; the only failure is Supply-chain (cargo-deny), which flags `lru 0.12.5` (RUSTSEC-2026-0253) from this branch's lockfile. That pin was removed on main by #1894 (8/11), after this branch's merge-base, and this PR does not touch `crates/Cargo.lock` — so the rebase onto current main requested above clears the advisory with no change needed here. No action on the advisory itself; just rebase.

PR #1809, opened 2026-08-09
build: restore verification-only fleet targets

### LINKED ISSUE #1785: build: restore a build-only fleet verification pair (fleet-build / fleet-check) in the Makefile
The Makefile currently exposes only `local`, which builds AND installs in one step. That ordering means the first verification that the freshly built binary loads the full pack set happens after the previously installed daemon has already been replaced: a bad build is discovered only once the serving binary is gone.

Restore a two-step pair:

- `fleet-build` (or equivalent): build the release binary and run the pack-set verification against the BUILT artifact in place, without touching the installed binary.
- `fleet-check`: the verification step alone, runnable against either the built artifact or the installed binary.

`local` can then compose as build-verify-install, with the install step gated on the verification passing.

Acceptance: a deliberately broken pack registration fails `fleet-build` while the installed daemon keeps serving, and the happy path leaves `local` behavior unchanged. Measured 2026-08-06 by Makefile rule enumeration: no build-only verification target exists today.

ohdearquant:

## Summary

- add `fleet-build` and verification-only `fleet-check` Make targets without
  installing artifacts or stopping the daemon
- verify the receipt-bound Cargo artifact by default, with an explicit
  `FLEET_ARTIFACT` override for an exact existing executable
- carry arbitrary valid artifact paths through Make and the shell without
  evaluating or mangling `$`, quotes, backticks, or `$(shell ...)`
- align ADR-026 and build documentation with the two verification modes

## Verification

  `060d8f190fb5513abb16784223bce6a5704cc781`
- full exact-head CI: [run 31291670836](https://github.com/ohdearquant/khive/actions/runs/31291670836) — success
- 31 Python build-verifier tests, including real Make invocations with hostile
  literal paths and fail-before-install controls
- real `make fleet-check FLEET_ARTIFACT=$HOME/.cargo/bin/kkernel` probe
- `cargo test --workspace`
- `cargo check --workspace`

## ADR

- ADR-026

Fixes #1785


ohdearquant:
Two notes from a pass over the Makefile changes:

1. `fleet-check` interpolates `$(FULL_PACKS)` into shell source, which technically admits command injection via a make-variable override. This matches the existing pattern in the other recipes (the same interpolation ships on main in `verify-local-artifact` and the smoke recipe), and a caller who can override make variables already has shell, so it is not treated as a defect introduced here. For consistency it would still be reasonable to route it the same way `FLEET_ARTIFACT_VALUE` is routed (env-exported, shell-quoted), which this PR already does for the documented caller-facing knob — that hardening can land as a follow-up covering all three recipes.

2. `scripts/tests/test_verify_local_artifact.py:1203-1209` slices the recipe text between exact `fleet-check:` and `clean:` markers; harmless target reordering would break it or silently inspect the wrong block. Target-aware parsing (or asserting behavior instead of text) would make it robust.

ohdearquant (commented):

Verdict on head 4d36a9ff6ea9b9ba175bc5724c65bd7ac2ce1409: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
CI at the current head fails in `MakefileGateContractTests` for a reconcilable reason: the Makefile in this branch ships the hardened target forms (`verify-local-artifact: validate-full-packs build-local`, `fleet-check: validate-full-packs`), while two contract assertions still expect the earlier ungated spellings — `scripts/tests/test_verify_local_artifact.py` asserts `"verify-local-artifact: build-local\n"` and `"fleet-check:\n"`, and the slice anchor `self.makefile.index("fleet-check:\n")` will raise once only one site is updated. Updating those three sites to the gated spellings should make the gate green; the macOS shard that was cancelled by fail-fast will also need its rerun at the new head.

The target design itself looks right: the `$(value)` literal capture plus env-only transport and the `validate-full-packs` allowlist close the shell-splice hole for both the default and command-line-override paths, and the ADR-026 wording change correctly carves the explicit-artifact mode out of the "never reads the installed binary" claim instead of contradicting it.

ohdearquant (commented):

oceanwaves630 (approved):
Approving at ab06bf3fe7d7f9974192cf81ad90b3122d3d1e71.

Mechanical checks at that exact commit: head unchanged at the moment of this
approval, check rollup has 19 successful and 7 skipped with no failing and no
pending runs, the PR is out of draft, and auto-merge is not armed.

This approval records the automated review's clean result at this same commit
together with those checks. It is not an independent line-by-line read of the
diff.

If the head moves, this approval no longer applies to it.

PR #1810, opened 2026-08-09
fix(db): drop transaction span before writer reply

### LINKED ISSUE #1790: Writer task acknowledges a completed transaction before deregistering its tx-registry span
## Summary

The writer task replies to its caller before dropping the RAII guard that holds the
transaction's `tx_registry` span. A caller that has already awaited its write can
therefore observe a committed transaction while the registry still reports an open
span for it.

## Mechanism

`crates/khive-db/src/writer_task.rs` registers the span at the top of the guarded
block (`register_scoped`, ~line 698) and the guard is dropped only when the blocking
closure returns (~line 726). The reply is sent from inside that block: for the commit
path, `reply.send(Ok(value))` in `execute_and_reply_reporting_terminal` (~line 195),
and separately on the `BEGIN IMMEDIATE` error path (~line 715). The caller resumes
from `reply_rx.await` (~lines 414-419) as soon as that send lands, which is before
deregistration.

`crates/khive-storage/docs/api/tx-registry.md` defines the registry as tracking
currently-open SQL transaction spans, so for the duration of that window the registry
contradicts its own documented contract.

## Why it matters beyond tests

`db_diagnostics` reports writer-contention and WAL-pin figures derived from registry
state. Spans that outlive their transaction inflate those figures, and the inflation
is proportional to write throughput, which is exactly when the diagnostics are being
consulted.

It also makes aggregate registry queries unreliable for callers reasoning about a
specific span: a write span lingering past its commit is visible to any aggregate
predicate over the same backend. That is observable today in the graph store, where
the read path registers `graph_traverse_read` and the write paths register
`graph_upsert_edges` / `graph_upsert_edge_guarded` on the same backend.

## Suggested direction

Defer every reply, on both the success and the `BEGIN IMMEDIATE` error path, until
after the transaction guard has dropped.

## Scope note

This is a source-level ordering defect and is independent of any individual test's
choice of assertion. It is filed separately for that reason.

ohdearquant:

## Summary

- keep the writer-task transaction registry span alive through commit/rollback processing
- drop that span before the typed oneshot reply can wake the caller
- preserve the same ordering when `BEGIN IMMEDIATE` fails without running the request
- add deterministic reply-wake regressions for both successful writes and failed `BEGIN`

## Verification

  `1a63fc6e5164562831c2f47e215cfe589b32f535`
- full exact-head CI: [run 31291706085](https://github.com/ohdearquant/khive/actions/runs/31291706085) — success
- focused writer reply/transaction-registry lifecycle regressions
- `cargo test --workspace`
- `cargo check --workspace`

Fixes #1790

### DIFF
--- a/crates/khive-db/docs/api/writer-task.md
+++ b/crates/khive-db/docs/api/writer-task.md
@@ -51,6 +51,10 @@ classification table in `writer_task.rs`'s module docs still opens its own
 writer; strict routing per ADR-136 D1 has not landed) replies the request's
 error via `AnyWriteRequest::reply_error` without ever invoking the
 request's operation closure via `AnyWriteRequest::execute_and_reply`.
+For transaction-wrapped requests, the scoped `writer_task_tx` registry span
+is dropped before the oneshot reply wakes the caller, both after a completed
+transaction and after a failed `BEGIN`. A caller that has observed its reply
+therefore cannot still observe that request as an open SQL transaction.
 There is no watchdog/retry story for a failed `BEGIN` (ADR-067
 Component D remains future work); the connection simply tries
 `BEGIN IMMEDIATE` fresh on the next request.
--- a/crates/khive-db/src/writer_task.rs
+++ b/crates/khive-db/src/writer_task.rs
@@ -88,6 +88,7 @@ mod sealed {
         fn execute_and_reply_reporting_terminal(
             self: Box<Self>,
             conn: &rusqlite::Connection,
+            tx_span: Option<khive_storage::tx_registry::TxHandle>,
         ) -> Option<khive_storage::error::WriterTaskRequestState>;
 
         fn execute_and_reply_top_level_reporting_terminal(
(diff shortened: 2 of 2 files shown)


ohdearquant:
The ordering change is right for the outcomes where the transaction actually closed, but the
guarantee added to the docs is unconditional and does not hold on the unknown-state paths.

`crates/khive-db/src/writer_task.rs:265` drops `tx_span` for every outcome of
`execute_wrapped_transaction`, ahead of the reply at `:268`. Not every outcome has closed the
transaction:

- `:195-205`, where `COMMIT` returns `Ok` but `conn.is_autocommit()` is false. The log line
  there says it outright: the connection is still in a transaction. Returns `SideEffectsUnknown`.
- `:214-220`, `:226-232`, `:238-240`, where `rollback_after_failure` yields
  `SideEffectsUnknown`. That helper classifies those cases, it does not close them.

On those paths the span is deregistered while the connection has not been retired, so
`any_open_labeled(.., "writer_task_tx")` reads false with a writer connection still in an open
or unknown transaction. That is the false-negative direction: the diagnostic that exists to find
a stuck writer reports clean exactly when there is something to find.
`crates/khive-db/docs/api/writer-task.md:54-57` now states the guarantee with no qualification,
that a caller which has observed its reply cannot still observe the request as an open SQL
transaction, and on these paths it is inverted.

Two ways out. Hold the span through terminal teardown until the connection is actually retired,
or scope the documented guarantee to the outcomes where the transaction is known closed. The
first is the stronger fix. Either way, a regression driving a real span through a terminal or
unknown outcome would pin it: the added tests exercise `Some(tx_span)` only on the normal path,
and the direct terminal tests pass `None`.

Separately, and lower priority. `ParkedWake::wake` performs an unbounded `recv()` inside
`Wake::wake`, while the registry wait beside it carries a five-second bound. Blocking in a waker
violates the contract even where the current two-worker scheduling survives it, and it means any
failure that leaves the sender alive wedges the test instead of failing it. A waker that records
entry and returns immediately, with a separate bounded barrier for the observation window, keeps
the oracle and removes the hang.

PR #1811, opened 2026-08-09
test(storage): cover bounded note-store admission

### LINKED ISSUE #1803: Bounded write admission through SqlNoteStore::with_writer_tx_storage is untested
`SqlNoteStore::with_writer_tx_storage` routes its writer request through
`WriterTaskHandle::send_bounded`, so a saturated write queue is meant to fail on the
configured admission deadline (`PoolConfig::write_admission_deadline_ms`) and surface
`StorageError::WriteQueueFull` rather than wait unbounded.

No test asserts that behaviour through the note store.

Enumeration over `crates/` for `WriteQueueFull|write_admission_deadline_ms|send_bounded`
(positive control in the same pass: `git grep -c "with_writer_tx" -- 'crates/*'` returns
`stores/note.rs:10`, `stores/note_tests.rs:4`, so the query surface is live) puts every
existing assertion in one of two places:

- `crates/khive-db/src/writer_task.rs` (~1050, ~1116, ~1180) — handle-level unit tests that
  call `WriterTaskHandle` directly and never enter a note store.
- `crates/khive-mcp/tests/integration.rs:1640` — a synthetic pack whose `dispatch`
  manufactures `StorageError::WriteQueueFull { timeout_ms: 175 }` to exercise the error
  shape. No database, no note store.

`crates/khive-db/src/stores/note_tests.rs` has write-queue routing coverage (`queue_depth`,
`write_queue_enabled: Some(true)`) but no admission-deadline or `WriteQueueFull` assertion —
it does not appear in the enumeration above at all.

Consequence: changing `send_bounded` back to `send` at `crates/khive-db/src/stores/note.rs:325`
reddens nothing in the suite. The mechanism is protected only by the call site being written
correctly, and a regression there would fail in the silent-wait direction, which is the
direction that does not announce itself.

This is pre-existing rather than new — `with_writer_tx` had the same untested property before
the bounded call was introduced, so nothing regressed. Filing so the gap is tracked rather
than rediscovered.

Suggested coverage: a note-store-level test that saturates the writer queue against a
short `write_admission_deadline_ms` and asserts `WriteQueueFull` with the configured
`timeout_ms`, mirroring `configured_enqueue_timeout_rejects_only_unaccepted_request` but
entering through `SqlNoteStore`.

ohdearquant:

## Summary

- add deterministic store-level coverage for `SqlNoteStore`'s bounded writer-task admission
- saturate the real file-backed writer queue with one executing request and one queued request
- prove a third `upsert_note` returns the configured `StorageError::WriteQueueFull` deadline
- prove the rejected note is never executed after queue capacity becomes available

## Verification

  `2a71b78560b9c86585b75c4f99f353c284154be6`
- full exact-head CI: [run 31291710079](https://github.com/ohdearquant/khive/actions/runs/31291710079) — success
- focused `SqlNoteStore` bounded-admission regression
- `cargo test --workspace`
- `cargo check --workspace`

Fixes #1803

### DIFF
--- a/crates/khive-db/src/stores/note_tests.rs
+++ b/crates/khive-db/src/stores/note_tests.rs
@@ -1836,3 +1836,110 @@ async fn upsert_note_routes_through_writer_task_when_flag_enabled() {
         "note must be committed and readable after queuing behind the occupier"
     );
 }
+
+/// #1803: the note store's transaction-wrapped write path must preserve the
+/// writer task's bounded admission contract. With one request executing and
+/// the sole queue slot occupied, `upsert_note` must return the configured
+/// `WriteQueueFull` error without accepting or later executing the note write.
+#[tokio::test]
+async fn upsert_note_reports_configured_write_queue_admission_deadline() {
+    let dir = tempfile::tempdir().unwrap();
+    let path = dir.path().join("note_store_bounded_admission.db");
+    let pool_cfg = PoolConfig {
+        path: Some(path),
+        write_queue_enabled: Some(true),
+        write_queue_capacity: 1,
+        write_admission_deadline_ms: 100,
+        ..PoolConfig::default()
+    };
+    let pool = Arc::new(ConnectionPool::new(pool_cfg).unwrap());
+    {
+        let writer = pool.writer().unwrap();
+        writer.conn().execute_batch(NOTES_DDL).unwrap();
+    }
+
+    let store = SqlNoteStore::new(Arc::clone(&pool), true);
+    let writer_task = pool
+        .writer_task_handle()
+        .unwrap()
+        .expect("writer task must be enabled for the file-backed pool");
+
+    // A is dequeued and blocks inside the real writer-task transaction,
+    // leaving the bounded channel empty but preventing the drain loop from
+    // accepting another request.
+    let (started_tx, started_rx) = tokio::sync::oneshot::channel::<()>();
+    let (release_tx, release_rx) = std::sync::mpsc::channel::<()>();
+    let a_task = {
+        let writer_task = writer_task.clone();
+        tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


ohdearquant:
Two things worth addressing before this lands.

**The "rejected note is never executed" assertion cannot fail for the reason it exists.**

After the third `upsert_note` returns `WriteQueueFull`, the check that the note was never
written goes through `get_note`, which reads from the connection pool
(`crates/khive-db/src/stores/note.rs:1122` dispatches through `with_reader`). A wrongly
accepted write would settle on the writer queue, and nothing in the test orders the queue
drain against that read.

So a `None` result is consistent with two different worlds: the write was correctly rejected,
or it was wrongly accepted and simply has not been applied yet. The assertion passes in both.
It therefore cannot detect the admission defect it is meant to guard, and it would keep
passing if bounded admission regressed.

There is already a working pattern for this in the same crate.
`crates/khive-db/src/writer_task.rs:1107-1119` hands the operation closure an `AtomicBool`
that the closure sets when it actually runs, then asserts the flag is still `false`. That is
an execution oracle rather than a state read, so it is unaffected by drain timing. Either
that, or an explicit drain barrier ahead of the read, closes the gap.

**Test isolation looks file-wide rather than incidental.**

`crates/khive-db/src/writer_task.rs:834-842` documents why tests touching the shared
transaction registry need `#[serial(tx_registry)]`, and the sibling test there carries it.
The new test drives a real file-backed writer task and carries only `#[tokio::test]`.

Counting across the whole file rather than the surrounding lines: `note_tests.rs` contains
zero `serial(` attributes. So this is not one test missing an attribute, it is a file with no
isolation at all against shared writer and registry state, and the remedy is file-scoped
rather than line-scoped.

PR #1817, opened 2026-08-09
fix(mcp): close the request envelope contract

### LINKED ISSUE #1813: ADR-045 and ADR-078 document an include_full_id envelope override that is not implemented, and unknown envelope fields are silently discarded
## Summary

ADR-045 and ADR-078 both document an envelope-level `include_full_id=true` override that
callers can pass to retain `full_id` in a response. That parameter is not implemented.
Passing it has no effect and produces no error.

## Evidence

At `origin/main`:

- `include_full_id` appears 7 times, all in ADR prose: 4 in
  `docs/adr/ADR-045-verb-response-presentation.md` and 3 in
  `docs/adr/ADR-078-output-format-shape-aware-rendering.md`.
- It appears 0 times anywhere under `crates/`. A control search for `presentation` over the
  same path returns 519 hits, so the search is live against those files and the zero is a
  real absence rather than a dead query.

`RequestParams` in `crates/khive-mcp/src/tools/request.rs` is the envelope struct, and the
MCP tool schema is generated from it. Its full field set is:

```
ops, presentation, presentation_per_op, save_to, format, format_per_op, request_id
```

There is no `include_full_id`. The struct carries no `#[serde(deny_unknown_fields)]`, so an
unknown envelope field is silently discarded rather than rejected. A caller who follows the
ADRs gets no override and no diagnostic.

## What the ADRs currently claim

ADR-045 has a section titled ``### `include_full_id` override (envelope-level)`` stating
that "callers may pass `include_full_id=true` at the envelope level (alongside
`presentation`) to force full UUIDs in the response". Its ADR-078 note repeats that
`full_id` is "retained in `format=json`, in `PresentationMode::Verbose`, and when
`include_full_id=true` is set".

ADR-078 §7.1 cites the same override, and its Neutral section states that it "continues to
work and takes effect before the `full_id` suppression rule in §7.1".

Two of the three retention conditions are real and verifiable in `presentation.rs`. The
third is not.

## Impact

The ADRs are the normative contract. A caller reading them to work out how to keep
`full_id` under `format=auto` or `format=table` will reach for a parameter that silently
does nothing, and the silence makes it look like the field was accepted. The two mechanisms
that do work are `format=json`, which is the builtin default, and `presentation=verbose`.

## Suggested resolution

Either implement the override, or amend both ADRs to remove it and state only the two
conditions that hold. Amending is the smaller change and matches the code as it stands.

If the override is removed rather than implemented, ADR-078's Neutral section needs the
matching edit, since it asserts continuity of a behaviour that was never present.

## The general class, which is the part worth fixing

The specific parameter is one instance. The reusable point is that the envelope does not
reject unknown fields, so *any* parameter that gets documented but never implemented
becomes silently accepted rather than loudly refused. The caller cannot tell the difference
between a parameter that worked and a parameter that was discarded, because both look
identical from the outside: a successful response.

There is no rejection at the schema layer either. The JSON Schema generated for the
`request` tool lists `properties` and `required: ["ops"]` but does not set
`additionalProperties: false`, so an unknown field is not refused before it reaches serde,
and serde without `deny_unknown_fields` discards it.

That makes documentation drift undetectable at runtime. A misspelling of a real parameter
has exactly the same shape and is equally silent: `presentaton=verbose` is discarded, the
call succeeds, and the response is rendered in the default mode.

Adding `#[serde(deny_unknown_fields)]` to `RequestParams` would turn this whole class into
an immediate error at the call site. It is a breaking change for any caller currently
sending fields that are being discarded, which is worth knowing about on its own, since
every such caller is already not getting the behaviour it asked for.

If that is too aggressive for a point release, a middle option is to accept unknown fields
but include them in the response as an explicit `ignored_fields` list, so a caller can see
that what it sent did not land.

## Related open question

Separately, and worth deciding rather than inheriting: should `format=auto` and
`format=table` drop `full_id` at all?

- One reading: they are human-facing render formats, `full_id` is view-only noise there,
  and any caller that needs a chaining handle belongs on `format=json` or
  `presentation=verbose`. This is what ADR-078 §7 already decided.
- The other reading: `full_id` exists precisely so a result can be chained into a strict
  by-id verb without a second lookup, and dropping it in some formats means the same verb
  returns a chainable result or an unchainable one depending on a rendering flag, which is
  a shape difference callers have to special-case.

Both readings are defensible and the current behaviour follows the first. Recording the
question here so the answer is a decision rather than a default.

ohdearquant:

## Summary

- reject unknown fields in the outer MCP `request` envelope instead of silently
  discarding misspellings or undocumented parameters
- close the generated request-tool JSON Schema with `additionalProperties: false`
- withdraw the never-implemented `include_full_id` override from ADR-045 and ADR-078
- document the two canonical-ID retention paths that actually ship: lossless
  `format=json` and `presentation=verbose`

## Verification

- focused serde rejection and generated-schema closure regressions
- independent exact-head review: READY with zero findings at `ebaf2786d85f551a72a79ddb972a0bb2bde80855`
- exact-head full CI: PASS ([run 31295264893](https://github.com/ohdearquant/khive/actions/runs/31295264893))
- `cargo test --workspace`
- `cargo check --workspace`

## ADR

- ADR-045 Amendment 1
- ADR-078 Amendment 1

Fixes #1813

### DIFF
--- a/crates/khive-mcp/README.md
+++ b/crates/khive-mcp/README.md
@@ -12,7 +12,8 @@ lifetime.
 ## Features
 
 - **One tool, `request`** — `RequestParams { ops, presentation, format, save_to, .. }` is
-  the entire MCP-visible surface (ADR-016); verb-specific schemas live in packs, not here
+  the entire MCP-visible surface (ADR-016); its outer envelope is closed to undeclared
+  fields, while verb-specific schemas live in packs
 - **Pluggable transports** — `Transport` trait + `TransportRegistry`; ships `StdioTransport`,
   open for more (e.g. Streamable HTTP) via `TransportRegistry::register`
 - **Daemon-aware dispatch** — `compute_config_id` fingerprints a resolved `RuntimeConfig`
--- a/crates/khive-mcp/src/tools/request.rs
+++ b/crates/khive-mcp/src/tools/request.rs
@@ -128,7 +128,7 @@ mod tests {
 
     #[test]
     fn request_tool_schema_is_closed() {
-        let schema = schemars::schema_for!(RequestParams);
+        let schema = rmcp::schemars::schema_for!(RequestParams);
         let value = serde_json::to_value(schema).expect("serialize request schema");
 
         assert_eq!(
--- a/crates/khive-mcp/src/tools/request.rs
+++ b/crates/khive-mcp/src/tools/request.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 /// Input for `request` — a DSL string (function-call or JSON form) plus
 /// optional presentation controls (`presentation` and `presentation_per_op`).
 #[derive(Debug, Default, Serialize, Deserialize, schemars::JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct RequestParams {
     /// One or more operations as a function-call DSL or JSON-form string.
     ///
--- a/crates/khive-mcp/src/tools/request.rs
+++ b/crates/khive-mcp/src/tools/request.rs
@@ -105,3 +105,35 @@ pub struct RequestParams {
     )]
     pub request_id: Option<u64>,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::RequestParams;
+    use serde_json::json;
+
+    #[test]
+    fn request_params_reject_unknown_envelope_fields() {
+        let error = serde_json::from_value::<RequestParams>(json!({
+            "ops": "verbs()",
+            "presentaton": "verbose"
+        }))
+        .expect_err("misspelled request-envelope fields must be rejected");
+
+        assert!(
+            error.to_string().contains("presentaton"),
+            "the validation error must name the rejected field: {error}"
+        );
+    }
+
+    #[test]
+    fn request_tool_schema_is_closed() {
+        let schema = schemars::schema_for!(RequestParams);
+        let value = serde_json::to_value(schema).expect("serialize request schema");
+
+        assert_eq!(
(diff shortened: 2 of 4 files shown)


ohdearquant:
The envelope closure checks out. `RequestParams` is not deserialized on the daemon path, where
the frame is constructed and reconstructed field by field against a separate frame type, so the
rejection stays at the outer tool boundary as the amendment says. The schema the new test asserts
is also the advertised one rather than a look-alike: `#[tool_router]` generates
`schema_for_input::<Parameters<RequestParams>>()` and `Parameters` delegates `JsonSchema` to
`RequestParams`, and with a single `schemars` 1.2.1 in the lockfile and no direct dependency in
`khive-mcp`, derive and macro resolve to the same copy.

One gap, and it is in the amendments rather than the code. The two halves of this change interact
in a way neither ADR states. `include_full_id=true` was previously accepted and silently ignored;
it is now a hard parameter-validation failure at the MCP boundary, before the daemon is reached.
A caller who followed the documented ADR-045 envelope therefore moves from a harmless no-op to a
broken call. Each amendment states one half, that the field is withdrawn and that unknown fields
are rejected, but neither connects them into a migration instruction.

One sentence in each amendment would cover it: legacy `include_full_id=true` calls now return
invalid parameters, and callers needing canonical IDs should select `format=json` or
`presentation=verbose`. The sweep for that field finds only the two ADR mentions being edited,
with no skill, example, fixture, or pack sending it, so the blast radius is small. It is just
that the affected reader is precisely the one who followed these ADRs.

Issue #1426, opened 2026-07-31
tests: provision cross-namespace contract suites for fail-closed authorization


ohdearquant:
After a prior test-directory reorganization moved Python CI test files without updating `scripts/ci.sh` and the scripts' own default binary paths, contract tests still issue cross-namespace writes as an unprovisioned principal. The runtime correctly denies those writes under the fail-closed authorization default, leaving the suites unable to pass end-to-end.

## Current behavior

With the paths repaired, running the four Python CI phases gives:
- `deno-tests`: PASS
- `smoke-tests`: PASS
- `vector-smoke`: PASS
- `contract-tests` (`tests/core/contract_test.py`): FAIL at the first cross-namespace write —
  ```
  permission denied for verb "create": principal 'anonymous:local' holds no capability for namespace 'ns-alpha'
  ```
  The `test_namespace_isolation` section (the namespace-blindness regression guard) writes into `ns-alpha`/`ns-beta` from an unprovisioned principal.
- `contract-suite` (`tests/core/khive-contract`): 75 failed / 17 passed / 2 xfailed, every failure the same denial against per-test random namespaces.

These failures predate the path breakage: an unprovisioned `anonymous:local` principal could once write into arbitrary namespaces, and the fail-closed authorization default now correctly refuses it.

## Expected behavior

Define and implement a test-harness provisioning posture that permits the intended fixture setup without weakening the authorization behavior under test — for example, an explicitly permissive test runtime, or minting a capability for the test namespaces at boot. That decision needs to interact carefully with the namespace-isolation regression guard, whose purpose is to catch reintroduction of by-ID namespace filtering: its own setup path must be provisioned without weakening what it guards.

## Acceptance
- Contract fixtures are explicitly authorized, or assert denial where that is the intended behavior.
- The namespace-isolation regression remains covered.
- The suite runs from its documented entry points, and CI can pass end-to-end.

Issue #1376, opened 2026-07-31
Use a canonical server configuration for authorization roster resolution


ohdearquant:
The gate caller roster is discovered from the daemon's working directory at boot. When clients share a server and database, a daemon can start with a roster from the first client's project configuration instead of the deployment configuration. That makes authorization policy depend on startup order and can deny callers that should be authorized.

## Current behavior

Configuration discovery is cwd-anchored: with no explicit database path set, the resolver reads the project-local configuration in the process's current working directory as the first match. When multiple clients share one server and one database but are launched from different working directories, whichever client's directory happens to provide the daemon's boot configuration determines the roster everyone else gets.

This is a known gap against [ADR-129](docs/adr/ADR-129-fail-closed-gate-default.md) Amendment 2's requirement that the caller roster live in configuration common to all spawn paths: no single canonical location currently satisfies that requirement for a shared-database deployment. In practice the roster must be duplicated into every configuration file a daemon might spawn under, and any new project directory can silently reintroduce a rosterless boot — a roster present only in a fallback tier has been observed to produce a daemon that denies every principal it should have authorized.

## Proposed direction

Define a canonical, working-directory-independent anchor for gate roster resolution. Candidate approaches, in rough order of preference:

1. Resolve the roster section from the same configuration tier the database path resolves from, so every process serving one database agrees on one roster file (matches existing backend-resolution precedent).
2. Overlay/merge semantics for the roster section across configuration tiers (project-level overrides, a home-level file supplies the deployment floor).
3. A dedicated roster file stored alongside the database, read unconditionally at boot.

Clients should still be free to resolve their own local attribution settings independently of the roster anchor. A startup diagnostic that logs the resolved roster configuration path together with the number of configured principals (without exposing configuration contents) would make "which roster did this daemon boot with" answerable from the log line rather than reconstruction.

## Acceptance

- Starting clients from different working directories cannot change a running daemon's gate roster.
- A partial client configuration cannot bootstrap a daemon with a partial authorization policy.
- Startup diagnostics identify the resolved configuration source and configured-principal count without exposing configuration contents.
- Regression tests cover shared-database clients with distinct working directories.

ohdearquant:
One sentence in "Current behavior" does not hold at the current revision and should be
re-derived rather than carried forward:

> a roster present only in a fallback tier has been observed to produce a daemon that
> denies every principal it should have authorized

That behavior is not reachable at `f74c5461f`. The caller roster has no reader there:
[...]

PR #1818, opened 2026-08-09
fix(gate): fail loudly on unsupported policy config

### LINKED ISSUE #1642: Capability gate ([gate] caller enrollment) absent from 0.7.0 — config section silently ignored
Binaries from the interim 0.6.x packaging line enforced a caller-enrollment capability gate: a `[gate]` config section carried a principal roster, and a caller without a roster entry was denied at dispatch with a named error ("no principal entry").

This tree contains no capability-gate implementation. Deployments upgrading to 0.7.0 with an existing `[gate]` section get no error and no warning: the section is silently ignored and every caller is admitted. Verified post-release with a two-sided probe from a neutral working directory — an enrolled principal and a deliberately unenrolled principal both succeeded against the 0.7.0 daemon, where the prior binary denied the unenrolled half.

The failure direction is permissive, which is the wrong direction to fail silently.

Requested: port the capability gate (enforcement at dispatch, deny-with-named-error for unenrolled principals, fail-closed when a roster is present but unreadable), or explicitly reject/deprecate the `[gate]` section so a config that requests enforcement cannot be ignored without notice.

### LINKED ISSUE #1738: Gate refusal must precede id resolution in update paths
With a config declaring `[gate] granted_actors=[]` and `grant_unattributed=false`, an `update()` from a non-granted actor against a nonexistent id returns "not found" rather than an authorization refusal. Two defects in one ordering:

1. The gate does not actually refuse the operation — "not found" is only reachable after the store was consulted, so authorization is being evaluated after (or not before) id resolution.
2. The response leaks existence information across a supposed deny boundary: a denied caller can distinguish present ids from absent ones by the error shape.

Expected: authorization is evaluated first, and a refused caller receives the same refusal regardless of whether the id exists.

Test note: the regression matrix should include the unread-default arm where a permissive-looking pass is actually the gate never being consulted (a false pass observed in the motivating measurements).

ohdearquant:
## Summary

- reject every top-level `[gate]` configuration section, including an empty table, instead of silently ignoring requested authorization policy
- pin gate-before-handler ordering with present/absent `update` targets so a denied caller cannot obtain an existence oracle
- align the authorization ADRs, configuration guide, example config, and runtime design notes with the current interim posture

## Scope

Issue #1642 explicitly permits loud rejection as the alternative to porting the retired interim roster. All configuration discovery tiers use the same loader and now return `UnsupportedGateSection` before typed config deserialization or server startup.

For #1738, the mutation-sensitive regression proves both present and absent identifiers consult the gate, return the same named `PermissionDenied`, and invoke no existence-resolving handler. A permissive control proves the test would fail if the handler were never reached.

This does not implement or claim #1376's superseded steady-state config roster. It also does not claim #1426's contract-test authorization provisioning or the broader unshipped store-grant/fail-closed runtime. Those gaps remain documented.

## Verification

- independent exact-head review: READY with zero findings at `e5c41ebc2342087b9353904ffb6df93e498fa1a3`
- exact-head full CI: PASS ([run 31295297687](https://github.com/ohdearquant/khive/actions/runs/31295297687))
- `cargo test --workspace`
- `cargo check --workspace`

Fixes #1642
Fixes #1738

### DIFF
--- a/crates/khive-runtime/docs/design.md
+++ b/crates/khive-runtime/docs/design.md
@@ -126,6 +126,9 @@
 ### Gate Authorization (ADR-029)
 
 - `RuntimeConfig::gate` defaults to `AllowAllGate`; production deployments plug in a policy-backed impl
+- Operator `[gate]` configuration is reserved and rejected until ADR-143's
+  store-held caller grants ship; an unenforced enrollment policy must not be
+  accepted silently
 
 ### Layered Retrieval Architecture (ADR-030)
 
--- a/crates/khive-runtime/src/engine_config.rs
+++ b/crates/khive-runtime/src/engine_config.rs
@@ -1270,6 +1270,30 @@ default = true
     }
 
     #[test]
+    fn home_gate_config_is_rejected_while_explicit_empty_config_is_hermetic() {
+        let project_dir = tempfile::tempdir().unwrap();
+        let home_dir = tempfile::tempdir().unwrap();
+        std::fs::create_dir_all(home_dir.path().join(".khive")).unwrap();
+        std::fs::write(
+            home_dir.path().join(".khive/config.toml"),
+            "[gate]\ngranted_actors = [\"lambda:enrolled\"]\n",
+        )
+        .unwrap();
+
+        let err = KhiveConfig::load_with_roots(project_dir.path(), Some(home_dir.path()), None)
+            .expect_err("an unsupported home gate policy must still fail loud");
+        assert!(matches!(err, ConfigError::UnsupportedGateSection));
+
+        let empty = project_dir.path().join("empty-khive-config.toml");
+        std::fs::write(&empty, "").unwrap();
+        let isolated = KhiveConfig::load_with_home_fallback(Some(&empty), None)
+            .expect("an explicit empty fixture must isolate config discovery")
+            .expect("the explicit config exists");
+        assert!(isolated.engines.is_empty());
+        assert!(isolated.actor.id.is_none());
+    }
+
+    #[test]
     fn test_load_with_home_fallback_explicit_path() {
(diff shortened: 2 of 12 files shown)


ohdearquant:
Two observations on this change.

**The `[gate]` rejection is not total for malformed TOML.**

`crates/khive-runtime/src/engine_config.rs:473-486` parses the file into a generic
`toml::Value` first, and only looks for the `gate` key once that parse has succeeded. A config
whose `[gate]` table is syntactically broken, for example:

```toml
[gate]
granted_actors = [
```

returns the pre-existing parse error and never reaches `UnsupportedGateSection`.

The case this misses is the one where the new error is most useful: an operator who has
enrolled callers in a `[gate]` table and made a syntax error gets a generic parse failure with
none of the enrollment explanation the new error was written to carry. Two ways to resolve it,
either is fine. Define the contract as covering syntactically valid TOML only and test that
boundary explicitly, or add a syntax-aware header precheck so the stated total rejection
actually holds.

**The new test helpers shadow the production wrappers, so the unit tests moved off the
production wiring.**

`crates/kkernel/src/code_ingest.rs:128-163` keeps the production path, which passes
`config: None` at `:139`. The test-local functions added at `:753-776` reuse those same names
and always pass `Some(&config)`. Because they shadow, the existing unit tests now call the
test versions rather than the production ones.

Every assertion is unchanged, so nothing about the suite looks different, but the coverage
moved: a future regression in the production wrapper's `None` forwarding would leave these
tests green. The binary-level test at `crates/kkernel/tests/code_ingest_cli.rs:231-246` still
exercises ordinary invocation, so this is a coverage regression rather than a break in
behaviour.

Renaming the test helpers so they no longer shadow, plus one test through each production
wrapper, restores what was covered before.

PR #1819, opened 2026-08-09
fix(memory): make fresh-tail replay single-snapshot

### LINKED ISSUE #1807: memory fresh-tail no-index leg reads across three snapshots, breaking ADR-118's single-snapshot guarantee
## Contract

ADR-118 states the fresh-tail read guarantee directly:

> The read runs in a single snapshot, so a write committing mid-query is either entirely
> visible or entirely invisible — never a torn log/row pair.

and names why that is load-bearing rather than incidental:

> This is load-bearing on pool-backed readers, where separate method calls are not guaranteed
> to share one connection even when bracketed by textual `BEGIN`/`COMMIT`.

ADR-144 carries the matching acceptance condition: the coverage proof "must be read from one
consistent snapshot to avoid proving coverage with two clocks."

The same ADR-118 passage describes the knowledge implementation as satisfying this with **one**
SQL statement covering the own-row guard, pair minimum, optional live count, selected log
suffix, and joined current embeddings.

## What the memory no-index leg actually does

`crates/khive-pack-memory/src/ann.rs` issues **three separate reads** for one logical fresh-tail
resolution:

1. `memory_ann_fetch_tail` — the CTE that selects the live count and the bounded raw-log suffix.
   This one statement is internally consistent.
2. `memory_ann_tail_point_read` — a per-subject loop, one `query_all` per upserted subject.
3. `memory_ann_tail_live_notes` — a separate chunked `IN (...)` query against `notes`.

Each is a distinct `SqlReader::query_all` call. In `crates/khive-db/src/sql_bridge.rs`,
`PoolBackedReader::query_all` acquires its reader **inside** the call:

```rust
async fn query_all(&mut self, statement: SqlStatement) -> StorageResult<Vec<SqlRow>> {
    let pool = Arc::clone(&self.pool);
    tokio::task::spawn_blocking(move || {
        let guard = pool.reader()?;          // <-- fresh pooled reader, per call
        execute_query(&guard, &statement)
    }).await?
}
```

So reads 2 and 3 are not pinned to the snapshot that read 1 observed. This is precisely the
mechanism ADR-118 names when it calls the guarantee load-bearing.

## Failure scenario

A concurrent commit lands between read 1 and read 2 for a subject `A` the CTE already selected:

- If the commit **updates** `A`, the point read scores the newer embedding while the log suffix
  describes the older state — a torn log/row pair, which is exactly what the ADR promises cannot
  happen.
- If the commit **deletes** `A`, the point read finds no vector row and the leg returns the
  documented contradiction `Err`, which the caller treats as Cold.

The selection bound itself stays safe in both cases, and the delete path degrades in the safe
direction. The defect is that the returned result is not the all-visible-or-all-invisible pair
the contract specifies, so a consumer relying on the stated guarantee is relying on something
the implementation does not provide.

## Scope

This predates the corpus-relative bound work. The per-subject point-read loop and the separate
notes query are unchanged by recent contract-alignment changes; only the count-and-suffix
statement was touched, and that half is correct. The three query labels are present in the tree
both before and after that change.

## Suggested fix

Either is acceptable:

- Hold one pinned transaction/reader across the whole no-index leg, so all three reads observe
  one snapshot; or
- Fold the selected tail, current embeddings, and live-note predicate into a single joined
  statement, matching what ADR-118 describes the knowledge consumer as doing.

A concurrent-writer regression test should accompany the fix — the current tests cannot
distinguish one snapshot from three.

## Not weakening the ADR

Relaxing ADR-118's single-snapshot sentence to match current behaviour was considered and
rejected: the ADR names the guarantee load-bearing, and weakening a load-bearing contract to
ratify an accidental implementation turns the document into a description of the code rather
than a specification for it. The code should catch up.

ohdearquant:
## Summary

- read the memory fresh-tail log suffix, current vector state, and live-note predicate in one SQLite statement
- retain the cheap empty-tail probe, then use one snapshot statement for every non-empty no-index replay
- add a deterministic interleaving regression that advances the simulated writer immediately after suffix selection
- serialize both pathless fresh-tail readers with the environment-disable tests to eliminate their exact CI race
- document the single-statement snapshot guarantee for no-index recall

## Contract

ADR-118 requires a concurrent commit to be wholly visible or wholly invisible to the exact fresh-tail leg. `PoolBackedReader` may acquire a different connection for each method call, so textual `BEGIN`/`COMMIT` cannot make a multi-call suffix/vector/note sequence atomic. This change follows the accepted knowledge-pack pattern: one statement selects the bounded suffix and left-joins the matching vec0 and note rows, then Rust coalesces the final operation per subject.

The regression models the old failure deterministically. A legacy multi-query reader observes the pre-commit log row and post-commit vector, while the new statement returns the coherent pre-commit pair and proves that exactly one query was issued.

## Verification

- independent exact-head review: READY with zero findings at
  `b15686d33ec24c10e8f1aae5f8aaa4e40461d774`
- exact-head full CI: PASS
  ([run 31295600006](https://github.com/ohdearquant/khive/actions/runs/31295600006))
- standalone Rust formatting check
- Deno documentation formatting check
- ADR-reference, stub-marker, JSON-data, and diff checks

Fixes #1807

### DIFF
--- a/crates/khive-pack-memory/docs/api/ann-lifecycle.md
+++ b/crates/khive-pack-memory/docs/api/ann-lifecycle.md
@@ -91,7 +91,12 @@ absence or closed state, so another process cannot make a retired cache serve st
 For a pathless bridge captured during the initial install-before-activation window, the guard first
 waits on the per-model publication lock and revalidates the pending row. A completed activation may
 continue; a row that remains closed, disappears, or errors still evicts the candidate. No-index
-recalls do not take this wait, preserving their bounded readiness timeout.
+recalls do not take this wait, preserving their bounded readiness timeout. Their capped exact leg
+first performs the cheap log-only empty-tail probe required by ADR-118. For a non-empty tail, one
+SQLite statement counts the live corpus, selects the corpus-relative retained-log suffix, joins each
+subject's current vector, and evaluates live-note membership. That replay statement is the snapshot
+boundary on both standalone and pool-backed readers: a concurrent commit is wholly visible or wholly
+invisible, never a log row from one state combined with a vector or note row from another.
 File-backed checkpoint writers serialize the segment, UUID sidecar, registry transition, and mmap
 re-adoption with `<segment-dir>/.bridge-checkpoint.lock`. After acquiring that lock, a writer
 revalidates that its candidate watermark is not behind the durable row before touching the segment;
--- a/crates/khive-pack-memory/src/ann.rs
+++ b/crates/khive-pack-memory/src/ann.rs
@@ -3315,6 +3315,7 @@ mod tests {
     /// for the pathless publication lock and revalidate the now-active row,
     /// not evict the candidate that the checkpoint is about to publish.
     #[tokio::test]
+    #[serial(adr118_fresh_tail)]
     async fn pathless_pending_reader_waits_for_checkpoint_activation() {
         const MODEL: &str = "pathless-pending-publication-wait";
         let rt = KhiveRuntime::memory().expect("runtime");
@@ -3374,6 +3375,7 @@ mod tests {
     /// loses its registration while the reader is blocked, revalidation must
     /// still evict and return an empty replacement leg.
     #[tokio::test]
+    #[serial(adr118_fresh_tail)]
     async fn pathless_pending_reader_evicts_after_registration_loss() {
         const MODEL: &str = "pathless-pending-publication-loss";
         let rt = KhiveRuntime::memory().expect("runtime");
--- a/crates/khive-pack-memory/src/ann.rs
+++ b/crates/khive-pack-memory/src/ann.rs
@@ -3319,6 +3319,7 @@ mod tests {
     async fn pathless_pending_reader_waits_for_checkpoint_activation() {
         const MODEL: &str = "pathless-pending-publication-wait";
         let rt = KhiveRuntime::memory().expect("runtime");
+        provision_test_vector_store(&rt, MODEL, 4);
         let ann = new_shared();
         let key = AnnKey::new(MODEL);
         register_consumer(&rt, MODEL)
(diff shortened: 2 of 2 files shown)


ohdearquant:
The single-statement rewrite of the tail hydration reads correctly, and the added serialization does remove a real test-fixture race: the two new pathless tests read the production environment variable and were previously exposed to the fixture tests that disable it. Both spawned readers are joined under a timeout, so nothing is left unawaited.

Three things to work through before this is done, two in the new test and one that predates this change.

**1. The regression does not execute a database, so it cannot observe the interleaving it is named for.** `InterleavingTailReader` returns preselected rows from Rust. There is no connection, no writer, no commit, no barrier and no scheduling interleave anywhere in the test, and the SQL assertions match string fragments rather than execution or result semantics. The mock is a faithful one-row imitation of the old call sequence, which is a useful thing to have, but it is not the old implementation and it does not establish that a real cursor sees a coherent snapshot.

Concretely: a changed query can still contain every expected join, return the mock's hard-coded coherent row, and pass, while a real sqlite-vec cursor observes different shadow-table state or a real pool-backed writer commits between the reads. Closing this needs a two-connection SQLite/vec0 regression with an explicit interleaving barrier, or instrumentation of the actual backend with a control that still exercises the old multi-call sequence.

**2. "Exactly one SQLite statement" is counted at the wrong layer.** The counter records one `SqlStatement` label per `SqlReader::query_all` call. The bridge does prepare and step one top-level statement per call, but that is not the same measurement: it counts neither prepares nor connection checkouts, and it cannot see anything outside the mock seam. Underneath, vec0 prepares and steps shadow-table rowid statements and reads shadow vector blobs through its own connection during virtual-table filtering and column production, so a one-entry call vector cannot certify one statement at the virtual-table layer.

The cheap fix is to rename the assertion to the layer it actually measures, which is one reader call through this seam. The stronger claim needs backend-level instrumentation or a real concurrent test.

**3. Predates this change, so not a request against it, but it bounds what the change can claim.** The serving path still calls `begin_read_snapshot`, then the registry minimum, then `fetch_final_tail_on`, then `end_read_snapshot` as separate calls. `PoolBackedReader::query_all` acquires its reader inside the method and drops it on return, so on the pool-backed path the same reader object does not pin a connection across those calls. The decision record requires the registry minimum and the tail coverage to come from the same snapshot.

The failure that allows: the registry read observes `m <= s`, another consumer then raises the registry minimum and compacts rows through `m`, and the later `> s` tail statement reads the post-compaction log and silently omits rows in the compacted interval while the earlier guard makes the path look coherent.

The base has the same call sequence, so this change does not regress it. It does mean the single-statement work closes the hydration half and leaves the guard-plus-tail half open. There is already a pattern for the closed form in this repo: the knowledge vamana path puts the registry guard and the selected tail in one statement. The alternative is a connection-pinned transaction API plus a pool-backed interleave test.

Worth noting for scope: the local `main` ref is stale in a way that produces a false several-hundred-file range. The real base is the merge base against `origin/main`, and against that this is two files.

PR #1820, opened 2026-08-09
fix(git): converge noncanonical digest anchors

### LINKED ISSUE #1708: git.digest anchor resolution silently excludes anchors with a non-canonical repo_slug
## Summary

`resolve_or_create_project` (crates/khive-pack-git/src/handlers.rs) resolves a repo anchor in three tiers: exact `properties.repo_slug` match, exact legacy `repo_url` match restricted to `repo_slug IS NULL` rows, and normalized legacy `repo_url` match, also restricted to `repo_slug IS NULL`. An existing `project` entity that carries a `repo_slug` in a non-canonical spelling (for example `ohdearquant/lattice` without the host, or any hand-written variant) is therefore excluded from every tier: tier 1 misses on exact string inequality, and tiers 2-3 exclude it by the `IS NULL` guard. The next `git.digest` run then creates a second anchor for the same repository, and the duplicate-anchor warning cannot see the pair because the existing anchor never enters any candidate set.

## Impact

Duplicate project anchors split the `annotates` corpus (PRs/commits/issues census answers from one arbitrary half). A live pair with identical `name` and identical `repo_url` has been observed in a production store; the resolution keys were absent or non-canonical on the existing anchor at mint time, and were backfilled later, which makes the pair look impossible against the current resolver until the property-mutation history is considered.

## Proposed direction

A candidate whose `repo_slug` is present but not equal to the derived identity should be surfaced (e.g. re-derive identity from its stored `repo_url` and, on a normalized match, report it as a conflicting-slug anchor in the ingest report) rather than silently excluded. Creation over such a candidate should at minimum carry a warning naming the excluded near-match.

## References

- Resolver: `resolve_or_create_project`, `find_projects_by_slug`, `find_projects_by_legacy_repo_url`, `find_legacy_projects_without_slug` in `crates/khive-pack-git/src/handlers.rs`
- Prior identity work: #1173, ADR-088 Amendment 2

ohdearquant:

## Summary

- make `git.digest` reuse present anchors whose stored `repo_slug` is
  noncanonical when their normalized repository identity matches
- preserve canonical-slug winners, repair the selected legacy anchor, and emit
  stable deduplicated warnings for URL-equivalent conflicting anchors
- converge HTTPS/SCP, unsluggable remote, and remote-less local identities while
  retaining literal local `?`/`#` paths and redacting credentials on every write
- include soft-deleted noncanonical anchors in the orphan convergence scan and
  align ADR-088 plus public API documentation

## Verification

- focused resolver and real `git.digest` regressions for remote, SCP, local,
  tombstone, canonical-winner, conflict-warning, redaction, and deduplication paths
- independent exact-head review: READY with zero findings at
  `8f59bfb731709a4f6d08008935e164447629d55f`
- exact-head full CI: PASS — https://github.com/ohdearquant/khive/actions/runs/31295939061
- `cargo test --workspace`
- `cargo check --workspace`

## ADR

- ADR-088 Amendment 2

Fixes #1708


ohdearquant:
One change requested before this can land:

**Canonicalize host case in the unsluggable-HTTPS fallback.** `remote_url_to_slug` lowercases the host (`source.rs:296`), but `canonical_remote_identity` — the fallback identity for accepted HTTPS URLs that do not satisfy the slug grammar — reassembles the authority verbatim (`format!("{scheme}{authority}{path}")`). Two case-variant spellings of the same host therefore still mint distinct repository anchors on the fallback path, which is the divergence class this PR exists to converge. Lowercase the host portion of the authority in the fallback (leaving userinfo redaction and the path untouched, matching the slug path's case rules), and add a case-variant fallback test alongside the existing normalization table.

Non-blocking follow-ups worth a look while in the area: the canonical-slug resolution pass normalizes every live project with a null/different slug even after the winner is found (O(P) per digest, and the changed SQL sorts without a JSON-property index), and absolute local-path candidates are normalized serially with one awaited `git remote get-url` child each. Bounding or batching both would keep reconciliation cost flat as anchors accumulate.

ohdearquant (commented):

Verdict on head f0e1311e76421929a58abdaeaf9c849e9f315111: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #1821, opened 2026-08-09
test(search): enforce symmetric row-shape parity

### LINKED ISSUE #1676: Multi-backend search rows drop kind, name and created_at, silently breaking field-keyed checks
Split out of #1674, where it was found but is not the same defect.

## Measured

Key sets for row 0 of the same query, same store, same binary:

- daemon (multi-backend coordinator path): `[entity_kind, id, score, snippet, source, title]`
- `KHIVE_NO_DAEMON=1` (handler path): `[created_at, entity_kind, id, kind, name, score, snippet, source, title]`

`kind`, `name`, and `created_at` are absent from the coordinator path. `kind` and `name` were added to the handler row shape in #1174 to match `list()` / `get()`; the coordinator's row assembly in `khive-mcp/src/server.rs` was not updated with it, and its comment still claims the shapes match exactly.

## Why this is worth a fix rather than a doc note

The missing field breaks instruments pointed at these rows. A check keyed on `.name` against the multi-backend path reports a value absent that is plainly visible in the response, and returns a clean zero with no error — the failure is silent and reads as a real negative result. That was caught only because the check was run alongside a known-present control on the same invocation; the same check run bare would have produced a confident wrong answer.

## Scope note

This is not fixed by the single-contributor passthrough in #1675. The fields are absent at the type level before the merge is reached: the coordinator carries `SearchHit` / `NoteSearchHit`, which have no `kind`, `name`, or `created_at`, and `khive-mcp/src/coordinator.rs` currently plumbs only a per-id kind map alongside them.

## Fix shape

`CoordinatorService::fan_out_search` already fetches the owning backend's entity/note to populate the kind map, so name and created_at are available at the same call site without an extra round trip. Widening that map (or carrying the fetched record) and emitting the three fields would close the gap. A regression test should assert the two dispatch paths produce the same key set for the same query, which is the check that would have caught this at #1174.

ohdearquant:

## Summary

- preserve the production multi-backend search row repair already shipped in merged PR #1629
- add the missing symmetric acceptance regression required by #1676
- drive direct and coordinated entity and note searches over the same primary store and compare their complete, dynamically derived key sets
- catch fields added or removed on either route, rather than checking only that the coordinator contains today's known fields

## Verification

- independent exact-head review: READY with zero findings at `55a5950ff716482b9cec05703ee2ad516e7d9a97`
- `rustfmt --edition 2021 --check crates/kkernel/src/coordinator/tests.rs`
- `git diff --check`
- exact-head full CI: PASS ([run 31296113222](https://github.com/ohdearquant/khive/actions/runs/31296113222))

Fixes #1676

### DIFF
--- a/crates/kkernel/src/coordinator/tests.rs
+++ b/crates/kkernel/src/coordinator/tests.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeSet;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


ohdearquant:
Route independence holds, which is the load-bearing part. The coordinated path validates, fans
out, hydrates, and rebuilds rows itself rather than delegating to the direct handler, and it
still runs its own fan-out and RRF merge when only one backend returns hits. So the comparison is
not true by construction, which is the way a test like this usually fails. The extraction helper
also fails loudly rather than quietly on zero hits, a non-array result, or a per-op error
envelope, so a total search outage cannot slip through as two equal empty key sets.

Two coverage limits worth closing while this is fresh.

Both calls pass `presentation: None`, which resolves to Agent, and Agent drops non-lifecycle
nulls and empty strings, arrays, and objects. If one route emits a field as null or empty while
the other omits it entirely, Agent erases exactly the difference this test exists to catch and
the assertion passes. Running the same assertion under `presentation=verbose` closes that, since
Verbose passes canonical values through.

The fixture also seeds both records only on the primary backend and inspects only
`results[0].result[0]`. A shape difference that appears only once two ranked lists are genuinely
merged, or only on a later row whose metadata hydration failed, is invisible at that vantage.
Either seed a matching row on both backends and compare every returned row's key set, or narrow
the test name and doc comment to first-row single-contributor coverage, so the name does not
promise more than the fixture checks.

ohdearquant:
**VERDICT: APPROVE**

Independent review at 55a5950f: clean approve, no blocking findings. The new test drives both search routes over the same primary runtime and requires exact row key-set parity for entity and note substrates, which is the right acceptance shape for #1676 (a one-way containment assertion would miss a field added to only one route). One minor non-blocking note: the response helper indexes into the first hit without a dedicated emptiness assertion; the panic message covers it adequately for a test helper.

PR #1822, opened 2026-08-09
fix(mcp): fail closed on unreachable daemon sockets

### LINKED ISSUE #1242: kkernel: unreachable daemon socket misdiagnosed as stale PID, client attempts daemon respawn from restricted context

ohdearquant:
## Summary

- distinguish a genuinely absent/refused daemon socket from an inaccessible or otherwise indeterminate socket
- keep `ENOENT` and `ECONNREFUSED` on the existing stale-socket self-heal path
- fail closed with structured `daemon_unreachable` diagnostics for `EACCES`, `EPERM`, and every other connect error
- prove an unreachable socket cannot trigger kill, spawn, retry recovery, or local-dispatch fallback

## Contract

The old forward path treated every Unix-socket connect failure as “no daemon.” That was safe for a missing path or a stale listener, but unsafe for permission and other reachability failures: recovery could manipulate a daemon the caller could not inspect, then silently execute against another runtime.

The new classification is deliberately narrow. Only OS `NotFound`/`ENOENT` and `ConnectionRefused`/`ECONNREFUSED` are recoverable `NoSocket`. Every other failure retains its error kind/code as `Unreachable`; the initial and post-recovery forward paths return the structured error immediately, while recovery probes classify it as uncertain and never kill.

`KHIVE_NO_DAEMON` remains the explicit caller-controlled local-dispatch opt-out. A real stale Unix socket remains recoverable.

## Verification

- independent exact-head review: READY with zero findings at `40aeac1304f115045a48f0f1ba0880312d7db331`
- exact-head full CI: PASS ([run 31296770316](https://github.com/ohdearquant/khive/actions/runs/31296770316))
- errno classification table for `ENOENT`, `ECONNREFUSED`, `EACCES`, and `EPERM`
- public forward-path regression with zero kill/spawn/local fallback
- stale dropped-listener recovery regression
- Rust and Deno formatting, ADR-reference, stub-marker, JSON-data, SQL, and diff checks

## ADR

- ADR-049 Amendment 4
- daemon lifecycle API documentation

Fixes #1242

### DIFF
--- a/crates/khive-mcp/docs/api/daemon-lifecycle.md
+++ b/crates/khive-mcp/docs/api/daemon-lifecycle.md
@@ -156,13 +156,21 @@ regardless of which task drives it.
 ## `forward_or_spawn` — the `None` contract (#644)
 
 Returns `None` only when nothing was ever written to the daemon and local
-dispatch is therefore safe: `KHIVE_NO_DAEMON` is set, or no daemon socket
-could be reached (`NoSocket`). It never returns `None` after the real frame
-has been written — `Some(Ok)`/`Some(Err)` both mean the request's fate is
-already decided at the daemon and the caller must not dispatch locally.
+dispatch is therefore safe: `KHIVE_NO_DAEMON` is set, or the socket is
+definitively absent/refused (`NoSocket`). A connect error that does not prove
+absence (`Unreachable`) returns `Some(Err)` immediately because the client
+cannot know whether a healthy daemon is serving other processes. It never
+returns `None` after the real frame has been written — `Some(Ok)`/`Some(Err)`
+both mean the caller must not dispatch locally.
 Under `KHIVE_DAEMON_STRICT=1`, the `NoSocket` case becomes `Some(Err(..))`
 instead (see `fallback_or_reject`) — `KHIVE_NO_DAEMON` itself is unaffected,
 since it is the caller's explicit, unconditional opt-out (nothing is ever
 recorded or counted for it). Once the real frame IS fully written
 (`ParseFailure`/`ProtocolMismatch`), this returns a hard error immediately
 instead of killing/respawning/retrying or falling back locally.
+
+Connection classification is intentionally narrow (#1242): `ENOENT` and
+`ECONNREFUSED` are `NoSocket`, preserving first-spawn and stale-socket
+self-heal. `EACCES`, `EPERM`, and every other indeterminate connect failure
+are `Unreachable`; they return the structured `daemon_unreachable` error and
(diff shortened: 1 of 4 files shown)


ohdearquant:
The classifier change and the contract reading behind it hold up: the recovery set is closed at the `ErrorKind` level rather than by raw errno, `KHIVE_NO_DAEMON` is still checked before any forward attempt so an inaccessible socket cannot defeat the opt-out, and an unreachable probe maps to uncertainty rather than `Dead`, so it never authorizes a kill or a spawn. The decision record and the lifecycle doc agree with the implementation.

One gap worth closing before this lands, in the test rather than the source.

**The "no retry, no recovery" claim is not fully observable by the regression.** The test checks `KILL_COUNT`, `SPAWN_COUNT`, and `fallback_total()`. Those are real positive oracles for kill and spawn, but none of them observes whether a forward attempt or a recovery entry happened. A regression could classify `EPERM` as `NoSocket`, enter a probe or retry path, perform no kill and no spawn, and eventually return the same `Some(Err)` after the test's unbounded await. Every assertion still passes while the property the test exists to protect is false.

The same shape applies at the caller boundary: a direct forward return does not positively establish that the public request path never invokes local dispatch. Nothing currently records that seam.

Suggested shape:

- add a test-only forward-attempt or recovery-entry recording seam, and assert exactly one initial classification with zero subsequent attempts, so the absence claim has a counter behind it rather than an inference from two unrelated ones;
- add a request-boundary test that records the local dispatch seam while the forced error is active.

Both are additive and test-only. The source behaviour under review already appears correct; the point is that the test cannot currently tell correct from a specific class of incorrect.

One note on the pre-existing surface, not a request against this change: the base already mapped `write_frame` errors to `NoSocket`. That broader post-connect behaviour sits outside this change's connect-time classification and is untouched here.

PR #1833, opened 2026-08-09
fix(git): reap abandoned cache staging directories

### LINKED ISSUE #1826: git cache: a killed clone leaves an orphaned .staging-* directory that nothing reclaims
## Summary

`install_fresh_clone` in `crates/khive-pack-git/src/cache.rs` stages each clone into a uniquely
named `.staging-<uuid>` directory and renames it into place on success (`cache.rs:354`). Every
in-process error path removes the staging directory first (`:342`, `:345`, `:348`, `:352`, `:355`).

A process kill (SIGKILL, OOM, host shutdown) skips all of those cleanup paths, and nothing else
ever reclaims the result. The code already states this, at `cache.rs:339`:

> `evict_lru` deliberately never touches non-owned names (`is_owned_entry`), so nothing else
> would ever reclaim this on its own.

Because the staging name is a fresh `Uuid::new_v4()` per attempt (`:334`) and no code path
enumerates existing `.staging-*` entries, an orphan is unreachable by construction. It is never a
retry candidate, never LRU-evicted, and never counted by the size accounting that `cap` exists to
bound.

## Impact

Unbounded disk growth outside the configured cache cap, accumulating silently.

Observed on one workstation: 29 orphaned staging directories totalling 14.4 GiB across 524,589
files, all created within a single day, after a series of clone-bearing processes were terminated.
The cache's own live entries totalled roughly 0.4 GiB over the same period, so the orphans were
about 36x the size of the real cache.

`cap` does bound each individual clone (`:347`), but it is evaluated against the staging directory
currently being installed. Orphans from previous runs are outside that measurement, so they neither
count toward the cap nor trigger eviction.

A further consequence for anyone diagnosing disk use: the orphans are dot-prefixed, so a shell glob
of the cache root does not list them and a `du` of the visible children understates the directory by
more than an order of magnitude.

## Reproduction

1. Invoke `git.digest` against a repository large enough that `install_fresh_clone` spends real time
   in `clone`.
2. `kill -9` the process mid-clone.
3. A `.staging-<uuid>` directory remains under the cache root holding the partial clone.
4. Repeat. Each kill adds another orphan. None is reclaimed, no verb reports them, and the cache cap
   is unaffected.

## Suggested fix

In-process cleanup structurally cannot cover a kill, so the sweep belongs at cache initialization:
when the cache root is opened, remove `.staging-*` entries that are direct children of the root and
older than a conservative age threshold.

Two details worth preserving from the existing code:

- Confine the sweep to the `.staging-` prefix and to direct children of the root, matching the
  containment discipline already in `remove_owned_entry` (`:363-372`).
- Keep an age threshold rather than removing every orphan unconditionally, so the sweep cannot race
  a clone legitimately in flight in another process.

Optionally, count orphaned staging bytes toward the reported cache size, so the condition surfaces
as a number rather than as unexplained disk consumption.

ohdearquant:
## Summary

- reclaim killed-clone `.staging-<uuid>` residue whenever the git-digest scratch cache is opened
- preserve in-flight clones and unrelated operator data through exact name, containment, file-type, and age checks
- document the crash-recovery boundary separately from ordinary LRU and per-clone cap enforcement

## Cleanup contract

Every cache mutation now initializes the scratch root through a bounded stale-staging sweep. A candidate must be a direct child, a real directory, and have the exact lowercase canonical `.staging-<uuid>` name generated by the clone installer. It is removed only when its directory mtime is strictly more than 24 hours old; fresh, future-dated, nested, symlink/file, and prefix-lookalike entries are retained.

Removal uses the cache's bounded retry helper and tolerates another process winning the same cleanup race. Other inspection or deletion failures remain visible rather than silently leaving unbounded data outside both configured cache caps.

## Verification

- TDD RED/GREEN implementation at exact head `b4df88d7130aca0262525eae5b1adccce7cdc08e`
- independent exact-head review: READY (0 findings)
- exact-head full CI: [run 31304945936](https://github.com/ohdearquant/khive/actions/runs/31304945936) — passed
- direct Rust and Deno formatting, ADR-reference, JSON-data, SQL, stub-marker, and diff checks
- focused deterministic stale deletion plus fresh/foreign/nested/non-directory preservation regressions

Fixes #1826

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2160,6 +2160,7 @@ dependencies = [
  "khive-storage",
  "khive-types",
  "lattice-embed",
+ "libc",
  "serde",
  "serde_json",
  "tempfile",
--- a/crates/khive-pack-git/Cargo.toml
+++ b/crates/khive-pack-git/Cargo.toml
@@ -26,6 +26,7 @@ tokio = { workspace = true }
 anyhow = { workspace = true }
 clap = { workspace = true }
 blake3 = { workspace = true }
+libc = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["full"] }
--- a/crates/khive-pack-git/docs/api/cache.md
+++ b/crates/khive-pack-git/docs/api/cache.md
@@ -195,10 +195,14 @@ namespace, then runs `reap_stale_staging` at most once per
 `REAP_THROTTLE_INTERVAL` (see above). `reap_stale_staging` enumerates only
 the namespace's direct children. A deletion candidate must be a real
 directory (never a symlink, file, or nested path) whose name is exactly the
-lowercase canonical hyphenated spelling of a UUID; its liveness is then
-decided by `staging_liveness` — `try_lock`-acquirable (or missing its lock
-file _and_ older than the 24h age fallback) means abandoned, anything else
-means live and untouched regardless of age. Removal uses the same bounded
+lowercase canonical hyphenated spelling of a UUID (an in-flight clone
+wrapper) or `trash-` followed by one (an interrupted
+`delete_verified_owned_entry`, whose renamed slot would otherwise be
+unreclaimable residue); its liveness is then decided by `staging_liveness` —
+`try_lock`-acquirable (or missing its lock file _and_ older than the 24h age
+fallback) means abandoned, anything else means live and untouched regardless
+of age. Trash residue never carries a lock file, so the age fallback alone
+governs it: a fresh entry (a recursive delete still in flight) survives. Removal uses the same bounded
 retry helper as owned cache eviction and tolerates another process winning
 the same cleanup race; other I/O failures surface instead of silently
 leaving disk growth unobservable.
--- a/crates/khive-pack-git/docs/api/cache.md
+++ b/crates/khive-pack-git/docs/api/cache.md
@@ -45,15 +45,39 @@ process-wide `EVICTION_LOCK`), and that pass defers — rather than blocks on
 — any candidate whose per-slot lock is currently held. See `slot_lock` and
 `evict_lru` below.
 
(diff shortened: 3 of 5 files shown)


ohdearquant (commented):

Verdict on head 3c535da09cb3ebf97ecd07ac9b95f091fe506a1b: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Two changes requested, one note.

1. The ADR-088 crash-residue rider added here still describes the earlier design: `.staging-<uuid>` entries as direct children of the scratch root, reclaimed purely by a 24-hour mtime fence. The implementation in this same PR ships the namespace + lock-liveness design instead (`.khive-git-staging/<uuid>/`, age only as the missing-lock-file fallback, and a fresh abandoned wrapper is reclaimed — your own test pins that). Please rewrite the rider to match the shipped mechanism; the cache.md prose is already correct and can be adapted.

2. `delete_verified_owned_entry` renames a doomed slot to `<namespace>/trash-<uuid>` before the recursive delete, but the sweep only admits bare canonical-UUID names — so a process killed during that delete leaves a `trash-*` directory nothing ever reclaims, reintroducing the unreclaimable-residue class this PR closes. Admitting `trash-<uuid>` names as sweep candidates under the same liveness/age-fallback logic closes it (an in-progress delete keeps the directory mtime fresh through child unlinks, so the age fence protects it).

Minor: `is_owned_entry_via_fd` treats any `.git` entry (including a symlink or file, since the stat is NOFOLLOW) as ownership evidence, where the pathname check requires a directory — matching S_IFDIR would make the fd re-check genuinely equivalent.

The liveness-over-age design itself is right, and the fd-verified deletion path with the (dev,ino) identity confirm is careful work.

PR #1857, opened 2026-08-09
fix(db): stop checkpoint telemetry amplifying pinned WAL

### LINKED ISSUE #1838: Checkpoint task writes a per-attempt outcome event into the store it is draining, amplifying WAL growth exactly while pinned
## Summary

Every checkpoint attempt records a lifecycle-outcome event into the same database the checkpoint is trying to drain. Under normal conditions this is invisible: the frames are reclaimed as fast as they are written. Under a WAL pin — the exact condition the checkpointer is fighting — those event writes accumulate in the WAL at the attempt rate, so the checkpointer becomes a principal author of the backlog it cannot clear.

The failure-handling path writes its failure into the store whose failure it is handling. That is the same shape as #1831 (a retry loop whose retries are the load), one layer down.

## Measured

During a live pin (checkpointed_frames frozen while log_frames grew):

- Event plane, 20-minute window: `counts_by_actor` shows `daemon:checkpoint_task` = **1825**, all `checkpoint_outcome_recorded` — ~1.5 events/second, dwarfing every other actor combined (the next largest wrote 68).
- WAL growth over the same period: ~2,000 frames/minute, from ~8,800 to ~39,000 log frames, WAL file 67 MB → 161 MB.
- At roughly 20 frames per event write (row + indexes), the event traffic accounts for approximately the entire observed growth rate. No other writer of comparable volume appears in the event census.
- The loop is self-accelerating in the wrong direction: pin → attempt fails → outcome event written → log grows → pressure warnings → more attempts (escalation to TRUNCATE, also recorded) → more events.

Secondary observation, same window: the daemon logged `checkpoint lifecycle event dropped because ...` — the lifecycle queue (capacity 1) already drops under pressure, so the recording is *already* lossy; completeness is not a property the current design delivers, which weakens the case for paying a store write per attempt.

## Why this matters

- It converts a bounded incident into an unbounded one. A pinned reader alone freezes reclaim; with this amplifier, WAL growth continues at ~120 MB/hour even when every other writer is quiet.
- It contaminates diagnosis. The event plane is the natural instrument for "who is writing during this incident," and the answer it gives is dominated by the instrumentation of the incident itself.

## Suggested direction

1. **Do not write per-attempt outcomes to the primary store while above the pressure threshold.** Aggregate in memory and write one summary row on state transition (below→above, above→below), which is also when the WARN logs already fire — the logs are edge-triggered; the event writes should be too.
2. Alternatively (or additionally), route checkpoint lifecycle events to the log/sidecar rather than the database. The consumer of these events during an incident is an operator reading diagnostics, not a query over the store — and during the incident is precisely when the store write is harmful.
3. Whatever is chosen, the invariant worth a test: **total frames written by the checkpoint task during a simulated pin must be O(state transitions), not O(attempts).**

## Reproduction shape

Open a read transaction and hold it (any long-lived reader). Generate modest write traffic to elevate wal_pages past the warn threshold. Observe the event plane: outcome events accrue per attempt at the tick rate, and WAL growth continues after the external write traffic stops.

## Related

- #1831 — retry loop whose retries are the load (channel ingest); this is the checkpoint-task instance of the same class.
- #1828 — long-lived reader lifetime is what creates the pin this amplifies.
- #1836 — where the long-lived readers come from.

ohdearquant:
Fixes #1838.

## Summary

- replace one persisted checkpoint-lifecycle event per elevated tick with two bounded transition records: episode-open and recovery
- retain per-episode tick count and peak WAL size in task memory, then include the aggregate in the recovery record
- expose honest diagnostics counters for observations, episodes, append failures, and bounded-state drops without writing those observations back into the WAL being diagnosed
- version the durable payload and align ADR-091 plus operator documentation with the transition-only contract

This removes the feedback loop where a pinned WAL caused the checkpoint task to append another durable row on every unsuccessful attempt.

## Verification

- independent exact-head review at `60d390702d0af385684d0863f32b515502eb9f5b`: **READY, zero findings**
- exact-head full CI: [run 31325837970](https://github.com/ohdearquant/khive/actions/runs/31325837970) — **passed**
- local source-only checks: `rustfmt --check`, Deno formatting, ADR reference lint, and `git diff --check`

## Scope

This PR changes checkpoint lifecycle telemetry only. It does not change checkpoint cadence, PASSIVE/TRUNCATE policy, WAL-pin attribution, or database write semantics.


ohdearquant (commented):

Verdict on head aec346ffa09f09fcfa804178a40ef007fd059198: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head f46754fa5b6b7df979c2d85124995eab080f3142: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

ohdearquant (commented):

PR #1858, opened 2026-08-09
fix(db): disable writer WAL autocheckpoint

### LINKED ISSUE #1848: storage(wal): disable autocheckpoint on writer-capable connections
**Defect.** Every writer-capable connection currently applies SQLite's per-connection `wal_autocheckpoint` threshold (default 4000 pages, roughly 16 MiB) in addition to running through a dedicated, scheduled checkpoint task. Because `wal_autocheckpoint` is a per-connection setting, it re-enables itself on any writer connection that does not explicitly disable it, regardless of a dedicated checkpoint task existing elsewhere in the process. A commit that crosses the page threshold on such a connection triggers an implicit PASSIVE checkpoint synchronously inside that commit, outside the scheduler's control.

**Mechanism.** A production system built around a single dedicated checkpoint connection assumes that connection is the only source of checkpoint I/O. If ordinary writer connections still carry a nonzero `wal_autocheckpoint`, every threshold-crossing commit becomes a second, uncoordinated checkpoint source — adding checkpoint I/O latency directly onto the commit path it was supposed to be removed from, and making checkpoint timing unpredictable under load. This is a well-documented SQLite behavior: `wal_autocheckpoint` is scoped per database connection handle, not per database file, so a dedicated checkpoint connection with the pragma disabled does not suppress it on sibling connections opened against the same file.

**External evidence.** A prior-art survey of production SQLite embedders converges on disabling per-connection autocheckpoint everywhere and routing all routine checkpoint I/O through one dedicated, long-lived connection running PASSIVE mode only — the only checkpoint mode that never acquires SQLite's writer lock. The survey found no comparison-set system that runs both a dedicated checkpoint connection and leaves autocheckpoint enabled on writer connections; systems either disable it universally or accept per-commit autocheckpoint as the sole mechanism (trading isolation for simplicity). SQLite's own documentation describes `wal_autocheckpoint` as connection-scoped and describes PASSIVE mode's writer-lock-free contract.

**Acceptance criterion.** After process initialization, `PRAGMA wal_autocheckpoint` reads `0` on every writer-capable connection in the pool, verified programmatically (not just at config default — including connections created after startup). A test demonstrates that a commit sequence which would have crossed the default 4000-page threshold no longer triggers an implicit checkpoint on a writer connection.

A related but separable question — detecting multiple independent checkpoint executors contending for the same database file — is intentionally out of scope here; it belongs with holder-census work rather than this one-pragma change.

ohdearquant:
Fixes #1848.

## Summary

- disable SQLite's connection-local `wal_autocheckpoint` on every writer-capable pooled and standalone connection
- keep checkpoint ownership with the dedicated ADR-091 task instead of letting an arbitrary commit synchronously checkpoint
- remove the public/environment override so callers cannot silently restore split checkpoint ownership
- retain read-only registry inspection and cover pooled, writer-task, standalone, checkpoint reopen, and sustained multi-thousand-frame WAL behavior

This makes commit latency independent of SQLite's default 1,000-page autocheckpoint threshold and leaves PASSIVE/TRUNCATE policy with one observable owner.

## Verification

- independent exact-head review at `e53d05d529e7b41a1f5c5dbbc2b92d03f985c5ab`: **READY, zero findings**
- exact-head full CI: [run 31326270830](https://github.com/ohdearquant/khive/actions/runs/31326270830) — **passed**
- local source-only checks: `rustfmt --check`, Deno formatting, ADR reference lint, and `git diff --check`

## Scope

This PR changes automatic-checkpoint ownership only. It does not alter the dedicated checkpoint cadence, PASSIVE/TRUNCATE thresholds, WAL-pin attribution, writer routing, or transaction semantics.


ohdearquant (commented):

Verdict on head 145e6cb6d89860ad5707eb8bb64f4bceb3650a0b: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head cc34774c212400175c8a1473b588b74ebda248a9: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
On the remaining request — a hard WAL bound under a pinned read snapshot: the concern is real, but it is a property of SQLite's checkpoint mechanics that this PR's scope (checkpoint ownership) does not change, and that `main` already exhibits identically. Evidence, executed:

**Engine control** (4096-byte pages, `wal_autocheckpoint=4000`, 1500 committed 32KB inserts):

| Arm | WAL size after workload |
| --- | --- |
| No concurrent reader | 16,517,112 bytes (~= the 4000-page bound) |
| Reader holding an open snapshot throughout | 62,582,832 bytes (~3.8x the bound, growing linearly with writes) |

**`main` reproduction**: at bc280a0f, `crates/khive-db/src/pool.rs:25` sets `DEFAULT_WAL_AUTOCHECKPOINT_PAGES = 4000`, applied unconditionally at `pool.rs:1483`. So every writable pool on `main` today has exactly the posture this PR's unclaimed-pool fallback preserves. The claimed-pool path hands checkpointing to the dedicated task instead, which is equally unable to advance past a pinned frame — no autocheckpoint setting, 0 or 4000, bounds the WAL when a reader pins the snapshot (same failure class as #580).

Bounding WAL growth under a pinned reader therefore needs a different mechanism — write-admission guard, reader-pin eviction, and/or a disk-space floor (the disk guard suggested earlier in this review is folded in there too). That design work is now tracked as #1876, which states the invariant to build and test against. This PR keeps its checkpoint-ownership scope: claimed pools disable autocheckpoint in favor of the checkpoint task; unclaimed pools retain the bounded fallback, with a regression test (`unclaimed_pool_retains_bounded_autocheckpoint_reclamation`) proving reclamation stays live in that posture.

ohdearquant (commented):

Verdict on head 8848f4c3b9d351299d3c6a0cd20be948ec2d3e4b: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head a10456a09e2ab0104b7028fb3211b8f747469a05: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #1861, opened 2026-08-09
fix(db): scope cached reader permits to active operations

### LINKED ISSUE #1828: sql_bridge reader permit is held for the handle's lifetime, exhausting the pool in long-lived processes
## Summary

A `sql_bridge` reader permit is released when the connection handle is dropped, not when the read operation finishes. In a long-lived process the permit is therefore effectively permanent, so a process that stays up for hours or days exhausts its bounded reader pool and every later read fails with `storage: timeout during sql_bridge.reader_handle`.

## Where

`crates/khive-db/src/sql_bridge.rs:601` and `:1565` both acquire a slot:

```rust
let handle_slot = acquire_handle_slot(
    pool.sql_bridge_reader_slots(),
    pool.config().checkout_timeout,
    "sql_bridge.reader_handle",
)
.await?;
```

The permit travels into `StandaloneHandle { conn, _slot }`. In `SqliteWriter::query_row` / `query_all` / `query_page` the handle is put back on the writer after the query:

```rust
self.handle = Some(handle);
```

so `_slot` is retained for the lifetime of the `SqliteWriter`, not the lifetime of the query. The pool is `Semaphore::new(max_readers.max(1))` (`crates/khive-db/src/pool.rs:522`), and `max_readers` defaults to `available_parallelism().clamp(1, DEFAULT_READER_CAP)` with `DEFAULT_READER_CAP = 8`.

## Impact

Eight lazily-opened reader handles retained anywhere in a process permanently exhaust that process's pool. Every subsequent acquisition then waits out `checkout_timeout` (default 5s) and fails. Because the error is raised at slot acquisition rather than during query execution, the symptom does not track query cost: verbs that never take a standalone reader keep working while verbs that do fail uniformly. That makes the failure read like a store-wide outage when it is per-process state.

Observed in a running deployment: identical calls failed in processes that had been up for hours and succeeded in a freshly started process against the same database in the same minute.

## Aggravating interaction

The same deployment had a WAL roughly 20x past `DEFAULT_WAL_AUTOCHECKPOINT_PAGES` (4000). Long-lived readers block checkpoint reset, an oversized WAL slows every read, slower reads hold slots longer, and slot exhaustion arrives sooner. The permit-lifetime bug is not the only factor but it is the one that converts pressure into a hard stop and makes recovery impossible without restarting the process.

## Suggested fix

Scope the permit to the operation. Acquire before the query, release when the query returns, and keep only the connection cached on the handle if caching the connection is desirable. If a retained connection must keep a permit, the pool needs to distinguish retained-idle from in-use so idle retention does not count against concurrent readers.

Raising `max_readers` is not a fix. It defers exhaustion and increases the number of readers pinning the WAL, which makes checkpoint starvation worse.

## Test that should fail today

Open more than `max_readers` `SqlReader`s through the bridge, run one query on each, drop none, then attempt one more read. Today that read times out; with the fix it succeeds.

ohdearquant:
## Summary

- scope file-backed raw-SQL reader permits to active opens and operations while retaining idle cached connections without consuming the bounded reader budget
- retain exactly one permit across an explicitly admitted `BEGIN DEFERRED` snapshot until `COMMIT`, `END`, `ROLLBACK`, cancellation, or handle drop proves the connection is back in autocommit
- reject prefixed or nested transaction-control spellings that could bypass the cached-reader state machine or batch atomicity
- add long-lived cached-reader and checkpoint regressions while explicitly keeping broader multiprocess WAL-pin issues out of scope

## Storage-safety invariants

- an operation permit is released only after the blocking SQLite call ends
- cancellation cannot return admission while a detached query or transaction still owns the connection
- poisoned or non-autocommit readers are rolled back or discarded before admission is released
- idle cached readers no longer exhaust `max_readers` or pin a WAL snapshot

## Verification

- exact independent review against current `main`: READY, no findings
- direct `rustfmt --edition 2021 --check` on all changed Rust files
- Deno formatting, ADR-reference lint, stub-marker lint, and `git diff --check`
- full GitHub Actions CI: green at exact head `1243b9e286ccdb6ca84128ffb85743cbe5e19b09` — https://github.com/ohdearquant/khive/actions/runs/31334281322

Fixes #1828


ohdearquant (commented):

Verdict on head f6ea86b3c800735cd8dc613f66966734ff4c2dd6: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head a741fc3ce6a67d5ae6d7d8254a335ddad8b7463e: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #1867, opened 2026-08-10
docs(adr): fix self-contradictory pipeline order in ADR-078 shape detection

ohdearquant:
Two adjacent Implementation subsections of ADR-078 stated the rendering pipeline order in opposite directions: the shape-detection subsection said detection runs before the redundancy-reduction pass, while the redundancy-reduction subsection said the reductions are applied before shape detection. The orders are observably different — the reductions drop keys (full_id, echoed properties children, default namespace) that the record-array detector keys on, so the same payload can classify differently under the two readings.

The shipped implementation applies the reduction pass first: `render_format` in `crates/khive-runtime/src/presentation.rs` reduces the value (unless the pass is skipped for `format=json` or verbose presentation) and then dispatches shape detection on the reduced value. This change corrects the shape-detection sentence to match the implementation and cross-references the skip condition. No code changes.

### DIFF
--- a/docs/adr/ADR-078-output-format-shape-aware-rendering.md
+++ b/docs/adr/ADR-078-output-format-shape-aware-rendering.md
@@ -471,8 +471,9 @@ resolution, when `default_output_format` is still `None`.
 
 ### Shape detection sequence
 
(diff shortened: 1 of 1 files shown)


ohdearquant:

Issue #1893, opened 2026-08-11
knowledge.compose does not honor request cancellation; abandoned requests spin and pin the WAL


ohdearquant:
Observed on a production database under normal load.

**Behavior**: three `knowledge.compose` requests whose callers had disconnected kept executing inside the daemon for over 3.5 hours at roughly 370% CPU combined. The tasks remained runnable after SIGTERM was delivered to the process and only stopped on SIGKILL.

**Impact**:
- Each stuck request held a long-lived read transaction. The oldest pinned the WAL checkpoint, growing the WAL past 20,000 pages; it collapsed to 21 pages immediately after the process was killed.
- The stuck requests saturated the reader pool, producing `sql_bridge.reader_open` timeouts across concurrent clients while the writer path stayed healthy — degraded reads with no obvious cause at the client.

**Asks**:
1. Compose (and any long-running read verb) should check for cancellation between pipeline stages so an abandoned request releases its read transaction promptly.
2. Graceful shutdown should bound how long it waits for in-flight requests before aborting them, so SIGTERM cannot hang behind a spinning task.

Related: #1846 (bounding read-transaction age) would independently cap the WAL-pin blast radius of this class.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ohdearquant:
Resolved on main by bec18a86 (PR #1897): request cancellation propagates through SQLite interruption and releases reader permits, with compose-stage and daemon-disconnect coverage.

Issue #1846, opened 2026-08-09
storage(sqlite): enforce a maximum read-transaction age across all connection types


ohdearquant:
**Defect / gap.** Nothing in the connection layer enforces a general upper bound on how long a read transaction may stay open. Every WAL-bounding guarantee (checkpoint completes, WAL file stays small, writers make progress) depends on readers eventually releasing their snapshot. Today that release is opportunistic: it happens when a caller finishes its work, not because any policy forces it. A single reader that never returns its connection — a stuck client, a leaked handle, a slow downstream consumer — pins the WAL indefinitely and defeats every other mitigation applied on the write side.

Two existing mitigations are adjacent but narrower in scope:
- one closes a permit-lifetime leak specific to a particular reader/writer pooling path, not a general age cap on any open read transaction;
- one addresses idle-process exit for one specific server transport, not an in-process transaction-age limit that applies regardless of transport.

Neither establishes a transport-agnostic, connection-type-agnostic maximum read-transaction age. A pinned reader arriving through any other code path is unaffected by both.

**Mechanism.** SQLite's WAL mode allows checkpointing to proceed only up to the oldest active reader's snapshot. A reader that never commits or rolls back holds that snapshot open forever, and checkpoint progress stalls at exactly that point — independent of how many other mitigations (bounded admission, disk-space guards, backup coordination) are layered on top of the writer path. This is a structural property of SQLite's WAL locking contract, not a bug in any one component: no combination of writer-side controls can bound WAL growth while a reader is allowed to hold a snapshot indefinitely.

**Evidence (prior-art survey).** A production incident report from an open-source embedded-SQLite replication project documents this exact failure mode directly: a long-lived-reader accommodation in their checkpoint logic caused their write path to block on readers, and the fix was removing that accommodation entirely because it produced a 100% application failure rate during the incident. SQLite's own WAL documentation states plainly that an open read transaction prevents checkpoint from reclaiming WAL pages older than that transaction's snapshot, and that there is no automatic timeout on that hold — the application is responsible for bounding it.

[...]

PR #1864, opened 2026-08-09
fix(db,runtime): bound writer-transaction work in blob GC and entity/note merges

ohdearquant:
Audits the write-transaction paths for unbounded work while the single writer connection is held, and bounds the ones that had none.

- Blob GC no longer runs inside writer transactions; its claim sweep is bounded and recoverable, so an interrupted sweep cannot strand claim rows.
- Entity and note merges now enforce per-transaction row and aggregate-byte budgets, charged inside the merge's own `BEGIN IMMEDIATE` transaction before each unit of state is retained: merge records, incident edge collection, endpoint-contract resolutions, recursive conflict-cascade rows, and config-scaled FTS/vector/pack fanout. Exceeding a limit rejects the merge with the observed counts and rolls back; a hub-node merge can no longer materialize unbounded state or hold the writer indefinitely. `MergeSummary` reports the observed usage, and the post-commit log carries the same numbers without extending the writer hold.
- The migrations test derives its expected version from `MIGRATIONS` itself instead of a hardcoded number.

Tests cover the row cap on incident-edge collection, the cascade-row cap, the byte cap on record reads, rollback on rejection, and the post-commit budget log for both entity and note merges.

Closes #1850

PR #1897, opened 2026-08-11
fix: interrupt abandoned SQLite request reads

### LINKED ISSUE #1895: Abandoned and timed-out requests keep executing their SQLite statements, exhausting the reader pool within minutes
Follow-up to #1893 with new evidence that narrows the defect and raises severity: this recurs within minutes on a freshly restarted daemon under ordinary traffic, and no index maintenance can mitigate it.

**Defect**: when a caller gives up on a query — the dispatch layer's 5s backend-search timeout, an abandoned `knowledge.compose`, or an MCP client disconnect — nothing interrupts the underlying SQLite statement. The statement keeps executing on its blocking thread, holding its reader connection permanently. Each slow search therefore converts into a leaked reader plus a core of CPU. Pool exhaustion follows within minutes; the writer path stays healthy, so the failure presents as fleet-wide read timeouts with no obvious cause.

**Evidence** (production daemon, three occurrences in one morning):
- Daemon log: `backend search task timed out backend=main timeout_ms=5000` repeatedly, while process CPU climbs (170–370% observed) and `sql_bridge.reader_open` timeouts cascade across unrelated subsystems (ANN builds, channel polls, schedule ticks).
- Shutdown log of a killed daemon: `knowledge.compose request abandoned before completion` with `phases=[("suggest", 1322942), ...]` — a suggest query still executing after 22 minutes.
- A 5s thread sample of a spinning daemon shows one thread with 1496/1496 samples inside a single `sqlite3_step`: FTS5 cursor initialization (`fts5MultiIterNew` → `fts5SegIterSeekInit`), ~75% of samples blocked on the SQLite page-cache mutex (`pcache1Fetch` → `psynch_mutexwait`) — concurrent stuck queries convoy on that mutex, which is where the CPU goes.
- Index maintenance is ruled out: live segment counts are healthy (15 segments on the largest index, ~1.0 GB, 271k blocks), so an FTS optimize would not help.

**Asks**:
1. Wire statement interruption (`sqlite3_interrupt` via the connection's interrupt handle) into every abandonment path: the dispatch backend timeout, compose phase abandonment, and MCP request cancellation. The reader slot must return promptly when the caller stops waiting.
2. As a backstop, bound statement runtime with a progress handler tied to a per-request deadline, so a statement no caller is waiting on cannot run unbounded even if an abandonment path is missed.
3. Regression coverage: a deliberately slow query fixture asserting the reader connection is back in the pool within a bounded interval after caller timeout, and that daemon CPU returns to idle.

Related: #1893 (cancellation for compose), #1846 (read-transaction age bound — caps the WAL-pin blast radius of the same class).

ohdearquant:
## Summary

- propagate one request cancellation/deadline context through MCP, daemon, coordinator fan-out, knowledge compose/search, memory recall, diagnostics, and every typed/raw SQLite read path
- interrupt abandoned SQLite statements with the exact connection interrupt handle plus a composed progress-handler deadline backstop, while quarantining connections whose cleanup cannot be proven
- keep admitted writes and reads inside writer transactions completion-preserving and non-interruptible
- return reader permits promptly after cancellation, bound daemon drain/transport EOF behavior, and stop late backend results from extending the advertised deadline
- keep the request context substrate-neutral in `khive-storage`; `khive-db` owns only SQLite-specific interruption and cleanup

## Root cause

Dropping or timing out the async request future did not stop its `spawn_blocking` SQLite work. The statement could continue indefinitely, retaining a reader connection, pinning WAL state, and consuming CPU after its caller had gone away. Several higher-level paths also swallowed cancellation as degraded success or created child work without inheriting the caller's deadline.

## Impact

Abandoned reads now stop at SQLite bailout points and release or quarantine their exact connection before reuse. A single absolute request deadline is preserved across fan-out and compose phases. MCP EOF/cancellation and daemon peer loss propagate to the same context. Admitted mutations are still awaited to their normal commit/rollback boundary and are never interrupted by the read deadline.

## Validation

- `CARGO_TARGET_DIR=/private/tmp/cargo-target-1895 cargo check --manifest-path crates/Cargo.toml --workspace --all-targets --all-features`
- direct `rustfmt --edition 2021 --check` on all changed Rust sources
- `deno fmt --check` on changed documentation
- ADR-reference, stub-marker, SQL, JSON, TOML, dependency-boundary, and `git diff --check` source gates

The new runtime regressions are compiled by the all-targets/all-features check; the full test suite has not been executed locally.

Fixes #1895.
Refs #1893.
Refs #1846.


ohdearquant:
Allocation note: this PR's ADR-091 amendment is numbered 11, but #1864 (already in flight, ahead in the merge queue) also carries an Amendment 11 ("write-transaction external-work audit"). Whichever of the two merges second needs to renumber against the merged document — numbers are allocated against the tree at merge time, and in-flight branches count. Assuming #1864 lands first, this one becomes Amendment 12.

ohdearquant:
Reproduction data from a production-scale store (~173k notes) that may sharpen the acceptance criteria: read failure under pool starvation has two independent axes, not one.

- Concurrency width, same query, same minute: width 1 and 2 succeed; width 5 fails 5/5 with `search_incomplete`, `missing_backends=["main"]`, `retryable=false` (reproduced twice; boundary somewhere between 2 and 5).
- Result-set size, single op, same minute: `search(kind=entity, ..., limit=25)` fails with the same signature while the identical query at `limit=3` returns real entities.

So a fix validated only on the concurrency arm can still leave low-concurrency high-limit reads failing, and that would read as a regression in before/after data. Suggested acceptance matrix: {width 1, width 5} x {limit 3, limit 25}, before and after, on a store large enough that limit=25 hydration is non-trivial.

One measurement caveat for any before/after comparison: the reader pool recovers on its own as load drains (observed: median probe latency 10.3s -> 2.2s over ~54 min with this fix not deployed), so a baseline captured during drain-down will falsely credit the fix. The degraded-window baseline needs to be captured under induced or naturally sustained load.

ohdearquant:
Close reading of the cancellation path at cc1ae2ff7 turns up one substantive gap and two smaller ones, plus a scope question.

**1. Post-grace detach keeps the failure mode the issue describes (`read_cancellation.rs:568-581`).** When an interrupted read does not settle within `KHIVE_SQLITE_INTERRUPT_GRACE_MS`, the code calls `worker.abort()` and returns a timeout — but `spawn_blocking` cannot be force-aborted once started (the code's own comment says so), so the worker keeps the pooled connection, admission, and any WAL read snapshot until SQLite actually honors the interrupt. In the common case the `InterruptHandle` has already fired and the window is short; in the worst case (hostile/slow UDF, long busy segment) the caller sees a timeout while the slot and snapshot stay live — which is exactly the "stop reporting, keep running" behavior of #1895, now bounded in caller latency but not in resource lifetime. ADR-091 Amendment 11 ("abandoned reads release WAL snapshots ... after its interrupt settles", `ADR-091:1370-1377`) reads as a stronger guarantee than this; ADR-005:590-599 candidly describes the actual detach semantics, so the two amendments currently disagree. Suggested resolution: either keep the operation joined until the worker finalizes (accepting the hostile-UDF latency cost), or track detached workers in a small retirement registry so their eventual release is observable and enforceable — and align the Amendment 11 wording with whichever guarantee ships. A regression that forces a worker past grace and asserts reader admission plus `wal_checkpoint(PASSIVE)` recovery would pin this.

**2. Raw SQL prepares before the interrupt target registers (`sql_bridge.rs:658-681`).** A cancellation arriving while `prepare_bound_statement` is still running has no interrupt target, and an unregistered raw operation takes the completion-preserving branch (`read_cancellation.rs:584-590`) — so an abandoned raw read's admission can be unbounded. Whether preparation is ever long-running in production is unverified from static review, but a cancellable preflight boundary (or a statically read-only raw API) would close it.

**3. Cached read transactions still bypass `tx_registry` (`sql_bridge.rs:862-875`, `1150-1165`).** The permit lifecycle is sound, but a post-grace detached reader holding an explicit transaction is invisible to the transaction-age diagnostics that ADR-091:220-240 expects. This preserves the pre-existing #1903-class observability gap; worth fixing here or explicitly deferring.

**Scope:** commit 4092e3f99 (config-ledger event attribution in `pack.rs`) is unrelated to read cancellation per its own diff — split it out or note why it rides along.

The width-axis admission work (`pool.rs:659-714`, predicate-aware checkout) and the deadline inheritance through the coordinator (`dispatch.rs:533-537`, `718-867`) both look right, and the cooperative-cancellation test coverage is solid; the items above are what keeps this from closing #1895's acceptance as stated.

PR #1920, opened 2026-08-11
fix(mcp): preserve bounded search backend errors

### LINKED ISSUE #1829: search reports which backend failed but never why, including at debug log level
## Summary

When a search backend fails, `search` correctly refuses to report a no-match, but it discards the underlying error. The response names which backend failed and never why, and the cause is not recoverable from logs either, including at `KHIVE_LOG=debug`.

## Observed

```
$ kkernel exec 'search(kind="entity", query="...", limit=2)'
{"error":{"kind":"search_incomplete",
          "message":"no-match was not established because selected backends failed",
          "missing_backends":["main"],
          "retryable":false}}
```

Re-running the same command with `KHIVE_LOG=debug` produces no additional line naming the failure. The operator learns that `main` failed and has no way to learn what happened to it.

## Why this matters

The refusal itself is right and should not change. A search that returned `[]` here would manufacture a fabricated absence, and callers downstream would record a confident "nothing found" for a question that was never actually asked. That behavior is worth keeping.

The problem is that the diagnosis stops there. `retryable: false` tells a caller not to retry but gives an operator nothing to act on, so a failing backend cannot be distinguished from a misconfigured one, a locked one, or one whose index is unavailable. In practice this turns a recoverable condition into an investigation that has to proceed by elimination from outside the process.

## Suggested fix

Carry the per-backend error into the response, alongside `missing_backends`, for example a `backend_errors` map from backend name to the error kind and message that backend produced. At minimum, log the cause at `warn` when a backend is dropped from a search, so the reason survives even if the response shape stays fixed.

A useful invariant: any response that reports a partial or refused result should be able to answer "why" from its own payload, without requiring a separate reproduction with different logging.

ohdearquant:
## Summary

- preserve bounded, masked per-backend search failure causes alongside missing backend IDs
- keep exact missing/error parity for partial success and all-failed `search_incomplete` envelopes
- bound diagnostic count, keys, messages, and per-operation wire budget
- warn with the same sanitized causes without exposing unbounded backend detail

## Validation

- bounded/masked/parity tests: 2/2 passed
- real degradation envelope tests: 6/6 passed
- frame omission tests: 2/2 passed
- Rust formatting and `git diff --check`: passed

Stacked on #1897.

Fixes #1829.


ohdearquant:
The branch is currently conflicting with main (main has moved under the stack — at least the checkpoint cadence and Windows walpin validation changes landed after this branch's base). Please rebase onto current main; that will also collapse the stale-base diff to the real ~10-file delta and trigger a fresh CI run at the new head, which this PR is missing (only the bench and pipeline-gate workflows ran at b768ff6). The bounded backend_errors design itself reads well — the ADR-130 amendment, the wire-budget const assert, and the masked coordinator warnings all line up with the implementation.

PR #1922, opened 2026-08-11
fix(runtime): attribute config ledger events

ohdearquant:
Split from #1897 so the read-cancellation change stays single-purpose. One-file fix: config ledger events carry the acting identity instead of the default attribution. Original change authored on the #1897 branch; no functional delta from the version reviewed there.


ohdearquant:
Superseded: the ConfigLocked verb misattribution this fixes landed on main in #1897 (merge commit bec18a86e), where the drained events now carry a config.lock pseudo-verb — same defect, dotted-verb naming consistent with the verb namespace. The change here edits the same lines and no longer applies.

PR #1964, opened 2026-08-15
fix(retrieval): guarantee untruncated recall pool and harden scratch-path handling

ohdearquant:
## Summary

Two defects in the retrieval evaluation harness.

**Recall@100 was computed from a possibly-truncated pool.** `memory.recall` applies a response character budget after ranking: below the requested `top_k` the response stays a bare array, just shorter, with surviving hits stamped `"truncated": true`. The harness accepted bare lists of any length, so budget-truncated pools were silently scored as Recall@100 — and the committed gold was itself affected (overall Recall@100 0.9972 at the default 4000-token budget; 1.0 on the full pool). Fixes:
- `HERMETIC_RECALL_CONFIG` now pins `default_token_budget` to the fixed server-side maximum (16000), keeping the config hermetic while giving the 400-note corpus ~2x headroom.
- Response validation now raises if the pool is shorter than `min(top_k, corpus_size)` or any hit carries a truncation marker, naming the query and observed depth.
- Gold regenerated on the untruncated pool (only the two Recall@100 values change).

**Scratch-path handling was validate-then-use.** Path checks and later opens/writes/deletes were separate steps, so components could be swapped for symlinks between them. The scratch root is now opened once with a component-by-component `O_DIRECTORY | O_NOFOLLOW` walk (honoring the macOS `/private` firmlink mapping), and all scratch access goes through that fd: `O_CREAT | O_EXCL | O_NOFOLLOW` claims, fd-relative reads, fd-relative recursive cleanup that unlinks symlink entries as leaves, and — where a subprocess needs a path string — a device+inode match between the fd-opened leaf and the resolved path, recomputed immediately before each spawn.

## Tests

Self-test suite extended 9 → 11 checks: a symlink swapped in after the claim is refused at use with the victim untouched, and a swapped-directory (non-symlink) substitution is refused by the device+inode recheck. `--self-test`, `make eval-retrieval-gold-check`, and the repository data-file guard all pass.

### DIFF
--- a/benches/retrieval/README.md
+++ b/benches/retrieval/README.md
@@ -62,6 +62,24 @@ baseline.
   detects — but does not prevent — that residual race for paths the
   consumer only reads. Do not read this harness's path verification as
   full race-resistance for the lifetime of a returned path.
+
+  **Residual TOCTOU window in cleanup (by design, not closed):**
+  `cleanup_scratch` walks the scratch tree through a TOCTOU-guarded fd and,
+  at every recursion depth, checks each directory's identity two ways: its
+  `fstat`ed (device, inode) must match the inode `os.scandir()` observed
+  for that name moments earlier (catching a directory swapped in *after*
+  this run started walking the tree — same mechanism regardless of nesting
+  depth, not just at the top level), and, for `home`/`tmp`, against the
+  identity recorded when this run created them. Neither check can see a
+  substitution that already happened *before* this run's own listing of a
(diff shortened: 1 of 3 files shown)


ohdearquant (commented):

Verdict on head 075a0cba5345dee65274fa7513bd7e370ca3d30e: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #1967, opened 2026-08-15
feat(retrieval): offline SPLADE sparse-fusion experiment sidecar

ohdearquant:
## Summary
Adds `benches/retrieval/splade_offline.py`, an additive-only sidecar that computes a SPLADE-family sparse-retrieval leg offline (precomputed sparse vectors, no new Rust dependency, no serving code) and RRF-fuses it against the committed `A_fused_direct` condition's own ranking with the runtime's RRF rule (`DEFAULT_RRF_K = 60`, matching `crates/khive-fusion/src/strategy.rs`). No committed condition, gold file, or `CONDITIONS` entry is modified; `evaluate.py --check-gold` still passes after the addition.

## Measured (fixed 400-note corpus, 40 graded queries, paired bootstrap n=40)
- Overall nDCG@10: 0.5921 → 0.7001 (Δ +0.108, 95% CI [0.092, 0.124], p < 0.0001, Cohen's dz 2.1)
- Positive and significant in every query class; Recall@100 / MRR@10 already saturated at 1.0 in baseline, unchanged.
- Robust to RRF k (k=60 vs k=10: Δ < 0.012).

## Methodology notes (also in README)
- Two-stage offline RRF (SPLADE leg fused against an already-fused ranking) is a conservative approximation of a native three-leg fusion; bias direction documented in the README section.
- Encoder: `naver/splade-cocondenser-ensembledistil` (ungated; `splade-v3` is gated on Hugging Face). Sparse-vector fixture (~1MB) is gitignored — over the repo's 512KB large-file cap and regenerable in ~17s via `--encode`.
- The harness's synthetic corpus derives target content from query text, so it cannot exhibit a low-recall `fresh_directive` scenario; this benchmark therefore measures whether the SPLADE leg is additive and safe, not whether it closes a recall gap on harder data. A follow-up hardening the `fresh_directive` corpus class is planned.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,6 +93,11 @@ implementation_notes.md
 # Real-data recall benchmark corpus (extracted from khive-graph.db, never commit)
 crates/khive-pack-memory/tests/fixtures/
 
+# Locally-regenerated SPLADE sparse-vector fixture (~1MB, over the 512KB
+# pre-commit large-file cap) — regenerate with
+# `benches/retrieval/splade_offline.py --encode`
+benches/retrieval/fixtures/
+
 uv.lock
 pyproject.toml
 # Exception: the khive-contract integration suite is a committed uv project
--- a/benches/retrieval/README.md
+++ b/benches/retrieval/README.md
@@ -172,6 +172,50 @@ drift. Root-causing that tie-break is out of scope for this eval-only PR
 gate meaningful for real regressions without chasing a single flipped rank
 at an exact score tie.
 
+## Experimental: offline SPLADE sparse-retrieval fusion
+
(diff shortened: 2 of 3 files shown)


ohdearquant:
Closing: this sidecar was an offline retrieval experiment, and the evaluation it supported has concluded — the measured production baseline does not justify integrating sparse-fusion into the serving path, so this code will not move toward merge. Branch retained for reference/reproducibility. If the offline comparison is revisited later, the known methodology gaps to fix first are: pinning the encode workflow's third-party dependencies and model artifacts, correcting the recall denominator when fewer than 100 valid candidates exist, and tie-breaking fused rankings by backing document ID.

Issue #1982, opened 2026-08-15
knowledge.search: documented 0.46/0.42 score bands are stale since the s/(s+1) squash; full-scan fallback scores off-topic queries in the on-topic band


ohdearquant:
## Observed

The knowledge.search verb description and docs state score bands (>=0.46 reliably on-target, 0.42-0.46 mixed, <0.42 off-target). Measured today:

- Known-present on-topic atoms score 0.34-0.47 (many under 0.46 under their own exact titles).
- A deliberately nonsense query ("qlkjz xwryb mnopv ...") returns atoms at 0.432-0.453 — inside the documented mixed/on-target band.

The bands do not separate on-topic from off-topic in the current pipeline.

## Root cause (source-read)

1. The bands were documented 2026-07-01 against raw TF-IDF scores (range roughly 2-10 for on-topic hits). The s/(s+1) squash introduced 2026-08-04 (79e6e4ffb) compresses all scores into [0,1), so even a perfect reviewed hit lands near 0.5 and typical good hits land 0.35-0.47. The threshold text was never recalibrated.
2. When FTS finds zero candidates (nonsense tokens), the full-scan fallback feeds the entire atom table through the TF-IDF scorer, producing irreducible 0.43-0.45 noise that no absolute floor separates from genuine hits.

## Impact

Presence/coverage checks thresholding on the documented bands read the corpus as emptier than it is and generate research/backfill requests for material that exists.

## Proposed

- Correct the verb description and docs: scores are squash-normalized to [0,1); absolute-score floors are not a presence signal. Recommend rank-based admission for presence checks.
- Separately consider bounding the full-scan fallback (e.g. mark fallback-sourced hits, or floor them at 0) so no-match queries do not emit on-target-looking scores.

ohdearquant:
Resolved by #2611. Search documentation now describes scores as request-relative and uncalibrated, and recommends rank plus provenance for coverage checks. Genuine FTS misses no longer rank unrelated recent atoms; ANN-only results are explicitly labeled. The lexical-miss regression and fusion provenance tests cover both paths.

PR #1983, opened 2026-08-15
fix(knowledge): preserve atom finalized flag across upserts

### LINKED ISSUE #1980: knowledge.search: finalized flag is never true on atom results, inverting corpus-presence checks
## Observed

Filtering knowledge.search results to finalized=true — the natural reading of "the finished material" — returns zero atom records and drops every real, finalized corpus atom.

Two independent measurements:

1. Over 25 hits from 5 queries against the atom corpus: all 13 finalized=true records resolved to no atom in the corpus (join-independent exact-title re-checks on 4 of them returned 0), while all 12 finalized=false records resolved to real atoms with 86-202 body lines. Zero exceptions.
2. A separate 6-result query for a known canonical atom (a3-g.0312 parameter-binding, ~190 body lines) returned it with finalized=false; the only finalized=true records in the result set were kind=domain, not atoms.

## Mechanism hint

The second measurement shows a clean kind split: every atom-kind result carried finalized=false and every finalized=true result was a domain record. The flag may be populated at the domain level and never propagated to (or read from) atom rows, rather than being stored inverted.

## Impact

Corpus-presence checks driven by this flag return false negatives for material that exists, which has already generated research/backfill requests for atoms that are present in the corpus.

## Expected

finalized on a search result reflects the finalization state of that record itself, for atoms as well as domains; a finalized corpus atom returns finalized=true.

### LINKED ISSUE #1985: knowledge.upsert_atoms: UPDATE path clears source_type/source_uri on rows that carry them
The atom upsert payload (\`{slug, name, content, tags?, properties?, finalized?}\`) has no source-attribution fields, and the UPDATE path writes \`source_type\`/\`source_uri\` unconditionally from the (absent) input. The result: re-upserting any existing atom by slug NULLs both fields on a row that previously carried them.

## Reproduction (measured against a live store)

1. A row created by an ingest path carries \`source_type\` and a \`source_uri\` identifying its origin.
2. One \`knowledge.upsert_atoms\` call on the same slug, supplying only name/content (and \`finalized\`).
3. Read the row back: \`source_type\` and \`source_uri\` are both null. \`created_at\` survives, properties shallow-merge correctly — only the source attribution is destroyed.

## Impact

Any maintenance or backfill pass that re-upserts existing atoms strips provenance from every row it touches. Same defect class as #1980 (a field the payload does not carry is overwritten with its zero value instead of preserved).

## Expected

The UPDATE path preserves existing \`source_type\`/\`source_uri\` when the input does not carry replacements, consistent with the fix shape in #1983. Refs #1980, #1983.

ohdearquant:
## Summary

Three defects in the knowledge pack, filed as #1980, #1982, and #1985:

1. **`finalized` flag erasure on re-upsert.** The `upsert_atoms` UPDATE
   path unconditionally wrote `finalized=?7`, bound from
   `finalized.unwrap_or(false)`. Any re-upsert against an existing slug
   that omitted the `finalized` field silently reset it to `false`,
   even for atoms that had previously been finalized. The INSERT path
   was already correct (no prior row to erase).
2. **Stale score-band guidance.** The `knowledge.search` and
   `min_score` parameter descriptions cited `score>=0.46` /
   `0.42<=score<0.46` bands as separating on-topic from off-topic
   results. Those bands were calibrated against raw TF-IDF scores
   before a later `s/(s+1)` score-squash normalization was introduced;
   the squash compresses all scores into `[0,1)`, and on-topic hits now
   land in the same 0.4-0.5 band as off-topic queries served through
   the full-scan fallback. The bands no longer separate anything.
3. **`source_uri`/`source_type` erasure on re-upsert.** Same class of
   bug as #1980, same UPDATE statement: `source_uri=?5,
   source_type=?6` were bound `NULL` whenever the caller's payload
   omitted the fields, and the plain assignment then overwrote any
   previously-recorded source attribution. A re-upsert of an existing
   slug supplying only `name`/`content` cleared both fields on a row
   that had them set from a prior insert.

## Fixes

1. `crates/khive-pack-knowledge/src/knowledge/crud.rs`: the UPDATE SQL
   now uses `finalized = COALESCE(?7, finalized)`, with `?7` bound as
   `NULL` when the caller omits the field (instead of coercing to
   `0`). An omitted field now preserves the existing value; explicit
   `true`/`false` still set/clear it as before. The
   draft→reviewed status-promotion `CASE` is unaffected, since
   `NULL = 1` evaluates falsy in SQLite. The INSERT path and the
   hardcoded `upsert_domains` mirror-atom path are unchanged — both
   were already correct.
2. `crates/khive-pack-knowledge/src/vocab.rs`: replaced the stale
   score-band sentence in both the `knowledge.search` description and
   the `min_score` parameter description with guidance to use result
   rank, not absolute score, for presence/coverage checks. No scoring
   behavior changed.
3. `crates/khive-pack-knowledge/src/knowledge/crud.rs`: the same UPDATE
   statement now uses `source_uri = COALESCE(?5, source_uri)`,
   `source_type = COALESCE(?6, source_type)` — same shape as the
   `finalized` fix. An omitted field preserves the existing
   attribution; an explicit value still replaces it. The INSERT path
   is unchanged.

Other `0.46`/`0.42` band references were found in
`crates/khive-pack-knowledge/README.md`, `docs/guide/api-reference.md`,
and `docs/adr/ADR-047-knowledge-pack.md` (prose docs, not code
descriptions) — left for a separate documentation change.

## Test plan

- [x] `cargo test -p khive-pack-knowledge` — full crate green across
      the unit-test binary and all integration suites (fixes.rs,
      param_strictness.rs, search_candidate_refill.rs, and the rest)
- [x] `cargo clippy -p khive-pack-knowledge --all-targets -- -D
      warnings` — clean
- [x] Added three regression tests in `tests/fixes.rs` for the
      `finalized` fix: omitting `finalized` on re-upsert preserves the
      existing flag (proved red on the pre-fix UPDATE SQL, green after
      the fix); explicit `finalized=false` still clears the flag;
      explicit `finalized=true` on a draft still sets the flag and
      promotes status to `reviewed`.
- [x] Added two regression tests in `tests/fixes.rs` for the
      `source_uri`/`source_type` fix: omitting both fields on
      re-upsert preserves the existing values (proved red on the
      pre-fix UPDATE SQL, green after the fix); explicit new values on
      re-upsert still replace the existing ones.

Fixes #1980
Fixes #1985
Refs #1982

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/crud.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/crud.rs
@@ -162,8 +162,11 @@ impl KnowledgeHandlers {
                     .execute(SqlStatement {
                         // Promote draft -> reviewed when this upsert finalizes the atom.
                         // Never demote an already reviewed row, and leave status
-                        // untouched when not finalizing.
-                        sql: "UPDATE knowledge_atoms SET name=?1, content=?2, tags=?3, properties=?4, source_uri=?5, source_type=?6, finalized=?7, status = CASE WHEN ?7 = 1 AND status = 'draft' THEN 'reviewed' ELSE status END, updated_at=?8 WHERE id=?9 AND namespace=?10".into(),
+                        // untouched when not finalizing. finalized=COALESCE(?7, finalized)
+                        // preserves the existing flag when the caller omits it; SQLite's
+                        // `NULL = 1` evaluates to NULL (falsy), so an omitted field also
+                        // leaves the status CASE on its ELSE branch.
+                        sql: "UPDATE knowledge_atoms SET name=?1, content=?2, tags=?3, properties=?4, source_uri=?5, source_type=?6, finalized=COALESCE(?7, finalized), status = CASE WHEN ?7 = 1 AND status = 'draft' THEN 'reviewed' ELSE status END, updated_at=?8 WHERE id=?9 AND namespace=?10".into(),
                         params: vec![
                             SqlValue::Text(atom_in.name.clone()),
                             SqlValue::Text(content.clone()),
@@ -171,7 +174,7 @@ impl KnowledgeHandlers {
                             props_json.as_ref().map_or(SqlValue::Null, |p| SqlValue::Text(p.clone())),
                             source_uri.map_or(SqlValue::Null, |s| SqlValue::Text(s.to_string())),
                             source_type.map_or(SqlValue::Null, |s| SqlValue::Text(s.to_string())),
-                            SqlValue::Integer(atom_in.finalized.unwrap_or(false) as i64),
+                            atom_in.finalized.map_or(SqlValue::Null, |f| SqlValue::Integer(f as i64)),
                             SqlValue::Integer(now),
                             SqlValue::Text(id),
                             SqlValue::Text(ns.clone()),
--- a/crates/khive-pack-knowledge/src/knowledge/crud.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/crud.rs
@@ -165,8 +165,10 @@ impl KnowledgeHandlers {
                         // untouched when not finalizing. finalized=COALESCE(?7, finalized)
                         // preserves the existing flag when the caller omits it; SQLite's
                         // `NULL = 1` evaluates to NULL (falsy), so an omitted field also
-                        // leaves the status CASE on its ELSE branch.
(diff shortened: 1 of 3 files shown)


ohdearquant:

PR #1970, opened 2026-08-15
docs(adr): supersession chain canonicalization for memory recall (ADR-157)

ohdearquant:
Proposes ADR-157: treat verified supersession as truth canonicalization rather than a general ranking signal.

A production-derived labelled evaluation of `memory.recall` (42 real queries, 22 with a labelled stored answer) measured nDCG@10 0.673 with the correct answer in top-10 for 91% of answerable queries. Both complete misses shared one mechanism: a terse correction note that intentionally replaced an earlier note, recorded only in text (no `supersedes` edge), losing lexically to the richer note it replaced.

The ADR adopts three bounded pieces, activation-gated on replaying the frozen evaluation:
1. Optional atomic `supersedes` capture on `memory.remember` (explicit UUIDs only, cycle-checked, all-or-nothing).
2. Reviewed offline backfill of historical text-recorded supersession claims (dry-run proposals, never automatic edges).
3. One batched serve-time chain-head substitution before scoring, replacing the current per-candidate (N+1) supersedes filtering with a single bounded closure query.

It explicitly rejects cluster recency, reranker features, and text-derived automatic edges, with implementation fences and a degradation contract in the document.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1989, opened 2026-08-15
docs(adr): durable edge-governance provenance for supersession canonicalization (ADR-159)

ohdearquant:
Defines the durable edge-governance mechanism that supersession chain canonicalization (ADR-157, #1970) references but does not define: which `supersedes` edges are authorized to substitute one memory note for another in recall output.

## Design

- **Append-only decision log** (`edge_governance_decisions`): every authorization or rejection, bound to the exact edge preimage (namespace, endpoints, relation), naming the authorizer and the authority evidence.
- **Serving projection** (`edge_governance_active`): one row per currently governed edge, repeating the bound preimage so the recall join fails closed even if an invalidation path regresses; covering index makes closure one indexed lookup plus one unique join, no JSON parsing, no hot-path crypto.
- **SQLite invalidation triggers** at the mutation seam: endpoint rewrites, merges, deletes, and same-key resurrection all clear the marker — handler-only clearing is insufficient because merge rewires issue raw `UPDATE graph_edges` statements.
- **Activation receipt** (`edge_governance_state`) with a short single-writer `BEGIN IMMEDIATE` cutover fence, closing the inventory-to-cutover write gap with a durable disposition for every late arrival.
- **Single-principal vs multi-actor activation modes**: multi-actor fails closed until an endpoint-scoped authority provider exists.

Alternatives (metadata reserved key, dedicated columns, UUID-keyed sidecar, event-derived, HMAC) are each rejected with the decisive failure recorded, including the resurrection failure trace that kills a UUID-only sidecar.

Includes the measured migration census (625 `supersedes` edges store-wide, 234 memory-to-memory) and the authority-coverage baseline (99% shared-namespace, no owner signal) that motivates fail-closed single-principal activation.

Docs-only; no code paths touched. Contract deltas required from a forthcoming ADR-046 amendment are stated as an explicit interface (§5, §"amendment" fences).

### DIFF
--- a/docs/adr/ADR-159-edge-governance-provenance.md
+++ b/docs/adr/ADR-159-edge-governance-provenance.md
@@ -119,27 +119,51 @@ can join cheaply.
 
 ### 1. Storage shape
 
-Three objects, added by versioned migration (ADR-015):
+Four objects, added by versioned migration (ADR-015):
 
 **`edge_governance_decisions`** — append-only record of every authorization
 or rejection:
 
 - `decision_id` (PK), `edge_id`;
 - bound preimage: `edge_namespace`, `source_id`, `target_id`, `relation`
-  (constrained to `supersedes`);
+  (constrained to `supersedes`), `target_backend`;
 - `disposition`: `authorized` | `rejected`;
-- `via`: `owner_direct` | `reviewed_proposal` | `migration_review`;
+- `via`: `owner_direct` | `reviewed_proposal` | `migration_review` |
+  `cutover_fence`;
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Approving at 06152cb9.

Every point previously raised against this PR has been re-checked against the
reworked text at this exact commit. Each was located by subject rather than by
the line number it was originally cited at, since the rework moved the lines.
All four are discharged at source. An independent read of the whole delta was
made at this same commit by a second reader.

The per-point checks above are my own read of this commit, not a
countersignature on someone else's.

PR #1991, opened 2026-08-16
docs(adr): ADR-046 amendment — edge-governance authority for memory supersession

ohdearquant:
Amends ADR-046 with the two contracts ADR-159 (durable edge-governance provenance, #1989) delegates to the proposal lifecycle:

- **Endpoint-scoped reviewer authority** for governance-bearing changesets: an `AuthorityReceipt` for `memory.supersede` is required at `review(approve)` (typed `ReviewerNotAuthorized` refusal before any event), revalidated at apply; self-approval is unconditionally forbidden for these changesets; multi-actor deployments without a real authority provider fail closed.
- **`ClassifyExistingEdgeGovernance`** — a new closed changeset arm classifying an existing memory-supersession edge in place: mandatory full-preimage equality (the last-writer-wins stale-target default explicitly does not apply), no `graph_edges` mutation, byte-for-byte preservation, decision + active-marker rows written in one transaction. Governance-bearing `AddEdge` applies map to `governed_link` and never degrade to a plain ungoverned link.

Everything outside governance-bearing changesets is byte-for-byte the base ADR. Verification checklist included (§A6).

Merge order: after #1989 (this amendment cites ADR-159's storage contract). Docs-only.

### DIFF
--- a/docs/adr/ADR-046-event-sourced-proposals.md
+++ b/docs/adr/ADR-046-event-sourced-proposals.md
@@ -116,7 +116,7 @@ pub enum ProposalChangeset {
         edge_id: Uuid,
         expected: ExpectedEdgePreimage,
         disposition: GovernanceDisposition,
-        reason_code: Option<String>,
+        reason_code: Option<GovernanceReasonCode>,
     },
     /// Compound: an ordered sequence of the above, applied atomically.
     /// NOTE: as of PR #517, only single-step Compound is accepted at propose-time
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Approving at 136cc9b4.

Every point previously raised against this PR has been re-checked against the
reworked text at this exact commit. Each was located by subject rather than by
the line number it was originally cited at, since the rework moved the lines.
All twelve are discharged at source. An independent read of the whole delta was
made at this same commit by a second reader.

One non-blocking observation remains, on the Compound arm, and it is routed as a
follow-up PR rather than a change here so that this head does not move.

The per-point checks above are my own read of this commit, not a
countersignature on someone else's.

PR #1998, opened 2026-08-16
docs(adr): recursion and invalidation-write verification arms; ADR-157/159 accepted

ohdearquant:
Docs-only follow-up to the ADR-157/159/046 series, adding the two verification arms identified during that series' review as landed-without-a-named-test:

**ADR-046 §A6 — recursion pair.** A1 specifies that governance-bearing classification recurses through `Compound` nesting, but no arm exercised it. The new arm names two distinct proposals (A1's rejected shapes and its governance-bearing class are disjoint): a nested rejected shape must refuse at propose time as its flat form does, and a nested governance-bearing step must require reviewer authority at review time. A shallow outer-variant classifier must redden both, proving the recursive evaluation is load-bearing.

**ADR-159 §Verification — invalidation-write mutation control.** §3 states the invalidation ledger row is load-bearing for same-key resurrection, but the existing rebuild arm only exercises the path where the append works, and the activation test covers trigger drop/alter through the serving-side gate — under which a rebuild-path assertion can become unreachable rather than red. The new clause requires an isolating mutant (monitored trigger DDL untouched; the append to `edge_governance_invalidations` alone no-opped, e.g. via a `BEFORE INSERT ... RAISE(IGNORE)` trigger on that table — the decision log is a separate table, so setup-phase decision inserts are unaffected) and requires the rebuild test to fail by spent-decision reactivation on its rebuild step, with the expected reddened arm stated before the run.

**Status flips.** ADR-157 and ADR-159 headers move to accepted. The acceptance events occurred in the order the documents' ordering clauses mandate, and are on the public record as the sequence of merges to main: ADR-159 first (merge commit c807de1c), the ADR-046 amendment second (merge commit 6ecd5374), ADR-157 last (merge commit 3a96f1b1). These flips record that completed sequence — the pre-flip state, in which the merged, in-effect amendment referenced a still-Proposed ADR-159, was the inconsistency.

No code changes.

### DIFF
--- a/docs/adr/ADR-046-event-sourced-proposals.md
+++ b/docs/adr/ADR-046-event-sourced-proposals.md
@@ -1139,14 +1139,17 @@ reviewer test.
   natural-key upsert selects an actively governed incumbent fails the
   apply with the stale-preimage contract, appends no decision row, and
   leaves the incumbent's projection row untouched.
-- A recursion pair on the A1 classification: a proposal whose only
-  governance-bearing step is nested at depth two or more —
-  `Compound([Compound([AddEdge{supersedes, …}])])` — is treated as
-  governance-bearing by every propose-time and review-time check: the
-  propose-time rejection shapes fire on the nested form, and a
-  provider-denied reviewer's approve is refused exactly as for the flat
-  form. Mutation control on the same pair: a classifier restricted to the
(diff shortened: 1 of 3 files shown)


ohdearquant (commented):

PR #2002, opened 2026-08-16
docs(adr): propose agent process lineage (ADR-161)

ohdearquant:
Proposes ADR-161, extending ADR-142's agent process model with durable lineage.

- `parent_agent_id` + `lineage_depth` on the process record, derived by the runtime from its own dispatch context (never caller-supplied), immutable, acyclic by construction, with a bounded maximum depth.
- Two read-only enumeration verbs, `agent.list` and `agent.descendants`, that never wake a process, carry an explicit `complete` discriminant, and apply the existing lifecycle-record authorizer per record so enumeration and `agent.observe` can never disagree about visibility.
- `agent.kill` gains an optional `descendants` parameter (default false, preserving ADR-142 semantics exactly): parent-first, per-record-authorized, per-record-reported cascade.
- Children survive their parent; lineage is history, not supervision, and confers no authority.

Docs-only change: one new ADR plus its catalog row. No code, no contract changes to any existing surface (additive fields, additive verbs, default-false parameter).

### DIFF
--- a/docs/adr/ADR-161-agent-process-lineage.md
+++ b/docs/adr/ADR-161-agent-process-lineage.md
@@ -324,7 +324,9 @@ with a published default. Total records visited by a single enumeration or by on
 is capped independently of `limit` by `lineage_visit_limit`, because a walk can visit far more
 records than it returns once filters and authorization are applied; exceeding it is a
 per-operation error naming the bound, never a silently short answer — a truncated-looking
-success here would be indistinguishable from a small tree. Cascade passes are capped by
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

Verdict on head 968da48747f61f12721e354a1c032e4e478f0555: REQUEST-CHANGES, 5 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2001, opened 2026-08-16
fix(runtime): fail closed on gate errors

ohdearquant:
AI-assisted: Implemented by OpenAI Codex for maintainer review.

## Summary

Implements ADR-160 Phase 0 / ADR-129 Stage 1a at the two `VerbRegistry` dispatch seams:

- return typed `RuntimeError::GateUnavailable { verb, reason }` when `Gate::check` fails
- refuse before normal pack handlers or intercepted operations can run
- emit `decision="gate_unavailable"` with `EventOutcome::Error`, no `deny_reason`, and empty obligations
- keep audit persistence best-effort without allowing an append failure to reopen dispatch or replace the typed refusal
- preserve explicit `Deny` as `PermissionDenied` / `EventOutcome::Denied`
- align public runtime, gate, Rego, and dependent ADR prose with ADR-129's accepted fail-closed contract

## Root cause

Both normal and intercepted dispatch converted a `GateError` into “no gate decision” and continued execution. A fallible authorization backend could therefore become an authorization bypass.

## Regression guards

Tests pin:

- zero handler invocation on normal Gate failure
- zero closure entry on intercepted Gate failure
- typed error and outage audit shape at both seams
- explicit Deny/outage separation
- one structured `gate.check` trace
- audit append failure cannot reopen dispatch
- Gate failure takes precedence over the `git.digest` receipt path
- stable `gate_unavailable` audit wire encoding

## Scope boundary

This PR does not flip the default `AllowAllGate`, implement later ADR-160 phases, or broaden into the already-fail-closed `authorize*` token-minting APIs.

## Validation

- `cargo test --workspace`
- `cargo check --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace --no-run`
- `cargo fmt --all --manifest-path crates/Cargo.toml -- --check`
- `deno fmt --check` on all changed Markdown
- `sh scripts/lint-adr-refs.sh` (376 files / 206 titled references)
- `git diff --check`
- `gitleaks protect --staged --redact --no-banner`

Three independent read-only review passes (dispatch/resource safety, public wire/provenance, and regression coverage/spec drift) returned PASS after the final doc corrections.

PR #2004, opened 2026-08-16
feat(storage): add bounded verified blob reads (ADR-160 phase 1)

ohdearquant:
## Stack / merge gate

- Depends on #2001 (ADR-160 Phase 0).
- Base is `codex/adr-160-phase-0-gate-fail-closed`; this PR cannot merge ahead of Phase 0.
- Implements **only ADR-160 Phase 1**.
- ADR-160 was accepted on `main` in `135c3d46` (#2005). Keep this PR draft until its parent-stack order and Claude review are satisfied.

## What changed

- Add the required `BlobStore::get_bounded_verified` contract, the portable 64 MiB envelope, and typed too-large / size-mismatch / digest-mismatch failures.
- Fs: resolve the configured root once, open root/shards/leaf without following links, pin one regular-file handle, read at most `max + 1`, then enforce size-before-BLAKE3 precedence.
- S3: one GET, zero HEAD, one absolute GET+body deadline, metadata early refusal, pre-append cumulative bounds, EOF size consistency, then BLAKE3.
- Update the read-only decorator and test fixtures without migrating any production consumer.
- Add shared fixture conformance, adversarial Fs/S3 cases, and a focused Windows execution lane.

## Resource-safety invariants

- `max_bytes > 64 MiB` fails before backend/provider work.
- Metadata is only an early-refusal hint and integrity witness, never authority for returned bytes.
- Oversize, truncated, grown, path-swapped, symlinked, or digest-mismatched objects never escape as successful bytes.
- Legacy unbounded `get` remains unchanged in this phase; runtime admission, cancellation supervision, consumer migration, and deletion of `get` are intentionally deferred to Phases 2–3.

## Validation

- `cargo check -j4 --workspace --all-targets`
- `cargo clippy -j4 --workspace --all-targets -- -D warnings`
- affected-crate Clippy rerun after final review fixes
- Fs bounded suite: 9/9
- S3 bounded suite: 9/9
- shared blob conformance: 5/5
- storage/runtime/moodboard targeted suites
- full workspace run reached one unrelated `khive-mcp` Unix-socket timing failure; its exact isolated rerun passed (403 sibling tests passed in that package run)

Two independent static review passes report no remaining merge blockers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -623,6 +623,11 @@ jobs:
       # Keep this still-new lane fast; widen alongside the `cargo check` step above.
       - name: khive-db walpin Windows FFI tests
         run: cargo test -p khive-db --lib walpin
+      # ADR-160 Phase 1: compile-only coverage cannot validate the Windows
+      # no-reparse, handle-pinned bounded blob read. Keep this lane focused on
+      # one normal-path test until the broader Windows suite is affordable.
+      - name: khive-db bounded blob read Windows test
+        run: cargo test -p khive-db --lib bounded_verified_get_accepts_exact_and_portable_maximum_limits
 
   coverage-ratchet:
     name: Coverage ratchet
--- a/crates/khive-db/src/stores/blob.rs
+++ b/crates/khive-db/src/stores/blob.rs
@@ -12,14 +12,16 @@
 
 use std::collections::HashMap;
 use std::fs;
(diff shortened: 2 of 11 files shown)


ohdearquant:
Replaced by #2094, which rebuilds this change directly on the current mainline (the prerequisite branch this PR was based on was closed after its content reached main separately).

PR #2007, opened 2026-08-16
feat(runtime): state the id contract once on every uuid parameter

ohdearquant:
`describe_verb` appends a single shared `ID_PARAM_CONTRACT` to every parameter declared `param_type=uuid`, instead of the contract being restated per parameter across each pack's `HandlerDef` or left implicit.

The contract it states: a full UUID is namespace-agnostic (the caller already knows the specific record), while a short hex prefix is a namespace-scoped resolution — a search — that can match nothing or match ambiguously across a namespace the caller does not share. Saying this once explains why some verbs accept a prefix and others require a full UUID: the asymmetry tracks whether the operation can assume the caller's namespace, rather than being an inconsistency between verbs.

Entity creation additionally echoes `entity_type_normalized` when an alias substitution was applied, so a caller can see both what it requested and what was actually stored. Case-only folding (case/hyphen/space folding to the same word) is deliberately not reported as a substitution.

Coverage: the shared contract on uuid params, single-entity and bulk create alias echo, and the negative case-folding arm.

Local gates at this head: `cargo fmt --check` clean, `cargo check -p khive-pack-kg -p khive-runtime --all-targets` clean, `cargo clippy` same scope clean, `cargo test -p khive-pack-kg -p khive-runtime` 1766 passed / 0 failed.


ohdearquant (commented):

Verdict on head 8c24c5d97f9891e8b976c946b32c19f67fc015d7: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):

PR #2014, opened 2026-08-17
feat(storage): implement attachment cutover (ADR-160 Phase 4b)

ohdearquant:
> AI-assisted: Codex implemented and prepared this PR; independent agents performed storage, GC/boot, moodboard-evidence, contract, and documentation review.

## Summary

- Implement ADR-160 Phase 4b's coordinated V21 attachment cutover: durable Pending/Incomplete/Complete state, resumable verified backfill, and atomic final removal of `entities.content_ref`.
- Add the first-class attachment capability and storage/runtime seams, with entity/note hard-delete cleanup and main-backend-only attachment authority.
- Migrate moodboard visual/model publication and reuse to attachment role `content`, plus authenticated model role `fann-network`; legacy backfill verifies the bundle, immutable event, FANN digest/bytes, and includes soft-deleted models even when the moodboard pack is disabled.
- Add the async host coordinator and config-aware schema-admin topology planner: distinct secondaries are inventoried/finalized before main, serving remains closed while cutover is incomplete, and exact-current admin paths do not resolve an unused BlobStore.
- Switch transactional GC to completed-V21 attachment liveness while preserving Phase 4a's early epoch gate, evidence validation, functional fence probe, bounded claim batches, and DB-owner → root-lock ordering.
- Align ADRs, public API docs, runtime/MCP/kkernel guidance, multi-backend operations, and migration runbooks.

The database schema, legacy evidence hydration, attachment publication, GC fences, and final column drop form one crash-resumable cutover. Splitting those pieces into independently mergeable intermediate states would create a dual-source or unprotected-liveness window, so this PR is the smallest coherent Phase 4b unit.

## Deployment gate — required

**Do not deploy Phase 4b until #2013 (Phase 4a) has merged and reached fleet convergence.**

Before starting this cutover:

1. upgrade every process that can invoke transactional blob GC to Phase 4a or newer;
2. drain every pre-Phase4a process and confirm no old sweep remains in flight;
3. quiesce Phase4a application readers/writers for the cutover;
4. only then start a Phase4b fleet.

The shared GC owner makes new processes wait; it cannot restore a legacy FANN blob that an older V20 sweep already deleted.

## Test plan

Post-stack verification:

- [x] `RUSTC_WRAPPER= cargo test -p khive-db -j4 stores::blob::tests::` — 63 passed
- [x] `RUSTC_WRAPPER= cargo clippy -p khive-storage -p khive-db --all-targets -j4 -- -D warnings`
- [x] `cargo fmt --all -- --check`
[...]

PR #2015, opened 2026-08-17
feat(retrieval): add bounded ranked-prefix materialization

ohdearquant:
## Stack and governance gate

- Stacked on #2014 (`codex/adr-160-phase-4-attachment-cutover`).
- Implements ADR-160 Phase 5 only.
- ADR-160 was accepted on `main` in `135c3d46` (#2005). Keep this PR draft until the prerequisite Phase 4a/4b operational rollout gates are satisfied and Claude completes review.

## What changed

- Add a root-exported, runtime-independent `khive-retrieval` ranked-prefix materialization controller with bounded inputs, batches, drop taxonomy, counters, and diagnostics.
- Preflight limits, taxonomy, uniqueness, and strict ordering before any callback or loader I/O.
- Validate each returned loader batch structurally before classification, preserve caller errors, stop classification at `K`, and validate the remaining tail without loader/classifier I/O.
- Replace moodboard's pack-local namespace merge with `khive_fusion::union_fusion` and migrate candidate materialization to the shared controller with `loader_batch_size = 1`.
- Preserve moodboard's existing score/self/entity/scope/kind/subtype/content-ref/blob-existence precedence, one-shot overfetch, compact ranks, and honest underfill.
- Document the opaque caller-owned score boundary and publish-order dependency.

## Regression guards

- Controller tests pin exact ceilings, invalid taxonomies, duplicate/non-monotonic input, malformed loader rows and deterministic error precedence, non-Clone scores, zero-output behavior, `K`/tail behavior, and bounded diagnostics.
- Moodboard tests pin canonical union ties and global truncation, self-hit zero I/O, post-`K` validation-only behavior, earlier loader-error precedence, typed drop accounting, honest underfill, soft-deleted entities, and malformed compatibility-projected refs.
- TDD RED was an intentional compile failure on the missing shared/moodboard seams; GREEN is the implementation in this PR.

## Validation

- `RUSTC_WRAPPER= cargo check --workspace -j4`
- `RUSTC_WRAPPER= cargo clippy --workspace --all-targets -- -D warnings`
- `RUSTC_WRAPPER= cargo test --workspace -j4`
- `deno fmt --check` on all six modified Markdown files
- `bash -n scripts/publish.sh`
- `git diff --check`

Independent contract, controller, resource, and final-diff reviews found no implementation Blocker/High/Medium issue. The remaining blocker is intentionally the spec/rollout gate above.

No Phase 6+ work is included.


ohdearquant:

PR #2019, opened 2026-08-17
feat(storage): stage dormant embedding-space registry shadow

ohdearquant:
## Stack

- Depends on #2017.
- Implements the independently safe dormant preparation slice of accepted ADR-160 Phase 7.

## Scope

- Adds canonical V22 migration `embedding_space_shadow_stage` with private shadow, legacy-provenance, and cutover-state tables.
- Stages deterministic legacy embedding-space identities while preserving the canonical `_embedding_models` registry as the only serving authority.
- Bounds row count, per-field bytes, and aggregate staging payload before materialization; reuses prepared inserts inside one atomic migration transaction.
- Revalidates the complete V21 attachment/GC fence before V22 staging and on exact-current startup.
- Advances V20/V21 attachment coordination to V22 without holding the DB-GC owner across blocking migration work.
- Lets destructive GC admit only canonical known-safe V21 or V22 ledgers; foreign V22 names and V23+ fail closed.

## Explicit non-goals

- No live provider registry replacement.
- No vector-table, ANN, cache, segment, watermark, pending-log, or daemon identity cutover.
- No activation of shadow rows. Existing provider and retrieval paths remain byte-for-byte authoritative.

## Verification

- `RUSTC_WRAPPER= cargo test -p khive-db -j4 v22_`: 16 passed.
- `RUSTC_WRAPPER= cargo clippy -p khive-storage -p khive-db -p khive-runtime -p khive-mcp --all-targets --all-features -j4 -- -D warnings`: passed.
- `RUSTC_WRAPPER= cargo check --workspace --all-targets -j4`: passed, including kkernel.
- `cargo fmt --all -- --check`, Deno formatting, staged diff, and conflict-marker checks: passed.
- Independent schema and resource-safety reviews: PASS, no Blocker/High/Medium findings.

## Follow-up gates before Phase 7 activation

1. Lattice must expose a prepared service sealed to a complete attestation of effective artifacts, transforms, execution backend, and cache behavior.
2. ADR lineage mapping must resolve legacy `_embedding_models.engine_name` versus the operator `[[engines]].name` slot.
3. Phase 7b must rebuild fresh vector/ANN/cache state and publish it atomically; it must not adopt legacy operational state by label.

This stays draft and stacked until #2017 and the prerequisite gates are resolved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Core migration V22, `embedding_space_shadow_stage`, adds dormant
+  embedding-registry shadow, exact legacy-tuple provenance, and cutover-state
+  tables without changing the live registry or vector-serving paths (ADR-160
+  D6 Phase 7a).
 - `khive_storage::EmbeddingSpaceIdentity`, a validated immutable physical-vector
   fence derived from a protocol-owned fingerprint and dimensions (ADR-160 D6).
 - ADR-149 Moodboard pairwise preference learning: actor-attributed randomized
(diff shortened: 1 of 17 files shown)


ohdearquant:

PR #2021, opened 2026-08-17
fix(db): count writer-task BEGIN refusals

ohdearquant:
## Problem

`writer_task.rs` classifies a failed `BEGIN IMMEDIATE` into a retryable busy
refusal (`StorageError::WriterTaskBusy`) or a non-retryable error, and returns
that classification to the caller. Neither outcome is recorded anywhere.

The writer contention diagnostics surface (`db_diagnostics`) therefore carries
no writer-side BEGIN failure count. A run in which callers are receiving
retryable refusals is indistinguishable on that surface from a run in which
nothing failed, because the absent counter and a zero counter read the same.
The doc comment on `WriterContentionDiagnostics` previously described this gap
as deliberate on the grounds that a shared counter would mislabel the failure;
that reasoning argues against reusing an existing counter, not against having
one.

## Change

- `WriterAcquisitionCounters` / `WriterAcquisitionSnapshot` gain
  `writer_task_begin_busy` and `writer_task_begin_errors`.
- The single production call site increments from the classifier's **own
  output** rather than re-deriving the busy condition, so the counter and the
  error the caller receives cannot disagree about which of the two occurred.
- `WriterContentionDiagnostics` exposes both fields; the doc comment is
  rewritten to state what the counters mean.

`writer_task_acquisition_timeouts` is untouched and keeps its current meaning.
The new counters are additive: no existing field changes value or name.

## Test

`contended_begin_increments_its_own_failure_counter` creates real lock
contention by holding a `BEGIN IMMEDIATE` from a separate writer while the
writer task attempts its own, then asserts:

1. `writer_task_begin_busy` goes 0 → 1 and `writer_task_begin_errors` stays 0
   (the classifier's two branches are distinguished, not merged);
2. `writer_task_acquisition_timeouts` is unchanged across the same window, so
   a later change cannot quietly fold the two failure modes together;
3. a subsequent **succeeding** acquisition leaves the failure counter at 1
   while `writer_task_acquisitions` advances, so the counter is shown to count
   failures rather than attempts.

The assertions were verified load-bearing by reverting the increment: the test
fails, and restoring it returns to green with no residue.

## Gates

`cargo fmt --all --check`, `cargo clippy -p khive-db --all-targets -D warnings`,
and `cargo check --workspace --all-targets` (119 crates, 0 warnings) all pass.

### DIFF
--- a/crates/khive-db/src/diagnostics.rs
+++ b/crates/khive-db/src/diagnostics.rs
@@ -439,7 +439,11 @@ pub fn wal_pin_attribution(_db_path: &Path, _sweep_interval: Duration) -> WalPin
 /// classes below. `writer_acquisition_timeouts` remains specific to the
 /// finite-wait pool-mutex stage; standalone SQLite failures and writer-task
 /// `BEGIN` failures have different ADR-135 F6 stages and are not mislabeled as
-/// pool checkout timeouts. `audit_append_failures` is supplied by the runtime
+/// pool checkout timeouts. Those stages now carry their OWN failure counters
+/// (`writer_task_begin_busy`, `writer_task_begin_errors`) rather than being
+/// absent: refusing to mislabel a failure is not a reason to omit it, and an
+/// omitted failure counter fails toward looking healthy, which is the reading
+/// an operator believes. `audit_append_failures` is supplied by the runtime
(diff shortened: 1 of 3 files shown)


ohdearquant (commented):

PR #2041, opened 2026-08-18
fix(daemon): forward resolved pack selection to auto-spawned daemon

ohdearquant:
## Summary

- The daemon auto-spawn path (`kkernel mcp` / `kkernel exec` spawning `kkernel mcp --daemon` when no warm daemon is reachable) only forwarded an explicit `--config`/`--db` override to the freshly spawned child, relying on ambient OS environment inheritance for everything else — including pack selection.
- That works when `KHIVE_PACKS` happens to be a plain OS environment variable, but a client whose pack selection came from an explicit `--pack` CLI flag, or from a discovered `[runtime].packs` entry in a project/home `khive.toml`, has nothing for the child to inherit.
- Result: the spawned daemon silently falls back to the built-in default pack set. Its config fingerprint then disagrees with the client that spawned it, so every subsequent request from that client falls back to local, in-process dispatch instead of the warm daemon — multiplying writer contention on the shared SQLite store.

## Root cause

- `crates/khive-mcp/src/server.rs` — the primary auto-spawn site hit by every interactive MCP stdio client — called the daemon-forward seam with no pack information at all.
- `crates/kkernel/src/exec.rs` — the `kkernel exec` daemon-forward seam — already forwarded `--config`/`--db` in narrow cases, but never forwarded packs either.
- `Command::new(exe)` inherits the parent's full environment by default (no `env_clear`/`env_remove` on this path), so `KHIVE_PACKS`-as-env-var cases mostly worked by accident; CLI-flag and config-file-derived selections did not.

## Fix shape

Forward the spawning client's already-resolved pack list explicitly as `--pack <name>` arguments on the spawned child, using the existing highest-precedence CLI-flag tier the config resolver already defines (`RuntimeConfigInputs::packs`) — no new precedence tier is introduced, and the daemon's own boot-time resolution is otherwise untouched.

- `crates/khive-mcp/src/daemon.rs`: `spawn_daemon_with_exe_and_config` and `forward_or_spawn_with_config` gained an `packs: Option<&[String]>` parameter, appended as `--pack` args when present. The bare `spawn_daemon`/`forward_or_spawn` helpers keep their old (no-packs) behavior unchanged.
- `crates/khive-mcp/src/server.rs`: the primary MCP stdio auto-spawn call site now forwards the loaded verb registry's own pack list (`self.registry.pack_names()`).
- `crates/kkernel/src/exec.rs`: the `kkernel exec` daemon-forward seam now forwards the client's resolved `cfg.packs` unconditionally, alongside its pre-existing conditional `--config`/`--db` forwarding.

## Test plan

All commands run from `crates/`:

- `cargo fmt --all -- --check` → exit 0
- `cargo clippy -p kkernel -p khive-runtime --all-targets -- -D warnings` → exit 0
- `cargo test -p kkernel` → all test binaries green, including two new regression tests:
  - `env_khive_packs_reaches_daemon_spawn_seam` — a client with `KHIVE_PACKS` set forwards exactly that resolved list to the daemon spawn seam.
  - `no_env_control_forwards_built_in_default_packs_to_spawn_seam` — control: with no pack-selection input, the client forwards exactly the built-in default set (proving the fix is additive and the unconfigured default path's effective outcome is unchanged).
- `cargo test -p khive-mcp` → 142 passed, including new/updated `daemon::tests::spawn_daemon_with_exe_and_config_*` cases proving `--pack` flags reach the spawned child's actual command line (and are omitted when no packs are given).
- From repo root: `git diff --check origin/main...HEAD` → exit 0


ohdearquant (commented):

Verdict on head c0d623128dd2437ad3826229304ddd5091b4e789: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2044, opened 2026-08-18
fix(runtime): mask gate-error text before log emission

ohdearquant:
## Summary
Supersedes #2001, carrying its full change set (fail-closed gate-error handling with stable classified wire reasons) plus one additional hardening: gate-error text is now bounded and masked before any warn-level log emission.

## Root cause
Gate backend errors can embed connection strings or credentials in their Display text. #2001 sanitized the wire path, but the new handlers logged the raw text — and log records are shipped and aggregated, so the log surface has the same disclosure requirements as the wire.

## Fix shape
- New `secret_gate::bounded_masked_log_text`: truncate, mask with the canonical detector set, truncate again; any truncation appends an ellipsis so the record declares its own incompleteness.
- Applied at all four gate-error log-emission sites: `gate_unavailable_error`, `authorize`, `authorize_with_visibility`, and `authorize_namespace` (the last now also returns the stable classified reason like its sibling authorize paths, since its only consumer logs the returned error).
- The cross-backend regression now asserts the canary connection string is absent from server logs with the masked marker present. Helper unit tests cover the connection-string and pathological-length arms.

## Test plan
- `cargo fmt --all -- --check`: exit 0
- `cargo clippy -p khive-runtime -p kkernel -p khive-mcp --all-targets -- -D warnings`: exit 0
- `cargo test -p khive-runtime` / `-p kkernel` / `-p khive-mcp`: all suites pass
- Mutation check: reverting the masked call at the authorize log site reddens the cross-backend regression on its log assertion; restoring re-greens it. Fresh compiles were observed on both runs.


ohdearquant (commented):

Verdict on head 4291f999e1a7730e6e47122428869648b44482ba: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Superseded by #2050.

PR #1837, opened 2026-08-09
fix(vcs): preserve imported edge metadata

ohdearquant:
## Summary

- classify generic JSON edges only from canonical `source`+`target`, preserving `from`/`to` entity metadata and rejecting complete dual signatures as ambiguous
- reject whitespace-only names/endpoints and malformed present timestamps instead of silently importing or replacing them with import time
- preserve edge properties and independent creation/update provenance across adapter, runtime, VCS cache, snapshot hashing, and three-way merge paths
- finish deterministic validation before the first VCS/runtime target write

## Integrity contract

The JSON adapter now treats `kind`+`name` as the complete entity signature and canonical `source`+`target` as the complete edge signature. A record carrying both is fatal at eager construction; `from`/`to` alone remain ordinary entity properties. Required strings are nonblank and present `created_at`/`updated_at` values must be RFC 3339 strings.

The tracked-NDJSON sync validate-first gate rejects blank names before creating its temporary database. `KhiveRuntime::import_kg` validates every entity kind/name and every edge weight before opening the entity store or issuing the first write; mutable endpoint existence and namespace checks remain deferred to the target-state phase. Kkernel's compatibility conversion defaults only absent timestamps and carries malformed-field context.

`ExportedEdge` now retains properties plus independent `created_at` and `updated_at` values throughout local and remote VCS flows. Snapshot hashes recursively canonicalize edge properties, so key order is irrelevant but metadata changes are visible. Three-way merge classifies governed edge identity, weight, and properties against the base: independent property edits compose, same-key and identity divergence return typed conflicts, one-sided edits survive intact, and timestamp-only drift does not manufacture a content conflict.

This completes #1755 and is the KG/VCS adapter-validation tranche of #1758. The findings-identity and tombstone-preservation requirements remain open and will land separately; this PR must not close the umbrella issue.

## Verification

- independent exact-head review: READY at `88eca222bb18e42dfc4c460e443d7f664ae0c313` (0 findings after the Clippy-only repair)
- exact-head full CI: [run 31307054387](https://github.com/ohdearquant/khive/actions/runs/31307054387) — passed
- direct Rust formatting and Deno documentation formatting checks
- ADR-reference, JSON-data, SQL, stub-marker, manifest/lock, static-contract, and diff checks
- full Deno suite: 507 passed, 0 failed, 1 ignored
- focused ambiguous-shape, metadata-discriminator, blank-name/endpoint, malformed-timestamp, validate-before-write, snapshot-hash, and three-way-merge regressions

[...]

PR #2070, opened 2026-08-19
fix(runtime): preserve imported edge metadata through the portability core

ohdearquant:
## Summary

Rust-core successor to #1837 (the TypeScript CLI importer slice follows separately). Carries the portability core and the semantic edge merge together because CI compiles the workspace-excluded merge crate against the extended `ExportedEdge` shape.

- `ExportedEdge` gains portable `properties` and independent `created_at`/`updated_at` provenance fields in the ADR-020 wire shape. Legacy archives without timestamps deserialize with import-time defaults; a present value is never replaced.
- Export, import, archive encode/decode, local sync, remote hash, and the JSON array adapter carry the metadata losslessly; edge properties participate in snapshot identity after recursive key canonicalization.
- Deterministic import validation (entity kind, blank name, reserved secret-gate property key, edge weight) completes before the first write, so a rejected archive can no longer partially import.
- JSON adapter edge detection now requires the complete canonical `source`+`target` signature; records ambiguous between entity and edge shapes are rejected (ADR-036).

## Testing

- `cargo check --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`
- `cargo test`: khive-runtime portability (18), khive-vcs (81), khive-vcs-adapters (48), kkernel archive (21) — all passing
- Semantic edge merge carries `properties` and provenance timestamps across added, modified, and reconciled edges; chained identity collisions are recorded and dropped rather than silently duplicating an earlier edge's durable id, with the regression test pinning exact surviving edges and conflict fields (khive-merge: 98 tests, clippy and doc build clean).

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2570,6 +2570,7 @@ dependencies = [
 name = "khive-vcs-adapters"
 version = "0.7.0"
 dependencies = [
+ "chrono",
  "khive-types",
  "serde",
  "serde_json",
--- a/crates/khive-merge/README.md
+++ b/crates/khive-merge/README.md
@@ -44,9 +44,9 @@ going through the `MergeEngine` trait object.
 `MergeConflict` enumerates what `Auto` can detect: `NameConflict`, `KindConflict`,
 `PropertyMismatch`, `ModifyDelete` (one branch edited, the other deleted),
 `DuplicateAddition` (both branches added the same UUID with different content),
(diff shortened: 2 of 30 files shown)


ohdearquant:
Split into #2083 (portability core and wire shape) and #2084 (semantic three-way edge merge, stacked on #2083) to keep review scope bounded. All changes from this branch are carried across the two.

Issue #2085, opened 2026-08-20
kg: list(kind="edge", target_id=…) cannot match an edge whose target is another edge — silent empty result, 4,349 affected


ohdearquant:
## Summary

`kg.link` accepts an edge whose **target is another edge** — this is how a judgment edge gets its
annotating note (`annotates`, note → edge). But `kg.list(kind="edge", target_id=<edge_uuid>)`
cannot match those edges. It returns an empty result set with no error and no warning.

So the relationship is writable but not queryable through the natural filter, and the failure is
silent and biased toward **"this edge does not exist"**.

## Measured

Against a production store (91,229 edges):

| | count |
|---|---|
| edges whose target is another edge | **4,349** (100% `annotates`) |
| edges whose target is an entity | 60,391 |

All 4,349 are unreachable via `list(kind="edge", target_id=…)`.

## Reproduction

```
# 1. create an edge, note its id
link(source_id=<A>, target_id=<B>, relation="supports", weight=0.8)   # -> edge_id E

# 2. annotate that edge — this SUCCEEDS
link(source_id=<note>, target_id=E, relation="annotates", weight=1.0) # -> ok

# 3. read it back with the natural filter — returns EMPTY, no error
list(kind="edge", source_id=<note>, target_id=E)                      # -> []

# 4. the data is present; these both find it
list(kind="edge", source_id=<note>)                                   # -> the annotates edge
neighbors(node_id=<note>, direction="both")                           # -> the annotates edge
```

## Why this is a data-integrity issue rather than an inconvenience

The natural way to verify a write is to read it back with both endpoints. That check returns
"absent" for an edge that is present, so the caller's next action is to **create it again**. The
failure direction produces duplicates in exactly the workflow that was being careful.

Measured consequence: an acceptance pass over 10 freshly-written edges reported **0 of 10 present**
when all 10 were present. The pass was only caught because it also ran a must-MATCH control on a
known-good edge of a different shape; without that control the natural reading is "the writes
failed", and the natural response is to redo them.

## Note on the current help text

`list`'s parameter help reads *"Filter edges by target node complete UUID"* — so node-only is
arguably the documented behaviour. The defect is the **asymmetry**: `link` creates these edges
happily and `list` cannot see them, with nothing at either call site saying so.

## Possible resolutions (not a preference, just the space)

1. `target_id` resolves an edge id as well as a node id when `kind="edge"`.
2. A separate `target_edge_id` filter, leaving `target_id` node-only and unambiguous.
[...]

PR #2086, opened 2026-08-20
docs(adr): ADR-167 service provenance, classification criteria, kind migration

ohdearquant:
Proposes three related changes to the entity/edge ontology, each motivated by something measured against a production store rather than inferred.

**1. One derivation endpoint pair: `Service introduced_by Document`.**
No pair currently admits a `Service` into a derivation or provenance relation with a `Document`, so a service cannot record the specification or ADR that introduced it. Every sibling kind can. The gap looks incidental rather than deliberate: the 2026-07-08 `introduced_by` amendment was reasoning about documents and orgs and did not consider services. The ADR explicitly rejects the wider forms (`Service introduced_by Person`/`Org`, any `derived_from` involving a service) because no observed refusal asked for them. The relation set stays at 17.

**2. Written service-vs-concept classification criteria.**
The two kinds have materially different endpoint contracts, so the choice at creation time silently decides what edges a record may ever carry, and there is no ADR stating what makes something one rather than the other. Proposed as two tests rather than a category list, because the observed failure is a borderline record being classified on its name: a liveness test (is it meaningful to say the thing is currently down) and an independence test (would it exist if every deployment were removed). Where they disagree there are usually two records, joined by `Service instance_of Concept`, which the contract already admits.

**3. Kind migration: a written procedure, not a mechanism.**
Kind is immutable, so a misclassification is correctable only by delete-and-recreate, which destroys edges. Measured over the 19-service population: 16 carry at least one edge, and a whole-population migration would destroy 62 edges, 42 of them `annotates`. Because `annotates` runs note to entity, each destroyed one leaves a note whose subject no longer exists.

The ADR proposes **not** building a kind-migration mechanism. Such a mechanism would have to re-point both directions, preserve edge ids, and carry annotations across the substrate boundary, which is substantial runtime surface for a rare operation on a small population. It proposes a five-step re-anchor procedure instead, and states the trigger that would reopen the mechanism decision as a countable thing: the number of times the procedure gets run.

Step 3 of that procedure is worth flagging for review. Re-creating an edge yields a new edge id, so notes annotating the record's *edges* are orphaned even when the record's own annotations were handled correctly, and that population is exactly what #2085 makes invisible: the target-side filter returns an empty result rather than an error. The procedure enumerates it from the source side.

Related: #2085.

### DIFF
--- a/crates/khive-runtime/src/operations.rs
+++ b/crates/khive-runtime/src/operations.rs
@@ -747,6 +747,9 @@ pub const BASE_ENTITY_ENDPOINT_RULES: &[(&str, EdgeRelation, &str)] = &[
     ("concept", EdgeRelation::IntroducedBy, "document"),
     ("concept", EdgeRelation::IntroducedBy, "person"),
     ("artifact", EdgeRelation::IntroducedBy, "document"),
+    // ADR-002 amendment (ADR-167): service provenance — the document that
+    // introduced a service (its ADR or design record).
+    ("service", EdgeRelation::IntroducedBy, "document"),
     ("document", EdgeRelation::IntroducedBy, "person"),
     ("document", EdgeRelation::IntroducedBy, "org"),
     ("concept", EdgeRelation::IntroducedBy, "org"),
--- a/docs/adr/ADR-001-entity-kind-taxonomy.md
+++ b/docs/adr/ADR-001-entity-kind-taxonomy.md
@@ -3,6 +3,11 @@
 **Status**: accepted\
 **Date**: 2026-05-22\
 **Authors**: khive maintainers
+**Amended 2026-08-21 ([ADR-167](ADR-167-service-provenance-and-kind-classification.md))**:
+step 5 of the classification decision tree gains a record-observable service/concept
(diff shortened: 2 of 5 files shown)


ohdearquant (commented):

Verdict on head 99c2dc36c14c2ea47503098f87d816fc6e97d615: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head b05c8e158997fea171b621eaf2c6410df99e4115: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Superseded by #2109, which lands the same ADR-167 change as a single commit covering every surface that states the endpoint table (the ADR-002 contract table, the runtime rule row, and the khive-pack-kg help mirror whose omission broke this PR's head). Closing per the supersession; #2109 carries the full gate evidence.

PR #1975, opened 2026-08-15
feat(kg-editor): copy bounded investigation evidence briefs

ohdearquant:
## Summary

- add a pure deterministic Markdown evidence-brief model with a 49,152-character hard bound on the final escaped output, robust variable-backtick code spans, and an exact optional-block omission disclosure
- export selected-module topology/SCC/history/ownership evidence plus the URL-focused hidden-coupling candidate, while refusing any bundle whose captured module revisions do not all match the recorded HEAD
- preserve source honesty (materialized khive DB snapshot vs curated static fallback), full SHA/time/exporter, canonical current URL, applicable aggregate windows, raw coverage/truncation disclosures, and source-role caveats
- add a module-inspector **Copy evidence brief** action with visible clipboard/model failure and stale-async completion suppression

The brief never calls captured data live. Retained graph history is exported only while Structure Graph is the active canonical view. It only states that a direct dependency is absent when the structure-edge page is complete; otherwise absence stays unknown. Candidate interpretations remain distinct from observed rows and every brief ends with fixed verify / confirm-or-refute guidance.

## Stack

- base: #1973 (`a5713d5e9862e0d2fea6afb3b442a7efd6742861`)
- this PR: `4245e3cd8b16a224271085029f22fbc116a7f300`

## Validation

- TDD RED/GREEN for the model, inspector action, source-revision guards (selected, focused endpoint, and non-selected SCC member), off-view retained pair, final escaped cap, and omission/footer reservation
- focused Vitest: 56 passed across model, controller, clipboard/race, and focused-pair integration
- focused Playwright Chromium: 2 passed (curated static fallback + materialized khive DB route, clipboard readback)
- `npm run typecheck`
- focused ESLint
- `npm run lint:tokens`
- `next build --webpack`
- pre-commit hooks and `git diff --check`

Closes #1974
Refs #1881
Refs #1887

PR #2095, opened 2026-08-21
feat(kg-editor): investigation briefs with source-schema allowlist contracts

ohdearquant:
Replaces #1975. The predecessor bounded bundle-derived text with Markdown
escaping and truncation at the point of rendering; a rendering control is
not an instruction/data boundary, so every dynamic source that reaches the
model-facing brief is now constrained at the source schema instead:

- exporter identity and SCC cycle ids: bounded identifier tokens (length
  cap, closed charset), rejecting at parse
- module language and page ordering: closed enums derived from the values
  the exporter actually emits
- module paths: the same addressable-path contract source paths already had
- free-text reason fields: length and control-character bounds at the
  schema plus delimited rendering (documented as residual free text — the
  emitted values have no stable code vocabulary to enumerate)
- page cursors: bounded opaque-token contract
- the caller-supplied canonical URL: validated with the public-repository
  URL contract before rendering, with a bounded placeholder on failure

Regressions inject imperative sentences, Markdown structure, control
characters, and oversized payloads into each source and prove either the
parse rejects the bundle or no raw value reaches the emitted brief.


ohdearquant:

This pull request sits in a stack rooted on `codex/kg-editor-analysis-drilldown-1881`, which was never merged. The stack is being closed as a single abandoned line of work rather than reviewed and rejected one by one.

The root's work has since landed by another route: all 17 files that branch introduced relative to its merge base are present on `main` today. One is a byte-identical blob; where they differ, `main`'s copies have moved further. The root's merge base is from 18 August and rebasing it onto current `main` would now revert roughly 69,000 lines, so carrying the stack forward would cost more than it could return.

PR #1958, opened 2026-08-15
feat(kg-editor): connect analysis results to inspector

ohdearquant:
## Summary
- make module-bearing results in history, topology, hotspot, coupling, treemap, ownership, API surface, and scorecard open the canonical shared module inspector
- preserve the active analysis while updating the typed investigation URL and browser history through the existing single controller
- distinguish truncated, unavailable, and complete-page missing module references without inventing navigation
- add keyboard, focus, responsive, and Back/Forward coverage

Depends on #1956.
Refs #1881 (does not close the remaining graph/package/commit selection work).

## Validation
- focused Vitest: 54 passed
- typecheck and focused ESLint: passed
- showcase byte gate and git diff check: passed
- Next production webpack build: passed
- focused Chromium analysis-result drilldown / Back-Forward flow: passed
- desktop and 375px visual QA: passed; no horizontal overflow
- two independent reviews: APPROVE

Full Vitest on this parent: 177 passed, with one inherited KG Studio duplicate Clear filter failure from the stack root; no drilldown failure.

PR #2106, opened 2026-08-22
feat(kg-editor): connect analysis results to the module inspector

ohdearquant:
Supersedes #1958, transplanting the analysis-inspector feature directly onto main: the stacked branch had absorbed older trees whose conflict resolutions dropped behaviors main has since gained, so this branch starts from main and carries only the feature.

Aggregate analysis results connect to the module inspector: fan-in, fan-out, and hidden-coupling entries render as inspection links that focus the inspector and update the module URL state; the command palette, treemap, and cycle views share the same inspection control; structure-graph subtree counts replace joined name strings. Preserved main-side behavior is pinned by tests, including bearer-token snapshot authorization with a non-retaining private cache, sanitized share links that never carry repository-supplied query or fragment data, page-metric reason precedence, cursor-aware history and series accounting, and layout reuse across metric-only changes.

Test suite: 24 files, 238 tests passing, with typecheck and lint clean.

### DIFF
--- a/apps/kg-editor/e2e/showcase.spec.ts
+++ b/apps/kg-editor/e2e/showcase.spec.ts
@@ -94,7 +94,8 @@ test("restores a shared investigation and follows browser back and forward", asy
   );
 
   await page.getByRole("searchbox", { name: "Find a module or path" }).fill(writer);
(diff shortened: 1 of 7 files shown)


ohdearquant (commented):

PR #2110, opened 2026-08-22
fix(blob): byte-safe GC canonicality checks and broadened fence probe

ohdearquant:
## Problem

Two hardening gaps in the blob GC transactional sweep path (reachable once the V21 attachment schema lands; unreachable on databases at V20):

1. **NUL-blind canonicality validation.** SQLite `length()` counts characters before the first embedded NUL and `GLOB` stops scanning there, so a value of 64 hex characters followed by a NUL and arbitrary bytes passes both the `length(content_ref) = 64` and hex-only GLOB checks while full-byte exact-equality anti-joins cannot match it against a clean digest. A referencing row carrying such a value would not be recognized as live.
2. **Narrow fence probe.** The attachment fence probe exercised only the fixed all-zero sentinel digest under entity/content shape, so a same-named trigger rewritten to fence only that digest or that shape passed the probe while other claimed digests or attachment shapes remained writable during deletion.

## Fix

- Both evidence probes in `validate_blob_gc_evidence` gain a `length(CAST(content_ref AS BLOB)) <> 64` arm: characters = 64 and bytes = 64 together force NUL-free canonical values.
- `blob_gc_fence_probe` now exercises four arms: the existing all-zero-sentinel INSERT/UPDATE arms plus a per-run random claimed digest under note/evidence shape on both INSERT and UPDATE paths. A restricted trigger rewrite now fails the probe and the sweep refuses with a typed error.

## Tests

- NUL-embedded claim ref refused by the validator (and the test proves the value passes the schema-level CHECK first, so the validator is the load-bearing backstop).
- NUL-embedded attachment ref refused.
- Digest-restricted trigger rewrite fails the probe (healthy fence probed first as positive control).
- Shape-restricted trigger rewrite fails the probe.
- Existing fence-probe collision/retention tests updated for the four-arm probe.

`cargo test -p khive-db --locked` and `cargo clippy -p khive-db --all-targets --locked -- -D warnings` pass.

## Sequencing

This must merge before the attachment-cutover work that registers migration 021, which is what makes the probed path reachable on real databases.

### DIFF
--- a/crates/khive-db/src/stores/blob.rs
+++ b/crates/khive-db/src/stores/blob.rs
@@ -1254,18 +1254,37 @@ async fn validate_blob_gc_evidence(sql: &dyn SqlAccess) -> StorageResult<()> {
     // length() and GLOB both stop at an embedded NUL, so a value of 64 hex
     // characters followed by a NUL and arbitrary bytes passes them while
     // failing the exact-equality liveness anti-join. The byte-length arm
-    // closes that class: chars = 64 AND bytes = 64 forces a NUL-free,
-    // all-single-byte value.
+    // closes that class: chars = 64 AND bytes = 64 * the encoding's bytes
+    // per ASCII character forces a NUL-free canonical value. CAST(TEXT AS
+    // BLOB) yields the database text encoding's bytes (1 per hex char in
+    // UTF-8, 2 in UTF-16), so the width is derived from the same database
+    // rather than assumed, and an unrecognizable answer fails closed.
+    let canonical_bytes = match reader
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

Verdict on head 0a1c30fe92505d04c4f26cd76bf7055460df12ae: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2115, opened 2026-08-22
fix(kg-editor): bound catalog discovery and strip nested credentials from history writes

ohdearquant:
Three hardening fixes on the showcase catalog path:

- Catalog discovery now aborts after a named 5-second bound, with expiry landing in the existing degraded path; the abort signal covers both the fetch and the body read.
- Repository values are truncated at the first query or fragment delimiter before any history or clipboard write, at the single choke point every write site routes through; validation still sees the original value, so parse verdicts are unchanged. The contract comment now states the guarantee it delivers.
- When Content-Length is absent, the body is read through a bounded stream reader that cancels past the 256 KiB cap instead of materializing the full body first.

Audit note: every revision of the shipped showcase registry and bundled analysis data was checked at value granularity; no repository URL ever carried a query or fragment component, so the credential-strip change is purely preventive — no shipped data needs scrubbing.

### DIFF
--- a/apps/kg-editor/src/components/showcase/showcase.test.tsx
+++ b/apps/kg-editor/src/components/showcase/showcase.test.tsx
@@ -417,7 +417,7 @@ describe("materialized repository lookup", () => {
     const url = new URL(window.location.href);
     expect(url.searchParams.has("access_token")).toBe(false);
     expect(url.hash).toBe("");
-    expect([...url.searchParams.keys()]).toEqual(["repo"]);
+    expect([...url.searchParams.keys()]).toEqual([]);
   });
 
   it("treats an empty repo parameter as invalid instead of falling back to the default entry", async () => {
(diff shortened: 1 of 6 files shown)


ohdearquant:
Closing this in favour of a scoped replacement rather than continuing here.

The repository-URL handling on this branch is a hand-rolled canonicalizer hardened by rejecting
particular characters. That shape does not terminate: each fix closes one input class, and the
next class is then found by the same method. One further class was found after the most recent
fix, which is the evidence for the claim rather than an argument about taste.

The replacement will state the property directly instead of enumerating rejections. A canonical
value is re-parsed and its segments are required to equal the ones it was built from, and
normalization is required to be idempotent, so that `normalize(normalize(x))` equals
`normalize(x)`. The input classes closed on this branch carry over as fixtures on that invariant,
so the invariant is demonstrated to subsume them rather than asserted to.

The single-normalizer consolidation from this branch (one canonicalizer shared by the reader, the
bundle validator and the share writer, rather than three implementations of the same rules) carries
over as well.

PR #2116, opened 2026-08-22
docs(adr): scope ADR-167 to the two amendments it specifies

ohdearquant:
Follow-up to #2112.

ADR-167 as drafted decided three things: one derivation pair, a service/concept tie-break inside ADR-001's decision tree, and a kind-migration closure procedure. This change removes the third and leaves the ADR scoped to the two amendments it is named for.

**Why the procedure comes out rather than takes another correction.** A closure rule for delete-and-recreate has to be a total function over the product of five independent dimensions: the substrate of each incident record, the relation kind, the endpoint role the migrating record occupies in that relation, the liveness of the incident row, and the namespace it lives in relative to the migrating caller's visibility. The procedure stated points on that function rather than specifying it. Each correction closed a case and left the shape unchanged, and the distance between a set of points and the function is exactly where an edge is destroyed without appearing in any enumeration that preceded it.

**What stays.** The measured blast radius remains in the context as the evidence that keeps the question open: across the 19-service population a hard delete-and-recreate would destroy 62 edges, 42 of them `annotates`, each leaving a note whose subject no longer exists. So does the enumeration-scope finding. The purge is `DELETE FROM graph_edges WHERE source_id = ?1 OR target_id = ?1`, with no namespace predicate and no tombstone predicate, while every listing path is visibility-scoped and filters to live rows. A completeness check scoped narrower than the destructive statement it guards is not a completeness check.

**What replaces it.** A deferral that states the shape a correct procedure must have and requires a successor ADR specifying that function before any implementation depends on kind migration. Until such an ADR is accepted, kind stays immutable in practice and a misclassified record is corrected by delete-and-recreate at the operator's own risk, with the blast radius above as the standing estimate of what that costs. Nothing accepted here creates or implies a supported migration path.

The acceptance criteria for the removed decision go with it; Decisions 1 and 2 keep theirs unchanged.

Also corrects two self-references that described the ADR-001 amendment as living somewhere other than ADR-001, and updates the ADR index title to match.

### DIFF
--- a/docs/adr/ADR-167-service-provenance-and-kind-classification.md
+++ b/docs/adr/ADR-167-service-provenance-and-kind-classification.md
@@ -226,10 +226,17 @@ instead — each of which must be written down per edge:
   against a tombstoned row is not refused downstream — only this check stops it. The
   edge keeps its id while the endpoint fields referencing the moved subject are updated (the
   primitive takes one or both fields): for an incident edge, every endpoint field referencing the
-  migrating record moves onto the new record — one field ordinarily, both for a self-loop; for a
+  migrating record moves onto the new record; for a
   closure member, the target moves to its annotated subject's mapped replacement — a closure
   member is never moved onto the new record. A closure
   member whose subject is itself preserved needs no move at all: its target id is still valid.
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

PR #2118, opened 2026-08-22
fix(kg-editor): replace repository-URL character checks with a round-trip invariant

ohdearquant:
Replaces #2115, which was closed rather than continued.

## Why

`normalizeRepositoryUrl` decoded each path segment, validated the decoded segments, and then built
the canonical value by joining those decoded segments back together with `/` without re-encoding
them. Any character that is structural in a URL path therefore changed the meaning of the canonical
value relative to the segments that had been validated: the object that was checked and the object
that was emitted were not the same object.

The previous attempt addressed this by rejecting particular characters in decoded segments. That
shape does not terminate. Each fix closed one input class and the next class was then found by the
same method, so the approach is replaced here rather than extended.

## What changed

The function now states the property instead of enumerating rejections. It builds the canonical
value, re-parses it with the same parser, and requires the re-parsed path segments to equal, element
for element, the segments the value was built from. A structural character shifts where a segment
boundary falls on the next parse, so the round trip fails and the input is rejected. No character
list is involved, and the check does not need to be extended when a new such character is found.

Idempotence follows from the same property and is asserted directly: for any input that normalizes
successfully, normalizing the result again returns the same value.

## Tests

Six inputs that motivated the previous approach are included as named fixtures, so the invariant is
demonstrated to cover them rather than asserted to. Five are refused by the round-trip check
(a decoded path separator, and single, double and triple percent-encoded query introducers, and a
decoded backslash). The sixth, credentials in the authority, is refused by the pre-existing
username/password check rather than by the new invariant, and its test says so — it is included for
completeness of the closed set, not as evidence for the round trip.

There is also a property arm over 500 generated inputs from a seeded deterministic generator,
asserting that each input is either refused or normalizes idempotently. It asserts that both
branches are non-empty, so it cannot pass by refusing everything.

## Also in this change

The reader, the bundle validator and the share writer now route through the one normalizer instead
of carrying three implementations of the same rules:

- `publicRepositoryUrlIssue` judges by calling the normalizer rather than reimplementing it.
- `parseRepository` drops its own raw-length check and returns the canonical value.
- The length limit applies to the canonical value rather than the raw input.
- Two byte-identical origin-and-path helpers are merged into one shared function.

One existing test expectation moves from a raw value to the canonical value as a consequence.

`npm test` (284 passing), `tsc --noEmit`, and `npm run lint` all pass.

### DIFF
--- a/apps/kg-editor/README.md
+++ b/apps/kg-editor/README.md
@@ -114,6 +114,17 @@ both be unique; one invalid entry makes the entire catalog unavailable. The cata
 sorted by analysis ID and exposes only those two public fields. It does not scan the
 analysis root or read a report.
 
+A repository URL is normalized before it is compared or stored, and normalization is
+deliberately strict about what a path segment may contain. A segment must survive a
+re-parse of the canonical URL unchanged, which rejects any character that is structural
+in a URL path — an encoded `/`, `\`, `?` or `#` — however many layers of encoding it
+arrives under. As a consequence a **literal percent in a repository name is refused**,
+because the canonical form joins decoded segments without re-encoding them and a bare `%`
+is ambiguous under that join. No major forge permits `%` in a repository name, so this
+costs nothing in practice, and it keeps the rule statable: no literal percent. A URL that
+is validly encoded but does not normalize to a stable value — a doubled `.git` suffix, for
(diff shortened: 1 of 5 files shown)


ohdearquant (commented):

PR #2120, opened 2026-08-22
docs(adr-137): amendment 1 — close the wire contract before the first consumer

ohdearquant:
## Why now

`khive-wire-protocol` implements framing, envelope grammar, and handshake sequencing decisions
that ADR-137 left unstated or stated only in prose. Each one is an interoperability commitment: a
second implementation written from the ADR alone has no way to reproduce it, so it would diverge
silently rather than fail loudly.

The crate has no in-tree consumer yet. That is the whole reason this lands now — every behaviour
named here can still be corrected for free. Once a consumer merges, each correction becomes a
compatibility break instead.

## What the amendment does

It promotes each decision to specified behaviour, marking those where the implemented behaviour is
correct as **RATIFIED**, and those where the contract differs from the code as **CHANGED**, with
the crate required to move.

Ratified: initial version and floor; frame ceiling and payload-only accounting; non-empty
operation ids; handshake sequence violations as connection-terminal; topic and timestamp
validation as a producer obligation rather than a decoder one.

Changed: envelope duplicate-member rejection must run before field validation, otherwise
acceptance depends on member order; an unknown error code arriving with no operation id is a
malformed frame rather than a request-terminal error, because there is no request to terminate;
received opaque subvalues must be preserved byte-exactly on relay rather than semantically.

## The frame-shape divergence

While confirming the others, one more surfaced and is resolved here rather than deferred. ADR-137
specifies `event.payload` as "a topic-specific JSON object"; the crate types it as an arbitrary
JSON value, so a scalar or array payload is accepted on the wire while an implementation written
from the ADR would reject it. The object requirement is retained: "field-by-field" presupposes
fields, and an object is the only shape a topic catalog can extend compatibly.

An amendment claiming to close the wire contract while leaving a known frame-shape divergence open
would misrepresent its own completeness, which is why it is in scope rather than filed separately.

## The fence

No first consumer of the crate may merge while any decision marked CHANGED still implements the
superseded behaviour. The amendment names those decisions explicitly so the condition is checkable
rather than a matter of interpretation.

## Verification

By a published vector matrix, not by inspection: positive and negative vectors for every rule,
run through the Rust crate and at least one independent JSON implementation, which must agree on
accept/reject classification, on whether a rejection is request-terminal or connection-terminal,
on consumed frame length, and on byte-exact opaque-subvalue round trip.

## Scope

Documentation only. No code changes in this PR. It authorizes no transport server, identity
mapping store, or deadline executor; those remain separately tracked.

### DIFF
--- a/docs/adr/ADR-137-tailnet-wire-transport.md
+++ b/docs/adr/ADR-137-tailnet-wire-transport.md
@@ -196,13 +196,22 @@ does today and the crate must move to match.
    outside the receiver's supported range is rejected rather than negotiated downward.
 
    **The rejection layer is stated normatively because two layers could plausibly own it.** A
-   version field carrying any non-negative integer, `0` included, is syntactically well-formed and
-   MUST decode: the decoder's job is grammar, and a receiver that rejected `0` as
-   `malformed_frame` would close the connection on a frame it could have answered. Admission is
-   where the range is enforced: a handshake naming a version outside the receiver's supported
-   range, `0` and `99` alike, is rejected at handshake admission with `unsupported_version`.
-   Version `0` is therefore not special-cased on the receive path; it is out of range like any
-   other unsupported value, and the floor of `1` is what makes it out of range.
+   version field carrying any non-negative integer **representable as an unsigned 32-bit value**,
+   `0` included, is syntactically well-formed and MUST decode: the decoder's job is grammar, and a
+   receiver that rejected `0` as `malformed_frame` would close the connection on a frame it could
+   have answered. Admission is where the range is enforced: a handshake naming a version outside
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

Verdict on head 4569d67b9574077c68fb41da1c64bef681301a71: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Superseded by #2124, which carries the same amendment restructured so the precedence list and the changed-decision set follow from the per-decision entries rather than standing beside them.

Issue #2117, opened 2026-08-22
Audit admission: a commit-deadline timeout is reported as QueueAdmissionExhausted, leaving the write outcome indeterminate


ohdearquant:
## Summary

`AuditBatchControl::submit` returns `AuditTerminalReason::QueueAdmissionExhausted` from two
conditions that mean opposite things. One is a genuine pre-enqueue refusal; the other is a
post-enqueue commit timeout whose write outcome is indeterminate at the caller. They are
indistinguishable in the error surfaced to a client.

## Mechanism

`crates/khive-runtime/src/audit_batch.rs`:

- **Path A, line 643-645** — a true admission refusal. `state.pending.len() >= max_pending_rows`
  is checked *before* the row is enqueued, so nothing was submitted and nothing will be written.
- **Path B, line 662-666** — `tokio::time::timeout(self.config.admission_deadline, rx)` elapsed.
  This is reached only *after* the row has been pushed onto `state.pending` and `submitted_rows`
  has been incremented (line 646-651). The timeout arm performs no removal, so the row remains
  queued and is drained by the generation driver (line 490) independently of the caller's error.

Both arms return the same variant. The variant's own documentation (line 105) describes only
Path A: "`AuditBatchConfig::max_pending_rows` was reached."

## Why the two are not interchangeable

Defaults are `max_pending_rows = 4096` and `admission_deadline = 5s` (line 173, 177).

A client issuing a modest batch cannot plausibly reach 4096 queued rows, so a
`QueueAdmissionExhausted` seen at that scale is Path B, not Path A. The error nonetheless names
queue exhaustion, which points a caller at batch size — a dimension that is not the cause and
that they can reduce without effect.

The consequential difference is the write outcome:

- Path A: the audit row was not written, and will not be.
- Path B: the audit row was enqueued and will most likely be written, after the caller has already
  been told the operation failed.

A caller therefore cannot use this error to decide whether to retry. Retrying a Path B failure
risks a duplicate obligation; not retrying a Path A failure drops one.

## Observed shape

Reported from a batched client call of 15 independent read ops against the same store: 8 failed
with this error and 7 succeeded, and the failures were the *first* 8 in the batch rather than the
last. That ordering is inconsistent with queue-depth exhaustion, which would starve the tail, and
is consistent with early submitters waiting on a cold generation driver and crossing the 5s
deadline while later ones landed on a warm one. The same ids succeeded immediately when re-issued
in smaller batches, so the ids and addressing were not implicated.

[...]

PR #2121, opened 2026-08-22
test(runtime): pin that an audit-obligation failure reports a write that already committed

ohdearquant:
## What this adds

One test in `khive-runtime`, no production code change.

## Why

The audit obligation row is written after the handler returns, from its return value, on a writer
acquisition separate from the handler's. When that obligation commit fails, `fold_audit_obligation`
turns the dispatch's `Ok` into an `Err` — but the handler's effect has already committed, and a
separate transaction cannot roll it back. The caller is told `ok=false` for work that was done.

The existing obligation test exercises this flip through `list`, a read, where the distinction does
not arise. Nothing covered it for a write verb, which is the case that decides caller behaviour: a
caller that reads this error as "it did not run" and retries applies the effect twice.

## What the test asserts

It registers a pack whose handler records each committed effect, dispatches it against an event
store whose appends fail, and checks both halves:

1. The effect is present in the recording vector while the returned error names the obligation
   failure.
2. A retry of the same dispatch leaves **two** committed effects, not one.

The second assertion is the point. The error message is the only caller-visible signal separating
this case from a failure where the verb never ran, so a change to that message, or to the fold, is a
change to what callers can safely retry. This test makes that change fail loudly instead of
silently.

## Scope

Pins current behaviour; does not change it. Whether the underlying flip is the right contract is a
separate question this PR takes no position on.

## Gates

- `cargo clippy -p khive-runtime --all-targets --all-features -- -D warnings` — clean
- `cargo test -p khive-runtime --lib obligation_failure_reports_a_write_that_already_committed` —
  1 passed, 0 failed

The test was also checked for discrimination rather than assumed: flipping the retry expectation to
a single element reddens exactly that one assertion and prints the two-element list, so it fails for
the reason claimed.

### DIFF
--- a/crates/khive-runtime/src/pack.rs
+++ b/crates/khive-runtime/src/pack.rs
@@ -5685,26 +5685,35 @@ pub(crate) mod tests {
     impl MemoryEventStore {
         /// Record a submission attempt. Call before any failure check.
         ///
-        /// The projection carries the verb and whether the row's resource has a
-        /// `cost_unit` key, not just kind and outcome. That key is the observable
-        /// consequence of the row having been built FROM the handler's return
-        /// value: `cost_unit::resource_payload` derives it from `ok_val`, while
-        /// the error path's `base_resource_payload` documents that it omits it.
-        /// Recording only the outcome would leave a submission that ignores
-        /// `ok_val` while still stamping `Success` indistinguishable from one
-        /// that sources it, and those are different implementations.
+        /// The projection carries the verb and the row's `resource.cost_unit`
+        /// value, not just kind and outcome, and not merely whether the key is
+        /// present.
+        ///
+        /// Presence alone is too weak to pin what it looks like it pins.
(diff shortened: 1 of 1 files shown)


ohdearquant:
Field evidence that this behaviour is reachable from the `request` MCP surface today, not only from
the runtime internals this test pins: a `create` call returned
`ok: false` / `internal: audit obligation commit failed for verb "create": QueueAdmissionExhausted`
and **committed the entity anyway**. The natural retry — the correct response to an
admission-pressure error — produced a second, complete, readable row 20 seconds later.

Full reproduction with both record timestamps is in #2117. Recording it here because it bears on what
this test should assert: the caller-visible consequence is a **duplicate**, and the caller has no
field in the response that would let it choose a different retry policy. A test that pins "the write
committed" is exactly right; a companion assertion that the response distinguishes
record-committed-audit-failed from nothing-committed would close the loop for callers.

ohdearquant (commented):

Issue #2104, opened 2026-08-22
git.digest: issues-only ingest requires a full clone and enforces the size cap only after checkout


ohdearquant:
Reproduction (any https source whose clone exceeds the cap; observed against a ~1.9 GiB repository): `git.digest(source="https://github.com/<owner>/<large-repo>", include=["issues"], max_items=50)`

Observed: the operation performs a full clone with a complete working-tree checkout (tens of thousands of files updated), then fails with `clone exceeds the per-clone size cap (1956946351 bytes > 1073741824 bytes); the clone was removed`. Wall time is spent on the entire transfer and checkout before the cap refuses the work.

Three defects:

1. **An issues-only ingest should not require repository objects.** `include=["issues"]` is served entirely from the GitHub API surface; requiring a clone makes issue ingest impossible for any repository whose history exceeds the cap, even though no git object is read. Decoupling metadata-only includes from git source resolution is a larger change than defect 2 and is tracked here as its own question.

2. **The clone is already blob-filtered; a working-tree checkout defeats the filter.** `clone()` passes `--filter=blob:none` (`crates/khive-pack-git/src/cache.rs:708`) but not `--no-checkout`, so `git clone` checks out the default branch and lazily backfills every blob reachable at `HEAD` — which is the "tens of thousands of files updated" above. The warm path repeats this: `advance_to_fetched_tip()` runs `git reset --hard refs/remotes/origin/HEAD` (`cache.rs:748`), called after fetch at `cache.rs:269` and after refetch at `cache.rs:319`.

   Nothing reads that working tree. The cache slot's only production consumer is `git.digest`'s remote mode (`handlers.rs:140`), and the commands it issues against the slot are `rev-parse`, `remote`, and `log`, all of which read the object database. `advance_to_fetched_tip`'s own doc comment already records that nothing writes into the working tree either.

   Remedy: pass `--no-checkout` at the clone, and advance `HEAD` with a ref update rather than `reset --hard`. The blob filter then holds for the slot's whole lifetime, and `dir_size` measures a filtered clone rather than a filtered clone plus a materialized tree. The size reduction follows from the mechanism and is stated here as predicted, not measured; a before/after byte measurement on a large repository is the acceptance criterion for the fix.

[...]

Issue #1644, opened 2026-08-02
git.digest remote-URL mode walks the cache clone's stale HEAD after fetch — empty walk reports done:true


ohdearquant:
The scratch-clone cache's `fetch` helper runs `git fetch --prune` (crates/khive-pack-git/src/cache.rs:419), which updates remote-tracking refs but never advances the clone's checked-out HEAD. `walk_commits` then runs `git log --reverse {cursor}..HEAD` against that stale HEAD.

Reproduced on a cache slot whose checkout was ~3 weeks old while FETCH_HEAD was minutes old: with an ingest cursor NEWER than the stale HEAD (advanced by earlier local-path ingests of the same project), `{cursor}..HEAD` is an empty range, git exits cleanly, and the pass returns zero ingested / zero skipped / zero refused with `done:true` — indistinguishable from a fully-ingested repo. Even without a cursor, remote-mode ingest silently covers only history as of the clone's last checkout advance, not the fetched tip.

Fix direction: walk the fetched tip, not the checkout — either resolve the walk range against `origin/HEAD` (and use the same ref for the file-listing pass), or hard-advance the disposable cache clone to the fetched tip after every successful fetch. A guard worth adding either way: if the walk range resolves empty while the remote tip is not an ancestor of the cursor, that is an inconsistency to surface, not a clean completion.

Related: #1617 (completion semantics), and the companion issue on `done`/cursor-stall reporting.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ohdearquant:
Exposure census, measured on a production deployment: **every existing cache
slot is stale.** Five slots, each frozen at its own clone date (2026-07-09
through 2026-07-30), consistent with the fetch-never-advances-HEAD mechanism —
the blind spot starts at a different date per repo, so no single coverage
statement covers all anchors.

One measured case adds nuance. On a 718-commit repo, a local-path pass
ingested exactly `718 − 644 = 74` commits, where 644 is what is reachable from
the stale slot's HEAD — confirming prior store state was bounded by the frozen
checkout. But the cursor was NOT sitting on the stale HEAD: it sat 33 commits
(5 days) behind it, with all 33 already in the store. That is explained by
designed stall semantics, not a new defect: a transient create failure freezes
the cursor at the last contiguous success while later records in the same pass
still land (idempotent by natural key). Corollary worth recording: a frozen
cursor value **dates the last clean pass** — if a later clean pass had run, it
[...]

PR #2122, opened 2026-08-22
fix(pack-git): stop materializing a working tree in the digest cache slot

ohdearquant:
Addresses defect 2 of #2104.

## The defect

The scratch clone passes `--filter=blob:none`, but nothing prevented the checkout. `git clone`
checks out the default branch, and that checkout lazily backfills every blob reachable at `HEAD` —
so the filter bought nothing. `dir_size` measured a filtered object store plus a fully materialized
tree, and the per-clone size cap refused work only after paying for all of it.

Nothing reads that tree. Every command this crate issues against a cache slot is `rev-parse`,
`log`, or `remote`, all of which read refs and the object database.

## Both halves are required

1. `clone` passes `--no-checkout`.
2. `advance_to_fetched_tip` moves a ref instead of running `reset --hard`.

The second is not optional, and this is the part worth reviewing. `reset --hard` repopulates the
working tree on the next `ensure_clone` regardless of how the slot was cloned, so `--no-checkout`
alone hands the bytes straight back on the first refresh.

`HEAD` is advanced by updating the branch its symref names, falling back to updating `HEAD` itself
when the slot is already detached. Both shapes occur — the first is what `git clone` produces, the
second is reachable through repair paths and through slots older than this code — and `rev-parse
HEAD` and `log` resolve identically either way.

## Measured

On this repository:

| slot state | size |
|---|---|
| `--filter=blob:none`, with checkout (before) | 61.7 MiB |
| `--filter=blob:none --no-checkout` (after) | 5.6 MiB |
| the 5.6 MiB slot after one `reset --hard` | 62.5 MiB |

The third row is why change 2 exists.

## Test

`cache_slot_never_materializes_a_working_tree` pins both halves against a local upstream. Its
second assertion — the one after the re-ensure — is what catches a regression to `reset --hard`;
the assertion on the fresh clone alone would not, because with the reset restored the slot
repopulates on refresh even though it was cloned `--no-checkout`.

Both arms were verified by mutation, each reddening only its own predicted assertion:

- dropping `--no-checkout` reddens the fresh-clone assertion, printing `["tracked.md"]`
- restoring `reset --hard` reddens the post-re-ensure assertion, printing
  `["tracked.md", "second.md"]`

The test's helper excludes `.git` and the cache's own marker file by name rather than by a dotfile
rule, so a checked-out dotfile would still count as a materialized tree. It also asserts the
upstream has a tree of its own, so an empty result from the helper is a real absence.

The existing `reensure_advances_checkout_to_fetched_tip` test passes unchanged, so the #1644
stale-HEAD guarantee is preserved.

## Scope

Existing slots cloned by older code keep their materialized tree until they are evicted or refused
by the cap; nothing here reaches back to clean them.

Defect 1 (an issues-only ingest should not require a clone at all) and defect 3 (the cap should
bound the transfer rather than refuse after it) are unaffected and remain open.

## Gates

- `cargo test -p khive-pack-git` — 255 + 93 passed, 0 failed
- `cargo clippy -p khive-pack-git --all-targets --all-features -- -D warnings` — clean

## Legacy slots, and which claims here carry arms

Slots created before `--no-checkout` still carry a worktree, and nothing on the ordinary path
would ever remove it. Such a slot is replaced whole rather than stripped in place: the removal
goes through the existing ownership-guarded remover, which opens the slot with
`O_DIRECTORY | O_NOFOLLOW`, re-checks ownership against that descriptor, renames by descriptor
into a private staging namespace, and verifies the moved inode before deleting anything. This
change therefore adds no new destructive traversal of its own.

Two claims in this PR are supported differently, and the difference is worth stating plainly.

**Carries arms.** The migration behaviour is pinned by tests: a legacy slot reached through the
ordinary path and a legacy slot reached through the repair path both lose their worktree and
index. `cargo clippy -p khive-pack-git --all-targets -- -D warnings` is clean;
`cargo test -p khive-pack-git` passes 93; `cache::tests` passes 39, including both migration
tests.

**Carries eyes only.** After removing a legacy slot, the code no longer re-reads the slot
pathname to decide what to do next; the removal reports whether it happened and the caller
branches on that boolean. The hazard it closes is a window between the removal and a re-read in
which an external writer could occupy the pathname, and the per-key lock is same-process only, so
it does not exclude that writer. No test reaches this: the window is internal to the function and
there is no injection seam, so both migration tests observe the same outcome with or without the
fix. It is verified by reading, not by a reddened arm, and it should be reviewed on that basis.

Migration is gated to unix. It removes through a helper whose non-unix body is a pathname-based
recursive delete with no ownership check. That body is pre-existing and already reachable on
Windows through size-cap eviction; this change declines to widen its reach with a second caller.
On a non-unix target a legacy slot keeps its worktree until ordinary LRU or cap eviction retires
it, which is what that target did before this change.

## Non-Unix child-`.git` pinning (tracked in #2149)

The Unix path pins the child `.git` by descriptor (`openat` `O_NOFOLLOW` +
`fchdir` + `--git-dir .`), closing the symlink-swap TOCTOU on the `.git` entry.
Non-Unix targets cannot pass a directory descriptor to git, so `git_at_slot` and
`revalidate_owned_slot` fall back to resolving `--git-dir <repo>/.git` by
pathname, which stays vulnerable to a `.git` swapped for a symlink after
validation. Both fallback sites now carry an at-site comment stating the
limitation. A Windows-safe handle-pin needs platform handle APIs that cannot be
exercised on this project's macOS/Linux CI, so it is tracked as a follow-up in
#2149 rather than shipped unverified.

### DIFF
--- a/crates/khive-pack-git/src/cache.rs
+++ b/crates/khive-pack-git/src/cache.rs
@@ -261,39 +261,62 @@ fn ensure_clone_locked(root: &Path, canonical_url: &str) -> Result<PathBuf, Cach
     let repo_dir = root.join(&key);
     let cap = clone_max_bytes();
 
-    // Slots created before `clone` gained `--no-checkout` still carry a
-    // worktree, and nothing else on this path would ever remove it. Such a slot
-    // is replaced whole rather than stripped in place: the removal re-derives
-    // ownership from a descriptor it opens itself, so this change adds no new
-    // destructive traversal.
-    let mut migrated = false;
-    if repo_dir.join(".git").exists() {
+    // Slot state is decided exactly once, in the same breath as the
+    // ownership check; the arms below key on this decision and never ask
+    // the filesystem again. A second `.git` read would answer a question
+    // about whatever occupies that pathname NOW — after our own removal,
+    // or after a foreign process's write in a shared scratch root, that is
+    // not necessarily anything this process owns, and the fetch arm would
+    // then mutate an unowned repository (refs and ownership marker).
+    // `slot_lock` is in-process only, so a foreign writer is not excluded.
+    enum SlotState {
+        /// `.git` present at the decision point, ownership verified,
+        /// no legacy worktree.
+        Owned,
+        /// Legacy slot (pre-`--no-checkout` worktree) replaced whole by our
+        /// own act; the pathname is vacant as far as this process is
+        /// concerned. The removal re-derives ownership from a descriptor it
+        /// opens itself, so migration adds no new destructive traversal.
+        Replaced,
+        /// No slot at the decision point.
+        Absent,
+    }
+    let slot = if repo_dir.join(".git").exists() {
         if !is_owned_entry(&repo_dir) {
             return Err(CacheError::UnsafeToReplace(repo_dir));
         }
-        migrated = migrate_legacy_slot(root, &repo_dir)?;
-    }
-
-    // `!migrated` and not a second `.git.exists()` read on its own. Asking the
-    // filesystem again would answer a question about whatever occupies that
-    // pathname NOW, which after a removal is not necessarily anything this
-    // process owns; the fetch arm below would then operate on it unchecked.
-    if !migrated && repo_dir.join(".git").exists() {
-        fetch(&repo_dir)?;
-        advance_to_fetched_tip(&repo_dir)?;
-        // `repo_dir` was just fetched into and its ownership already
-        // confirmed above; it vanishing here is a real problem (`slot_lock`
-        // excludes a concurrent `ensure_clone`/`refetch_clone`/`reclone` on
-        // this same key, so nothing else in this crate should be touching
-        // it), not a maybe-absent slot, so propagate rather than swallow.
-        let size = dir_size(&repo_dir)?;
-        if size > cap {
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

Verdict on head 7d42b6994f79aaf1d97144df9205a35176d290bd: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 6da62972b995269845440bfe4078f748209554f9: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 72cf61050a6de0e52f2e67101a3a1f9bca3ef5b1: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2124, opened 2026-08-22
docs(adr-137): amendment 1 — close the wire contract before the first consumer

ohdearquant:
Amendment 1 to ADR-137, closing the wire-level decisions the parent left unstated or stated only in prose. It lands before `khive-wire-protocol` has an in-tree consumer, which is the window in which any of these behaviours can still change without a compatibility break.

## What it settles

Ten numbered decisions. Five are recorded as contract because the implemented behaviour is already correct for it. Five require the crate to move: envelope member semantics, the id-less unknown-error-code case, handshake sequence coverage across both endpoint roles, opaque subvalue byte fidelity, and the outermost type of `event.payload`.

Each decision states the behaviour normatively rather than describing what the code does, so a non-Rust client can implement it from the ADR.

## Why the precedence list is in here

The parent makes the crate's own documentation a second normative authority. Six of the ten decisions contradict that documentation as it stands, so the amendment names each superseded passage. Three of those displacements are bounded rather than wholesale, and each entry says where its bound falls.

That list is explicitly a debt. The implementation that ratifies these decisions updates the crate documentation in the same change and deletes the section.

## Structure

The superseded parent paragraphs carry inline notes pointing at the decisions that bound them, so a reader who stops at the parent does not leave holding a replaced rule.

Supersedes #2120, which carried the same amendment with the precedence list and the changed-decision set stated independently of the per-decision entries. Here both are consequences of the entries themselves.

Docs only. No code changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,6 +60,12 @@ repos:
         language: system
         files: ^(docs/.*\.md|crates/(.*/docs/.*\.md|.*/design[^/]*\.md))$
         pass_filenames: false
+      - id: lint-adr-137-displacements
+        name: ADR-137 amendment 1 derived passages match their table
+        entry: uv run scripts/lint-adr-137-displacements.py --check
+        language: system
+        files: ^(docs/adr/ADR-137-.*\.md|scripts/data/adr-137-amendment-1-displacements\.json|scripts/lint-adr-137-displacements\.py)$
+        pass_filenames: false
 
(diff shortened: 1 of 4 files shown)


ohdearquant (commented):

Verdict on head d5c90888648ff3a53537e2749d815a7a38b55de3: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2127, opened 2026-08-23
docs(adr-137): condition the client's pre-handshake admissions on an outstanding handshake

ohdearquant:
Follow-up to #2126. A review at that PR's head raised one blocking correctness gap and several
smaller ones; #2126 had already merged when the finding landed, so this carries them forward against
the merged text rather than reopening anything.

## The correctness gap

Correction 2 stated decision 6's pre-completion admissible set per receiving role. That closed the
role-blindness, but it stated the client's half as a property of the frame kind alone: inbound
`handshake_ack` is admissible at a client, with no precondition on the client's own state.

So a conforming client could accept an unsolicited `handshake_ack` on a fresh connection it had sent
nothing on, and could accept one naming a version it never offered. Either lets a peer fix the
connection's version unilaterally, which is the negotiation the parent's first-frame rule exists to
force. Decision 1 makes the second case concrete rather than theoretical: version `0` decodes and
reaches admission, so a parseable version field is not evidence the version was negotiated.

Both client admissions are now conditioned on having a handshake outstanding, and the
`handshake_ack` must be validated against it. Recorded as correction 3, with the shape of the three
corrections named: the original rule forbade the frames that perform the handshake, correction 1
exempted them without regard to role, correction 2 scoped the exemption by role but not by
connection state. Each repair was correct about what it fixed and silent about the next dimension.

## Checker and table

- **`english_list([])` raised `IndexError`.** Reachable whenever every displacing row is
  `wholesale` — a state the table only became able to reach in #2126, which introduced `wholesale`
  as a statable value. Reproduced by mutating all four bounded rows to `wholesale`: traceback
  before, `rc=1` with a normal complaint after. The empty case now returns the empty string and the
  precedence closing branches on it with its own sentence, rather than a degenerate rendering of one
  that presupposes a member.
- **The precedence closing names one decision as its illustrative example.** That is a claim about
  membership sitting inside a generated passage, and nothing checked it. If that decision stopped
  being bounded, the passage would publish a sentence contradicting the membership list beside it.
  Generation now fails with a message saying to rewrite the sentence rather than change the number.
  Verified by mutating only that decision to `wholesale`: `rc=1`, guard message.
- **A row carrying a scope note and no scope kind, while displacing a passage, produced two
  complaints for one defect with one fix.** Measured against the pre-fix script in the same tree,
  same mutated table: 2 complaints before, 1 after, both still rejecting.
- **`_scope_kinds` described a rule set the code had stopped implementing.** It said the checker
  rejects a kind without a note; the checker requires a note only for the bounded kinds, and rejects
  a note on `wholesale`. The sentence predates `wholesale` and was not updated when it was added, in
  the same change that added it. Rewritten to describe what the code does.

## Verification

`uv run scripts/lint-adr-137-displacements.py --check` exits 0 on the real table, with all 14
citations resolving and both derived passages present verbatim. Every mutation above was restored
from a snapshot copy and proved byte-identical afterwards, never inverse-edited, and each mutation
asserted that it had actually taken before its result was read.

Documentation and one script only; no crate code changes, so no build is required.


ohdearquant (commented):

Verdict on head e3ad7b6100cc1b3592a97fd4dae305846176175f: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2128, opened 2026-08-23
fix(gtd,mcp): reject unrepresentable due dates and fingerprint the display timezone

ohdearquant:
Two defects on main in the date-only due-date path that arrived with the
configured display timezone.

## 1. Unchecked arithmetic on a caller-supplied date (availability)

`%Y` accepts the full signed year range chrono can represent, so a caller can
name a date whose anchor window runs off the end of `NaiveDateTime`. The
skipped-midnight branch then did bare `-` / `+` on local midnight, and chrono's
operators panic on overflow rather than returning `None`. A date-only due value
near the end of the range therefore panicked inside a public verb instead of
returning a validation error, repeatably.

The window is now computed with `checked_sub_signed` / `checked_add_signed` and
clamped to the representable range rather than failing. Near the ends of the
range the window is truncated, not absent, and the instants left in it are
exactly the ones a search could return.

Clamping alone is not enough, because a date at the bottom of the range can
still be valid. In a zone east of UTC the minimum date has no representable
local midnight, but later local times on that same date do exist, so the date
has a least representable instant and the resolver's contract can be honoured.
The search returns the lower bound directly when that bound already carries the
date.

Measured at `NaiveDateTime::MIN`, rather than argued: `Pacific/Apia` +12:33:04,
`Pacific/Midway` +12:37:12, `America/Adak` +12:13:22 and `America/Anchorage`
+14:00:24 all resolve to that instant, while `America/New_York` -4:56:02,
`America/Los_Angeles` -7:52:58 and `Pacific/Honolulu` -10:31:26 have a
representable local midnight and resolve through the direct branch. A zone's
name is not evidence of the sign: `America/Adak` reads as west of UTC, but its
earliest offset is positive because it crossed the international date line.

Clamping also makes the obvious comparison unsafe, so it is not used. Calling
`date_naive()` on a zone-local datetime panics when the offset pushes the local
wall clock outside `NaiveDateTime`'s range. A positive-offset zone reaches that
at the top of the range, a negative-offset zone at the bottom. The search path
uses a total comparison instead: read the offset, try checked arithmetic, and
when the local datetime cannot be built take the ordering from the overflow
direction, since past the top is later than every representable date and below
the bottom is earlier.

The resolver therefore has a single failure mode: no representable instant
carries this local date. It returns `Option` accordingly, and it is total over
every date `%Y` can parse, so no caller needs a range check of its own.

## 2. Display timezone missing from the warm-daemon configuration fingerprint

Date-only resolution reads `display_timezone`, but the configuration
fingerprint did not hash it, so two runtimes differing only in that setting
produced the same configuration id. A warm daemon could anchor a client's
date-only due value under the wrong zone and persist the wrong instant.

The component is included unconditionally. The default is the host's zone
rather than UTC, so "differs from the default" is host-dependent and cannot be
the condition.

The rollout cost, stated as it actually happens: a daemon already warm when
this lands keeps the identity it computed at startup, so a client built from
this code sends an id that daemon does not recognise. The daemon answers with a
configuration mismatch and the client falls back to local dispatch. It does not
respawn. Until that daemon is restarted, every request pays a failed forwarding
round trip and cannot use the daemon's warm indexes or embedders, and each one
increments a counter documented as never expected on a correctly configured
fleet. Restarting the daemon ends it.

## Tests

Each fix was checked by reverting it and reconciling against the arms predicted
before the run:

- Removing the fingerprint component reddens exactly the differing-zone test;
  the matching-zone test stays green.
- Reverting the resolver's checked arithmetic restores the panic.

Boundary coverage names which path each zone takes, because the two are
different code: east of UTC the minimum date exercises the lower-bound return,
and the assertion is that the resolved instant IS `NaiveDateTime::MIN`. The
usual one-second-earlier check cannot be written there, because one second
earlier does not exist. West of UTC the same date resolves through the direct
branch. A separate test exercises the two cases where the total comparison and
`date_naive()` disagree, so the helper is not mistaken for ceremony.

The searched window has one definition: the all-zone sweep takes its probe
bounds from the same helper the resolver uses, so a change to the radius cannot
leave the sweep checking the old window while still reporting zero violations.
That sweep walks every zone in the pinned database over 1850-2100 and asserts
the least-instant property directly.

The two fingerprint tests each build their configs from one base rather than
from two constructor calls. `RuntimeConfig::default` reads `HOME` to derive
`db_path`, `db_path` is folded into the id, and other tests in this binary set
and restore `HOME` around their own work, so two calls read a mutable global at
two different instants. That perturbation pushes the two ids apart, which fails
the equality test and, more importantly, would let the inequality test pass for
a reason unrelated to timezones.

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -600,8 +600,24 @@ pub(crate) fn compute_config_id_with_runtime_policies(
     // Included unconditionally rather than only when non-default. The default
     // is the HOST's zone, not UTC, so "differs from the default" is itself a
     // host-dependent predicate and would make identity depend on where the
-    // fingerprint was computed. The cost is that every existing warm daemon
-    // takes a new identity once, which is a respawn, not a correctness event.
+    // fingerprint was computed.
+    //
+    // The cost, stated as it actually happens: a daemon already warm when this
+    // lands keeps the identity it computed at startup, so a client built from
+    // this code sends an ID that daemon does not recognise. The daemon answers
+    // `config_mismatch` and the client falls back to LOCAL dispatch. It does
+    // not respawn — `FallbackReason::ConfigMismatch` is classified
+    // `FallbackSeverity::Illegitimate`, and the kill-and-respawn path
+    // (#644/#539) governs the protocol/parse reasons, not this one. So until
+    // that daemon is restarted, every request pays a failed forwarding round
+    // trip and loses the daemon's warm indexes and embedders, and each one
+    // increments a counter documented as never expected on a correctly
+    // configured fleet.
+    //
+    // Spelled out because "the daemon takes a new identity" invites the reading
+    // that it restarts itself. It does not, and nothing here makes it: this is
+    // a one-time operational cost that ends when the daemon is restarted, by
+    // whoever restarts it.
     let base = format!(
         "packs=[{}];db={};embed={};extra=[{}];fresh_tail={};blob_hydration_bytes={};backend={};outbound=[{}];git_write={};display_tz={}",
         packs.join(","),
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

ohdearquant (commented):

Verdict on head fcdb878c01f427ec6c61641357fdc23bea72ab75: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2130, opened 2026-08-23
test(gtd): trigger the all-zone timezone sweep when the chrono-tz pin moves, and pin its probe granularity

ohdearquant:
The all-zone timezone sweep asserts that date-only resolution returns the least
instant carrying a local date, across every zone in the bundled `chrono-tz`
database. It is `#[ignore]`d because it walks the whole database, which is the
wrong cost on an ordinary pull request and the right one exactly when the
bundled zone data changes under the resolver.

That made it a remembered duty rather than a triggered one. This adds the
trigger.

## The bundled zone data is an input, not an inert dependency

Moving the `chrono-tz` pin can change which dates have no local midnight and
where a gap begins, so it can change the resolver's answers without changing a
line of the resolver. A workflow now runs the sweep on any pull request
touching a manifest or lockfile that can carry that version, and on the pack
holding both the resolver and the sweep. `make tz-audit` runs the same thing
locally.

The job prints the resolved `chrono-tz` version before running, read from the
lockfile and with no `|| true`, because a step that cannot fail cannot report:
the sweep's result only means something against a named version of the zone
data. It runs `--ignored` rather than `--include-ignored`, so a failure in an
ordinary test cannot be reported as a failure of the sweep.

This does not gate ordinary resolver correctness. The crate's normal tests
cover that on every pull request through the main job.

## The probe granularity is pinned by a test that can fail

The sweep's skipped-day branch probes the search window for an instant carrying
the requested local date. The step size matters: a local-date interval shorter
than the step can sit between two probes, so a coarse probe reports that no
instant carries the date without having looked at the instants that would. That
failure is invisible against the real database, which contains no interval of
that shape.

So the step is pinned against a fabricated calendar in which one local date
exists for thirty seconds. A one-minute probe steps over it and finds nothing; a
one-second probe finds it. The fixture guards itself, asserting that the
interval really is shorter than the coarse step, so the test cannot pass for a
reason unrelated to granularity. Synthetic data is the right fixture precisely
because the real table has no such transition: an assertion that cannot be made
to fail is not an assertion, and the alternative was shipping the finer step on
argument alone.

Pinning it needed the probe extracted into a helper that takes the calendar as a
parameter, which is what lets the test substitute a fabricated one.

## Note on the search window

The helper takes the window as bounds rather than a centre and a radius. The
bounds come from `anchor_search_window`, the same function the resolver uses, so
the searched window has a single definition and a change to the radius cannot
leave this probe checking the old window while still reporting no violations.
That also keeps the helper free of bare `-` and `+` on `NaiveDateTime`, which
abort at the ends of the representable range rather than returning `None`; the
step is checked for the same reason, so overflowing past the end of the range
ends the search instead of the process.

## The workflow declares its own token scope

With no `permissions:` block a workflow inherits the repository's default
`GITHUB_TOKEN` scopes, so its actual authority is set by a repository setting
the file neither states nor tracks, and changing that setting would widen it
with no diff for a reviewer to see. This job checks out, builds, and runs one
test, so it declares `contents: read` and nothing else.

That default is currently read-only, so this is about making the property
reviewable rather than closing an open hole. Three other workflows are in the
same position and are left alone here, since they are pre-existing and unrelated
to this change.

## Verification

The sweep runs green at this tree: 597 zones over 91311 dates each, 4299 gap
dates, 818 folds, 7 whole days a zone's calendar skips, and zero violations.
The granularity test's coarse arm fails and its fine arm passes, which is the
discrimination the test exists to provide.

### DIFF
--- a/.github/workflows/tz-database-audit.yml
+++ b/.github/workflows/tz-database-audit.yml
@@ -8,11 +8,15 @@ name: Timezone Database Audit
 # where a gap begins.
 #
 # The all-zone sweep exists to re-derive that. It is `#[ignore]`d because it
-# walks every zone and costs ~50s, which is the wrong trade on every ordinary
-# PR and the right one exactly when the pin moves. This workflow supplies the
-# trigger, so the duty is enforced rather than remembered: it runs on any PR
-# touching a manifest or lockfile that can carry a chrono-tz version, and on
-# the gtd pack that holds both the resolver and the sweep.
+# walks every zone and costs 48.28s in the debug profile CI uses (measured
+# 2026-08-23; the same sweep is 3.08s with --release, and the job still runs
+# debug because the trigger is a lockfile change, which invalidates the build
+# cache, so a release run would pay a cold optimized rebuild to save 45s of
+# sweep). That is the wrong trade on every ordinary PR and the right one
+# exactly when the pin moves. This workflow supplies the trigger, so the duty
+# is enforced rather than remembered: it runs on any PR touching a manifest or
+# lockfile that can carry a chrono-tz version, and on the source of the gtd
+# pack, which holds both the resolver and the sweep.
 #
 # What this does NOT gate: ordinary correctness of the resolver, which the
 # crate's normal (non-ignored) tests cover on every PR through the main CI job.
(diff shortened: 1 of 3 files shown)


ohdearquant (commented):

Verdict on head beaa02accae5658f902e87b6690f00a3d84eae49: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2151, opened 2026-08-24
docs(adr): ADR-170 — dedicated events daemon

ohdearquant:
Proposes splitting event persistence into a dedicated events daemon owning `events.db`.

Motivation: SQLite serializes writers per file, and a measured 24-hour window shows the event plane writing ~76,600 rows against ~2,540 domain-mutation dispatches (~30:1) into the same store — ~96% of write traffic by dispatch count is observational telemetry queueing ahead of domain writes, surfacing as `writer_task_begin_busy` refusals under load.

Design in brief:
- events daemon (same binary, new subcommand) owns `events.db`, reusing `SqlEventStore` and the existing Unix-socket framing/admission pattern
- domain process forwards appends fire-and-forget through a bounded queue with an explicit drop policy (the loss-tolerant durability class made concrete); never blocks the domain path
- reads open `events.db` read-only (WAL one-writer/many-readers)
- domain availability never depends on events-daemon liveness

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2152, opened 2026-08-24
feat(runtime): events-daemon split — audit lane leaves the domain store (ADR-170)

ohdearquant:
Implements ADR-170 (#2151): a dedicated events daemon owning `events.db`, with event persistence routed by append class.

## What moves, what stays

- The ADR-133 idempotent audit-batch lane — the measured bulk of event write volume (~82% of rows in a 24h production window, and the one-per-dispatch class that scales with load) — persists to `events.db`, owned by a supervised events daemon on its own Unix socket (same framing and peer-uid admission as the main daemon socket).
- Plain event appends stay on the domain store. The legacy `events` table has raw-SQL consumers whose correctness depends on co-resident rows: the schedule drain's creator-provenance fence, the kg projection worker's guarded event INSERT (transactional with main-store state), and the graph query compiler's cross-substrate UNION. Moving the whole plane breaks them by construction; the first was caught as a hard test failure during implementation, the others would have failed silently.
- Trait-level reads (`get_event`, `query_events`, `count_events`) merge both stores, so consumers on the `EventStore` trait observe one event plane.

## Host classes

The shared config resolver emits embedded (socket-less) mode for every file-backed resolution — one-shot CLI hosts and tests have no events daemon. The resident daemon entrypoints upgrade the resolved config to socket forwarding and supervise the events daemon from that same resolved config, so forwarding clients and the supervisor cannot anchor at diverging paths. Multi-backend boots re-anchor the split beside the declared main backend. Socket paths derive beside the events database they serve, never from a process-global location. `KHIVE_EVENTS_SPLIT=0` is the deployment kill-switch back to legacy behavior.

## Testing

- New: split-store routing (plain→legacy, idempotent→lane, merged + windowed reads), socket round-trip idempotency dispositions, fire-and-forget delivery, queue-overflow drop accounting, dead-socket typed errors with local preflight, protocol-skew refusal, daemon guard exclusivity, direct mode, resolver/daemon-upgrade contract.
- Full workspace `cargo check --all-targets`; khive-runtime (1316), khive-mcp (427), kkernel, and all touched pack suites pass; clippy and rustdoc clean under `-D warnings`.

Stacked on #2151 (ADR); the two ADR commits at the base of this branch belong to that PR and this one reduces to the implementation once it merges.

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -3946,7 +3946,7 @@ pub fn resolve_runtime_config_with_db_anchor(
     };
 
     // ADR-170: events-daemon split. Every file-backed resolution routes event
-    // persistence to `events.db` beside the main store, in DIRECT (embedded)
+    // persistence to the events database beside the main store, in DIRECT
     // mode: this resolver serves one-shot hosts (`kkernel exec`, `reindex`,
     // ingest) and tests, which have no events daemon to talk to. Resident
     // daemon hosts upgrade the resolved config to socket forwarding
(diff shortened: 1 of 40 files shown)


ohdearquant:
Superseded by the stacked series #2190 → #2195 → #2202 → #2203, which carries the same change set split into reviewable parts, with the design doc in #2151. Deferred follow-ups are tracked in #2204.

PR #2153, opened 2026-08-24
docs(adr): ADR-171 — brain daemon: profile state leaves the domain process

ohdearquant:
Phase 2 of the store decomposition, building on the primitive ADR-170 (#2151/#2152) validated.

Design in one paragraph: the brain pack's profile posterior state (three pack-internal tables plus warm in-memory state) moves behind a brain daemon owning `brain.db`, using the same socket pattern (framing, peer-uid admission, flock guard, supervision, kill-switch). Feedback reaches it by tailing the ADR-170 events lane with a checkpointed cursor — no push protocol, catch-up after downtime for free. Recall keeps its existing verb-level seam (`brain.profile`) and receives only the few coefficient numbers it already consumes; when the daemon is unreachable, recall proceeds un-reweighted and never blocks — feedback rides the events lane, so nothing is lost during an outage.

Two verified seam facts distinguish this from the events split: the brain tables have no raw-SQL co-residency consumers outside the brain pack, and consumers reach brain state through verbs — so the whole state can move, where the events plane could not.

Draft for spec-gate review.


ohdearquant (commented):

Verdict on head d7067c089111bbb10eebe0416af5e1e597dbdacf: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2154, opened 2026-08-24
ci: run workflow contract tests

### LINKED ISSUE #2131: scripts/tests/test_ci_workflows.py is never executed, so its workflow contract cannot fail
`scripts/tests/test_ci_workflows.py` asserts a contract over `.github/workflows/`, including that the push guard holds only the write permissions it needs. Nothing runs it, so none of those assertions can fail.

## What runs today

`scripts/ci.sh` invokes two of the three files in `scripts/tests/`:

```
41:    python3 "$SCRIPT_DIR/tests/test_verify_local_artifact.py"
44:    python3 "$SCRIPT_DIR/tests/test_writer_census.py"
```

`test_ci_workflows.py` appears in neither `scripts/ci.sh`, `.github/`, nor the `Makefile`. A pickaxe over the whole history finds no reference to it outside the file that defines it:

```
git log --oneline -S'test_ci_workflows' -- . ':!scripts/tests/test_ci_workflows.py'   # empty
git log --oneline -S'test_verify_local_artifact' -- . ':!scripts/tests/test_verify_local_artifact.py'
# 5efa8a157 build: verify local artifact before install (#1770)
```

The second command is the control: the same query does return a result for a file that is wired, so the empty result for the first is a real absence and not a broken query.

It is also not picked up by a directory-level discovery run. The only pytest configuration in the repo is `tests/khive-contract/pytest.ini`, whose `testpaths = tests` is scoped to that directory.

## Why it matters beyond the missing coverage

A test that cannot fail reads as coverage. Someone changing a workflow's permissions, the weekly job's lockfile handling, or the component runner's target filtering would see a green pipeline and reasonably conclude the contract still holds, because a suite asserting exactly that exists in the tree. The gap is invisible from the pipeline output.

The same shape leaves changes to `.github/workflows/` ungated in general: no workflow linting runs on a pull request that adds or edits a workflow file, so the first validation any workflow gets is executing on a subsequent run.

## Current state of the suite

Run by hand at `ad58e596` the three tests pass, so wiring it is a one-line addition rather than a fix-then-wire. Suggested home is `phase_lint` in `scripts/ci.sh`, beside the two sibling suites already invoked there.

ohdearquant:
## Summary

- Run the existing CI workflow contract suite from the lint phase.
- Keep workflow trigger, permission, and benchmark-command assertions enforceable in normal CI.

## Verification

- Regression suite: `scripts/tests/test_ci_workflows.py`
- `python3 scripts/tests/test_ci_workflows.py`
- `sh -n scripts/ci.sh`
- `git diff --check`

Fixes #2131

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -42,6 +42,9 @@ phase_lint() {
 
     echo "=== Writer Census Contract Tests ==="
     python3 "$SCRIPT_DIR/tests/test_writer_census.py"
+
+    echo "=== CI Workflow Contract Tests ==="
+    python3 "$SCRIPT_DIR/tests/test_ci_workflows.py"
 }
 
 phase_no_stubs_scan() {


ohdearquant (commented):

PR #2155, opened 2026-08-24
fix(ci): declare benchmark workflow permissions

### LINKED ISSUE #2134: bench-1m and bench-pipeline declare no GITHUB_TOKEN permissions, relying on an untracked repository default
Follow-up to #2131. Three workflows declare no `permissions:` at any nesting level, so their `GITHUB_TOKEN` authority comes from a repository setting that no file in the tree states or tracks.

**Nothing is over-privileged today.** Measured before writing this, so that no one reads it as urgent:

```
$ gh api repos/<repo>/actions/permissions/workflow
{"default_workflow_permissions":"read","can_approve_pull_request_reviews":false}
```

The default is read-only and workflows cannot approve pull request reviews. The undeclared workflows therefore run read-only right now. The issue is that this is true by a toggle rather than by anything reviewable: flipping that setting to `write` would widen every undeclared workflow at once, with no diff, no commit, and nothing for a reviewer to see.

## The three, and they are not the same case

**`bench-1m.yml`** — triggers on `push`, `pull_request`, `workflow_dispatch`. Its only write-shaped step is `actions/upload-artifact`, which does not need token write. Should declare `permissions: contents: read`.

**`bench-pipeline.yml`** — triggers on `pull_request`, likewise only uploads an artifact. Should declare `permissions: contents: read`.

**`bench-component.yml`** — this one should probably be left alone, and it is worth stating why rather than declaring something on it for consistency. It is `workflow_call`-only, invoked four times from `bench-track.yml`, and reusable workflows receive the calling job's permissions. Each of those four caller jobs declares `contents: write` and `issues: write`, which is exactly what its failure path needs when it runs `gh issue reopen` and `gh issue comment` against the tracking issue. Adding a top-level block here would constrain callers rather than grant anything, so "undeclared" is the correct design for a reusable workflow and not an oversight.

## Consequence for the population check proposed in #2131

A directory-wide assertion that every workflow declares `permissions:` would flag `bench-component.yml` permanently and correctly-by-its-own-rule, which is how a new gate starts crying wolf on day one. Whatever that check ends up asserting should exempt `workflow_call`-only workflows, or assert against the calling job instead.

## Suggested scope

Two one-line additions (`bench-1m.yml`, `bench-pipeline.yml`), plus the exemption noted above folded into #2131 rather than done here. Deliberately not folded into the pull request where this was found: that one adds a workflow and fixed its own missing declaration, and these three are pre-existing and unrelated to it.

ohdearquant:
## Summary

- Declare explicit read-only repository-content permissions for the one-million-row benchmark workflow.
- Declare the same least-privilege permissions for the pipeline benchmark workflow.
- Leave the reusable component workflow's caller-provided permissions unchanged.

## Verification

- `python3 scripts/tests/test_ci_workflows.py`
- Direct permission-contract assertions for both changed workflows
- Pre-commit YAML validation
- `git diff --check HEAD^ HEAD`

No regression test was added because this change is limited to CI configuration.

Fixes #2134

### DIFF
--- a/.github/workflows/bench-1m.yml
+++ b/.github/workflows/bench-1m.yml
@@ -56,6 +56,9 @@ on:
         required: false
         default: ""
 
+permissions:
+  contents: read
+
 env:
   CARGO_TERM_COLOR: always
 
--- a/.github/workflows/bench-pipeline.yml
+++ b/.github/workflows/bench-pipeline.yml
@@ -29,6 +29,9 @@ on:
       - "scripts/perf/bench_pipeline_daemon.py"
       - ".github/workflows/bench-pipeline.yml"
 
+permissions:
+  contents: read
+
 env:
   CARGO_TERM_COLOR: always
 


ohdearquant (commented):

PR #2161, opened 2026-08-24
fix(tests): isolate contract harness config

### LINKED ISSUE #2107: contract_test.py is not hermetic: home-config backend declarations break every contract test
## Symptom

On a machine where the user-level config (config.toml under `~/.khive`) declares `[[backends]]`, the `make ci` Contract Tests stage fails on every test before any assertion runs:

```
RuntimeError: MCP server closed stdout unexpectedly
```

## Cause

`tests/contract_test.py::_start_server` spawns `kkernel mcp --db <tempfile>` with `env = {**os.environ, "KHIVE_NO_DAEMON": "1"}` — config discovery is inherited from the developer's environment. Discovery falls through the working-directory khive.toml, then the db-directory config.toml, then the user-level config; when that last file declares multiple backends, the binary (correctly, per its own error message) refuses the `--db` override as ambiguous and exits before the MCP handshake:

```
Error: --db "<tempfile>" (or KHIVE_DB) cannot be combined with [[backends]]: 2 backend(s) are already declared in the discovered config ...
```

CI runners pass only because they have no home config, so the gap is invisible to CI and hits exactly the developers running the gate locally.

## Verification

Pinning `KHIVE_CONFIG` to an empty file before the same `kkernel mcp --db <tempfile> --no-embed --log error` invocation makes the server initialize and answer the MCP handshake normally.

## Suggested fix

1. `_start_server` should pin config discovery itself — e.g. write an empty config to the test's temp dir and pass `--config` (or set `KHIVE_CONFIG`) so the contract suite is hermetic regardless of the developer's home config.
2. When the child exits before the handshake, the harness should include the child's captured stderr in the raised error — the refusal message above was captured by the `stderr=PIPE` pipe but never surfaced, which turns a one-line diagnosis into a debugging session.

ohdearquant:
## Summary

- pin the MCP contract harness to an empty per-test configuration file so ambient home configuration cannot affect startup
- add a regression contract that initializes successfully with a two-backend user configuration present
- include captured child stderr when the server exits before completing the MCP handshake

## Verification

- `test_home_config_isolation` failed before the fix and passes afterward
- `python3 tests/contract_test.py` (11/11 passed)
- `cargo build --release -p kkernel`
- `python3 -m py_compile tests/contract_test.py`
- `git diff --check HEAD^ HEAD`

Closes #2107

### DIFF
--- a/tests/contract_test.py
+++ b/tests/contract_test.py
@@ -1069,9 +1069,32 @@ def test_batch_outcome_status(proc: subprocess.Popen) -> None:
 # ---------------------------------------------------------------------------
 
 def test_home_config_isolation(proc: subprocess.Popen) -> None:
-    """An ambient multi-backend home config must not affect this test server."""
-    stats = _tool(proc, "stats", {})
-    assert isinstance(stats, dict), f"stats returned a non-object: {stats!r}"
+    """An ambient multi-backend home config must not affect this test server.
+
+    The harness points the server at a freshly-minted temp database, so the
+    entity listing must start empty: a server that resolved the ambient home
+    configuration instead would surface that backend's pre-existing rows. The
+    sentinel round-trip then pins the write path to the same store.
+    """
+    before = _tool(proc, "list", {"kind": "entity", "limit": 10})
+    assert before == [], (
+        f"fresh temp database lists pre-existing entities: {before!r}"
+    )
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

PR #2164, opened 2026-08-24
fix(runtime): preserve empty property strings

### LINKED ISSUE #1995: kg: update response echo omits property keys set to empty string, reading as deleted
Setting a property to an empty string via `update(id=..., properties={"summary": ""})` succeeds — a subsequent `get()` shows `summary: ""` — but the update call's own response echo returns `properties` without the `summary` key at all, which reads as the key having been deleted.

`update` cannot delete a property key (documented, fine), but its response renders the result as though it had. A caller who verifies their write from the response — the natural pattern, since the response is right there — concludes the opposite of the truth.

Fix either direction: render falsy values in the echo, or omit `properties` from the echo entirely so absence cannot be read as deletion. The former keeps the echo useful for verification.

ohdearquant:
## Summary

- preserve empty strings nested inside caller-owned `properties` during Agent presentation
- retain existing empty-field compaction for response metadata outside `properties`
- cover update echo and subsequent get readback with a regression test

## Verification

- `cargo clippy -p khive-runtime --all-targets -- -D warnings`
- `cargo clippy -p khive-pack-kg --all-targets -- -D warnings`
- `cargo test -p khive-pack-kg`
- `cargo test -p khive-runtime --lib`
- regression: `update_empty_string_property_survives_agent_echo_and_readback`

Fixes #1995

### DIFF
--- a/crates/khive-pack-kg/tests/integration.rs
+++ b/crates/khive-pack-kg/tests/integration.rs
@@ -7,6 +7,7 @@
 use async_trait::async_trait;
 use khive_pack_kg::KgPack;
 use khive_runtime::pack::{HandlerDef, PackRuntime};
+use khive_runtime::presentation::{present, PresentationMode};
 use khive_runtime::{
     arm_prefix_resolve_fail_scoped, EntityCreateSpec, KhiveRuntime, Namespace, NamespaceToken,
     ParamDef, RuntimeError, VerbCategory, VerbRegistry, VerbRegistryBuilder, VerifiedActor,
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

Verdict on head 5745213c6e13ce4361d414597f13c75172029c21: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Closing — the behaviour fix landed in #2211.

The regression test from this branch is preserved in #2275, re-cut against current main. The branch is retained.

PR #2166, opened 2026-08-24
fix(pack-comm): report mailbox-wide unread total

### LINKED ISSUE #1931: comm.inbox unread_count is bounded by limit — reads as a mailbox total but is not
Measured: `comm.inbox(limit=10)` returned `unread_count: 10` and `comm.inbox(limit=2)` returned `unread_count: 2` against the same mailbox state, with more unread mail present in both cases.

The field name reads as a mailbox-wide total, so a caller watching it drop after a `mark_read` batch can score the batch as successful when in fact the count only reflects the returned page. This nearly caused a wrong conclusion during an unrelated investigation.

Fix options: compute the count over the whole mailbox (one COUNT query, independent of limit), or rename to something page-scoped (`unread_in_page`) and add a real `unread_total`. The first matches what every caller already believes the field means.

ohdearquant:
## Summary

- make `comm.inbox.unread_count` reflect the caller's complete unread inbound mailbox
- reuse the existing actor and legacy-message visibility rules
- preserve page filtering, pagination, status selection, and sent-history behavior
- add regression coverage for totals larger than the requested page

## Verification

- `cargo fmt -p khive-pack-comm -- --check`
- `cargo clippy -p khive-pack-comm --all-targets -- -D warnings`
- `cargo test -p khive-pack-comm` (370 passed)
- regression: `i1931_inbox_unread_count_is_mailbox_wide`
- `git diff --check`

Fixes #1931

## Note on the index/fixture coupling

The new partial index on `notes` evaluates `json_extract` over `properties`, so SQLite refuses a
malformed-JSON row while it exists. A `khive-pack-git` acceptance fixture had been disabling that
constraint by dropping a single json-evaluating index *by name*, which a second such index silently
re-imposes. The fixture now drops them *by class* — it enumerates the json-evaluating indexes on
`notes` and drops all of them — so adding a further index of this kind does not break an unrelated
test. Index changes on `notes` generally require class-robust fixtures.


ohdearquant (commented):

Verdict on head 8fd97e199c41cf940ac8255849970c2cf124dc55: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head b6cb8e475151817031da68bc3e62d03d7f152888: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 6d3a5b65387329c3153546de96213199e7cc923b: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 9b3a949a2ff9e5f3712d873a3eaf4c1095d77d58: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Two objections were raised against this branch at `345b8279`. Adjudicating both at source, one
holds and one does not.

### Does not hold: the unbounded count on the `limit > 0` page path

The objection is that `handle_inbox`'s ordinary page path runs `SELECT COUNT(*) FROM notes`
without a limit, so `comm.inbox(limit=1)` pays work proportional to every matching row.

The observation is accurate and the code is not this branch's. `query_notes_filtered` and
`query_note_page_snapshot` are untouched here:
`git diff origin/main...HEAD -- crates/khive-db/src/stores/note.rs` contains no line matching
`query_notes_filtered`, `query_note_page_snapshot`, or `COUNT(*)`. The count construction is
byte-identical on `main` (`crates/khive-db/src/stores/note.rs:1367` there,
`:1422` here — the shift is this branch's insertions elsewhere in the file).

That makes it a latent property of the paging contract rather than a defect in this change, and
it deserves its own issue rather than blocking this one.

### Holds: `LIMIT 1001` bounds returned rows, not candidate work

Confirmed, and the mechanism is narrower than an index-shape problem.

`count_unread_messages` reaches the store through `query_notes_filtered_bounded`
(`crates/khive-db/src/stores/note.rs:1543-1595`), which is a **paging** query being used to
count. It unconditionally appends an order clause — `ORDER BY created_at DESC, id ASC` when the
filter names no sort key (`:1576`) — and it selects all thirteen note columns and hydrates a
full `Note` for every row (`:1581-1589`).

The recipient predicate is `ifnull(json_extract(properties,'$.to_actor'),'') IN (?, '')`, which
is two index ranges. Ordering *across* two ranges cannot be served from the index, so SQLite
must materialise and sort every candidate before it can apply the limit. Two arms against a
10,000-row legacy recipient-less population, run read-only on `sqlite3 ':memory:'` with the
index exactly as this branch defines it:

```text
ARM A — the query as shipped (ORDER BY, all columns):
  SEARCH notes USING INDEX idx_notes_unread_probe_recipient (namespace=? AND kind=? AND <expr>=?)
  USE TEMP B-TREE FOR ORDER BY

ARM B — same predicate, no ORDER BY, SELECT 1:
  SEARCH notes USING INDEX idx_notes_unread_probe_recipient (namespace=? AND kind=? AND <expr>=?)

CONTROL — a predicate no index can serve:
  SCAN notes
```

The control confirms the plan output discriminates. Arm B is the whole finding: with the sort
removed, `LIMIT` becomes a genuine work cap and the two-range `IN` costs nothing extra, because
nothing has to be ordered across the ranges.

So the fix is not to reshape the index or add a direction predicate to it. It is that a count
should not be issued as a page. A dedicated bounded-count query — `SELECT 1`, no order clause,
`LIMIT max_rows + 1`, count the rows returned — makes the bound real, removes the sort, and
removes the hydration of up to 1001 full note rows to produce a single integer. The saturation
contract at the cap is unchanged.

A regression test for this needs a large recipient-less legacy population; the added test at
`crates/khive-db/src/stores/note_tests.rs:2244-2352` grows another actor's recipient bucket
instead, which does not exercise it.

### Also noted

`EqOrMissing` now matches a *present* empty string, because the SQL is
`ifnull(expr,'') IN (?,'')` (`crates/khive-db/src/stores/note.rs:790-823`) while the operator
contract says equality-or-absent (`crates/khive-storage/src/note.rs:234-236`). That widening
reaches every `EqOrMissing` consumer, not just this query. If the count gets its own SQL as
above, the widening no longer needs to exist for this feature at all.

### Disposition

This goes back for rework rather than another pass. The count-as-page shape is a design defect
rather than a line fix, and the `EqOrMissing` widening was introduced to make an index serve a
query that should not have needed one.

oceanwaves630 (approved):

PR #2169, opened 2026-08-24
fix(kg-editor): preserve showcase navigation state

### LINKED ISSUE #2034: kg-editor: view switches lose scroll and panel state; unknown view params coerce silently
Observed in the repository showcase (khive bundle).

**Symptom.**
1. Switching views (sidebar) scrolls the page back to the top hero, forcing a full re-scroll to the content on every switch.
2. After a view switch, the graph panel resets to the repository root while the URL still carries the previous `module=` parameter — panel state and URL state diverge. (Cold-load deep-link restore handles the same URL correctly.)
3. An unknown `view=` value silently coerces to the default view with no notice.

**Expected.** View switches preserve scroll position (or scroll to the view's own heading); panel state stays consistent with URL parameters after in-page navigation; unrecognized `view=` values either surface a notice or drop the parameter from the URL.

ohdearquant:
## Summary

- route sidebar view changes through the existing analysis focus-and-scroll path
- preserve structure-graph subtree and node selection across view switches
- align explicit module navigation and history restoration with graph state while leaving the implicit repository overview at the root
- retain the existing unknown-view repair notice and canonicalize only explicit module parameters

## Verification

- focused RED/GREEN regressions for view-navigation scroll and module restoration
- `npm run lint`
- `npm run typecheck`
- `npm test` (20 token-contract tests and 289 Vitest tests)
- `git diff --check`

Fixes #2034

### DIFF
--- a/apps/kg-editor/src/components/showcase/repo-showcase.test.tsx
+++ b/apps/kg-editor/src/components/showcase/repo-showcase.test.tsx
@@ -405,6 +405,64 @@ describe("repository showcase", () => {
     expect(scrollIntoView).toHaveBeenCalledOnce();
   });
 
+  it("moves focus and scroll position to an analysis selected from the view navigation", async () => {
+    const bundle = golden();
+    const user = userEvent.setup();
+    const scrollIntoView = vi.fn();
+    Object.defineProperty(HTMLElement.prototype, "scrollIntoView", {
+      configurable: true,
+      value: scrollIntoView,
+    });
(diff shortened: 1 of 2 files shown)


ohdearquant:
Parking this one: the kg-editor view-state work is not on an active lane right now. The branch `codex/kg-editor-view-state-2034` is preserved unchanged and this will be re-cut against main the next time kg-editor is worked on. Closing to keep the open-PR set to what is actually moving.

PR #2170, opened 2026-08-24
fix(comm): report mark-read outcome status

### LINKED ISSUE #1888: comm.mark_read: success envelope masks per-id failures (marked_count:0 under lock contention reads as success)
Observed 2026-08-10 during database-is-locked contention: `comm.mark_read` returned an envelope of `ok:true` / `status:success` while the payload carried `marked_count: 0, failed_count: 2`. A consumer reading only the envelope records unread messages as read; the actual verdict lives in the payload counts, so the envelope misleads by omission.

Either the envelope should reflect payload-level failure (failed_count > 0 fails the op, consistent with how `--strict` treats partial batch failure), or the verb documentation should state explicitly that the envelope is transport-only and consumers must read `marked_count`/`failed_count`. The first option is safer: mark-read is a state claim, and a claim that silently covers zero rows invites exactly this misread under contention, which is when it matters most.

### LINKED ISSUE #1939: comm.read reports ok:true while the mark-read write failed (mark_error present, read still false)
Measured on a live store under writer contention.

A `comm.read` call returned `"ok": true` while the same result carried `"mark_error": "pool failure during writer_task_begin: database is locked"` and `"read": false` in properties. The operation's envelope claims success; the state change it exists to perform did not land.

A caller checking `ok` believes the message was marked read. Only a caller who inspects `properties.read` (or notices `mark_error`) learns otherwise, and nothing in the envelope signals they should look.

Expected: `ok` must be false when the mark write fails — the presence of `mark_error` and a success envelope in one result is a contradiction. If a read-without-mark result is intentionally useful, it needs a distinct, explicit degraded shape rather than a success code.

Related context: the failure window coincided with `database is locked` at `BEGIN IMMEDIATE` on other write verbs, so the underlying lock is a known contention class; this issue is about the success reporting, which is wrong regardless of why the write failed.

ohdearquant:
## Summary

- add explicit `success` / `failed` status to each mark-read result
- classify bulk outcomes as `success`, `partial`, or `failed` from their aggregate counts
- document the result statuses in comm discovery and user guidance

## Verification

- `lock_contention_reports_failed_mark_read_statuses` holds a real SQLite writer lock and verifies honest singleton and zero-mark bulk failure results
- `bulk_read_response_reports_success_partial_and_failed_statuses` covers all aggregate status states
- `cargo fmt --all`
- `cargo clippy -p khive-pack-comm --all-targets -- -D warnings`
- `cargo test -p khive-pack-comm` (371 passed)

Fixes #1939
Fixes #1888


ohdearquant:
Reviewed at head `bf07f6923eb2cd70341aadde9e012c1f06dce2c4`. This is a genuine honesty fix and most of
it is careful. One structural point is worth addressing before it lands, because it undercuts the
field being added.

## The aggregate ignores the per-result `status` it just introduced

`bulk_read_response` derives its counts from `read`, not from `status`:

```rust
let marked_count = results
    .iter()
    .filter(|result| result["read"].as_bool() == Some(true))
    .count();
```

So the response now carries two independent representations of the same outcome — a per-result
`status` written by each producer, and an aggregate computed from `read` — with nothing tying them
together. A producer that sets one without the other, or sets them inconsistently, yields a response
that contradicts itself, and every count in it will be wrong in a way that reads as authoritative.

The new test demonstrates the gap rather than closing it:

```rust
outcomes.iter().map(|read| json!({ "read": read })).collect()
```

Those fixtures have **no `status` field at all**, and the aggregate still produces success / partial /
failed. That only works because the aggregate does not consult the field this PR exists to add. A
fixture that omits the new field cannot detect a producer that gets the new field wrong.

Deriving `marked_count` from `status == "success"` makes the per-result field the single source of
truth and the aggregate a function of it, which is what the docs now describe. If `read` should stay
authoritative instead, then assert the two agree — but pick one, because right now the field is
decorative at the aggregate level.

Related: the fixture shape is not production-shaped. Real results carry `id`, `full_id`, `status`,
`read` and `properties`; `{"read": bool}` is a shape no producer emits.

## The same honesty standard, one line up

Both the singleton and atomic paths synthesize properties when the read-back fails:

```rust
_ => {
    let mut fallback = orig_props.clone().unwrap_or_else(|| json!({}));
    fallback["read"] = json!(true);
    fallback
}
```

That `_` arm catches `Ok(None)` — the note no longer exists — and `Err(_)` — the read failed —
identically, fabricates `read: true`, and the row goes out as `status: "success"` with properties
indistinguishable from ones actually read back.

To be fair on provenance: this is pre-existing, and the diff only adds the adjacent `status` line, so
it is not a regression. But it is the same question this PR is answering, in the same function. The
write genuinely succeeded, so `status: "success"` is defensible — what is not visible anywhere is that
the *properties* are synthesized. A degradation marker on that arm, and splitting the vanished-note
case from the read-failed case, would make the response say what it knows. The existing comment at the
site shows the fabrication is deliberate, so this is about disclosing it, not removing it.

## Checked and sound

- **Empty `ids` is rejected** by `validate_bulk_read_targets`, which closes what would otherwise be a
  nasty trap: with an empty `results`, `failed_count` is `0 - 0 = 0` and the status expression would
  return `"success"` for a call that marked nothing. Unreachable as written — worth keeping a test on,
  since the guard and the status expression are in different functions and only their combination is
  safe.
- **No requested id can silently vanish from the accounting.** `validate_read_target` is called with
  `?`, so an invalid id fails the whole call rather than being dropped, which means
  `unique_count` differs from `requested_count` only by deduplication.
- **The atomic path cannot claim success on a failed commit.** `patch_note_property_atomic(...).await?`
  precedes the loop that emits `status: "success"`, so the hardcoded value is only reachable after the
  patch committed.
- **No asymmetric field presence.** The best-effort success arm pushes `mark_read_target`'s value,
  which returns `read_response`, and all three of its arms (`Ok(true)`, `Ok(false)`, `Err`) now carry
  `status`. Successful and failed rows have the same shape.
- **The lock-contention test injects a real failure** — holding an actual SQLite writer lock rather
  than mocking the error — which is the right way to test this.

## Minor

`vocab.rs` now says "Best-effort responses carry status=success|partial|failed; atomic=true provides
all-or-nothing mutation." The atomic path also returns through `bulk_read_response` and therefore also
carries an aggregate `status` (always `success`, since every row it builds is `read: true`). Worth
saying so, otherwise a caller may branch on the field's presence to detect which mode ran.

named, and the verification list in the description was not relied on.

ohdearquant:
Requesting changes, but the gap here is a contract that was never written down rather
than a coding mistake, and I think the fix is smaller than it looks.

## What the code does today

`handle_read`'s storage-error arm returns a JSON value, not an `Err`:

```rust
Err(e) => {
    json!({ "status": "failed", "read": false, "mark_error": e.to_string(), ... })
}
```

and the MCP dispatcher writes `map.insert("ok".to_string(), Value::Bool(true))`
unconditionally for any registry `Ok`. So under writer contention the public surface
returns `{"ok": true, "result": {"status": "failed", "read": false, "mark_error": ...}}`.

That is the shape #1939 and #1888 call a contradiction: an envelope that says success
wrapping a payload that says the mark did not happen.

## Why I don't think the code is wrong

There is already a precedent in this repo for exactly this shape, ratified rather than
patched. A few lines above the `ok` insertion:

```rust
// ADR-130 §1: `status` is present on every successful search envelope
// (complete or partial); absent for every other verb.
```

Search degrades by staying `ok: true` and carrying a status the caller branches on. That
is a deliberate design, not an oversight, and a failed best-effort read mark is the same
category of event: the operation was attempted, the outcome is knowable, and the caller
needs to distinguish "did not happen" from "call failed".

So the useful question is not "should this be `ok: false`" but "which contract is comm
under, and where is that written".

## What is actually missing

`docs/adr/ADR-040-communication-and-schedule-packs.md:695` still states that the
degraded-mark amendment changes no `comm.read` response field, and the arms defined at
`:697`-`702` do not include `status`. This PR adds `status` in the handler and advertises
it in `vocab.rs`. No ADR file is touched by the diff.

That leaves the normative contract contradicting the shipped shape, which is the part
that will bite a downstream client.

The change I would want, and it resolves both halves at once: amend ADR-040 to define the
degraded arm — `status`, `read`, `mark_error`, and the bulk `marked_count`/`failed_count`
— and to state explicitly that the outer `ok` stays `true` for a degraded mark, with
consumers required to branch on the payload. That converts the envelope behaviour from an
undocumented contradiction into a documented contract, and it is consistent with how the
search degradation was handled.

Worth noting so it doesn't cause a detour: ADR-130's `status` is an *envelope* field
reserved to search, while the `status` here is inside `result`. They do not collide, and
nothing about the envelope needs to change.

## One test gap regardless of which way the contract goes

`crates/khive-pack-comm/tests/mark_read_failure_status.rs` drives the raw `VerbRegistry`
and asserts `single["status"]`, `single["read"]`, `bulk["status"]`, `marked_count` and
`failed_count`. It never asserts the outer `ok` field, so it cannot observe the behaviour
that the issues are actually about, and it would keep passing if the envelope changed in
either direction.

Whichever contract is chosen, this needs a wire-level assertion that pins `ok` to the
chosen value. That is the property under discussion, so it should be the property the
test holds down.

ohdearquant:
The reframe is right and the assertion requirement from the earlier pass is now met. One thing
still outstanding before this can land.

## The wire-level assertion requirement is satisfied

`crates/khive-pack-comm/tests/mark_read_failure_status.rs` asserts through the verb surface
rather than on the handler’s return types — it dispatches `comm.read` and `comm.mark_read` and checks
the returned JSON directly:

- `single["status"] == "failed"`, `single["read"] == false`, and `mark_error` populated
- `bulk["status"] == "failed"`, `bulk["marked_count"] == 0`, `bulk["failed_count"] == 2`
- every element of `bulk["results"]` carrying `status == "failed"` and `read == false`

That is the right level. A caller reading these fields is the thing the change exists to serve,
and the test now fails if the wire shape regresses even when the Rust types still compile.

## Outstanding: the documented contract has not been updated

This changes `comm.mark_read`'s response contract by adding `status`, `marked_count`,
`failed_count`, `mark_error`, and per-element `results[].status`. The contract document that
specifies this verb —
`docs/adr/ADR-040-communication-and-schedule-packs.md` — still describes `comm.mark_read` only
as:

> Default best-effort behavior matches `comm.read(ids=...)`; `atomic=true` commits every unique
> mark in one transaction or none.

There is no outcome-status shape in that description, so a reader working from the contract will
not know these fields exist or what they promise. The diff currently touches the pack README,
the message-lifecycle API doc, and the guide, but not the document that is authoritative for
this verb's contract.

What the amendment needs to state, because these are the questions a caller will actually have:

1. The exact field set on both the single and bulk paths, and which are always present versus
   conditional.
2. Whether `status` is per-request or per-element, and how the two relate when they disagree —
   the tests pin a case where the bulk status is `failed` while individual results are also
   `failed`, but not the mixed case.
3. Whether a partial success is `failed`, or a third value, and how that interacts with the
   existing `atomic=true` all-or-nothing guarantee, which is the sentence being extended.
4. Whether `mark_error` is caller-facing text with a stable shape or a diagnostic string, since
   the difference decides whether anyone may match on it.

Point 3 is the one worth settling first: `atomic=true` currently promises "every unique mark in
one transaction or none", and an outcome-status field only carries information for the
best-effort path. Whether the new fields are meaningful under `atomic=true` should be stated
rather than inferred.

Holding here until the contract document carries this, since the code and the document would
otherwise disagree about a verb that already ships.

ohdearquant (commented):

Verdict on head bf07f6923eb2cd70341aadde9e012c1f06dce2c4: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at 5fb8b3c5.

PR #2171, opened 2026-08-24
fix(pack-session): advance oversized line skips

### LINKED ISSUE #2010: session mirror: ingest livelocks on a complete line longer than max_line_bytes plus one buffer window
### Summary

`read_bounded_chunk` never advances past a **complete** JSONL line that is longer than
`max_line_bytes` by more than one `BufReader` window. The cursor is left at the line start and
every subsequent pass repeats the same bounded read, so ingest for that file stops permanently
at that offset. Observed live: three session files stalled for days, ~325 MB unreachable, and
16,822 warn lines emitted in a single daemon log generation.

### Mechanism

`read_line_bounded` (`crates/khive-pack-session/src/mirror/ingest.rs:234`) sets `oversized` once
`total + take > max_line_bytes`, and then returns `OversizedUnterminated` at the **first**
`fill_buf` window that contains no `\n`:

```rust
if oversized {
    // Already over the cap and this fill_buf window had no `\n`:
    // stop here rather than looping onward toward EOF
    return Ok(LineRead::OversizedUnterminated { bytes: total });
}
```

`read_bounded_chunk` (`:331`) then deliberately does not advance `new_offset`, on the stated
assumption that a later pass reaches the terminator:

```
// do NOT advance new_offset past line_offset. The next call
// re-reads from the same line_offset and is bounded the
// same way — cheap and repeatable, whether the file is
// still growing (a later pass will eventually see the
// terminator and fall into the `Oversized` skip-and-advance
// arm below) ...
```

That assumption does not hold. Because the scan stops at the first oversized window without a
terminator, the `Oversized` skip-and-advance arm is reachable **only** when the terminator
happens to fall inside the same window that first crossed the cap. Re-reading from the same
offset is deterministic, so a line that is more than roughly one window past the cap can never
reach that arm, no matter how many passes run. The file does not have to be corrupt, truncated,
or growing — an ordinary complete line that is simply large is enough.

### Evidence

Measured against one of the stalled files (static, mtime four days before the reading):

| Fact | Value |
| --- | --- |
| `max_line_bytes` | 8,388,608 |
| Length of the line at the stuck offset | 9,545,257 bytes |
| Terminated? | yes — 1,351 further newlines exist after it |
| Stuck offset | 18,185,805 |
| File size | 30,220,312 |
| Control: first line of the same file | 46,741 bytes |

The warn line reports `line_bytes=8396800`, which is `max_line_bytes` plus exactly one 8 KiB
`BufReader` window — the scan stops one window past the cap rather than at the real line length
of 9.5 MB.

Three files were stalled simultaneously. Their offsets are byte-identical across every retry
cycle, which is what distinguishes a livelock from slow progress:

| Stuck offset | File size | Unreachable |
| --- | --- | --- |
| 57,810,646 | 355,621,547 | 83.7% (~298 MB) |
| 38,178,962 | 53,607,365 | 28.8% (~15 MB) |
| 18,185,805 | 30,220,312 | 39.8% (~12 MB) |

Retry cost: 16,822 occurrences of the warn in one rotated log generation and 156 more in the
following one within ~78 minutes, roughly one cycle per file per minute. The volume is a
material contributor to log rotation.

### Impact

1. **Silent data loss at the tail.** Everything after the oversized line in an affected file is
   never ingested. The mirror reports no error state — only a repeating warn — so a reader of
   the mirrored data sees a truncated session with no marker saying where or why it stopped.
2. **Unbounded retry.** The blocking condition is invariant, so "bounded retry" is bounded per
   pass but unbounded in total.
3. **Recovery does not help.** A daemon restart re-reads the same offset and stalls identically.

### Suggested direction

The `PACKSESSION-AUD-003` bound (never buffer or read more than `max_line_bytes` in one call)
should be preserved; what is missing is forward progress across calls. One shape that keeps both:
persist a *mid-skip* state for the file — once a line is known oversized, later passes discard
from where the previous pass stopped rather than from the line start, still reading at most
`max_line_bytes` per pass, until the terminator is found, then resume normal parsing. Skipping
remains explicit and logged, and the per-pass memory and I/O bound is unchanged.

Whatever shape is chosen, the invariant worth testing directly is that a complete line of
`max_line_bytes * N` bytes is eventually skipped and the file's remaining lines are ingested.
The current test suite passes with the defect present, so a fixture with a line several windows
past the cap is the load-bearing addition. `crates/khive-pack-session/docs/api/mirror-ingest.md`
documents the current behaviour and needs the same amendment, including the code comment quoted
above, which states the assumption that does not hold.

ohdearquant:
## Summary

- checkpoint bounded discarded prefixes while skipping oversized JSONL lines
- resume persisted mid-line skips after restart without parsing line fragments
- add multi-window forward-progress coverage and update the ingest contract guide

## Testing

- `cargo fmt -p khive-pack-session -- --check`
- `cargo clippy -p khive-pack-session --all-targets -- -D warnings`
- `cargo test -p khive-pack-session`

Fixes #2010


ohdearquant:
Rebased onto current `main`, which carries the amended ADR-080. The cursor
contract objection against this branch is resolved by that amendment: a
persisted mid-line offset is now the specified behaviour rather than a departure
from the record.

The amendment also states a second requirement this branch does not yet
implement, and without it the wedge is only moved rather than removed.

ADR-080 §7 says an advance over a line already known to exceed the cap **claims
the path** and commits regardless of a sibling candidate's error. The reasoning
given there is that a checkpoint pass advances the cursor while inserting zero
rows, which is precisely the shape the empty-advance rule refuses to commit when
any candidate errored — so without the carve-out the checkpoint never persists.

That carve-out is not present. `crates/khive-pack-session/src/mirror/service.rs:1216-1224`
treats every zero-row advance identically: when `had_errors` is true it returns
`None` instead of calling `commit_empty_advance`, and the deferred path it
dispatches through is at `:1441-1458`. Nothing distinguishes an over-cap claim
from an ordinary blank-span advance.

Concrete failure: a complete over-cap line whose terminator lies several reader
windows past the cap, in an overlapping-root candidate set where a sibling
candidate errors on every pass. `ingest.rs:354-365` correctly returns
`inserted: 0, new_offset > start_offset`, but dispatch discards it, so the
persisted cursor advances by zero on every poll. The line stays the frontier and
every later record is unreachable — the same wedge, reached through the dispatch
layer instead of the read layer.

Note that `service.rs` is not in this branch's diff, so the change is incomplete
where it matters most, and no test here exercises the dispatch path with an
erroring sibling.

Suggested shape: carry "this advance is an oversized-line claim" through the
dispatch result, commit it even when `had_errors` is true, and keep the existing
error-vetoed commit for ordinary blank or unparseable empty advances. A
service-level test with an overlapping-root dispatch plus an erroring sibling,
asserting the *persisted* cursor strictly increases, would pin it — assertions on
the returned stats alone cannot, since the returned stats are already correct
today.

ohdearquant (commented):

Verdict on head 40ca60fb2f65e102278041bf1659ad0e65b25449: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at 82c390bd.

PR #2177, opened 2026-08-24
fix(db): make schema-terminal assumptions version-agnostic

ohdearquant:
## Problem

Several sites assumed V21 (`attachments_first_class`) is the last migration in the chain. Each
breaks the moment a V22 lands:

- `attachment_cutover_status` treated a `complete` marker beside a ledger above V21 as an
  impossible pair. Only a ledger **below** V21 is one: later migrations record on top of a
  completed cutover in the normal course.
- `validate_schema_is_current` ran the physical cutover-state check only when the ledger was
  exactly 21, so a later version would skip cutover validation entirely.
- Test fixtures pinned terminal-version literals (21, and 22 for the ahead-of-build fixture),
  so the ahead-of-build fixture stops being ahead the day a real V22 lands on that number.

Separately, the blob-GC cutover fence read only `MAX(version) = 21` plus the named V21 row. A
ledger missing rows below V21 satisfies both while carrying an incomplete migration history
this gate never validated.

## Change

Two directions, deliberately opposite, because the two predicates answer different questions.

**Relaxed, where the assumption was that V21 is terminal.** The ledger predicates in
`attachment_cutover_status` and `validate_schema_is_current` become `>= 21`. The physical
facts the reject matrix removes one at a time continue to carry the safety, and the
accept-path tests running at the latest schema are the positive control for the semantics.
Widening `validate_schema_is_current` fails *closed*: at V22+ it now runs a cutover check it
previously skipped.

New `migrations::latest_schema_version()` derives the terminal version from the migration
chain. Terminal-version assertions and the ahead-of-build fixture use it instead of literals,
so they stay correct when a migration is added.

**Tightened, where destructive work is admitted.** ADR-160 Phase 4a admits report-only and
destructive blob-GC sweeps only for an exact completed V21 epoch; ahead-of-V21 epochs are
unsupported. The fence now enforces that as the whole canonical ledger rather than a terminal
row: `version` is the table's PRIMARY KEY, so `COUNT(*) = 21 AND MIN = 1 AND MAX = 21` holds
if and only if the ledger is exactly the contiguous set {1..21}. The V21 literal became a
bound parameter in the same statement.

## Behaviour change at the current schema head

One, and it is the point of the blob.rs half: a database whose ledger is missing rows below
V21 while retaining a named V21 row at `MAX(version) = 21` previously passed the GC fence and
now fails closed. `gate_rejects_incomplete_ledger_behind_v21` is that case, and it asserts the
two facts the old predicate read are still intact so the rejection can only come from the new
contiguity clause.

Every other predicate in this change evaluates identically while V21 is the latest migration.

## Testing

- `cargo test -p khive-db` — all suites green (lib 817 passed).
- `cargo test -p kkernel` — all suites green (lib 444 passed, integration suites included).
- `cargo fmt --check` and `cargo clippy --all-targets --all-features -p khive-db -p kkernel` clean.

### DIFF
--- a/crates/khive-db/src/migrations.rs
+++ b/crates/khive-db/src/migrations.rs
@@ -1007,10 +1007,7 @@ pub fn inspect_schema_is_current(path: &std::path::Path) -> Result<u32, SqliteEr
 /// exact match performs no writes.
 pub fn validate_schema_is_current(conn: &Connection) -> Result<u32, SqliteError> {
     let current_version = read_schema_version(conn)?;
-    let latest_version = MIGRATIONS
-        .last()
-        .map(|migration| migration.version)
-        .unwrap_or(0);
+    let latest_version = latest_schema_version();
 
     if current_version < latest_version {
         return Err(SqliteError::InvalidData(format!(
@@ -1206,7 +1203,7 @@ fn run_migrations_locked(conn: &mut Connection) -> Result<u32, SqliteError> {
     // pre-consolidation V2..V22 ledger — or was written by a newer build. Either
     // way the baseline schema would be silently skipped, leaving the process on a
     // stale schema. Fail loudly instead of corrupting silently.
-    let latest_version = MIGRATIONS.last().map(|m| m.version).unwrap_or(0);
+    let latest_version = latest_schema_version();
     if current_version > latest_version {
         return Err(SqliteError::InvalidData(format!(
             "database schema version {current_version} is ahead of the latest known migration \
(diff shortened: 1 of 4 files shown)


ohdearquant (commented):

Verdict on head 3ecd3eddad484f376591a8b57cad7a2bd66c1e90: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 94ea1780864f5747c224c094cc166a79b950439b: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 389435c564f6e26e7f01a298df6aff85ee599028: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2193, opened 2026-08-24
docs(adr-045): Amendment 3 — caller-owned property strings survive agent-mode empty-drop

ohdearquant:
Amendment 3 to ADR-045: in Agent mode, empty strings nested under a record's `properties` object are preserved; the empty-string drop continues to apply everywhere else.

A key under `properties` exists only because a caller wrote it, so an empty string there is data — it distinguishes "set to empty" from "absent or deleted". Under the unamended rule, the echo of an update that sets a property to `""` omits the key entirely and a successful write reads as a deletion (#1995). The token-economy rationale is unaffected: nothing machine-generated is reintroduced.

Scope stated explicitly: empty arrays and objects under `properties` stay dropped; callers distinguishing those cases read `format=json` or `presentation=verbose`. Extending the carve-out to container values is a separate decision.

Companion code change: #2164.


ohdearquant (commented):

Verdict on head 05de7ea1a15676c5e0a855d92d687fc02658cc06: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head d808d349df1b4e2867448ea0482ec8039af93e68: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2205, opened 2026-08-24
fix(runtime): classify secret-gate source notation

### LINKED ISSUE #1988: secret gate: dense LaTeX trips high-entropy-token; published RFC test vectors trip hex-credential-token
Two distinct false-positive classes were measured during a large mathematics/cryptography corpus ingest through `knowledge.upsert_atoms`. Both refusals come from the write-path secret gate (`crates/khive-runtime/src/secret_gate.rs`). They fail in different places and want different fixes, so they are reported together but should be evaluated separately.

## Class 1 — `high-entropy-token` fires on dense LaTeX

Measured: 18 of the first 19,738 atoms attempted were refused, all by this one detector, zero other classes. 17 of the 18 sit in the mathematics portion of the corpus — the region with the densest LaTeX bodies. Every masked excerpt in the refusal messages is a LaTeX construct: `\operatorname`, `\lambda`, `H_0^1(`, `K_N(\t`, `\tau_{`, `\exp\l`, `$$d_H^`, `$(\mat`, `\|u_n\`, `[\omeg`.

Mechanism (read at source): a long whitespace-delimited LaTeX fragment has the character-class profile of a credential token — mixed case, backslashes, braces, sub/superscripts, no natural-language whitespace rhythm — so its Shannon entropy clears `ENTROPY_THRESHOLD`. The trigger-word requirement is satisfied *by the mathematics itself*: attention "key", "token", probability "seed", etc. are ordinary vocabulary in these texts. The `is_structured_identifier` exemption cannot rescue these because it is deliberately scoped to `!near_trigger`.

Suggested direction: a LaTeX-fragment shape exemption analogous to the existing file-path exemption — recognized LaTeX shape (backslash-command density, brace/caret structure) exempts the token *unless* `has_clause_credential_label` attaches an immediate credential label to it. Any such exemption needs the module's usual two-sided tests, including an adversarial control that a real credential embedded inside LaTeX-shaped text is still blocked (e.g. a 40-hex run wrapped in `\texttt{...}` beside a credential label).

## Class 2 — `hex-credential-token` fires on published RFC test vectors

A documentation atom describing EdDSA secret-material handling quotes RFC 8032's published Ed25519ph test vectors in a labelled code block, beside the expected public key and an assertion that key derivation reproduces it. The detector fires on the 32-byte hex values near the (inevitable) trigger words.

This class is different in kind: the detector is **right about the shape** — these are real 32-byte hex values in credential context — and **wrong about the secrecy**. They are internationally published constants whose entire purpose is to appear in documentation. A cryptography corpus hits these constantly (RFC vectors, NIST CAVP sets, BIP-39 test mnemonics); quoting them verbatim is the canonical way such documentation is written.

If exempted at all, the only sound shape is an **exact-match allowlist of the published constants themselves** — never a shape or context exemption. A byte-equal match against a published vector cannot leak anything (the value is public by construction), and anything not byte-equal still refuses, so the failure direction is preserved. A context exemption ("looks like documentation", "inside a code block") would silence the detector for exactly the input it exists to catch.

Observed rate for this class is roughly 1 in 6,000 atoms on this corpus, which is low enough that the alternative — leave the detector as-is and let such refusals get a human read — is also defensible. Filing so the trade-off is decided deliberately rather than by default.

## Scope note

Per the module doc, this gate is accidental-persistence hygiene on a single-principal host, not defense against a same-uid adversary. The false-positive cost therefore lands entirely on legitimate documentation/mathematics corpora — which is exactly the traffic measured here.

### LINKED ISSUE #2056: secret gate: refusal excerpt names a non-hex word for hex-credential-token, and pattern-name vocabulary self-triggers refusals
## Reproducer (current main, comm.send path; also reproduces via kkernel exec on a scratch store)

Writing this content is refused:

```
The hex-credential-token bucket held 18. Six are two values verbatim:
  0af7651916cd43dd8448eb211c80319c -- spec example identifier
```

Refusal, verbatim:

```
write blocked: content matches secret pattern hex-credential-token at masked excerpt verbat...8chars. If this is a real credential, remove it before writing. If it is not (e.g. a file path, UUID, or hash that happens to sit near a word like key/secret/auth/token), reword so the token is not glued directly next to that word — e.g. move it to its own line, or insert a space or word between them.
```

Control: the identical content with the first sentence removed (no occurrence of the pattern name, same hex value on its own line) is accepted. A second control with a credential-shaped value (`aws_secret_access_key=...`) is refused as expected, so the gate is live on the probe path in both arms.

The hex value is the W3C Trace Context specification's published example trace-id, not a credential.

## Defect 1: excerpt misattribution

The masked excerpt is `verbat...8chars`: an 8-character span beginning `verbat`, i.e. the English word `verbatim`. That word contains `v`, `r`, `t`, `i`, `m` — it cannot match a hex token pattern, so the excerpt does not name the span the pattern actually matched. The plausible hex candidate sits on the following line. Likely mechanisms: match offsets computed against a whitespace-normalized buffer but rendered against the raw content, or the trigger-word adjacency tokenizer selecting the wrong neighbor as the reported credential span. Either way the operator-facing advice ("move it to its own line") points at the wrong token — and in this reproducer the value already IS on its own line.

## Defect 2: pattern-name vocabulary self-triggers

The trigger vocabulary (`token`, here inside the pattern's own name `hex-credential-token`) sits roughly 45 characters before the hex value, across a sentence boundary and a newline. Consequence: any text that quotes a prior refusal, classifies refusal reasons, or documents the gate's patterns supplies the trigger vocabulary by construction, and is then refused whenever it also carries any 32-hex identifier — including spec-published example values. Reports about false positives become unwritable through the same gate that produced them, which suppresses exactly the feedback that would surface FP classes.

ADR-115's exact-content manifest covers the published-constant class (the example trace-ids), but neither the excerpt misattribution nor the vocabulary self-trigger is addressed by a manifest: both reproduce with values that would never be enrolled.

## Suggested directions

- Render the excerpt from the same buffer and offsets the matcher used, or re-derive the span against raw content before masking; assert in tests that the reported span actually matches the pattern's own character class.
- Bound the adjacency window at sentence or line boundaries, or exclude trigger words that are part of the gate's own pattern vocabulary when they appear in hyphenated pattern-name form.

### LINKED ISSUE #2076: secret_gate: high-entropy-token fires on boundary-anchored prose compounds while identical substrings inside camelCase identifiers pass
## Observed

During knowledge-note writes, two of six notes in one batch were refused with `write blocked: content matches secret pattern high-entropy-token`. The flagged content in both was a public repository-at-commit source citation with a file path — the shape every research-derived note carries. Nothing credential-shaped was present.

## Discriminator (measured, with in-batch controls)

- **Token length is not the trigger**: a note carrying the identical 73-character citation applied cleanly in the same batch; another passed carrying an 89-character one.
- **Token content is not the trigger**: same evidence.
- **Raw distance is not the rule**: one note passed with a matching keyword five characters from a long token while a refused one was blocked at nine.
- **The separating variable is boundary anchoring**: refused notes carried the matching keyword in hyphen- or word-delimited form; every note that passed carries the same substrings only inside larger camelCase identifiers, where no word boundary precedes them. Rewording only the boundary-anchored occurrences — leaving the embedded ones and the citations untouched — made both writes pass on the next attempt. Positive confirmation.

## Why this is inverted for the content the gate sees most

Research prose describes mechanisms in hyphenated English and cites code in camelCase. As implemented, the boundary rule treats the hyphenated English compound as a credential label and the camelCase identifier as not-a-label — flagging the prose and passing the code reference.

## Second defect: the remedy text cannot cure

The refusal's own guidance says to move the token onto its own line or insert a word. Following it exactly reproduced the refusal byte-identically, because the separation was already two characters — the guidance is not derived from the predicate that actually fired. Remedy text should describe an action that changes the predicate's outcome.

## Related

#1988 tracks false positives from dense LaTeX and RFC test vectors — same pattern family, different discriminator. This issue is specifically the word-boundary predicate and the mismatched remedy guidance.

ohdearquant:
## Summary

- Bound entropy-trigger context to the candidate sentence and report the exact contributing hex span, so refusal excerpts identify the actual candidate (#2056).
- Recognize canonical repository revision URLs and exact HTML commit hrefs while keeping direct credential labels fail-closed (#2076).
- Exempt structurally recognizable dense LaTeX only when it contains no credential-shaped inner run; published secret-shaped RFC vectors remain intentionally fail-closed (#1988).
- Make remediation guidance name sentence/paragraph boundaries that actually change the predicate, and document the resulting security contract.

## Verification

- RED: the new #2076 repository-reference/guidance cases and #1988 dense-LaTeX case failed before implementation; their must-block controls passed.
- GREEN: `cargo test --manifest-path crates/Cargo.toml -p khive-runtime --lib issue_1988 -- --nocapture` — 3 passed.
- `cargo test --manifest-path crates/Cargo.toml -p khive-runtime --lib issue_2056 -- --nocapture` — 2 passed.
- `cargo test --manifest-path crates/Cargo.toml -p khive-runtime --lib issue_2076 -- --nocapture` — 3 passed.
- `cargo test --manifest-path crates/Cargo.toml -p khive-runtime --lib` — 1,415 passed, 7 ignored.
- `cargo clippy --manifest-path crates/Cargo.toml -p khive-runtime --all-targets -- -D warnings`
- `cargo fmt --manifest-path crates/Cargo.toml --all`
- `git diff --check`

## Security boundaries

Known-provider prefixes and inline credential assignments are unchanged. Repository references still block behind a direct credential label. LaTeX only exempts recognizable notation without credential-length hex or independently high-entropy alphanumeric runs. Exact public test-vector values do not receive a publication-based bypass.

Closes #2056
Closes #2076
Closes #1988


ohdearquant:
CI is red at this head on a single root: `khive-pack-git::acceptance
ingest_masks_pr_body_hash_near_token_without_dropping_note`
(`crates/khive-pack-git/tests/acceptance.rs:1600`). The other two red checks are
downstream of it — the aggregate gate, and the coverage job, whose `Compute
coverage` step failed so its ratchet step never ran and produced no figure.

The failure looks like a genuine behaviour change rather than flake. Trigger
proximity is now evaluated within sentence bounds:

```rust
let near_trigger = contains_trigger(after_last_sentence_boundary(&window[..raw_start]))
    || contains_trigger(before_first_sentence_boundary(&window[raw_end..]));
```

The test fixture is:

```
Rotated the deploy token. Old hash was <64 hex chars> before rotation.
```

The trigger word `token` is in the first sentence and the hex run is in the
second, so the sentence break severs them, the run is no longer treated as
near a trigger, and the raw 64-hex value reaches stored content unmasked.

The boundary rules consider a period followed by an alphanumeric byte
intra-sentence, which keeps `v1.2` intact, but a label and its value can
straddle a sentence break — `... token. Old hash was <value>` is a common
shape in release and rotation notes. Narrowing trigger scope trades a false
positive for a false negative, and for a masking backstop the false negative
is the costly direction, so the assertion looks correct as written.

Suggested direction: keep the intra-token span refinement, which is the part
that improves reported match accuracy, and treat the sentence-boundary
narrowing of `near_trigger` separately — a label-to-value association that
survives one sentence break would keep this case masked without reintroducing
the broad windows the change is trying to tighten.

ohdearquant:
Six red boxes on this head, but they reduce to two independent roots — worth separating before
anyone chases the wrong one.

**Not signals:** the two cancelled shards (`ubuntu 1/2`, `macos 1/2`) completed 17 and 23 seconds
after `ubuntu 2/2` failed, so they are fail-fast siblings, not separate failures. `CI gate` started
after everything else finished and is a pure aggregator. And `Coverage ratchet` is red for a reason
that is not coverage: its `Compute coverage` step failed and `Check coverage does not regress` was
**skipped**, so no coverage was ever measured. That box is named for a judgment it did not make; it
will clear on its own once the build does.

**Root 1 — a real regression, and it is on this change's own subject.** `ubuntu 2/2` ran 1867 tests
and exactly one failed: `khive-pack-git::acceptance
ingest_masks_pr_body_hash_near_token_without_dropping_note`, panicking at
`crates/khive-pack-git/tests/acceptance.rs:1600` with

```
raw 64-hex hash must not survive into stored content:
"Rotated the deploy token. Old hash was <64 hex chars> before rotation."
```

That is the relaxation landing wider than intended. The fixture is a sentence containing the word
`token` and a 64-character hex value — a length that is in the credential-length set — and after this
diff the value survives into stored content unmasked. The test is pre-existing and was passing, so
this is not a fixture that needs updating to match new behaviour; it is the case that says the gate
must still mask a credential-shaped hex value when a credential label sits beside it. Whatever
narrowing makes accurate matches accurate needs to leave that pairing flagged.

Worth checking alongside this: another open change to `secret_gate.rs` also relaxes the same gate for
UUID-shaped values near token prose. Each relaxation is defensible read alone, and their combination
is not obviously the union of two safe changes — whichever lands second should be re-validated
against the merged result rather than against the base it was written on, because both are widening
the same exemption surface.

**Root 2 — the secret scan, which is almost certainly a fixture rather than a real credential.** The
job reports `leaks found: 1` and nothing more; CI runs `gitleaks detect --source . --redact
--no-banner` (`.github/workflows/ci.yml:385`) with `fetch-depth: 0` and there is no
`.gitleaks.toml`, so it scans full history under default rules and redacts its own output. The log
therefore names nothing.

What the tree shows, though, is quite specific. `secret_gate.rs` on `main` contains 52 hex runs of 32
characters or more and passes this scan, so hex length alone is not what trips it. What `main` has
**zero** of — and what the newest commit introduces exactly one of — is a binding whose *name* is a
credential keyword holding a long hex literal: `let credential = "<40 hex chars>";` at
`crates/khive-runtime/src/secret_gate.rs:3338`. One new occurrence, one reported leak. Default
generic-secret rules key on precisely that identifier-next-to-high-entropy-value shape, which is also
the shape this crate's own gate is built to catch, so the fixture is doing to gitleaks exactly what
`secret_gate.rs` exists to do to callers.

I could not run gitleaks here to confirm it flagged that line, and the redacted log does not say, so
treat the specific line as a strongly-evidenced inference rather than a confirmed root — the counts
match and the baseline discriminates, but the tool has not been asked.

Scope, if it is that line: it appears only in the newest commit. The preceding commit's copy of the
file has zero such bindings, `main` has zero. So amending that commit clears it from history;
no rewrite of earlier commits is needed. Renaming the binding to something that is not a credential
keyword is the smallest fix and costs the test nothing, since the value's role in the test comes from
its shape rather than its variable name. An allowlist entry would also work but spends a permanent
exception on a temporary problem.

ohdearquant:
This branch's head is currently turning the secret scan red on **every other open pull request in the repository**, not just this one. The mechanism is worth stating carefully because the finding itself is benign and the blast radius is not.

**What gitleaks is matching.** A single finding, `generic-api-key`:

```
f1c8e732e61b9ec2516b794dbb37afd5152cee59:crates/khive-runtime/src/secret_gate.rs:generic-api-key:3338
```

That line is a test fixture — a 40-hex-character constant bound to a `credential`-named local, used to exercise the secret gate's own detector. It is not a credential and nothing needs rotating. Reproduced locally with the same pinned version CI installs (`v8.21.2`), against a range containing only this branch's commit, with a must-match positive control in the same invocation and a commit-count assertion so an empty result could not read as clean.

**Why it reaches other pull requests.** `.github/workflows/ci.yml` checks out with `fetch-depth: 0` and then runs:

```
gitleaks detect --source . --redact --no-banner
```

With no `--log-opts`, the scan population is every commit reachable from every ref in the checkout — the whole repository, not the change under review. A `fetch-depth: 0` checkout carries all origin branches, so any commit pushed to any branch is scanned by every subsequent CI run on every pull request.

The timeline matches that exactly. This commit was pushed at `03:31:12Z`:

| secret scan started | result |
|---|---|
| 03:28:38Z | success |
| 03:36:03Z | failure |
| 03:39:34Z | failure |
| 04:00:57Z | failure |

Every scan that began before the push is green; every scan after it is red, across changes that share no files with this one. Runs on `main` are still green only because `main`'s most recent commit predates the push by two minutes — the same job runs on `push: [main]` and on the nightly schedule, so the next merge inherits it.

**Two independent things to fix.**

1. *This branch.* The constant needs a shape the `generic-api-key` rule does not match — built at runtime from parts, shortened below the entropy/length threshold, or moved behind a helper — rather than a `.gitleaksignore` entry. `.gitleaksignore` fingerprints are `commit:path:rule:line`, and the file's own comments record that this exact allowlist has already been re-pinned once when a later commit re-attributed the same two fixtures. A fingerprint pinned to a branch commit stops matching the moment the branch is rebased or squash-merged, so an allowlist entry here would need re-pinning at merge and would silently expire.

2. *The scan's scope.* A secret scan on a pull request should judge the commits under review. Passing `--log-opts` for the merge-base range would make each run's result a property of its own change; today a red scan on any pull request may be reporting a commit from an unrelated branch, with no way to tell from the log.

Worth adding `-v` at the same time. The current invocation prints only `leaks found: N` with no file, line, commit or rule, so every consumer of a red scan has to reproduce locally before learning anything — which is the only reason the cross-branch coupling above was not obvious immediately.

There is also a structural tension here that will keep recurring: the secret gate's tests must contain secret-shaped strings to test the detector, so this file is inherently adversarial to a repo-wide credential scanner. A scoped scan plus a deliberate fixture convention for this one module would settle it more durably than per-finding allowlisting.

ohdearquant:
## The red `CI gate` is subject-matter adjacent, not ambient

`CI (ubuntu-latest, shard 2/2)` fails one test:

```
FAIL khive-pack-git::acceptance ingest_masks_pr_body_hash_near_token_without_dropping_note
panicked at crates/khive-pack-git/tests/acceptance.rs:1600:5:
raw 64-hex hash must not survive into stored content: "Rotated the deploy token.
  Old hash was <64-hex fixture> before rotation."
```

`Coverage ratchet` also fails at `Compute coverage`, downstream of the same run.

This branch touches no file under `crates/khive-pack-git/`, which on several other branches
today has been a reliable sign that a failure is ambient. It is not that here, and the
distinction matters: that acceptance test exercises the masking behaviour this branch changes.
A 64-hex hash sitting near the word `token` is precisely the value-shaped-versus-label-shaped
case that source-notation classification decides, so an untouched crate says nothing when the
behaviour under test is the behaviour being reclassified.

Reading the assertion as a specification: the fixture wants the hash masked while the note is
still stored. The failure says the raw hash reached stored content, so the classification change
appears to have moved this input from masked to passed-through. Worth confirming against the
new notation rules directly rather than re-triggering, since a fresh run will reproduce it.

ohdearquant:
## This `Secret scan` red is real and specific to the change

The branch now carries the scoped scan, and the scope resolved correctly:

```
range="2e59ba274120389cf76f04a9cb2a97b9c841262a..370c6d6864e1fb04641770afe83f0cc5b05b931e"
2 commits scanned.
leaks found: 1
```

So this is not the unscoped whole-history form that was reporting the same ambient finding on
unrelated branches — it judged two commits, both of them this change's.

```
RuleID:  generic-api-key
File:    crates/khive-runtime/src/secret_gate.rs
Line:    3338
Finding: let credential = "REDACTED"
```

A literal on the right-hand side of `let credential = ...` matches `generic-api-key` on shape alone,
so a synthetic fixture trips it exactly like a real key would.

**Suggested direction: build the fixture at runtime rather than allowlisting it.** Something like
`format!("{}{}", "AKIA", "0123456789ABCDEF")`, or assembling it from bytes, keeps the test's
intent and leaves no literal in the tree for a scanner — or a reader — to match. That matters
independently of the check: this repository is public, and a committed string that reads as a
credential is a publication surface whether or not the value was ever live.

If you would rather allowlist it, one thing to know first: the fingerprint gitleaks emits is
commit-pinned —

```
f1c8e732e61b9ec2516b794dbb37afd5152cee59:crates/khive-runtime/src/secret_gate.rs:generic-api-key:3338
```

— so any amend, rebase, or squash of that commit changes the fingerprint and the entry stops
matching. It fails toward red rather than green, which is the safe direction, but it will look like
a new finding rather than a stale allowlist entry, and the line number pins it further.

ohdearquant:
## The sentence trim breaks label-to-value association, and an acceptance test already proves it

Narrowing trigger context to matched spans is the right idea. The sentence trimming goes one step
further than that, and it converts a masked value into a persisted raw one.

At this head both sides of the trigger context are trimmed to the current sentence:

```rust
// crates/khive-runtime/src/secret_gate.rs:969-971
let near_trigger = contains_trigger(after_last_sentence_boundary(&window[..raw_start]))
    || contains_trigger(before_first_sentence_boundary(&window[raw_end..]))
    || has_inline_credential_trigger(raw_token);
```

`origin/main` reads the whole window instead:

```rust
let near_trigger = contains_trigger(&window[..raw_start])
    || contains_trigger(&window[raw_end..])
    || has_inline_credential_trigger(raw_token);
```

Downstream there is no second chance. An off-trigger pure-hex token is allowlisted outright at
`:997-999` (`if !near_trigger && is_pure_hex(token) { continue; }`), and `collect_mask_spans` then
copies the value through unchanged, so the predicate is load-bearing for the whole masking call
rather than just for entry into it.

The existing fixture `"Rotated the deploy token. Old hash was <64 hex> before rotation."` puts the
label and the value on opposite sides of a period, which is precisely the shape the trim discards.
This is not hypothetical: the macos shard fails
`ingest_masks_pr_body_hash_near_token_without_dropping_note`, panicking at
`crates/khive-pack-git/tests/acceptance.rs:1600`, with 92 passed and 1 failed. The production path
that reaches it is `crates/khive-pack-git/src/ingest.rs:2251-2252`, where a fetched body is passed
through `mask_secrets` before persistence.

The asymmetry is what makes this worth holding on. A false positive masks something harmless and is
fixed by rewording. A false negative writes credential-shaped content into durable storage, silently,
and nothing downstream re-checks it.

The new local test establishes that a detector *name* should not cross a sentence boundary. It does
not establish that a credential label and its value may safely cross one, and the acceptance test
asserts the opposite.

**Suggested direction:** keep the matched-span improvement, but do not let sentence scoping break
label-to-value association in the masking predicate. Either exempt the credential-label lookup from
the trim, or keep a separate fail-closed path for credential-shaped hex so that trimming can only
ever add coverage, never remove it.

For transparency on how this was produced: an automated pass over this diff ran on a model from the
same family as the one that authored it, so its output was treated as input evidence only. The three
load-bearing facts here — the trimmed predicate at this head, the untrimmed base version, and the
named acceptance-test failure in the macos shard log — were each re-derived directly before posting.

ohdearquant (commented):

Verdict on head 08be572a1339d4cde9e80fb2bc0b75b82c2a6a6d: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Main is merged in and the branch is mergeable again. Two conflicts, both in the
entropy trigger path, resolved toward main's newer structure with this branch's
intent folded in rather than dropped:

- `check_entropy_heuristic` now calls main's `entropy_trigger`, which already
  carries the inline-credential check and adds the line-local rule for a bare
  Git-length value. The sentence-boundary narrowing this branch introduced moved
  inside that helper, so trigger context still stops at a sentence boundary. The
  preceding-label fallback stays exempt, since that line is already required to
  end in an assignment delimiter.
- The allowlist paragraph keeps both rules rather than either: the path, revision
  and LaTeX exceptions from this branch, and the line-local trigger rule for bare
  Git-length hex from main.

One test is red at this head and is deliberately not papered over:
`forty_hex_bridge_fragments_keep_cross_line_detection_and_full_masking` leaves
the second bridge fragment in the masked output. Three configurations were
measured, which narrows it usefully:

| configuration | result |
| --- | --- |
| as pushed | only the bridge test fails (332 passed, 1 failed) |
| sentence-boundary narrowing removed | bridge test **and** `issue_2056_detector_name_does_not_trigger_across_sentence_boundary` fail |
| step 6 redaction target switched back to the whole token | bridge test **and** `issue_2056_hex_bridge_masks_the_matched_candidate` fail |

So the narrowing is load-bearing and is not the cause, and step 6's return shape
is not the cause either: those are the two changes a reader would suspect first,
and each one only adds a second failure when reverted. The failure direction
matters here, since a surviving fragment in masked output is a leak rather than a
cosmetic difference, so this stays open until the cause is named.

PR #2211, opened 2026-08-25
fix(runtime): preserve empty strings under properties in Agent presentation

### LINKED ISSUE #1995: kg: update response echo omits property keys set to empty string, reading as deleted

ohdearquant:
Agent presentation dropped every empty string unconditionally, so an update setting a property to `""` echoed back without the key — a successful write read as a deletion (#1995).

A key under a record's `properties` object exists only because a caller wrote it, so `""` there is data: it distinguishes set-to-empty from absent. This guards the empty-string drop with the existing `inside_properties` context, at any nesting depth under `properties`. Empty arrays and objects under `properties` are still dropped, per ADR-045 Amendment 3's scope note (#2193).

Tests: preservation at two nesting depths, plus controls that the drop still applies outside `properties` and to empty containers within it.

Fixes #1995.

### DIFF
--- a/crates/khive-runtime/src/presentation.rs
+++ b/crates/khive-runtime/src/presentation.rs
@@ -687,6 +687,12 @@ fn transform_field_agent(
         {
             Some(value)
         }
+        // Preserve empty strings under a record's `properties` object: those
+        // keys exist only because a caller wrote them, so `""` there is data
+        // (set-to-empty vs absent, ADR-045 Amendment 3, issue #1995). Empty
+        // arrays and objects under `properties` are still dropped per the
+        // amendment's scope note.
+        Value::String(s) if s.is_empty() && context.inside_properties => Some(value),
         // Drop other empty strings, arrays, objects.
         Value::String(s) if s.is_empty() => None,
         Value::Array(a) if a.is_empty() => None,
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

Verdict on head 5aa8bb3c2e6a64861ad58f962be6f41f4a482276: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 9d516a568b0ccb9e6bd195bd29a55937cf7a77b1: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2221, opened 2026-08-25
fix(runtime): enforce authority and mode at the visibility and blob install seams

ohdearquant:
Two enforcement gaps at public runtime seams, each now held at the seam itself:

- **`authorize_with_visibility`**: the gate was consulted once, for the primary (write) namespace, and the minted token then carried every caller-supplied extra namespace unchecked. Each extra namespace grants read visibility, so each one now takes its own gate check before it may enter the minted set. Any deny or gate error refuses the whole mint and names the offending namespace, fail-closed.

- **`install_blob_store`**: a store installed after boot bypassed the runtime's declared access mode — boot-time resolution wraps stores for read-only runtimes, but the post-boot install seam accepted arbitrary writable stores. Incoming stores are now wrapped on read-only runtimes: every physical mutator refuses, the bounded read surface stays available. Wrapping an already-wrapped store is harmless.

Regression tests carry their own controls: the visibility test performs the same mint without the denied namespace (success proves the refusal comes from the per-extra check); the blob test seeds an object through the raw store first (proving the store is writable, so the refusal can only come from the seam).

Gates: khive-runtime lib suite 1332 passed / 0 failed; clippy clean; downstream crates compile.

### DIFF
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -2616,27 +2616,27 @@ async fn prepare_configured_storage_topology(
         // selection required for verified migration must fail without leaving
         // a new incomplete marker. The blob pack's backend mode governs
         // read-only wrapping during serving boot.
-        let blob_runtime_read_only = if base_config.packs.iter().any(|pack| pack == "blob") {
+        let blob_governing_backend = if base_config.packs.iter().any(|pack| pack == "blob") {
             match khive_cfg.packs.get("blob") {
                 Some(pack) => backends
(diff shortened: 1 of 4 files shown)


ohdearquant (commented):

Verdict on head 98d43546cd463dfd998bf78d33bd9392649b4d42: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

Issue #1836, opened 2026-08-09
MCP stdio server processes accumulate, each holding a live WAL connection for its lifetime


ohdearquant:
## Summary

MCP stdio server processes (`kkernel mcp`) accumulate on a long-running host. Each retains an open SQLite connection to the database in WAL mode for its entire lifetime, so the population of WAL participants grows without bound as sessions come and go.

This matters because the checkpointer's ability to reclaim the WAL depends on the read marks held by live connections. A growing set of long-lived connections is the necessary precondition for the "long-lived reader may be pinning an old snapshot that PASSIVE cannot reclaim" condition the runtime already warns about.

## Observed

Single host, one database, measured concurrently:

```
13  kkernel mcp        (stdio bridges)
 1  kkernel mcp --daemon
```

Ages of the bridge processes ranged from under two minutes to just over 40 hours. Eight were older than seven hours; three were older than 36 hours.

Every one of them held the database file, the `-wal`, and the `-shm`:

```
pid      db   wal  shm   age
79560    19   11   2     00:36:35     <- daemon (control)
22380    11   10   2     40:10:24
52403    11   10   2     38:38:15
577      11   10   2     38:20:58
20999    11   10   2     24:35:51
37223    11   10   2     24:19:41
38274    18   10   2     24:19:07
93508    11   10   2     13:00:53
13083    11   10   2     07:53:37
```

`-shm` is the WAL index shared-memory segment. Its presence means these are live connections participating in the WAL protocol, not leftover file descriptors.

The runtime's own holder census agrees exactly. `db_diagnostics()` returned 14 attributed holder PIDs, and that set is identical to the set enumerated independently at the OS level — the 13 bridges plus the daemon, with `census_pids_without_attribution: 0`.

## What is not being claimed

Presence is not pinning. In the same reading, all 14 holders were live and `checkpointed_frames == log_frames`, so the WAL was fully reclaimed while every one of these connections was open. Holding a connection is the precondition; holding a read transaction open across a write burst is the actual pin, and that was not measured here.

Nor is it established that each of these processes corresponds to a still-active client session. What is established is the population, its age distribution, and its resource footprint.

## Why it is still a defect

1. **Unbounded growth.** Nothing observed reaps a stdio server process. The count rises with session churn on a host that stays up, and each addition enlarges the set of connections the checkpointer must reason about.
2. **It widens the window for the pin condition.** More long-lived connections means more chances that one of them is sitting on an old read mark when a write burst arrives.
[...]

PR #2230, opened 2026-08-26
fix(mcp,runtime): idle-session timeout for stdio bridges; refuse duplicate daemon boot

### LINKED ISSUE #1874: kkernel daemon: no single-instance guard — two detached daemons can serve one store, parking stale readers that pin the WAL
## Observed

Two detached `kkernel mcp --daemon` processes coexisted on one machine for ~18 hours (started ~11h apart, both parented to PID 1, neither supervised). The newer one had re-bound the daemon socket — the socket file's inode matched only the newer process's FD, so all socket clients were served correctly — while the older one stayed alive holding open file descriptors and registered read marks on the primary store.

During a period of sustained write traffic, `checkpointed_frames` froze while the WAL grew past 60K pages (~280MB); the daemon's own checkpoint diagnostics reported a long-lived reader pinning an old snapshot that PASSIVE checkpoints could not reclaim. Terminating the stale duplicate daemon was followed by the checkpoint advancing on the next cycle and the WAL draining. (An alternative concurrent reader release cannot be fully excluded, but the stale daemon was the only anomalous holder in the census.)

## Root cause shape

Daemon startup does not verify exclusivity. Re-binding the unix socket succeeds by unlinking the path, which silently orphans a still-running older daemon: it keeps its DB handles, WAL read marks, and checkpoint loop, invisible to every socket client.

## Expected

At startup, before binding:

1. If the socket path exists, attempt a connection. A live response means an instance is already serving — refuse to start (loudly), do not unlink and re-bind over it.
2. A stale socket file (connect refused / no response) may be unlinked and re-bound, as today.
3. Optionally: a pidfile/flock held for the daemon's lifetime as a second arm, so exclusivity does not depend on socket probing alone.

The failure mode this closes is quiet by construction: both daemons run correctly in isolation, clients see no error, and the only symptom is WAL growth under a frozen checkpoint — visible only to whoever is watching checkpoint telemetry.

## Verification path

1. Start one daemon; start a second → second must exit non-zero naming the live instance.
2. Kill the first uncleanly (socket file left behind); start a second → must start (stale-socket arm).
3. With the guard in place, the two-daemon state should be unreachable without deliberately bypassing it.

### LINKED ISSUE #1921: MCP stdio server: no idle-session timeout lets abandoned client bridges hold reader-pool slots for days
## Observed

An lsof census of the primary store database's holders found 31 distinct `kkernel mcp` processes, with an age distribution of: 14 older than 4 hours, including three at 53 hours and four in the 45-47 hour range. All of the multi-day processes were stdio MCP bridges whose parent client applications (desktop UI sessions) were still alive but had long since stopped using the connection.

Each such bridge holds:
- a reader-pool admission slot,
- an open read fd on the store (visible to the WAL holder census),
- potentially a WAL read snapshot, contributing to checkpoint starvation (see #1895 for the in-process analogue).

During a concurrent incident window, the WAL grew monotonically at ~185 MiB/h to 2.6x `journal_size_limit` while the holder count stayed flat at 27-34 — consistent with a fixed population of non-releasing readers rather than load. Killing 17 stale bridges (all exited cleanly on SIGTERM) dropped the holder count from 31 to 16.

## Defect

`kkernel mcp` (stdio serve mode) has no idle-session timeout or server-side liveness policy. A client that stops talking but keeps the pipe open — a UI app with a dormant tab, a session whose parent never reaps children — pins server-side resources indefinitely. Connection lifecycle is distinct from #1895's abandoned-read reaping: these sessions are idle, not stuck mid-read.

## Suggested direction

- Idle timeout on stdio serve sessions: no request for N minutes (configurable, default generous, e.g. 60m) -> graceful shutdown of that bridge process. A client that returns simply respawns the bridge.
- On shutdown, release pool admissions and close read fds before exit (the clean-SIGTERM behavior observed suggests this path already works).
- Optionally, a startup census log line (bridge count + oldest age) so accumulation is visible before it starves the pool.

## Acceptance

Spawn a bridge, hold the pipe open with zero traffic past the timeout, and assert the process exits and the store holder count decrements; a live client with intermittent traffic must not be reaped.

ohdearquant:
> **Superseded in part — do not merge as-is.**
>
> The stdio-bridge session-lifetime half of this change now lives in #2239, where it was
> revised twice: the idle timeout became opt-in rather than on by default, and an outstanding
> request obligation now expires instead of deferring the idle close for the life of the
> session. The copy of that code on this branch predates both revisions, so merging this PR
> would reintroduce the behaviour those revisions removed.
>
> What remains uniquely here is the duplicate-daemon-boot refusal:
> `crates/khive-runtime/src/daemon.rs`, `crates/khive-mcp/src/daemon.rs`, and
> `crates/khive-runtime/tests/duplicate_daemon_refusal.rs`. Three of the nine commits on this
> branch touch both halves, so the daemon half is not separable at a commit boundary. It will
> be cut onto a fresh branch after #2239 lands, and this PR closed with a pointer to it.
>
> The boot behaviour it implements is specified in ADR-049 Amendment 6 (#2236).

---

Closes #1921, closes #1874; addresses #1836 via the same idle-timeout seam (an idle bridge now releases its pooled resources instead of holding a live WAL connection for its lifetime).

- `CancelOnEofTransport` gains an optional idle timeout (env `KHIVE_BRIDGE_IDLE_TIMEOUT_SECS`, default 3600s, 0 disables): a quiet window past the timeout is treated exactly like client EOF, driving the existing graceful-close path that already tears down the session and its pool handle. An admitted request with a response still being written — running long, or delivered slowly to a backpressured reader — defers that close instead of being cancelled out from under it; a request is only counted as "still awaiting its response" until its write has actually finished, not merely been handed off. Cancellation-safety of racing away the in-flight receive verified against the transport's buffering (state lives in `self`, not the polled-away future). Daemon mode is unaffected (it never reaches `serve_stdio`).
- That deferral is itself bounded by a second, independent knob: `KHIVE_BRIDGE_RESPONSE_DEADLINE_SECS` (default 300s; accepted range is 1 second and up — `0` is rejected at startup, naming the variable, the rejected value, and the accepted range) caps how long a single response write may stay pending before it is abandoned and the session closed anyway. Without it, a peer that admits a request and then simply stops reading its response — while leaving the connection open — would keep that write pending forever and the idle timeout would defer indefinitely, never reaping the session.
- A second `kkernel mcp --daemon` against a store already served by a live, responsive daemon now fails loudly: error names the incumbent pid and socket path, process exits non-zero. The stale-socket recovery path (dead pid or unreachable socket) is unchanged. The liveness probe requires an exact protocol-acknowledgement shape (not just any well-formed response) so it can't mistake an unrelated response — for example a metrics snapshot — from the same daemon for the acknowledgement it's looking for.

Tests: idle timeout cancels without EOF; intermittent traffic is not reaped; an admitted long-running request is not cancelled by wire idleness; a backpressured response write defers idle-close; a response write that never resolves is still reaped within the response-delivery deadline; a `0` response-delivery deadline is rejected at startup rather than accepted as an unbounded opt-out; duplicate boot refuses while the first daemon is live, including against a metrics-only response (verified to fail without the fix). Full khive-runtime suite green; khive-mcp/kkernel checks clean.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2463,6 +2463,7 @@ dependencies = [
  "khive-gate",
  "khive-gate-rego",
  "khive-query",
+ "khive-runtime",
  "khive-score",
  "khive-storage",
  "khive-types",
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -1152,8 +1152,18 @@ async fn probe_daemon_identity(config_id: &str, namespace: &str, timeout_ms: u64
             // in handle_conn produces ok=true + result=None + error=None.
             //
             // A normal successful dispatch always sets result=Some(_) (never None).
-            // A normal error dispatch sets ok=false. So the sentinel is unambiguous.
-            let is_probe_ack = resp.ok && resp.result.is_none() && resp.error.is_none();
+            // A normal error dispatch sets ok=false. `metrics_only` produces the
+            // exact same ok=true/result=None/error=None shape but carries
+            // metrics=Some(_); a request that itself echoed a request_id would
+            // carry request_id=Some(_). This probe frame sets neither, so both
+            // must come back None too — otherwise a metrics snapshot or an
+            // echoed request_id from an unrelated exchange would be misread as
+            // this probe's own acknowledgement.
+            let is_probe_ack = resp.ok
+                && resp.result.is_none()
+                && resp.error.is_none()
+                && resp.metrics.is_none()
+                && resp.request_id.is_none();
             if is_probe_ack
                 && !resp.version_mismatch
                 && !resp.namespace_mismatch
@@ -4769,6 +4779,64 @@ mod tests {
                  not None (which would cause silent fallback to local dispatch)"
             ),
         }
+
+        clear_daemon_env();
+        std::env::remove_var("KHIVE_LOCK");
+    }
+
+    // ── client-side probe rejects a metrics-only response (#2230) ────────────
+    //
+    // The daemon's `metrics_only` arm answers with `ok=true, result=None,
+    // error=None`, every mismatch flag false, the current protocol version,
+    // and a matching `served_config_id` — the exact same shape the probe-ack
+    // arm produces, differing only in carrying `metrics: Some(...)`. A
+    // well-formed metrics snapshot response must not be misread as this
+    // probe's own acknowledgement, or the client would defer to a peer that
+    // never actually answered the identity probe.
+
+    #[tokio::test]
+    #[serial]
+    async fn probe_daemon_identity_rejects_a_metrics_only_response() {
+        clear_daemon_env();
+        let dir = tempfile::tempdir().expect("tempdir");
+        let sock = dir.path().join("khived.sock");
+        let pid_file = dir.path().join("khived.pid");
+        let lock_file = dir.path().join("khived.recovery.lock");
+
+        std::env::set_var("KHIVE_SOCKET", &sock);
+        std::env::set_var("KHIVE_PID", &pid_file);
+        std::env::set_var("KHIVE_LOCK", &lock_file);
+        std::env::remove_var("KHIVE_NO_DAEMON");
+
+        let config_id = "packs=[kg];db=:memory:;embed=none;extra=[];backend=main";
+        let listener = tokio::net::UnixListener::bind(&sock).expect("bind fake daemon socket");
+
+        let response = DaemonResponseFrame {
+            ok: true,
+            result: None,
+            error: None,
+            namespace_mismatch: false,
+            config_mismatch: false,
+            served_config_id: Some(config_id.to_string()),
+            version_mismatch: false,
+            daemon_protocol_version: PROTOCOL_VERSION,
+            metrics: Some(khive_runtime::daemon::MetricsSnapshot::default()),
+            request_id: None,
+        };
+        let fake_handle = tokio::spawn(serve_one_response(listener, response));
+
+        let outcome = probe_daemon_identity(config_id, "test", 2_000).await;
+
+        tokio::time::timeout(std::time::Duration::from_secs(2), fake_handle)
+            .await
+            .expect("fake daemon listener task timed out")
+            .expect("fake daemon listener task panicked");
+
+        assert!(
+            matches!(outcome, ProbeOutcome::Dead),
+            "an otherwise-matching response carrying a metrics snapshot must not be treated \
+             as a probe acknowledgement; got {outcome:?}"
+        );
 
         clear_daemon_env();
         std::env::remove_var("KHIVE_LOCK");
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -1037,24 +1037,36 @@ fn stdio_bridge_idle_timeout_from_env() -> Option<std::time::Duration> {
 /// (an in-flight response always defers idle-close) and the session would
 /// never be reaped.
 ///
-/// Overridable via `KHIVE_BRIDGE_RESPONSE_DEADLINE_SECS`; `0` disables it
-/// (an explicit opt-out, matching `KHIVE_BRIDGE_IDLE_TIMEOUT_SECS=0`). An
-/// unparsable value falls back to the default. Default: 300s (5 minutes) —
+/// Overridable via `KHIVE_BRIDGE_RESPONSE_DEADLINE_SECS`, accepted range
+/// 1..=u64::MAX seconds. Unlike `KHIVE_BRIDGE_IDLE_TIMEOUT_SECS`, this bound
+/// cannot be disabled: `0` is a startup error naming the variable, the
+/// rejected value, and the accepted range, rather than a silent opt-out — a
+/// configuration that restores an unbounded pending write restores the
+/// defect this deadline exists to close (see the type doc above). An
+/// unparsable value falls back to the default rather than erroring, matching
+/// this codebase's other `_from_env` helpers. Default: 300s (5 minutes) —
 /// long enough that legitimately slow verbs and ordinary reader backpressure
 /// never trip it, short enough that a peer that has genuinely stopped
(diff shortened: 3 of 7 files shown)


ohdearquant:
The regression test at `crates/khive-mcp/src/daemon.rs:6110-6119` states its motivation as a
version-skew scenario that cannot occur at the current protocol version.

The stated premise needs a binary that both predates `probe_only` and carries the current
`PROTOCOL_VERSION`. No such binary exists.

`probe_only` first appears in `89db3ff6f`, and `PROTOCOL_VERSION` was `1` immediately before
that commit:

```
$ git log --oneline --reverse -S'probe_only' -- crates/ | head -1
89db3ff6f fix(mcp): surface daemon error on connection-reset mid-dispatch (#91)

$ git grep -n 'const PROTOCOL_VERSION' 89db3ff6f^ -- crates/
89db3ff6f^:crates/khive-runtime/src/daemon.rs:36:pub const PROTOCOL_VERSION: u32 = 1;

$ git grep -n 'const PROTOCOL_VERSION' main -- crates/
main:crates/khive-runtime/src/daemon.rs:48:pub const PROTOCOL_VERSION: u32 = 4;
```

Every bump above `1` postdates `probe_only`, by `git merge-base --is-ancestor 89db3ff6f <c>`:

| commit       | postdates `probe_only`? | version it carries |
| ------------ | ----------------------- | ------------------ |
| `da39378c1`  | no                      | 1                  |
| `dff4aafc6`  | yes                     | 2                  |
| `f27880356`  | yes                     | 3                  |
| `bbb5458b4`  | yes                     | 4                  |

`da39378c1` is the one commit in that set that predates `probe_only`, and it still carries
version `1`, so it is not a counterexample. Any binary carrying version >= 2 already has
`probe_only`.

Control for the ancestry predicate: `git merge-base --is-ancestor bbb5458b4 89db3ff6f` returns
false, so it is not answering true for every pair.

A version-`1` binary would not reach dispatch in any case. The serve-side mismatch arm at
`crates/khive-runtime/src/daemon.rs:1092-1103` is the first arm of the chain, ahead of the
`metrics_only` arm at `:1104` and ahead of ops dispatch, and it short-circuits with
`version_mismatch: true`.

**The test itself is fine and should stay.** It guards the `is_probe_ack` predicate
(`resp.ok && resp.result.is_none() && resp.error.is_none()`), and its fail-if-reverted claim at
`:6125-6127` does not depend on the skew story: drop `is_probe_ack` and an identity-matching
`ok=false` response classifies as Alive, `KILL_COUNT` stays `0`, the assertion fails.

What is inaccurate is the stated reason. As written, a reader is told the probe classifier
fences a live version-skew path. It does not, because the version handshake fences it first.
That misleads two ways: someone could weaken or drop the version handshake believing the probe
classifier covers the same ground, and the scenario reads as current when it was reachable only
in the window between the two commits above, while the constant was `1`.

Suggested reword of `:6112-6116`: say what the fixture actually is, a response matching every
identity field that is not a probe ack. Keep the fail-if-reverted paragraph verbatim. If the
historical motivation is worth keeping, name the window and note that the version handshake
closes it.

oceanwaves630:
Closing in favour of #2281, which carries the daemon half of this branch cut fresh from current `main`.

This branch mixed two independent changes — the duplicate-daemon boot refusal and a stdio-bridge idle-session timeout — across nine commits, and went conflicted against `main`. The halves do not depend on each other: `khive-runtime` sits below `khive-mcp` in the dependency order, so the daemon work needs nothing from the bridge work.

**What moved to #2281:** the duplicate-daemon refusal in `crates/khive-runtime/`, including the protocol-identity probe and its full-shape response check (the metrics branch is otherwise identical to the probe-ack branch, so `ok=true` alone is not a sufficient test), plus the `duplicate_daemon_refusal` integration test. Verified there against current `main`: fmt clean, clippy clean, 1535 tests passing.

**What did not move:** the stdio-bridge idle-session timeout, the in-flight request accounting, the response-delivery deadline, and the `KHIVE_BRIDGE_RESPONSE_DEADLINE_SECS=0` startup refusal. That work is not abandoned — it is simply not carried by #2281 and needs its own branch from current `main`. Nothing in #2281 blocks it.

One note for whoever picks the bridge half up: the commit subjects on this branch do not partition the work. `fix(mcp): scope the idle timeout to genuinely idle sessions and probe daemon identity` is labelled mcp-only but contributes 113 lines to `crates/khive-runtime/src/daemon.rs`. Splitting on subject lines alone drops it.

oceanwaves630:
Correction to my previous comment, which was wrong on one point.

I wrote that the stdio-bridge work here "is not abandoned — it is simply not carried by #2281 and needs its own branch from current `main`." It does not need a branch: **it is already on `main`**, landed by #2247 (`110e35d2`, "fix(mcp): bound the lifetime of an abandoned stdio bridge session"), and in a more developed form than this branch carries.

Verified at `origin/main` under `crates/khive-mcp/`:

- `KHIVE_BRIDGE_IDLE_TIMEOUT_SECS`, `KHIVE_BRIDGE_RESPONSE_DEADLINE_SECS`, `KHIVE_BRIDGE_MAX_OUTSTANDING_REQUESTS`, `CancelOnEofTransport` — all present.
- The `KHIVE_BRIDGE_RESPONSE_DEADLINE_SECS=0` startup refusal, which was this branch's final commit, is on `main` **with a regression test** asserting the value is rejected rather than silently accepted (`server.rs:8094-8115`).
- The idle timeout is scoped to a genuinely idle session — an admitted request whose response is still being written defers the close (`server.rs:1015`), which is what this branch's mid-series commit was reaching for.
- `main` additionally makes the timeout opt-in with a written rationale against ADR-091's rejected "kill long-lived reader sessions" alternative. This branch defaults it on. That is a deliberate difference and `main` is the considered side.

So both halves of this PR are resolved: the duplicate-daemon refusal moved to #2281, and the bridge half is superseded by already-merged work. Nothing here is stranded and no follow-up branch is needed.

How I got it wrong: I checked that the bridge commits were absent from #2281 and reported them as unlanded, without checking whether they were present on `main` by another route. "Not in this PR" is not "not in the tree."

PR #2233, opened 2026-08-26
fix(db,runtime,mcp,comm): guard six read-modify-write paths with compare-and-swap

### LINKED ISSUE #1753: Read-modify-write full-row upserts lose concurrent updates across gtd, kg, and schedule paths
Several write paths use read-modify-write with a full-row upsert whose guard does not cover the fields being replaced, so a concurrent update to unrelated fields is silently overwritten from a stale snapshot. All read at `f74c5461f6737b831f143d2dfc953d71348fd73a`.

- **gtd lifecycle transitions** — `crates/khive-pack-gtd/src/handlers.rs:537` (and the same shape at `:597`): the transition helper reads the full properties object, computes a replacement, and issues a conditional UPDATE whose predicate checks only `json_extract(properties,'$.status')`. A concurrent property update that leaves status unchanged is overwritten by the stale full-property write.
- **generic note update** — `crates/khive-runtime/src/curation.rs:1250`: help describes properties as a merge patch, but the implementation reads the note, merges in memory, and upserts the complete note with no version or conditional-row predicate.
- **atomic entity/note update plans** — `crates/khive-runtime/src/atomic_prepare.rs:805-810`: the plan uses `entity_upsert_statement` (full-row INSERT OR REPLACE) guarded only by `AffectedRowGuard::exactly(1)`, which detects row count, not concurrent modification.
- **schedule finalization** — `crates/khive-mcp/src/pending_events.rs:1014-1063`: finalization replaces the complete `properties` value while its CAS guards only status and the firing token.

**Expected:** a snapshot/version predicate (or field-scoped UPDATE) so concurrent writers fail, merge, or retry visibly.
**Actual:** last-writer-wins over the whole row while each caller reports success.

ohdearquant:
Fixes #1753.

Six production read-modify-write paths read a row, computed a new value from that snapshot, and wrote it back unconditionally. A concurrent writer landing between the read and the write was silently overwritten: no error, no counter, no log. The loser of the race simply lost, and the surviving row looked normal.

Each of the six now goes through a compare-and-swap write. The pattern is the one already established by `replace_note_if_unchanged` (`crates/khive-db/src/stores/note.rs`): the update carries its snapshot's revision in the `WHERE` clause, so a writer whose snapshot has gone stale affects zero rows and finds out, instead of clobbering.

## What changed

Two new shared primitives, mirroring the existing note primitive:

- `replace_entity_if_unchanged`
- `replace_edge_if_unchanged`

Both express the write as `UPDATE ... WHERE id = ? AND updated_at = ? AND deleted_at IS ?` with a monotonic-timestamp condition, and report affected rows so the caller can distinguish "applied" from "someone else moved it".

Threaded through the six sites, each using the guard form appropriate to its layer:

| Site | Crate | Guard form |
|---|---|---|
| Channel heartbeat | `khive-pack-comm` | store primitive |
| Entity update | `khive-runtime` (curation) | store primitive |
| Non-symmetric edge update | `khive-runtime` (operations) | store primitive |
| Atomic entity update plan | `khive-runtime` (atomic_prepare) | guarded statement builder, with an affected-row guard that turns a stale zero-row result into a whole-unit failure |
| Atomic non-symmetric edge update plan | `khive-runtime` (atomic_prepare) | same |
| Schedule finalization | `khive-mcp` | conditional update pinned to the dispatch claim |

Schedule finalization is deliberately not a plain revision check. Its update matches on the claim itself — status, firing timestamp, and dispatch invocation id, with an optional exact-properties equality — and returns success only on exactly one affected row. Pinning the claim rather than the revision is the stronger condition for that path, because it also rejects a finalize from a superseded dispatch.

## Not in scope

`khive-pack-code`'s ingest sites share the same shape and are deferred to their own change; they are a different subsystem with a different concurrency story, and bundling them here would put two independently reviewable subjects in one PR.

No accepted contract changes. Every guard here is an application of an already-ratified mechanism to a new call site, not new policy.

## Tests

Each site has a regression that deterministically reproduces two writers proceeding from one revision and asserts the stale writer is refused rather than merged. The `atomic_prepare` tests additionally assert that a zero-row result fails the enclosing unit rather than passing silently, which is the arm that would otherwise turn a detected race back into a silent one.

The per-site regressions are whole-guard tests, and it is worth saying what that does and does not buy. They cannot attribute a failure to one condition of the guard: both racers take their replacement revision from a wall-clock read, and the fixture pins the two expected revisions equal without pinning the two replacement revisions, so removing the revision-equality condition alone leaves the outcome dependent on which clock read won. What they do establish is that with the revision-equality condition intact they cannot observe the monotonic-timestamp condition disappear at all. Mutating that condition on its own reddened nothing in the crate, so it had no regression coverage. An isolating fixture now covers it directly, supplying a correct snapshot revision and deletion marker so it is the only predicate that can refuse the write, and under that same mutation it is the single test that reddens.

### DIFF
--- a/crates/khive-db/src/stores/entity.rs
+++ b/crates/khive-db/src/stores/entity.rs
@@ -104,6 +104,73 @@ pub fn entity_upsert_statement(entity: &Entity) -> SqlStatement {
     }
 }
 
+/// Full-entity compare-and-swap update used after caller-side normalization
+/// was derived from a read snapshot. Unlike [`entity_upsert_statement`], this
+/// never inserts and cannot overwrite a row whose revision or deletion
+/// marker moved after the snapshot was read. The replacement revision must
+/// also be strictly greater than the persisted snapshot revision; equality
+/// is a refused CAS, never a successful write with an unchanged concurrency
+/// token. Mirrors `note_replace_if_unchanged_statement`
+/// (`crates/khive-db/src/stores/note.rs`).
+pub fn entity_replace_if_unchanged_statement(
+    entity: &Entity,
+    expected_updated_at: i64,
+    expected_deleted_at: Option<i64>,
+) -> SqlStatement {
+    let properties_str = entity
+        .properties
+        .as_ref()
+        .map(|v| serde_json::to_string(v).unwrap_or_default());
+    let tags_str = serde_json::to_string(&entity.tags).unwrap_or_else(|_| "[]".to_string());
+    SqlStatement {
+        sql: "UPDATE entities SET \
+                namespace = ?1, kind = ?2, entity_type = ?3, name = ?4, description = ?5, \
+                properties = ?6, tags = ?7, updated_at = ?8, deleted_at = ?9, \
+                merged_into = ?10, merge_event_id = ?11 \
+              WHERE id = ?12 AND updated_at = ?13 AND deleted_at IS ?14 \
+                AND ?8 > updated_at"
+            .to_string(),
+        params: vec![
+            SqlValue::Text(entity.namespace.clone()),
+            SqlValue::Text(entity.kind.clone()),
+            match &entity.entity_type {
+                Some(t) => SqlValue::Text(t.clone()),
+                None => SqlValue::Null,
+            },
+            SqlValue::Text(entity.name.clone()),
+            match &entity.description {
+                Some(d) => SqlValue::Text(d.clone()),
+                None => SqlValue::Null,
+            },
+            match properties_str {
+                Some(p) => SqlValue::Text(p),
+                None => SqlValue::Null,
+            },
+            SqlValue::Text(tags_str),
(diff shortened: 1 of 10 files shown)


ohdearquant:
Closing in favour of two narrower changes.

This branch bundled two independent fixes. Splitting them keeps each test
attributable to the defect it covers, which the combined branch could not do.

**Guarded entity and edge replacement** — including the tombstone-resurrection
case and its isolating fixtures — is now #2238.

**The `pending_events` finalizer fix** will follow as its own change once #2238
lands. This branch touched one of the three write paths that need it; the
replacement covers all three, with a test per path. Those tests are only
meaningful once #2238 is on `main`, because failing against that head is what
demonstrates each test is bound to its own path rather than to the guard work.

No commit from this branch is being carried forward as-is.

PR #2236, opened 2026-08-27
docs(adr): ADR-049 Amendment 6 — incumbent classification at boot

ohdearquant:
Adds Amendment 6 to ADR-049. Documentation only; no code in this PR.

## Problem

Daemon boot answers "may I clean up the rendezvous and bind?" using a probe
predicate that returns a bare boolean. A boolean cannot distinguish "nothing is
serving this socket" from "something is serving it that this client may not
use", so three responses emitted by a demonstrably live daemon — a
protocol-version mismatch, a config mismatch, and a metrics reply — take the
same unlink-and-bind path as an absent endpoint.

Unlinking a live daemon's socket and pid file does not stop it. The result is a
second listener bound beside a running owner, which contradicts the convergence
requirement already stated in Amendment 3: exactly one live, identity-matching
daemon owning the rendezvous.

## What the amendment specifies

- Separates **liveness** (is anything serving this socket?) from
  **acceptability** (is it a peer I can share it with?), and states that
  unlink-and-rebind is licensed by a negative answer to liveness alone.
- Enumerates eleven incumbent states with a normative disposition per state.
  Only two of them permit binding. The other nine are refusals today collapsed
  into cleanup.
- Replaces the boolean with a `MayBind` / `Refuse { pid, reason }` disposition,
  since a single bit is what invites the collapse.
- Makes **exit codes normative**, because supervisors read them: a supervisor
  configured to restart only on unsuccessful exit treats exit 0 as a deliberate
  stop. Configuration-class refusals exit 0; state-class refusals exit nonzero.
- Specifies `--replace-incumbent` as the only path that may remove a live
  incumbent: operator-elected, signal then **wait for observed process death**,
  refuse on timeout with the socket intact, never an implicit `SIGKILL`.
- States test obligations that assert **outcomes** — incumbent still alive,
  socket still present, no second listener — rather than return values, since a
  return-value-only test passes while the socket is unlinked under a live
  process.

## Two readings recorded because they are easy to invert

- The Consequences section calls the daemon disposable in the sense that killing
  and respawning it is safe. That licenses replacement being safe; it does not
  make replacement something boot performs unprompted.
- Amendment 3 already anticipates multiple launch attempts and lets the
  daemon-side fence pick the sole owner, so losing that fence is legitimate
  rather than an error, and that state exits cleanly.

## Scope

One state in the table is reachable only defensively: the boot probe's request
frame does not set the metrics flag, and the daemon's metrics reply is gated on
that flag in the requesting frame, so boot cannot elicit a metrics-only reply.
It is retained and labelled as defensive, because a reply carrying metrics is
still unambiguous proof of life and misreading it as absence would be the same
defect.

Amendments 3, 4 and 5, the client-side recoverer lock, and the daemon-side boot
fence are unchanged. This amendment constrains only what boot may conclude from
a probe result and what it may do about it.

### DIFF
--- a/docs/adr/ADR-049-khived-daemon.md
+++ b/docs/adr/ADR-049-khived-daemon.md
@@ -328,3 +328,133 @@ is not consulted as request cancellation state, preserving repeated daemon-run
 tests. A blocking write that ignores async abort continues to own its SQLite
 connection until its blocking closure exits; the daemon never interrupts that
 write or reports a retryable read timeout for it.
+
+## Amendment 6 (2026-08-26): incumbent classification at boot
+
+Boot answers one question — "may I clean up the rendezvous and bind?" — and the
+convergence invariant above requires that the answer be governed by whether a
+daemon is **live**, never by whether it is one this process would choose to talk
+to. This amendment separates those two questions and enumerates the states boot
+must distinguish.
+
+### Liveness and acceptability are different questions
+
+- **Liveness**: is a process serving this socket right now? Answered by the
+  connect and by whether anything comes back at all.
+- **Acceptability**: is that process a peer this client can share the rendezvous
+  with? Answered by identity — protocol version and `config_id`.
+
+Unlink-and-rebind is licensed by a negative answer to **liveness only**. It is
+never licensed by a negative answer to acceptability, because an unacceptable
+peer that is alive still owns the socket. Removing its rendezvous files does not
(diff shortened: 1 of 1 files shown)


ohdearquant:
Superseded by #2240, which carries this amendment's final content unchanged. Closing so the change has a single home.

PR #2239, opened 2026-08-27
fix(mcp): bound stdio bridge session lifetime

ohdearquant:
## Problem

An abandoned stdio bridge session holds resources indefinitely. A client whose pipe never
closes but which stops sending requests pins its reader-pool admission slot and its DB
connection for as long as the parent process lives. Nothing in the session lifetime bounds
that: the only exit was real EOF.

## Change

Two independent bounds, plus the accounting that makes the first one safe.

- **Response-delivery deadline** (`KHIVE_BRIDGE_RESPONSE_DEADLINE_SECS`, default 300s, on by
  default). A response write that never resolves — a peer that admits a request and then
  stops reading while holding the pipe open — is abandoned at the deadline. Timing the write
  out drops it, which releases the writer lock that a later `close()` would otherwise block
  on, and closes the session. `0` is a startup error naming the variable, the value, and the
  accepted range, not a silent opt-out: a disabled deadline restores the unbounded
  pending-write defect the deadline exists to close.
- **Idle-session timeout** (`KHIVE_BRIDGE_IDLE_TIMEOUT_SECS`, **off by default**). When
  configured, a session quiet for that long closes.
- **Request-obligation TTL** (`KHIVE_BRIDGE_REQUEST_OBLIGATION_SECS`, default 3600s). How
  long an admitted request whose response has not been written keeps deferring the idle
  close.

### Why idle reaping is off by default

[ADR-091](docs/adr/ADR-091-wal-snapshot-lifetime.md) lists "kill long-lived reader sessions"
among its rejected alternatives, on the ground that long-lived stdio sessions are live
clients and that bounding what they hold underneath them is the better fix. Closing a session
after an hour of quiet is that rejected policy whatever the mechanism, because this transport
has no signal separating an abandoned pipe from a client that simply has not been asked
anything. Enabling it by default would reverse an accepted decision from inside an unrelated
change. It is an operator's explicit act instead: a supervised deployment, a CI harness, a
batch runner. Making it the default is an ADR-091 amendment, not a different number here.

An unparsable value falls back to *disabled* rather than to an hour, so a typo cannot start
closing live sessions.

### Why the obligation is timestamped rather than counted

The idle check must not treat a session with work outstanding as idle, or it cancels a
running handler out from under itself. But rmcp spawns each request handler with
`spawn_service_task` and drops the join handle, so a handler that panics never reaches the
response construction that clears its obligation. Deferring on an outstanding obligation with
no bound hands any panicking handler the power to disable idle reaping for the life of the
session — the unbounded lifetime the timeout exists to close, reintroduced through its own
guard.

So the transport records each request's admission instant and defers only while the newest
outstanding obligation is younger than the TTL. The TTL is deliberately not the idle window:
tying them together would make a handler that runs longer than one quiet window stop
protecting its own session, which is the guarantee the obligation provides.

### Why obligations are keyed by request id

The freshness rule reads the newest outstanding admission, so the entry it reads has to be
one that is genuinely still outstanding. rmcp spawns each handler and each response send
independently, so a newer request's response can finish before an older request's. Retiring
the oldest entry on any completion — the obvious FIFO reading of "a response arrived, so one
obligation is discharged" — would then retire the older request's admission for the newer
request's completion and leave a *completed* request's instant as the newest entry. The
freshness test would read that instant and keep deferring past the older obligation's TTL,
restoring the unbounded session lifetime this guard exists to close.

Entries are therefore `(RequestId, Instant)` and retirement removes the entry whose id
matches the response or error being written. `JsonRpcError.id` is optional, because a parse
error has no id to echo, so those retire nothing — which is correct: no obligation was ever
admitted for them.

Keying retirement changes what happens to a request that never produces a response at all: a
handler that panics, or a request the peer cancels. Oldest-first retirement removed *some*
entry on every completion, so the queue tracked admissions minus completions. Keyed
retirement removes only the entry whose response was written, so an unanswered request would
hold its slot for the life of the session and a long-lived session would accumulate one per
such request without bound. An entry past the TTL is already not evidence of life — the
freshness check will not defer on it — so keeping it buys nothing, and entries past the TTL
are dropped whenever the queue is touched. They are pushed in admission order, so the stale
ones are a prefix.

A request also counts as in flight until its response has finished *writing*, not merely been
handed to the transport send. rmcp spawns the response send and the framed write awaits the
transport writer, so a backpressured reader leaves the write pending well after the handler
returns; counting that as idle would cancel a session out from under an undelivered response.

The idle check races a single pinned `receive()` future rather than starting a fresh one per
deferred window, so a request arriving fragmented across a window is reassembled instead of
losing its already-read prefix.

## Known gap

The response-delivery deadline covers responses this transport writes. It does not cover
parse-error responses, which the underlying line transport writes directly through its own
framed writer without passing through the deadline. A peer that sends malformed input and
then stops reading can leave that one write pending. What bounds it depends on what else the
session has outstanding, and the idle window alone is not the answer:

- Idle timeout disabled: nothing bounds it.
- Idle timeout enabled, nothing else outstanding: the idle window bounds it.
- Idle timeout enabled with a request still awaiting its response: the idle close defers
  while that obligation is fresh, so the bound is the request-obligation TTL rather than the
  idle window. A peer can reach this deliberately by admitting a request and then sending
  malformed input.

Closing the gap requires replacing or adapting the line transport, which is out of scope here
and documented in `docs/configuration.md`.

## Tests

Ten tests, each asserting an outcome rather than a return value.

| Test | Asserts |
| --- | --- |
| `stdio_idle_timeout_cancels_root_without_eof` | the timeout alone ends the session |
| `stdio_idle_timeout_does_not_reap_client_with_intermittent_traffic` | traffic inside the window keeps it alive |
| `stdio_idle_timeout_does_not_cancel_an_admitted_long_running_request` | a slow handler is not idle |
| `stdio_idle_timeout_does_not_cancel_while_a_response_write_is_backpressured` | a pending write is not idle |
| `stdio_idle_timeout_reaps_a_peer_that_stops_reading_its_response` | the delivery deadline fires and closes |
| `stdio_idle_timeout_preserves_a_fragmented_request_while_another_is_in_flight` | the pinned receive keeps a partial frame |
| `stdio_idle_timeout_reaps_a_session_whose_handler_never_answers` | an obligation past its TTL stops deferring |
| `stdio_idle_timeout_reaps_on_the_older_obligation_when_a_newer_one_completes_first` | out-of-order completion does not extend the session |
| `stdio_obligation_queue_drops_entries_past_their_ttl` | an unanswered request does not hold its slot forever |
| `response_deadline_from_env_rejects_zero` | `=0` is a startup error naming variable, value, and range |

The negative tests are the substance. Without them, a timeout that fires during a slow
handler or a backpressured write passes a "does it time out" check while doing exactly the
damage the feature must not do. The never-answers test asserts the fresh case first, so it
cannot pass against a build that simply ignores in-flight work; wiring the freshness check to
a constant true reddens that test and only that test.

A test that asserts a non-event passes just as well against a build whose timer was never
armed, so each "does not cancel" test also asserts, through a shared helper, that once
nothing is protecting the session the timer does close it — with the pipe still open, so
nothing but the timer can be what ended the session. Forcing the transport to read its idle
timeout as unconfigured reddens every test that configures one and leaves the others green.

The out-of-order test picks timings that separate correct behaviour from the defect rather
than merely exercising the path: two requests admitted 1500ms apart against a 2000ms
obligation TTL, the first parked forever and the second answered promptly. Correct behaviour
closes about 2000ms after the first admission; FIFO retirement closes at about 3500ms, and
the assertion sits between them.

## Gates

At `64f9ff03de7a5b26c9c7b8648e28bed48c1435fd`: `cargo fmt --all --check` clean,
`cargo clippy -p khive-mcp --all-targets -- -D warnings` with no warning or error lines,
`RUSTDOCFLAGS="-D warnings" cargo doc -p khive-mcp --no-deps` clean, and
`cargo test -p khive-mcp --lib` at 438 passed / 0 failed.

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -7828,6 +7828,92 @@ mod request_read_cancellation_tests {
         );
     }
 
+    /// Regression: two outstanding obligations under one request id are
+    /// refused, not resolved by guessing.
+    ///
+    /// Retirement matches on request id, so two live entries sharing an id
+    /// make it ambiguous which one a completing response discharges — and both
+    /// resolutions are wrong in opposite directions. Removing the first match
+    /// can leave a *completed* request's instant as the newest entry, so the
+    /// freshness check defers past the older obligation's TTL: the unbounded
+    /// session this whole mechanism exists to close. Removing the last match
+    /// can leave the older instant, so the session closes out from under a
+    /// live handler. The transport cannot tell the two apart, so it refuses
+    /// the second admission instead.
+    ///
+    /// Idle reaping is off and the pipe is deliberately never closed, so the
+    /// refusal is the only thing that can end this session. A build that
+    /// admitted the duplicate would leave `root` uncancelled and this test
+    /// would exhaust its bound.
+    #[tokio::test]
+    async fn stdio_refuses_a_second_outstanding_obligation_under_one_request_id() {
+        use rmcp::transport::async_rw::AsyncRwTransport;
+        use tokio::io::AsyncWriteExt;
+
+        let first_admitted = Arc::new(tokio::sync::Notify::new());
+        let probe = OutOfOrderProbeServer {
+            first_admitted: first_admitted.clone(),
+            calls: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
+        };
+        let root = tokio_util::sync::CancellationToken::new();
+        let (server_io, client_io) = tokio::io::duplex(16 * 1024);
+        let (server_read, server_write) = tokio::io::split(server_io);
+        let transport = crate::transport::CancelOnEofTransport::with_idle_timeout(
+            AsyncRwTransport::new_server(server_read, server_write),
+            root.clone(),
+            // No idle timer: only the duplicate-id refusal can close this.
+            None,
+            None,
+            Some(Duration::from_secs(3600)),
+        );
+        let obligations = transport.in_flight_handle();
+        let running = rmcp::service::serve_directly_with_ct(probe, transport, None, root.clone());
+        let (_client_read, mut client_write) = tokio::io::split(client_io);
+
+        let request = b"{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"tools/call\",\
+                        \"params\":{\"name\":\"probe\",\"arguments\":{}}}\n";
+        client_write
+            .write_all(request)
+            .await
+            .expect("write the first request");
+        tokio::time::timeout(Duration::from_secs(2), first_admitted.notified())
+            .await
+            .expect("rmcp never admitted the first request");
+
+        // Premise: id 1 is genuinely still outstanding. Without this the second
+        // request would just be a reuse after retirement, which is legitimate
+        // and not what this test is about.
+        assert_eq!(
+            obligations.lock().expect("obligation queue poisoned").len(),
+            1,
+            "fixture premise: the first request must still be outstanding when the duplicate \
+             arrives, otherwise nothing is ambiguous"
+        );
+        assert!(
+            !root.is_cancelled(),
+            "fixture premise: a single outstanding request must not have closed the session"
+        );
+
+        client_write
+            .write_all(request)
+            .await
+            .expect("write the duplicate-id request");
+
+        let deadline = std::time::Instant::now() + Duration::from_secs(5);
+        while !root.is_cancelled() && std::time::Instant::now() < deadline {
+            tokio::time::sleep(Duration::from_millis(20)).await;
+        }
+        assert!(
+            root.is_cancelled(),
+            "a second outstanding obligation under an id that already has one must close the \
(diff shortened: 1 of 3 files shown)


ohdearquant:
Superseded by #2241, which carries the same fix in a reviewable scope.

This branch bundled the session-lifetime bounds with a set of timing tests that
assert the absence of an event within a wall-clock margin. Those tests cannot
fail fast and cannot distinguish a correct implementation from a slow machine,
so they were removed rather than tuned. #2241 keeps the three bounds and the
five tests that assert observed state transitions, at 957 insertions against
1682 here.

The deterministic-clock work those removed tests were reaching for is real and
follows in its own change: the transport currently mixes Tokio timers with
`std::time::Instant`, so pausing Tokio's clock does not move the obligation
freshness check. A clock seam is what makes those cases testable without
sleeping, and the regression test for the ordered-polling fix belongs with it.

No commits from this branch are lost; #2241 is built from them.

PR #2241, opened 2026-08-27
fix(mcp): bound the lifetime of an abandoned stdio bridge session

ohdearquant:
## Problem

A stdio bridge client whose pipe never closes but which stops sending requests is
indistinguishable from a live one. Nothing reaps it, so it holds its reader-pool
admission and its DB connection for as long as its parent process happens to survive.
Sessions have been observed alive at tens of hours this way.

## What this changes

Three bounds. Two are opt-in and disabled by default: the idle-session timeout and the
request-obligation TTL do nothing unless configured. The third, the response-delivery
deadline, is **not** optional — it defaults to 300 seconds when unset and rejects `0`
(`crates/khive-mcp/src/server.rs:1101-1118`), because a disabled deadline lets a peer that
stops reading pin the bridge's response write forever, which is one of the leaks this change
exists to close. So a server with no configuration does change behaviour: a response pending
longer than five minutes is now abandoned. `docs/configuration.md:298-302` documents that
default; this paragraph previously contradicted it and the code, and the code is right.

**Idle-session timeout.** A window with no request is treated as EOF: `receive()` yields
`None`, and rmcp's ordinary graceful-close path releases the pooled resources exactly as
a real disconnect does.

**Outstanding-obligation accounting.** Wire-idle is not session-idle, so the idle branch
closes only when nothing admitted still owes a response. Obligations are held in a queue
keyed by request id, each carrying the instant it was admitted. Keying by id is load
bearing: rmcp spawns handlers and response sends independently, so a newer request's
response can finish first, and retiring the oldest entry on any completion would leave a
completed request as the newest entry and make the freshness test read the wrong
timestamp.

An obligation is evidence of life only while it is fresh. A handler that panics never
reaches the response construction that would retire its entry, so a bare counter would
defer forever and reintroduce the unbounded lifetime through the guard meant to prevent
it. Entries past the obligation TTL are dropped, and a second outstanding obligation
under an id that already has one is refused rather than admitted, because two entries
sharing an id make retirement ambiguous.

**Response-delivery deadline.** rmcp serializes writes through a mutex held across the
pending write, so a peer that admits a request and then stops reading its response would
also block any later `close()` on the same transport. Timing out the write future drops
it, which releases the guard the way any other cancellation would.

**Ordered polling in the receive loop.** Tokio's `select!` picks a starting branch at
random for fairness. Here that means a request already sitting in the buffer can lose a
coin flip to an idle timer that has just elapsed, closing the session on a peer that did
answer inside the window. The read branch is polled first and must win that tie.
Starving the timer is the usual cost of biasing and is not a cost here: a read that is
continuously ready means the session is not idle, which is exactly when the timer must
not fire.

## Tests, and one gap stated plainly

The tests here assert observed state transitions rather than the absence of an event
within a wall-clock margin: the idle timeout cancels the root without EOF; a second
outstanding obligation under one id is refused; a reused id whose obligation is already
stale is still refused; entries past the TTL are dropped; a zero response deadline is
rejected at startup.

**No test in this PR catches the ordered-polling fix.** This was checked rather than
assumed: removing the `biased;` keyword and running the full library suite leaves all
433 tests passing. The reason is structural. Reproducing the defect requires a read that
is already ready at the same instant the idle timer elapses, and constructing that
instant deterministically requires control over the clock the transport reads. The
implementation currently mixes Tokio timers with `std::time::Instant`, so pausing Tokio's
clock does not move the obligation freshness check.

A deterministic clock seam is the follow-up, and the regression test for this fix belongs
with it. Shipping a timing test for it here would mean shipping a test with exactly the
defect the follow-up exists to remove.

## Verification

```
cargo fmt --all -- --check
cargo clippy -p khive-mcp --all-targets -- -D warnings
RUSTDOCFLAGS="-D warnings" cargo doc -p khive-mcp --no-deps
cargo test --no-fail-fast -p khive-mcp --lib
```

All clean; 433 passed / 0 failed, reproduced across three consecutive runs.

One unrelated flake was observed while checking that: `daemon::tests::
permission_denied_socket_fails_without_lifecycle_or_local_fallback` failed once under
load with `left: 1, right: 0`, then passed in three consecutive full-suite runs and in
isolation. It is not touched by this change and it also passes on the base commit. It is
noted here because an intermittent test that fails a correct implementation is worth
tracking separately.

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -995,6 +995,129 @@ fn stdio_serve_mode_for(resumed_generation: Option<u32>) -> StdioServeMode {
     }
 }
 
+/// Optional idle timeout for a stdio bridge session: when configured, no
+/// request for this long and the session closes (see
+/// [`crate::transport::CancelOnEofTransport`]), releasing its reader-pool
+/// admission and DB connection — a client that comes back simply respawns the
+/// bridge, seamlessly from its perspective.
+/// This closes only a genuinely idle session: an admitted request with a
+/// response still being written — running long, or delivered slowly to a
+/// backpressured reader — defers the close rather than being cancelled out
+/// from under it, up to the separate response-delivery bound documented on
+/// [`stdio_bridge_response_deadline_from_env`].
+///
+/// **Off unless configured, and that is a deliberate reading of existing
+/// repository law rather than caution.** ADR-091 enumerates "kill long-lived
+/// reader sessions" among its rejected alternatives, on the ground that
+/// long-lived stdio sessions are live Claude Code instances and closing them
+/// by age is a worse user experience than bounding what they hold underneath
+/// them. Closing a session after an hour of quiet is that rejected policy
+/// whatever the mechanism, because this transport has no signal that
+/// separates an abandoned pipe from a live client that simply has not been
+/// asked anything. Defaulting it on would reverse an accepted decision from
+/// inside an unrelated change, so the default is off and turning it on is an
+/// operator's explicit act — a supervised deployment, a CI harness, or any
+/// context where session churn is cheap and a pinned WAL connection is not.
+/// Making it the default requires amending ADR-091, not a different number
+/// here.
+///
+/// Set `KHIVE_BRIDGE_IDLE_TIMEOUT_SECS` to a positive number of seconds to
+/// enable it. `0`, absent, and unparsable all leave it disabled; unparsable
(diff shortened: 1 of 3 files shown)


ohdearquant:
Superseded by #2247, which carries the same work plus a scope correction to the rule itself: the exception for a repeatable write error is now conditioned on the message class as well as the writer's health, because a failed response send is only logged and leaves the peer waiting on an answer that is never coming.

PR #2246, opened 2026-08-27
perf(events): opt into pages without totals

ohdearquant:
## Summary

- Add a defaulted `EventStore::query_events_without_total` seam for callers that need event rows but not the page total.
- Make the SQLite implementation share its data query between total-aware and no-total reads, skipping `COUNT(*)` for the latter.
- Make the split implementation use the no-total path for both underlying stores while preserving merged ordering and the bounded prefix window.

## Design notes

The choice is expressed at the event read seam rather than by adding a field to `PageRequest`. The cursor remains unchanged, existing callers retain total-aware behavior through the default, and only the optimized paths opt out.

`ForwardingEventStore` remains on the default because the daemon `QueryEvents` frame has no separate no-total choice; changing that wire protocol is outside this change.

The merged offset refetch remains unchanged. It was observed as a separate cursor-paging concern and is left for a follow-up.

## Validation

- `cargo clippy -p khive-storage -p khive-db -p khive-runtime --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p khive-storage -p khive-db -p khive-runtime --no-deps`
- `cargo test -p khive-storage -p khive-db -p khive-runtime`

### DIFF
--- a/crates/khive-db/src/stores/event.rs
+++ b/crates/khive-db/src/stores/event.rs
@@ -165,6 +165,66 @@ impl SqlEventStore {
             .await
         }
     }
+
+    async fn query_events_page(
+        &self,
+        filter: EventFilter,
+        page: PageRequest,
+        include_total: bool,
+    ) -> Result<Page<Event>, StorageError> {
+        let namespace = self.namespace.clone();
+        let limit_i64 = i64::from(page.limit);
+        let offset_i64 = i64::try_from(page.offset).map_err(|_| StorageError::InvalidInput {
+            capability: StorageCapability::Events,
+            operation: "query_events".into(),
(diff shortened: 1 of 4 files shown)


ohdearquant:
Closing — this is no longer needed.

#2268 removed the discarded `COUNT(*)` from the paged event read unconditionally: `query_events` now returns `total: None` on every call, pinned by a dedicated regression test, and the merged event plane propagates `None` rather than inventing a number. `count_events` remains for callers that genuinely want a cardinality.

With no total computed on the paged path, there is nothing for an opt-in no-total variant to opt out of. The branch is retained.

PR #2247, opened 2026-08-27
fix(mcp): bound the lifetime of an abandoned stdio bridge session

ohdearquant:
An abandoned stdio bridge session — a client whose pipe never closes but which has
stopped sending — is otherwise indistinguishable from a live one, so nothing reaps
it and it holds its reader-pool admission and DB connection for as long as the
parent process happens to survive. This bounds that lifetime, and it bounds the
symmetric case where a session stays alive but can no longer answer.

## The rule

**A failed outbound write closes the session, unless the writer is still healthy
AND the loss is observable by someone.** Both halves are load-bearing and the
second one is the correction that this branch exists to carry.

A write error saying the operation may simply be repeated leaves the writer
usable. Tokio's blocking stdout restores its idle state and puts the writer back
before returning the flush result, and the retry macro that repeats interrupted
operations is not applied to that branch, so an interrupted flush arrives as an
error from a writer that can still carry the next message. Cancelling there would
trade a lost message for a healthy session.

That is not sufficient on its own, because the three outbound message classes are
not handled alike:

- a failed server-initiated **request** send is handed back to the local responder
  awaiting it (`rmcp-1.8.0` `src/service.rs:1066-1073`)
- a failed **notification** send does the same (`:1074-1093`)
- a failed **response** send is only logged (`:1095-1112`)

For the first two, something in the process learns the message was lost. For a
response, nothing does: no bytes reach the peer, no local caller is waiting, and
the serve loop keeps running, so the client that asked the question waits on an
answer that is never coming and cannot distinguish that from a slow one. A healthy
writer does not help it. Closing is what turns the loss into an end-of-input it
can act on.

So the exception is scoped to non-response writes.

## Also here

A response write that never resolves pins the session, because the underlying
transport serializes writes through a mutex held across the pending write, which
also blocks any later close. A response deadline bounds the write and drops the
future rather than merely cancelling a token, which releases that guard.

Idle reaping treats a quiet window as end-of-input only when no admitted request
is still awaiting its response, and an outstanding request counts as evidence of
life only while it is fresh — a handler that dies without answering would
otherwise defer the close forever, reintroducing the exact unbounded lifetime this
change exists to remove.

## Verification

The interrupted-response test differs from the interrupted-notification test in
the message class and nothing else, so the pair is what makes the scope
load-bearing rather than incidental. Short-circuiting the class check reddens
exactly that one test and leaves the other 439 green, which was the prediction
stated before the run.

`cargo fmt --all -- --check`, `cargo clippy -p khive-mcp --all-targets -D warnings`,
`RUSTDOCFLAGS="-D warnings" cargo doc -p khive-mcp --no-deps` and `deno fmt --check`
all clean; `cargo test -p khive-mcp` gives 440 unit and 145 integration passing.

## Known gap, stated

The response deadline covers responses this transport writes. It does not cover
parse-error responses, which the underlying line transport writes directly through
its own framed writer without passing through here.

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -995,6 +995,129 @@ fn stdio_serve_mode_for(resumed_generation: Option<u32>) -> StdioServeMode {
     }
 }
 
+/// Optional idle timeout for a stdio bridge session: when configured, no
+/// request for this long and the session closes (see
+/// [`crate::transport::CancelOnEofTransport`]), releasing its reader-pool
+/// admission and DB connection — a client that comes back simply respawns the
+/// bridge, seamlessly from its perspective.
+/// This closes only a genuinely idle session: an admitted request with a
+/// response still being written — running long, or delivered slowly to a
+/// backpressured reader — defers the close rather than being cancelled out
+/// from under it, up to the separate response-delivery bound documented on
+/// [`stdio_bridge_response_deadline_from_env`].
+///
+/// **Off unless configured, and that is a deliberate reading of existing
+/// repository law rather than caution.** ADR-091 enumerates "kill long-lived
+/// reader sessions" among its rejected alternatives, on the ground that
+/// long-lived stdio sessions are live Claude Code instances and closing them
+/// by age is a worse user experience than bounding what they hold underneath
+/// them. Closing a session after an hour of quiet is that rejected policy
+/// whatever the mechanism, because this transport has no signal that
+/// separates an abandoned pipe from a live client that simply has not been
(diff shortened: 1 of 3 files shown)


ohdearquant:
The complexity counter removed in this branch incremented once per method call rather than per element visited, so it measured call volume and could not detect the growth it was written to bound — the assertion held for any implementation. A guard that observes element access instead is designed in #2274. Capacity rejection and the fail-closed poisoned-lock behaviour remain covered here.

oceanwaves630 (approved):

Issue #2204, opened 2026-08-24
events-split follow-ups: per-page COUNT, per-read socket dial, duplicated freeze loop


ohdearquant:
Three follow-ups deferred from the events-daemon split series (ADR-170, parts 1-4: #2190/#2195/#2202/#2203, superseding #2152).

> **Items 1 and 3 were re-derived at `e76835c4` and the numbers below are the corrected ones.**
> Derivations, with controls: [item 3](https://github.com/ohdearquant/khive/issues/2204#issuecomment-5448244322),
> [item 1](https://github.com/ohdearquant/khive/issues/2204#issuecomment-5449982655).
> The original text of both items understated the scope; item 1 additionally named a consumer-side
> fallback that does not exist in production code.

1. **Unconditional `COUNT(*)` on paged reads** — every paged read runs a `COUNT(*)` with no `LIMIT`
   before its data query, so it scans the whole filtered set while the data query fetches only
   `offset + limit` rows. Six paths, not one: `khive-db/src/stores/event.rs:1281`,
   `graph.rs:2085` and `:2214`, `note.rs:1310` and `:1367`. `entity.rs:897` is already conditional
   and is the shape to follow (`skip_total` at `:883`, leaving `Page.total` as `None`).
   A merged read pays the cost twice, once per underlying store.

   The constraint that decides the design: the merged path folds the two totals as
   `(Some(a), Some(b)) => Some(a + b), _ => None` (`khive-runtime/src/events_split.rs:1829-1832`),
   so a `None` from either store erases the total for the merged read. The skip has to be decided
   where the caller's intent is known and applied to both sides together, not added per-store.

2. ~~**Fresh socket per read**~~ — **DECLINED 2026-08-28, measured.** Each forwarded daemon read
   dials a fresh Unix socket. That is deliberate and documented: `EventsSplitClient` states that
   round-trips each own a fresh connection so "concurrent reads and audit flushes never queue
   behind a single stalled call", i.e. it is a tail-latency property, and a single pooled
   connection reintroduces exactly that head-of-line blocking.

   The whole prize for giving it up is **57.3 µs per read operation** (not per row): AF_UNIX
   connect+close 52.0 µs, fresh-connection round-trip 62.8 µs, already-open round-trip 5.4 µs,
   3000 samples per arm. That is 91% of an *echo* round-trip, which overstates the real share
   because the production round-trip also runs a SQLite query — against plausible round-trips it
[...]

PR #2266, opened 2026-08-28
refactor(storage): extract the shared WAL sidecar freeze helper

ohdearquant:
Addresses item 3 of #2204.

## What

Five crates each carried a byte-identical copy of the fixture that makes a SQLite database's `-wal`/`-shm` sidecars read-only, and thirteen further call sites wrote the same sequence inline. This moves the helper into a single `test_support` module in `khive-storage`, behind an off-by-default `test-support` feature, and updates every caller.

Definitions removed (5): `khive-db/src/backend.rs`, `khive-mcp/src/components.rs`, `khive-mcp/src/serve.rs`, `khive-mcp/src/server.rs`, `kkernel/src/code_audit.rs`.
Inline sites converted (13): across `khive-runtime`, `khive-pack-knowledge`, `khive-mcp`, `khive-db`, `kkernel`.

## Why khive-storage

It is the only crate every consumer depends on. `khive-db` reads as the more natural home given the helper's subject, but `khive-pack-knowledge` does not depend on `khive-db`, so a helper there could not reach two of the inline sites. A `#[cfg(test)]` module cannot cross crate boundaries at all, which is why this is a feature rather than a test-cfg. Consumers take it as a dev-dependency with `features = ["test-support"]`.

## Behaviour

Unchanged. The helper body is the existing one; four of the five copies were byte-identical under whitespace normalisation and the fifth differed only in `&Path` versus `&std::path::Path`.

23 files changed, 54 insertions, 270 deletions.

### DIFF
--- a/crates/khive-db/Cargo.toml
+++ b/crates/khive-db/Cargo.toml
@@ -56,7 +56,7 @@ windows-sys = { version = "0.61", features = [
 tokio = { workspace = true, features = ["full", "test-util"] }
 rusqlite = { workspace = true, features = ["bundled", "column_decltype", "hooks", "limits", "functions"] }
 bytes = "1"
-khive-storage = { version = "0.8.0", path = "../khive-storage" }
+khive-storage = { version = "0.8.0", path = "../khive-storage", features = ["test-support"] }
 khive-types = { version = "0.8.0", path = "../khive-types", features = ["serde"] }
 uuid = { workspace = true }
 criterion = { workspace = true }
--- a/crates/khive-db/src/backend.rs
+++ b/crates/khive-db/src/backend.rs
@@ -912,27 +912,8 @@ mod tests {
     use super::*;
     use khive_storage::types::{SqlStatement, SqlValue};
 
-    /// A writable fixture backend can leave `-wal`/`-shm` sidecars behind at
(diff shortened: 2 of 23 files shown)


ohdearquant:
Gate results for this branch.

- `cargo clippy --workspace --all-targets -- -D warnings` — pass
- `cargo test --workspace` — 445 passed, 1 failed

The single failure is `exec::tests::build_local_fallback_server_installs_blob_store_single_backend`, and it is **not caused by this branch**. It is an environment interaction filed separately as #2267: a shell whose `KHIVE_PACKS` names `formal` or `moodboard` crosses a default-feature test build that does not contain them. Two-arm control run against the same build, both arms in one session:

- with `KHIVE_PACKS` naming the optional packs: exit 101, `unknown pack name "formal"`
- `env -u KHIVE_PACKS`, same test: exit 0

Both arms are recorded on #2267. The change in this PR is test-only and touches no pack wiring.

Verification of the extraction itself, re-derived in this worktree rather than taken from the brief: 0 remaining local definitions of the helper, 0 remaining inline sites, and 40 references made up of the 22 original call sites, the 13 converted inline sites, and 5 `use` imports.

PR #2270, opened 2026-08-28
fix(mcp): route pack-scoped rows through the pack's assigned backend

oceanwaves630:
Assigning the comm or knowledge pack its own backend via `[packs.<name>]` exposed three gaps:

## Knowledge concept tier wrote to the wrong backend

`learn`/`cite`/`topic` (and compose's KG-entity blend in `search.rs`) called entity read/write APIs on the pack runtime directly. On a secondary-assigned backend those concept entities and `introduced_by` edges landed in the pack's own file — self-consistent from inside the pack, but invisible to the shared graph. This is the same seam ADR-083 closed for the session pack; all concept-tier sites now route through `core()`.

## Outbox delivery loops were blind to a comm-assigned backend

The email/telegram outbox loops scanned and marked outbound `message` notes through the generic wire `list`/`update` verbs (kg-routed) and held the kg runtime for the `external_id` claim. With `[packs.comm]` assigned, `comm.send` writes the outbox row into the comm backend while the wire scan lists the main one — outbound delivery goes permanently silent. The loops now use non-wire owner-side runtime APIs (`list_undelivered_outbound_messages`, `mark_outbound_message_delivered`, plus the existing claim) on the comm pack's runtime, and delivery admission for both directions keys on that runtime's writability. The scan preserves the wire predicate exactly: outbound, absent-or-null `delivered_at`, newest-first, 200-row result cap, 10k scan cap.

## No way to exclude a pack's rows from embedding

New optional `[packs.<name>] no_embed = true` strips the embedder set from that pack's runtime: its rows get FTS and metadata only — no `vec_*` rows, no ANN participation. Intended for packs whose rows are structural rather than retrieval targets. The opt-out covers the runtime's `core()`-routed writes too (documented), since `core()` inherits the runtime's embedder set.

## Tests

- Two-backend concept-tier regression: entity must land on main and not the pack backend; listing/citing resolve through `core()`.
- Comm-secondary outbox end to end: the wire list is provably blind to the outbox row while the non-wire scan finds, claims, sends, and marks it — all in the comm file, zero rows on main.
- Scan predicate/ordering/limit unit test (null `delivered_at`, inbound, soft-deleted exclusions).
- Per-pack `no_embed` with a must-match control (kg runtime keeps its embedders).
- Mixed read-only topology admission updated to the comm-keyed contract.

Single-backend mode is zero-change: every pack runtime is the same runtime there, and `core()` is a self-clone.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2219,6 +2219,7 @@ dependencies = [
  "criterion",
  "inventory",
  "khive-brain-core",
+ "khive-db",
  "khive-fold",
  "khive-fusion",
  "khive-pack-brain",
--- a/crates/khive-mcp/src/serve.rs
+++ b/crates/khive-mcp/src/serve.rs
@@ -3735,10 +3735,12 @@ fn build_pack_runtime(
     main_backend: &Arc<StorageBackend>,
     main_runtime: &KhiveRuntime,
 ) -> KhiveRuntime {
-    let rt = KhiveRuntime::from_backend(backend, rt_config);
+    // Every pack runtime carries main's embedder wiring for core(): a
+    // main-assigned pack has no core pointer, but with `no_embed` its own
+    // registry is empty and core-routed concept writes must still embed.
+    let rt = KhiveRuntime::from_backend(backend, rt_config).with_core_embedders_from(main_runtime);
     if backend_name != BackendId::MAIN {
         rt.with_core_backend(main_backend.clone())
-            .with_core_embedders_from(main_runtime)
     } else {
         rt
     }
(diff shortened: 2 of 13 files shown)


ohdearquant (approved):
Reviewed the full diff: comm-routed outbound delivery via owner-side runtime APIs, core()-routed knowledge concept-tier writes (ADR-073/083), no_embed runtime opt-out, and docs. Regression coverage on secondary-backend topologies is in place. LGTM.

ohdearquant (approved):
Re-reviewed the fix delta at this head: config fingerprinting, core embedder wiring on both runtime arms, channel-scoped outbox scan, and terminal delivery outcomes all verified with regression coverage. LGTM.

PR #2279, opened 2026-08-29
docs(adr-130): amendment 2 — cause classification and server-paced retry

oceanwaves630:
## What

ADR-130 currently fixes every per-backend failure cause to a single `kind` value,
`backend_error`, and requires `error.retryable=false` unconditionally.

The single-constant vocabulary cannot distinguish a deadline from a fault, and
the record demonstrates this against itself: its own degraded-empty wire example
types a leg as `backend_error` while that leg's message reads `backend search
timed out after 5000ms`. One value is already carrying two meanings.

The unconditional `retryable=false` then instructs a compliant caller not to
retry a timeout that would likely have succeeded.

## The amendment

1. **`kind` becomes a closed two-value vocabulary** — `timeout` | `backend_error`.
   Classification is structural and MUST NOT be derived by matching text in a
   rendered message, and it is computed *before* Amendment 1's diagnostic
   truncation so an omitted cause cannot change the answer.
2. **Conditional retryability** — `error.retryable` MAY be `true` only when every
   retained *and omitted* failed leg classified as `timeout`. Any mixed or
   non-timeout failure keeps it `false`. `retryable=true` requires a positive
   `retry_after_ms`.
3. **A direct answer to the recorded load-amplification arithmetic**, not a
   restatement of the benefit — see below.
4. **Admission control as a requirement rather than advice**, with each
   obligation assigned to the party that can discharge it.
5. A corrected wire example for the all-timeout case.
6. Verification rows.

## On the amplification argument

The existing retry contract records that a naive client turns 50 logical
searches/sec into as many as 150 attempts/sec. That figure is not disputed here.
Three things are said about it instead:

- **It assumes immediate retry.** The tripling holds only if the three retries
  are issued without delay so that attempts for one logical request overlap.
  Under mandatory backoff they are spread across the backoff window and decay.
  With `retry_after_ms` the server — the party that knows it is degraded — sets
  that pace rather than the client.
- **A boolean was never the load control.** The section concedes the decisive
  point itself: `retryable=false` "reduces that risk but cannot eliminate it for
  clients that ignore the field." The client in the amplification scenario is
  precisely one that retries every `ok=false`, i.e. one ignoring the field. So
  the protection is unavailable exactly where it is needed, while the cost of
  the unconditional `false` falls entirely on well-behaved callers.
- **The outage case gets strictly better.** Circuit-breaker admission suppresses
  attempts *including first attempts*, so during a sustained outage offered load
  falls **below** the 50/sec baseline, where today it stays at 50/sec plus
  whatever field-ignoring clients add.

What this does **not** claim: that typed causes are more accurate and therefore
justify retrying (true, and it does not answer the argument), or that clients
will comply (which is why the pacing obligation sits on the server).

## Party assignment in the admission-control section

The three admission requirements are client-side behaviours, so stating them as
a condition on the emitting surface would have been unsound. They are split:

- **The retrying client** implements the bounded budget, jittered exponential
  backoff (never shorter than `retry_after_ms`), and the circuit breaker.
- **The emitting surface** MUST NOT advertise `retryable=true` unless it
  publishes that policy, and MUST emit a `retry_after_ms` derived from its own
  degraded state rather than a fixed constant. A surface that cannot publish all
  three keeps `retryable=false`.

The server cannot enforce client backoff and the amendment does not claim it
can; a non-conforming client is already ignoring the field today.

## Scope of body edits

The amended body passages carry pointers at their own location, not only inside
the amendment — the completeness contract, the retry contract, the
degraded-empty wire example, the compatibility note, and the retry-policy
verification row. A reader who lands mid-document does not copy a superseded
requirement.

`retryable=false` remains the default and the safe value throughout.

Docs only; no code changes.

### DIFF
--- a/docs/adr/ADR-130-search-response-completeness-and-ranking-evidence.md
+++ b/docs/adr/ADR-130-search-response-completeness-and-ranking-evidence.md
@@ -327,6 +327,13 @@ release":
   boolean. A caller still sending `min_score` receives an explicit
   unsupported-field error rather than a silent reinterpretation; a caller still
   reading `score` finds it absent rather than redefined.
+  **Amendment 2's wire changes ship in this release** — the two-value
+  `backend_errors[].kind` vocabulary and `retry_after_ms`. They are not eligible
+  for v0.8.0: Amendment 1 shipped in v0.8.0 with `kind` fixed to the single
+  constant `backend_error`, so a strict reader may have pinned that value.
+  Widening a closed vocabulary that a released reader validates is a breaking
+  change for that reader, and takes the next release rather than a point update.
+  Amendment 1 could be introduced additively; Amendment 2 cannot.
 
 Per consumer class:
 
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
Amendment 2 argument now scoped exactly: backoff bounds the burst, the breaker is the sustained-case control with its reach stated to the conforming population, and the release gating is recorded in three agreeing places with the fixture sets disagreeing by design.

PR #2280, opened 2026-08-29
docs(adr-080): widen the oversized-line rule for terminators past one read window

oceanwaves630:
## The gap

The session mirror's line-tail rules split oversized lines into two cases:

- a **complete** over-cap line (terminated by `\n`) — skipped, cursor advanced past it, so
  "ingestion cannot wedge on one oversized line";
- a line crossing the cap with **no terminator yet** — cursor left at that line's start, so
  the next pass repeats the same bounded read rather than an unbounded tail scan.

That split is exhaustive only under an assumption the record does not state: that a line's
terminator is either found within one bounded read window or is genuinely not there.

A third state exists — **a complete line whose terminator lies several reader windows past
the cap**. It belongs to the first case by its own definition, since it is terminated, but
it is indistinguishable from the second case at the moment of the read, because one bounded
window is the only evidence available.

Applying the recorded second-case rule to it never advances the offset: same read, same
conclusion, every pass, forever. That is precisely the wedge the first case exists to
prevent, reached through the branch written to be the safe one — and every later record in
the file is unreachable for as long as that line is the frontier.

## The change

Section 7 **widens** the second case rather than correcting it. Behaviour for a
still-growing final line and for a truncated tail is retained; what changes is that "leave
the cursor at the line's start" is replaced by a rule that also terminates for a complete
line spanning several windows:

- **Checkpointed skip.** On an over-cap read ending without a terminator, the cursor
  advances by the bounded bytes discarded in that call and is persisted, though no event was
  parsed. Progress is monotonic in every pass that reads bytes, which is what makes the case
  terminate.
- **The mid-line marker is derived, not stored** — a cursor whose preceding byte is not
  `\n`. Offset zero is a line boundary by definition and has no predecessor to read.
- **No remainder is promoted to a record.** The resumed pass starts in skip mode, so an
  under-cap remainder is not parsed as an independent JSON record.
- **Bounds are unchanged.** No pass reads to EOF searching for a terminator; a truncated
  tail with no new bytes is an EOF no-op.

## Two properties recorded deliberately

The derived marker is the load-bearing choice, not an optimization. A stored "mid-line" flag
and the stored offset are two facts that can disagree — after a crash between the two writes,
or after a later change that updates one path and not the other — and the disagreement is
silent, because both values are individually well-formed. Reading the byte before the cursor
cannot disagree with the cursor.

The skip-mode resume prevents a failure strictly worse than the wedge: without it the tail of
a discarded line could be admitted as a legitimate event, silently.

## Pointers at the superseded text

Rather than only describing the change in the new section, the amended passages carry the
reference where a reader lands:

- the general "advances the cursor to the last complete line boundary" statement now names
  the exception;
- the unterminated-line paragraph carries the amendment pointer and says why.

The cap itself, the `tracing::warn!` with file and byte offset, the never-parse rule, and the
complete-line skip-and-advance path are all unchanged.

Docs only; no code changes.

### DIFF
--- a/docs/adr/ADR-080-session-pack-oss-storage-mechanism.md
+++ b/docs/adr/ADR-080-session-pack-oss-storage-mechanism.md
@@ -222,8 +222,10 @@ assigned backend is a read-only snapshot.
 | claude.ai export | `claude_ai_export` | `<exports dir>/**/conversations.json` (JSON array) | whole-file     |
 
 **Line-tail** (append-only JSONL): each pass reads from the file's stored byte offset,
-parses complete lines, and advances the cursor to the last complete line boundary. A file
-whose length has not grown past the cursor is skipped without being opened.
+parses complete lines, and advances the cursor to the last complete line boundary — with
+one exception added by §7, where a line already known to exceed the cap checkpoints a
+bounded discarded prefix mid-line. A file whose length has not grown past the cursor is
+skipped without being opened.
 
 A single line is never buffered past a hard per-line byte cap (`MirrorLimits::max_line_bytes`,
 PACKSESSION-AUD-003): a complete line (terminated by `\n`) over the cap is skipped —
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
The oversized-skip advance is now a claiming advance with the strand-safety justification stated where the dependency lives, and the verification arm exercises both dispatch shapes so the empty-advance regression cannot pass vacuously.

Issue #560, opened 2026-07-04
ci: no-stub guard is a partial clippy check, not the requested placeholder scanner (from #228)


ohdearquant:
Closed-issue audit (2026-07-03, batch07). #228 asked for a CI gate against stub/placeholder code.

**What landed**: `scripts/ci.sh:20` adds a "No-Stub Guard"; `ci.sh:27-31` denies `clippy::todo`, `clippy::unimplemented`, and `clippy::dbg_macro` for lib/bin targets; `.github/workflows/ci.yml:58` runs it.

**What's missing**: the issue also asked for detection of `unreachable!("...stub...")` and other obvious placeholder markers — no such scanner exists. Shipping source still contains stub/TODO markers, e.g. `crates/khive-retrieval/src/replay/engine_replay.rs:27` and `crates/khive-retrieval/src/hybrid/cross_encoder.rs:39`.

**Secondary drift**: ADR-066 names this a "no-stub guard" under the CI matrix (`docs/adr/ADR-066-autonomous-merge-pipeline.md:62-65`) and elsewhere calls it a required status *context* (`ADR-066:239-241`), but the actual required contexts are the matrix `CI` jobs, not a separate "No-stub guard" context (`scripts/apply-autonomous-merge.sh:106-117`). Align the ADR's naming/context model with what's actually gated, and extend the marker coverage to placeholder-string `unreachable!`/`panic!` calls.

ohdearquant:
Resolved on main by 4f4c80e6 (#1102): the placeholder-message scanner (scripts/lint-stub-markers.sh) with self-test now runs as the first CI phase, complementing the clippy restriction lints.

PR #2283, opened 2026-08-30
test(harness): isolate ambient build configuration

### LINKED ISSUE #2249: smoke_test.py asserts the DEFAULT pack set while inheriting the caller's KHIVE_PACKS
`tests/smoke_test.py` asserts the default pack set is exactly 12:

```python
assert len(verbs_result["pack_counts"]) == 12, (
    "default config loads 12 packs; pack_counts must name each once: ..."
)
```

but it spawns `kkernel` with the caller's environment inherited wholesale:

```python
env = {**os.environ, "KHIVE_NO_DAEMON": "1"}
```

`{**os.environ}` appears at nine spawn sites in the file. Neither `HOME` nor `KHIVE_PACKS` is
overridden anywhere in it. Since `KHIVE_PACKS` is the documented way to choose the loaded pack
set, any developer or host that exports it fails this assertion with a message that reads like a
product regression:

```
AssertionError: default config loads 12 packs; pack_counts must name each once:
['blob', 'brain', 'code', 'comm', 'formal', 'git', 'gtd', 'kg', 'knowledge',
 'memory', 'moodboard', 'schedule', 'session', 'workspace']
```

That output is the environment's pack list, not a defect in the tree.

## Why this is worth fixing rather than documenting

The failure direction here is the benign one — it fails loudly and stops. The concerning property
is the inverse: the test asserts a claim about **default** configuration while reading whatever
configuration the caller happens to have. An environment that removed a pack, or that set
`KHIVE_PACKS` to exactly twelve different packs, would satisfy `len(...) == 12` and the assertion
would pass without having tested the default set at all. The assertion's subject is not the thing
it reads.

CI does not export `KHIVE_PACKS`, so this is invisible there by luck rather than by construction.

`tests/contract_test.py` already does the isolation this file lacks — it sets `HOME` to a
temporary directory around the run and restores it afterwards.

## Suggested fix

Neutralize the pack-selection inputs at the spawn sites rather than at the assertion: build the
child environment explicitly, dropping `KHIVE_PACKS` and pointing `HOME` at a temporary directory,
so the assertion measures the default set by construction. Asserting the exact twelve names rather
than only the count would also make the failure message say which pack is unexpected.

## Reproduction

```
export KHIVE_PACKS=kg,gtd,memory,comm,schedule,session,workspace,blob,git,knowledge,brain,code,formal,moodboard
python3 tests/smoke_test.py     # fails the len(...) == 12 assertion
env -u KHIVE_PACKS python3 tests/smoke_test.py   # passes
```

Both arms were run against a release build; the second passes end to end.

### LINKED ISSUE #2251: ci.sh cannot complete with a non-default CARGO_TARGET_DIR; contract-tests aborts and the last four phases are silently skipped
## Summary

`scripts/ci.sh` cannot complete when `CARGO_TARGET_DIR` points anywhere other than
the default `crates/target`. The `contract-tests` phase aborts, `ci.sh` exits
non-zero, and because `run_all` is a plain loop under `set -e`, every phase after
it is silently skipped.

## Mechanism

`tests/contract_test.py` resolves the binary under test as:

```python
BINARY = os.environ.get(
    "KKERNEL_BINARY",
    os.path.join(
        os.path.dirname(__file__), "..", "crates", "target", "release", "kkernel"
    ),
)
```

(`tests/contract_test.py:59-64`)

It consults `KKERNEL_BINARY` and then a hardcoded default. It never consults
`CARGO_TARGET_DIR`. So when `CARGO_TARGET_DIR` is set, the `release` phase builds a
perfectly good binary into that directory, and `contract-tests` then asserts on a
path where nothing was ever written:

```
=== Contract Tests ===
Traceback (most recent call last):
  File ".../tests/contract_test.py", line 1109, in main
    assert os.path.exists(BINARY), (
AssertionError: Binary not found at '.../tests/../crates/target/release/kkernel'.
Build with: cd crates && cargo build --release -p kkernel
make: *** [ci] Error 1
```

## Why this is worse than a plain failure

The four phases declared after `contract-tests` in `run_all` — `deno-tests`,
`smoke-tests`, `vector-smoke`, `contract-suite` — never run, and nothing in the
output says so. The run ends with **zero failing test summaries**, because no test
ever reported a failure; the harness aborted before the tests were reached.

That makes the failure read in two contradictory and both-wrong ways depending on
what is inspecting the log:

- Anything that counts test failures reads the run as **clean**.
- Anything that diffs the declared phase list against emitted phase headers reads it
  as a failure of the code under test, rather than of the harness configuration.

Observed twice on the same day against two different revisions, with identical
tracebacks, so it is deterministic rather than incidental.

## Reproduction

```bash
cd <repo>
CARGO_TARGET_DIR=/some/other/path make ci
# -> runs through `release`, aborts in `contract-tests`, exits 2,
#    and never runs the last four phases
```

Workaround, which is what makes the bug easy to miss: set `KKERNEL_BINARY`
explicitly to `$CARGO_TARGET_DIR/release/kkernel` and re-run the remaining phases
individually via `sh scripts/ci.sh <phase>`.

## Suggested fix

Have the default honour `CARGO_TARGET_DIR` when it is set, keeping `KKERNEL_BINARY`
as the explicit override:

```python
_target = os.environ.get("CARGO_TARGET_DIR") or os.path.join(
    os.path.dirname(__file__), "..", "crates", "target"
)
BINARY = os.environ.get("KKERNEL_BINARY", os.path.join(_target, "release", "kkernel"))
```

Two things worth doing alongside it, both independent of the path fix:

1. Check the other harness scripts `ci.sh` drives for the same hardcoded assumption,
   rather than fixing only the one that happened to abort first.
2. Make `run_all` report which declared phases did not run when it exits early. A
   gate that stops partway currently reports the one phase that failed and stays
   silent about everything it skipped, which is the property that made this
   misleading rather than merely broken.

## Related

Same class as #2249 (`smoke_test.py` inheriting the caller's `KHIVE_PACKS`): a test
harness reading ambient build/runtime configuration through an assumption that holds
only for the default invocation.

### LINKED ISSUE #2267: cargo test --workspace fails when KHIVE_PACKS names an optional pack
Since `khive-pack-formal` and `khive-pack-moodboard` became off-by-default features (#2259), `cargo test --workspace` builds without them while an ambient `KHIVE_PACKS` may still name them. The test binary is then asked for a pack it does not contain.

## Observed

`cargo test --workspace` at `e76835c4`, in a shell where `KHIVE_PACKS` names `formal` and `moodboard`: 445 passed, 1 failed —

```
exec::tests::build_local_fallback_server_installs_blob_store_single_backend
```

The test itself constructs `RuntimeConfig::default()` and `KhiveConfig::default()` and never mentions packs, so the pack list reaches it from the ambient environment.

## Suspected cause, NOT yet tested

Pack resolution falls back to `KHIVE_PACKS` when no `--pack` is given (`khive-mcp/src/args.rs:54`), so the ambient value reaches a test that set no pack list of its own. Combined with the default-feature build lacking those packs, the construction fails.

I have not run the isolating A/B, so treat the cause as unverified. The decisive control is two arms of the same test in one session:

```
KHIVE_PACKS="kg,gtd,memory,comm,schedule,session,workspace,blob,git,knowledge,brain,code,formal,moodboard" \
  cargo test -p kkernel exec::tests::build_local_fallback_server_installs_blob_store_single_backend
env -u KHIVE_PACKS \
  cargo test -p kkernel exec::tests::build_local_fallback_server_installs_blob_store_single_backend
```

If only the first fails, the cause holds.

## Prior art in the tree

`khive-mcp/src/pending_events.rs` already treats ambient `KHIVE_PACKS` as a test hazard and calls `std::env::remove_var("KHIVE_PACKS")` at five sites. The defence exists; it is just not applied on this path.

## Why it is worth fixing rather than documenting

The environment that triggers it is the normal one for anyone whose shell is configured to run the full pack set, which is the configuration the local build gate itself uses. The failure reads as a code defect on `main` and is not one, so it costs a diagnosis every time someone runs the workspace suite. Options, in rough order of preference: clear the variable in the affected test the way `pending_events.rs` does; make the fallback ignore packs the binary was not compiled with, with a warning; or have the test pin its pack list explicitly.

ohdearquant:
Closes #2249
Closes #2251
Closes #2267

## Summary
- resolve the contract-test binary from the explicit override or Cargo target directory
- isolate default-pack smoke children from ambient pack and HOME configuration
- pin the fallback-server unit test to built-in packs
- report failed and skipped phases when the full CI loop exits early

## Validation
- 7 CI workflow contract tests
- 11/11 contract tests with a non-default CARGO_TARGET_DIR
- full default-pack smoke suite under a hostile ambient KHIVE_PACKS value
- exact kkernel fallback-server regression test under optional ambient packs
- cargo fmt --all -- --check
- sh -n scripts/ci.sh
- git diff --check


ohdearquant:
Three things worth addressing before this goes further. The first is the one
that undercuts the stated goal.

**1. The pinned `HOME` is bypassable by a higher-precedence channel.**

`smoke_child_env` removes `KHIVE_PACKS`, sets `HOME`, sets `KHIVE_NO_DAEMON`, and
copies the rest of the environment through (`tests/smoke_test.py:54-59`).
`KHIVE_CONFIG` is not among them — it does not appear in the file at all.

But `KHIVE_CONFIG` is bound as the explicit `--config` input
(`crates/khive-mcp/src/args.rs:71`), and an explicit config path is Tier 1 in the
loader, returning before the HOME-anchored discovery tiers are consulted at all
(`crates/khive-runtime/src/engine_config.rs:629-652`). So on a runner where
`KHIVE_CONFIG` is set, every smoke child loads that config regardless of the
pinned `HOME`, and the isolation the change is for does not hold. If the path
exists and declares backends, the child can route elsewhere or reject its
explicit `--db :memory:`; if it does not exist, the loader fails loud with
`ExplicitConfigMissing`.

Removing `KHIVE_CONFIG` from the child environment — and auditing the other
config-selection variables the same way — would close it. Worth adding a hostile
`KHIVE_CONFIG` case alongside the existing hostile-`KHIVE_PACKS` one.

**2. The resolved binary path does not survive the step boundary.**

`scripts/ci.sh:298-305` exports `KKERNEL_BINARY` in the shell running
`ci.sh release`. In the workflow, `release` and `smoke-tests` are separate `run`
steps (`.github/workflows/ci.yml:253-264`), so that export is gone by the time
smoke runs, and `tests/smoke_test.py:30-33` resolves only `KKERNEL_BINARY` or
`crates/target/release/kkernel` — it never consults `CARGO_TARGET_DIR`. With a
non-default target dir the smoke gate therefore probes a path that may be absent
or stale. The contract harness got the new resolver; the smoke harness did not.
`test_release_exports_one_binary_path_for_later_harnesses` asserts on source
text, so it cannot see a process-boundary failure.

**3. `test_full_ci_reports_failed_and_skipped_phases` asserts on strings, not
behaviour.**

It checks that three literals appear in the script
(`scripts/tests/test_ci_workflows.py:138-142`), while the behaviour lives in
`scripts/ci.sh:419-448` — exit-status capture, trap removal, phase accounting.
Changing the production condition to `if false`, or moving `trap - 0` so the
reporting trap never fires, leaves all three literals in place and the test
green. That is the reporting regression the change exists to prevent, so the
test should drive `run_all` with an injectable phase runner that fails at a known
phase and assert the failed and skipped phase names plus a nonzero status.

Smaller: `tests/contract_test.py:65-69` and `scripts/ci.sh:299-304` build the
path as `${target-dir}/release/kkernel`. With a configured build target Cargo
places it under `${target-dir}/${triple}/release/kkernel`, which
`docs/adr/ADR-026-rust-binary-packaging.md:333-341` specifically calls out. That
one fails closed (missing binary rather than a wrong pass), so it is the least
urgent of the four.

ohdearquant:
CI is red at this head on one root: `No-stub placeholder scan (before any
cargo)` on the macOS shard. The other reds are fail-fast siblings and the
aggregate gate.

The failure is caused by this branch, and it is worth being careful about how it
gets fixed.

```
self-test FAILED: ci.sh run_all must run 'no-stubs-scan' first (before any cargo
phase); found 'currentphasephase'.
```

`scripts/lint-stub-markers.sh:772-782` extracts the phase order out of `ci.sh`
and asserts the placeholder scan is first, because a guard that runs after a
cargo phase could read a tree, an allowlist, or the guard script itself after a
PR-controlled build step had already rewritten them (#560, the mutable-checkout
bypass). It does that with:

```awk
/^run_all\(\)/{f=1} f&&/for phase in/{g=1;next} g&&/do$/{exit} g{gsub(/[^a-z-]/,"");if($0!="")print}
```

That parser assumes the phase names are the lines following `for phase in`. This
branch moves the list into `run_all_phases="\ … "` above the loop and collapses
the loop to `for phase in $run_all_phases; do`, so the parser instead reads
`current_phase=$phase`, strips it to `[a-z-]`, and yields `currentphasephase`.

Reproduced both sides: the same awk against `main`'s `ci.sh` yields
`no-stubs-scan`, `lockfile`, `forward-deployed`; against this head it yields
`currentphasephase`, `runphasephase`, `done`.

**The ordering itself is fine** — `run_all_phases` still begins with
`no-stubs-scan`. So the guard is failing closed on input it can no longer parse,
which is the correct behaviour for it, rather than reporting a real reorder.

The reason to flag it rather than just note the red: the cheapest way to make
this green is to relax the parser, and a relaxed parser can stop detecting the
thing it exists for. Whatever the fix, it should come with a control that
demonstrates the guard still fails when the phase order is *genuinely* changed —
move `no-stubs-scan` after a cargo phase in a scratch copy and confirm the
self-test reddens. Without that, this becomes a check that passes and no longer
verifies the ordering guarantee.

ohdearquant:
## `No-stub placeholder scan` failure: the ordering property is intact, the check that reads it is not

CI fails at `scripts/ci.sh no-stubs-scan` with:

```
stub-marker lint: 717 file(s) OK
=== No-Stub Guard (placeholder-string scanner self-test) ===
self-test FAILED: ci.sh run_all must run 'no-stubs-scan' first (before any cargo phase); found 'currentphasephase'. Moving the placeholder scan after a cargo phase reopens the mutable-checkout bypass (#560).
```

That message reads as an ordering regression, and it is not one. The phase order this branch
ships is unchanged — `no-stubs-scan` is still first in the list, and the list is still what the
loop executes:

```sh
run_all_phases="\
    no-stubs-scan \
    lockfile \
    forward-deployed \
    ...
for phase in $run_all_phases; do
    current_phase=$phase
    run_phase "$phase"
done
```

What broke is the check's parser. `scripts/lint-stub-markers.sh` extracts the order with:

```sh
first_phase="$(awk '/^run_all\(\)/{f=1} f&&/for phase in/{g=1;next} g&&/do$/{exit} g{gsub(/[^a-z-]/,"");if($0!="")print}' "$ci_sh" | head -1)"
```

It assumes the phase list is written inline immediately after `for phase in`. This branch hoists
the list into `run_all_phases` and iterates that, so the first line after `for phase in` is now
`current_phase=$phase`, which the `gsub(/[^a-z-]/,"")` reduces to `currentphasephase`.

Running that same `awk` against both versions of the file reproduces it without CI:

```
ci.sh (base)       first_phase=[no-stubs-scan]
ci.sh (this head)  first_phase=[currentphasephase]
```

So the check asserts a source shape while claiming to assert an execution order, and this branch
changed the shape while preserving the order. It failed closed, which is the right direction.

Worth being deliberate about the repair, because the cheap version leaves the check weaker than
it looks. Re-pointing the `awk` at `run_all_phases="` would go green here, but it would then be
reading a *declaration* rather than what the loop consumes — a later change that iterates a
different variable would pass while the order actually moved. Reading the order the loop really
consumes keeps a genuine reorder failing: resolve the variable named on the `for` line, or have
`ci.sh` emit its phase order and assert on that output.

ohdearquant (commented):

Verdict on head ca78210a929e4ebbc9e949c37b894194b795464e: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head c68f8cfe3885b549d94b9eeed764e48e14c42b23: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 0035605a2c4649e38ad1d952ae6c4bd7c077280e: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

ohdearquant (commented):

Verdict on head 1a648578314207fb91dac56be62500b4b20b9a29: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Approving.

Smoke-child environment isolation is now done by prefix: `smoke_child_env`
drops every `KHIVE_*` setting inherited from the parent and lets the child
resolve its own defaults, with the one value the harness genuinely needs set
explicitly on the returned mapping afterwards. That is stronger than removing
named settings one at a time, because it does not need updating when a new
`KHIVE_*` setting is introduced.

Verified at this head rather than carried over from an earlier read: the strip
runs ahead of the single deliberate re-add, and the remaining textual mentions
of specific settings are comments and docstrings, not code paths that
reintroduce them.

One non-blocking follow-up stands, in `scripts/lint-stub-markers.sh`:
`check_ci_phase_order` relies on its callers invoking it inside a conditional.
With `set -e` active, a future direct call could exit before the producer
status is captured. Guarding the assignment with `if/then/else` or `||` would
let the helper own its own failure semantics.

Stamping, not reviewing.

PR #2284, opened 2026-08-30
ci(wasm): cache and retry wasmtime setup

### LINKED ISSUE #2020: CI: wasm-parity gate reds on a transient wasmtime download, with no retry or cache
The `wasm-parity (khive-changeset)` job installs a pinned `wasmtime` release over the network as a required setup step, with no retry and no caching:

```yaml
curl https://wasmtime.dev/install.sh -sSf | bash -s -- --version v46.0.1
"$HOME/.wasmtime/bin/wasmtime" --version
```

A transient failure fetching that release archive fails the step, and the following line then fails with exit 127 because the binary was never installed. Because `wasm-parity` is one of the jobs the aggregate CI gate requires, the whole gate goes red for whatever pull request happened to be running, regardless of its content.

## Observed

```
Fetching archive for Linux, version v46.0.1
https://github.com/bytecodealliance/wasmtime/releases/download/v46.0.1/wasmtime-v46.0.1-x86_64-linux.tar.xz
curl: (35) Recv failure: Connection reset by peer
Error: Could not download Wasmtime version 'v46.0.1'
/home/runner/work/_temp/....sh: line 3: /home/runner/.wasmtime/bin/wasmtime: No such file or directory
##[error]Process completed with exit code 127
```

The parity check itself never executed. Every other job in the same run succeeded, and the change under test modified one Markdown file. Re-running the identical workflow on the next commit passed without any change to the job.

## Why this is worth fixing

The failure mode is indistinguishable, at the gate level, from a real parity regression: the gate reports `wasm-parity: failure` and blocks the merge. Diagnosing it requires opening the job log and recognising that the failure is in setup rather than in the check. That cost is paid by whoever is unlucky enough to be running at the time, and it recurs whenever the upstream release endpoint has a blip.

Exit 127 is also a weak signal on its own, since it reports only that the binary was missing, not why the download failed. The `-e` shell option means the `curl | bash` failure does not itself abort the step, so the error surfaces one line later than its cause.

## Suggested fix

Any of these would remove the failure mode; the first two are the smallest changes:

1. Retry the install with backoff, so a single transient reset does not fail the job.
2. Cache the pinned toolchain keyed on the version string, so the download happens once per version rather than once per run.
3. Fail the step at the download itself rather than at the subsequent version check, so the log names the cause directly.

Pinning the version is correct and should stay; the problem is only that a pinned artifact is re-fetched on every run over a network that occasionally resets.

ohdearquant:
## Summary

- cache the pinned Wasmtime toolchain by runner OS, architecture, and exact version
- retry the pinned release archive download and fail at the network error after bounded retries
- verify restored and newly installed binaries report v46.0.1 before exporting them to later parity steps

## Test plan

- `uv run --no-project --python 3.11 python scripts/tests/test_ci_workflows.py`
- `pre-commit run --files .github/workflows/ci.yml scripts/tests/test_ci_workflows.py`

Closes #2020

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -475,6 +475,8 @@ jobs:
     defaults:
       run:
         working-directory: crates
+    env:
+      WASMTIME_VERSION: v46.0.1
     steps:
       - uses: actions/checkout@v7
       - uses: dtolnay/rust-toolchain@1.95.0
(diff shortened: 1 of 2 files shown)


ohdearquant:
Two items.

**1. The new test does not protect the property it is named for.**

`.github/workflows/ci.yml:511` has no condition, so verification currently runs
after both a cache hit and a cache miss — which is right. But
`scripts/tests/test_ci_workflows.py:80-82` only asserts that the *install*
step's condition appears somewhere in the job. Adding
`if: steps.cache-wasmtime.outputs.cache-hit != 'true'` to the *verification*
step would satisfy every assertion in
`test_pinned_runtime_is_cached_retried_and_verified` and still be green.

That matters because the verification step is also what appends the binary
directory to `GITHUB_PATH` (`:521`). Skipping it on a cache hit would leave the
later parity steps either failing with `wasmtime: command not found` or silently
resolving some other `wasmtime` on `PATH` — an unverified toolchain running the
gate. Asserting the verification step has no `if:`, or that it runs on both
paths, would pin the case the test name claims.

**2. Cache key and archive target are decoupled.**

`:496` keys the cache by `runner.arch`, while `:501`, `:505` and `:510` always
select the `x86_64-linux` archive and member. On an ARM64 runner the key becomes
a distinct namespace while a cold run still downloads an x86_64 binary, and the
version step then fails with an exec-format error. The current `ubuntu-latest`
runner is x64 so this is latent, but the key does not describe the
artifact-selection rule. Either pin the job to an x64 runner and key the fixed
target explicitly, or derive archive, extracted member, and cache key from one
architecture-to-target mapping.

The retry, extraction member selection, and unconditional verification all look
right otherwise.

ohdearquant (commented):

Verdict on current head (175a4f6d298460bc135a9f7b95babb03136e6107): APPROVE, zero blocking findings.

Checked: the pinned version is now compared exactly (`.github/workflows/ci.yml:527`), the release asset is selected by runner architecture with the cache key carrying OS, architecture and exact version, and the restored binary is verified unconditionally before it is exported to PATH; `scripts/tests/test_ci_workflows.py` asserts each of those properties. Hosted CI is green at this head.

This is a comment, not an approval. A human reviewer decides whether to approve and merge.

oceanwaves630 (approved):
Approved at 175a4f6d2984.

PR #2285, opened 2026-08-30
fix(release): report semver gate work honestly

### LINKED ISSUE #2254: make publish-dry is void locally: cargo-semver-checks cannot parse rustdoc v60
`make publish-dry` fails before it validates anything, on any branch, because the
local `cargo-semver-checks` cannot parse the rustdoc JSON the current toolchain
emits.

## Observed

```
Parsing kkernel v0.8.0 (current)
error: unsupported rustdoc format v60 for file: <target>/doc/khive_bm25.json
(supported formats are v56, v57)
make: *** [publish-dry] Error 1
```

## Cause

- `cargo-semver-checks` 0.48.0 accepts rustdoc JSON **v56/v57**.
- `rustc 1.98.0 (88d9e12ae 2026-08-18)` emits **v60**.
- The repository has no `rust-toolchain.toml`, so the local toolchain is
  whatever is installed and can move underneath the pinned expectation.

The gate dies while *parsing* rustdoc output, before any version comparison
runs, so it cannot reach the check it exists to perform.

## Why this is worth fixing rather than just upgrading locally

The failure is indistinguishable, at a glance, from a real semver finding — and
it is most likely to be believed exactly when it is least likely to be true. It
surfaced here on a version-bump branch, where a red semver gate reads as "this
bump is incoherent". It was not: running the same
`cargo semver-checks --package khive-bm25` on a branch with no version change
produces the identical error.

CI is unaffected. `.github/workflows/release.yml` pins
`obi1kenobi/cargo-semver-checks-action` at a commit SHA and that action installs
its own tool, which is why this drift never appears in a CI run and can sit
unnoticed locally for as long as it likes.

## Suggested fix

Have `scripts/publish.sh` assert its `cargo-semver-checks` version, or the
rustdoc format it supports, against the toolchain in use, and fail with a
message naming the mismatch rather than surfacing a parse error. A minimum
version documented alongside the publish instructions would also close it.

Pinning the toolchain is a larger decision and is not proposed here.

### LINKED ISSUE #2255: publish.sh misdescribes the SemVer exclude list; two excluded crates do have crates.io baselines
`scripts/publish.sh` introduces the SemVer exclude list with this comment:

> The exclude list (crates with no crates.io baseline yet) lives in `scripts/lib/semver_excludes.txt`

That description does not match the file's contents or the file's own stated policy.

`scripts/lib/semver_excludes.txt` currently holds six crates. Querying `https://crates.io/api/v1/crates/<name>` for each, only four of them have no baseline (HTTP 404, "crate does not exist"): `khive-wire-protocol`, `khive-pack-agent`, `khive-pack-moodboard`, `khive-repo-showcase`. The other two, `khive-channel-telegram` and `khive-pack-blob`, both return HTTP 200 with `max_version` 0.7.0, so they do have a published baseline the gate could compare against. `khive-types` returns HTTP 200 as a control in the same pass.

The file's own header states a different and broader criterion, which the two entries do satisfy:

> Crates excluded from the release SemVer gate: pre-1.0 surfaces still moving under active development. A crate leaves this list when its wire/API surface freezes.

So the entries are legitimate under the policy the file declares; it is the `publish.sh` comment that is wrong. The concern is that a future reader trying to decide whether a crate belongs on the list gets a rule ("no baseline yet") that would evict two current entries.

Two things worth separating:

1. Correct the `publish.sh` comment to describe the list the way the file itself does, so the two documents state one criterion rather than two.
2. Separately, confirm that `khive-channel-telegram` and `khive-pack-blob` are still intended to be exempt now that they have baselines, and drop them from the list if their surfaces have settled.

No release impact: a `0.x` minor bump is the breaking bump for `0.x` crates, so 0.7 to 0.8 already signals incompatibility to a dependent's resolver whether or not the gate inspects those two.

### LINKED ISSUE #2265: Release SemVer gate passes without evaluating any lint on 0.x minor bumps
The release SemVer gate passes on every release so far without evaluating a single
lint, and the passing output is not visibly different from a gate that checked
something.

## Measured on the 0.8.0 release

`scripts/publish.sh` runs, before the publish loop:

```
cargo semver-checks check-release --workspace <--exclude x6>
```

Across all 38 checked crates the per-crate result is identical:

```
Checking khive-<name> v0.7.0 -> v0.8.0 (major change)
 Checked  0 checks: 0 pass, 254 skip
 Summary  no semver update required
```

38 crates reporting `no semver update required`, 0 reporting an update required, and
**0 of roughly 9,652 possible lint instances (38 x 254) actually ran.** The gate then
prints `SemVer gate OK`.

## Why

This is correct behaviour from `cargo-semver-checks`, not a bug in it. Under Cargo's
0.x semantics the leftmost non-zero component carries major significance, so
`0.7.0 -> 0.8.0` is a *major* bump. A major bump permits breaking changes, so every
lint is skipped as inapplicable.

Which means the gate has been vacuous on every release to date, because every release
to date has been a 0.x minor bump. It will stay vacuous until either a 1.0 or a
patch-level release. The 0.8.0 tag annotation documents two deliberate breaking
changes (`BlobStore::get` replaced by `get_bounded_verified`, `Entity::with_content_ref`
removed) and the gate had nothing to say about them, correctly.

## Why it is worth recording rather than shrugging at

Two things are true at once and only the first is obvious:

1. Nothing is wrong with this release. Breaking changes are legitimate here.
2. The gate is currently providing no signal, while reading in the surrounding script
   as though it provides one. `scripts/publish.sh` treats a zero exit as "SemVer gate
   OK" and proceeds to publish 44 crates.

That second point has a concrete consequence. A gate that cannot run and a gate that
runs and skips everything both produce a pass for the calling script. #2254 is exactly
the first case: a tooling mismatch made the gate unable to parse rustdoc at all. The
two failure modes are hard to tell apart from the outside, and the exclusion list adds
a third way to reach a pass without a check.

Relatedly, the `scripts/lib/semver_excludes.txt` exemptions cost nothing at a 0.x minor
bump, since excluded and non-excluded crates alike receive zero lint evaluation. That is
worth knowing before anyone reasons about how much those exclusions give up.

## Suggested direction

Not "make the gate strict" — the skipping is correct. Rather, make the gate report what
it did, so a vacuous pass is legible:

- Have `publish.sh` surface the aggregate (`N crates checked, M lints evaluated`) and
  say plainly when `M == 0`, instead of printing an unqualified `SemVer gate OK`.
- Consider whether a 0.x minor bump wants a baseline comparison against the previous
  *minor* under different rules, or an explicit acknowledgement in the release notes
  that the API delta was not machine-checked.

The immediate ask is only the first one: distinguish "checked and clean" from "checked
nothing", because right now the release process cannot tell them apart.

ohdearquant:
Closes #2254
Closes #2255
Closes #2265

## Summary
- reject cargo-semver-checks versions older than the current rustdoc-compatible floor
- preserve checker exit classes and label incomplete tool/runtime runs separately from API findings
- aggregate per-crate evaluated lint counts and call out a zero-lint vacuous pass
- align the publish-script comment with the shared pre-1.0 exclusion policy

## Validation
- 6 SemVer helper unit tests
- publish-guard fixture suite, all 6 cases passed
- bash syntax checks for publish and guard scripts
- Python compile checks
- git diff --check

### DIFF
--- a/scripts/lib/semver_gate.py
+++ b/scripts/lib/semver_gate.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+"""Validate and summarize the local cargo-semver-checks release gate."""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+
+# v0.50.0 adds rustdoc JSON v60/v61 support, covering the repository's
+# current Rust 1.98 toolchain. Future format changes must raise this floor.
+MIN_CHECKER_VERSION = (0, 50, 0)
+_VERSION_RE = re.compile(r"\bcargo-semver-checks\s+v?(\d+)\.(\d+)\.(\d+)\b")
+_ANSI_RE = re.compile(r"\x1b\[[0-?]*[ -/]*[@-~]")
+_CHECKED_RE = re.compile(r"\bChecked\s+\[[^\]]+\]\s+(\d+)\s+checks:")
+
+
+def parse_checker_version(output: str) -> tuple[int, int, int]:
+    match = _VERSION_RE.search(output)
+    if match is None:
+        raise ValueError(
+            "could not parse `cargo-semver-checks --version` output: "
+            f"{output.strip()!r}"
+        )
+    return tuple(int(part) for part in match.groups())
+
+
+def validate_checker_version(output: str) -> tuple[int, int, int]:
+    version = parse_checker_version(output)
+    if version < MIN_CHECKER_VERSION:
+        required = ".".join(str(part) for part in MIN_CHECKER_VERSION)
+        actual = ".".join(str(part) for part in version)
+        raise ValueError(
+            f"cargo-semver-checks {actual} is too old; khive requires >= {required} "
+            "for the current rustdoc JSON generation"
+        )
+    return version
+
+
+def summarize_log(text: str) -> tuple[int, int]:
+    clean = _ANSI_RE.sub("", text)
+    selected_checks = [int(match.group(1)) for match in _CHECKED_RE.finditer(clean)]
+    if not selected_checks:
+        raise ValueError(
+            "cargo-semver-checks completed without any recognizable per-crate "
+            "`Checked ... N checks:` records"
+        )
+    return len(selected_checks), sum(selected_checks)
+
+
+def _check_version(raw_version: str) -> int:
+    try:
+        version = validate_checker_version(raw_version)
+    except ValueError as exc:
+        print(f"ERROR: {exc}", file=sys.stderr)
+        return 1
+    actual = ".".join(str(part) for part in version)
+    print(f"    cargo-semver-checks {actual} compatibility floor OK")
+    return 0
+
+
+def _summarize(path: Path) -> int:
+    try:
+        crates, evaluated = summarize_log(path.read_text(encoding="utf-8"))
+    except (OSError, ValueError) as exc:
+        print(f"ERROR: cannot summarize SemVer gate: {exc}", file=sys.stderr)
+        return 1
+
+    if evaluated == 0:
+        print(
+            f"    SemVer gate completed: {crates} crates checked, 0 lints "
+            "evaluated — VACUOUS PASS for this version delta"
+        )
+    else:
+        print(
+            f"    SemVer gate completed: {crates} crates checked, "
+            f"{evaluated} lints evaluated"
+        )
+    return 0
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(dest="command", required=True)
+
+    version_parser = subparsers.add_parser("check-version")
+    version_parser.add_argument("version_output")
+
+    summary_parser = subparsers.add_parser("summarize")
+    summary_parser.add_argument("log", type=Path)
+
+    args = parser.parse_args(argv)
+    if args.command == "check-version":
+        return _check_version(args.version_output)
+    return _summarize(args.log)
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -104,7 +104,14 @@ DELAY=10  # seconds to wait for crates.io index between publishes
 echo ""
 echo "--- Publish ladder guard (CRATES membership vs cargo metadata) ---"
 METADATA_JSON=$(mktemp)
-trap 'rm -f "$METADATA_JSON"' EXIT
+SEMVER_LOG=""
+cleanup_publish_temp() {
+    rm -f "$METADATA_JSON"
+    if [[ -n "$SEMVER_LOG" ]]; then
+        rm -f "$SEMVER_LOG"
+    fi
+}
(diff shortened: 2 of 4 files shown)


ohdearquant:
The classification work here is solid — checker exits 100/101/other stay
distinct, an unrecognizable log fails closed, and per-crate errors are not
absorbed into the aggregate. Three things stop it short of the goal in the
title, and the first two are on the publish path.

**1. The vacuous pass is a message, not a gate.**

`scripts/lib/semver_gate.py:70-81` prints

```
SemVer gate completed: N crates checked, 0 lints evaluated — VACUOUS PASS for this version delta
```

and then both branches fall through to the same `return 0`. `scripts/publish.sh:174`
calls `summarize` and, under `set -e`, a zero return means execution continues
into the publish loop.

So a run where every selected crate emitted `Checked … 0 checks:` is
indistinguishable *by exit status* from a run that actually evaluated lints, and
publication proceeds either way. The text is honest and the status is not, which
leaves the gate exactly where it was for any caller that reads the status —
which is what a release pipeline does. Zero evaluated checks should be a nonzero
result, or a distinct outcome the caller explicitly refuses to publish on.

`scripts/tests/test_semver_gate.py:42-47` asserts on the returned `(2, 0)` tuple
rather than the CLI's exit status, so it passes with this behaviour intact.

**2. A failed log capture can be masked.**

```bash
set -o pipefail
cargo semver-checks check-release --workspace "${SEMVER_EXCLUDE_ARGS[@]}" 2>&1 | tee "$SEMVER_LOG"
SEMVER_STATUS=${PIPESTATUS[0]}
set +o pipefail
```

(`scripts/publish.sh:152-155`.) `pipefail` is enabled and then discarded —
indexing `PIPESTATUS[0]` reads only the checker's status. If `tee` fails partway
through (write error, full disk) while the checker exits 0, `SEMVER_STATUS` is 0
and `$SEMVER_LOG` holds a truncated log. If the surviving prefix contains at
least one recognized `Checked` record, `summarize` returns 0 and publication
proceeds on incomplete evidence. Checking `PIPESTATUS[1]` and failing closed
before summarizing would close it.

**3. The version parser accepts prereleases as the stable floor.**

`scripts/lib/semver_gate.py:15-27` matches the first three numeric components
only. `cargo-semver-checks 0.50.0-alpha.1` parses as `(0, 50, 0)` and passes the
floor check, although a prerelease precedes `0.50.0` — the check admits exactly
the builds it exists to exclude. `0.50.0.1` also parses as `(0, 50, 0)` despite
not being a three-component version. A `+build` suffix is correctly equivalent.
Parsing the full version and comparing prerelease precedence, rejecting
malformed suffixes, would fix it; the current tests only cover output with no
numeric prefix.

Smaller: the fixture at `scripts/tests/publish-guard-test.sh:57` now passes
`__empty_ladder_sentinel__`, which is a real ladder member, so it would not
redden a mutation that wrongly accepts a genuinely empty `CRATES` array. And
`.github/workflows/release.yml:168-170` still describes the old exclusion policy
("no crates.io baseline yet") that this PR updates elsewhere, so the two
consumers now disagree.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at f6f8d73f.

PR #2286, opened 2026-08-30
test(memory): await detached ANN build signals

### LINKED ISSUE #2258: flaky: recall_1657 arm 2 relies on a zero-duration tokio timeout always elapsing
`recall_1657_degraded_outcome_captures_reason_at_failure_site` arm 2 asserts a zero-duration `tokio::time::timeout` always elapses, which it does not.

## Observed

CI on `main` at `52364b96763e79803c7f1393285a0a403184face`, attempt 1, job `CI (macos-latest, shard 1/2)`:

```
FAIL [0.112s] (2535/4725) khive-pack-memory
  handlers::recall::tests::recall_1657_degraded_outcome_captures_reason_at_failure_site

thread '...' panicked at khive-pack-memory/src/handlers/recall.rs:1888:9:
#1657 a bounded-out readiness wait must mark the per-model outcome degraded
```

`CI (ubuntu-latest, shard 1/2)` was cancelled as a consequence of that failure and `CI gate`
failed as its aggregate, so this is one failing job, not three.

## Why it is a test defect rather than a product regression

Arm 2 (`crates/khive-pack-memory/src/handlers/recall.rs:1869-1898`) calls
`collect_model_ann_hits` with `ann_ready_timeout_ms = 0` against a cold model and asserts the
bounded wait expires.

The wait is `crates/khive-pack-memory/src/handlers/common.rs:1406-1410`:

```rust
tokio::time::timeout(
    std::time::Duration::from_millis(ann_ready_timeout_ms),
    done_rx,
)
```

`tokio::time::timeout` polls the inner future before it consults the deadline. If `done_rx` is
already resolved at that first poll, the call returns `Ok(..)` no matter how small the duration
is, including zero. So the `Err(_elapsed)` arm the test requires fires only when the ensure task
has not yet completed at first poll. That is a race on scheduler and machine speed, not a
property the test can rely on.

The same tree passed this test on other runs: CI succeeded at attempt 1 on
`138bf1fb090750d7911a2fb662404db8bcbb0d44` (run 33094976847) and on
`99e2733f777bce6e656aa1b71d6494598adc1b75` (run 33096378486). Three passes and one failure over
code that did not change between them.

Caveat on the mechanism: the zero-duration behaviour above is read from the source and from
`tokio::time::timeout`'s documented poll order. It has not been demonstrated with a targeted
reproducer here, so treat the mechanism as read-at-source rather than executed.

## Suggested direction

Do not widen the timeout — a larger number makes the race rarer without removing it. Make the
un-readiness deterministic instead, for example by holding the ensure task's completion behind
the existing test failpoint machinery in `handlers::common::retrieval_failpoints` (arm 1 of the
same test already uses that path successfully) so the receiver is provably unresolved when the
deadline is checked.

Arm 1 of this test is sound and should stay as it is.

## Scope

Test-only. Nothing here indicates the degradation path itself is broken: arm 1 exercises the
retrieval-failure branch and passes.

### LINKED ISSUE #2269: flaky: recall_836 asserts a detached ANN build finishes inside a fixed 3s wall-clock budget
`recall_836_self_build_timeout_detaches_build_instead_of_dropping_it` asserts a detached background
ANN build finishes inside a fixed 3-second wall-clock budget. On a contended runner it does not.

This is a sibling of #2258, not a duplicate: that issue is `recall_1657` arm 2, where a zero-duration
`tokio::time::timeout` returns `Ok` because `timeout` polls the inner future before consulting the
deadline. The mechanism here is different — a bounded poll loop with a wall-clock budget — so fixing
#2258 does not fix this one.

## Observed

CI attempt 1 of run `33160821962`, job `CI (ubuntu-latest, shard 1/2)`, on the head of PR #2268
(`6a33ccba35d71e23073ce5c81bedbbae5cb0810d`):

```
FAIL [   3.493s] (2633/4727) khive-pack-memory
  handlers::recall::tests::recall_836_self_build_timeout_detaches_build_instead_of_dropping_it

thread '...' panicked at khive-pack-memory/src/handlers/recall.rs:1973:9:
the detached build must eventually install a fresh ANN index for
recall-836-self-build-detach-model instead of being dropped on timeout (#836)
```

The rest of that attempt was green: `Summary [177.779s] 2636/4727 tests run: 2635 passed (1 slow,
1 leaky), 1 failed`. The other three shards succeeded, and 13 of the 14 gated jobs succeeded — `CI
gate` failed only as the aggregate of this one job.

**Attempt 2 of the same run passed the same test at the same commit.** Nothing was pushed between
the two attempts, so the code under test is byte-identical across them: step 20 `Workspace tests
(nextest shard 1/2)` of job `98819023161` completed `success`. Two disagreeing outcomes over an
unchanged tree is the finding — the test's result is not a function of the code alone.

## Why it is a test defect rather than a product regression

The assertion is a fixed budget, not a signal. `crates/khive-pack-memory/src/handlers/recall.rs:1965-1976`:

```rust
let mut warmed = false;
for _ in 0..300 {
    if crate::ann::is_current(&ann_handle, &key).await {
        warmed = true;
        break;
    }
    tokio::time::sleep(std::time::Duration::from_millis(10)).await;
}
assert!(warmed, "the detached build must eventually install a fresh ANN index ...");
```

300 iterations of a 10ms sleep is a 3-second ceiling, and the observed failure took 3.493s, which is
that ceiling elapsing rather than an early abort. The comment above the loop says it polls "rather
than sleeping a fixed amount", and polling is indeed better than one long sleep, but the loop still
encodes a wall-clock deadline: it asserts that a detached task which embeds a note and builds and
installs an ANN index completes within 3s on whatever machine runs it. That is a property of the
runner, not of the code under test.

The same job log shows the runner was heavily contended at that moment: another test in the same
shard was marked `SLOW [> 60.000s]` and finished at `118.205s`. A 3s budget is not safe against that
kind of scheduling pressure. `#[serial(background_tasks)]` serializes this test against other
background-task tests, not against the ~2600 other tests nextest is running concurrently in the shard.

The test passes routinely on unchanged code. CI succeeded on `main` at `e76835c4c293`,
`457e08a18fd2`, `0d5ba4ad2150`, `52364b96763e`, `38557eff9672`, `1f223bec4db1`, `2b1074329c97`,
`3f8f17b7b68e` and `5afc78c6817b` (runs 33129359406, 33126149298, 33118429965, 33100061642,
33098602591, 33075943700, 33070887322, 33061897891, 33058217514), all of which run this test.

The PR it failed on cannot reach it. `git diff --name-only origin/main...HEAD` on that branch returns
exactly `crates/khive-db/src/stores/event.rs` and `crates/khive-db/src/stores/event_tests.rs`, and the
change is the removal of an unconditional `COUNT(*)` on the events page path. The test builds a
`KhiveRuntime::memory()`, creates one note and polls the ANN cache; it never queries events. Stated
as read-at-source: no reproducer was run to demonstrate the absence of a path, so treat the
independence argument as resting on the diff scope and the repeated passes above, not on an executed
proof.

## Suggested direction

Do not raise 300 to a larger number. A bigger budget makes the failure rarer without removing the
dependence on runner speed, and it lengthens every green run.

Wait on the build's own completion signal instead of a clock. The detached `ensure_ann_for_model`
task already knows when it has installed the index; exposing that completion to the test — a
notify/join handle, or the existing `handlers::common::retrieval_failpoints` machinery that #2258
recommends for the sibling test — turns the assertion into "the detached build completed" rather than
"the detached build completed within 3 seconds". Keeping a generous timeout as an outer failsafe is
fine, as long as the pass condition is the signal.

## Scope

Test-only. Nothing here suggests the #836 detach behaviour itself regressed: the first half of the
test — that the timed-out recall still returns FTS results marked `degraded: ann_unavailable` —
passed, and it is the later warm-up assertion that failed.

ohdearquant:
Closes #2258
Closes #2269

## Summary
- add a model-scoped test-only hook around the real detached ANN ensure task
- hold the task unresolved while proving bounded-wait degradation
- release and await the build's own completion signal before asserting installed ANN state
- remove the zero-timeout readiness race and three-second polling budget

## Validation
- exact recall_1657 failure-site regression passed under hostile ambient packs
- exact recall_836 detached-build regression passed under hostile ambient packs
- cargo clippy -p khive-pack-memory --tests -- -D warnings
- cargo check -p khive-pack-memory
- cargo fmt --all -- --check
- git diff --check


ohdearquant:
The gating here is clean, and it was the first thing worth checking: the hook module is `#[cfg(test)]` at `common.rs:31-34` and both call sites are independently gated at `common.rs:1474-1476` and `common.rs:1484-1486`, so a release build has no hook symbols at all rather than inert ones. `Notify::notify_one` retains a permit, so an `entered` or `completed` signal that fires before its waiter starts is latched, not lost. Both hazards this shape usually carries are handled.

One thing to fix before merge.

**The direct call has no test-side deadline, so the regression this test exists to catch would hang it rather than fail it.**

`recall.rs:1872-1886` installs the hook and then awaits `collect_model_ann_hits` bare. The first `tokio::time::timeout` is at `recall.rs:1888`, which is reached only after that await returns. The hook holds the detached task at `common.rs:113-118` until it is released, so the only thing making the call return at all is the production readiness bound at `common.rs:1490-1493`.

That inverts the test's purpose. If the bounded `done_rx` wait is removed or bypassed — precisely the regression the new synchronization is meant to pin down — the hooked task never sends, the direct call waits forever, and the 30-second failsafes below it are unreachable. The job then dies on the harness wall clock with no failure site, which is strictly less diagnosable than the polling loop this replaces.

Suggested fix: wrap the `collect_model_ann_hits` call at `recall.rs:1872` in a test-level `tokio::time::timeout` with a diagnostic expectation, so a missing production bound reddens this test locally instead of consuming the job.

Worth naming the general shape, because it is the same one running through this batch of changes: the detection added is correct, and the outcome for the consumer is unchanged. The useful question to put to each of these is *what does the consumer do differently now* — here the consumer is CI, and on the regression path it still gets an unexplained timeout. That question catches this; "is the check present" does not.

oceanwaves630 (approved):
Approving on the owning seat's verification at this head under the standing lane. Static reviewer APPROVE at the exact head, CI green at that head (all checks success or skipped, none pending or failed, re-read at click time; the earlier runner-infrastructure timeout has cleared on rerun), and no daemon verdict at that head. Seat gates cited and accepted: 75 recall unit tests passing and clippy clean.

PR #2287, opened 2026-08-30
test(runtime): isolate process-global test state

### LINKED ISSUE #2224: Filtered parallel test runs flake: pack audit tests intermittently observe another test's audit rows
Running a filtered subset of khive-runtime lib tests in parallel intermittently fails pack audit assertions, on current main content as well as feature branches.

Reproduce (fails ~4 of 6 runs at a checkout matching current main):

    cargo test -p khive-runtime --lib pack::tests

Observed failures rotate across the subset, all one of two shapes:

- an exact-count assertion sees 2 audit rows where 1 was appended, e.g. `pack::tests::audit_event_persists_to_event_store_on_deny` ("one audit event persisted to EventStore on deny: left: 2, right: 1"), `dispatch_unknown_verb_allowed_by_gate_still_persists_audit_row`, `link_audit_bulk_links_get_no_enrichment`;
- a degradation-observability assertion fails, e.g. `db_diagnostics_with_audit_metrics_reports_batch_failure_and_degradation` ("the config-locked row must have degraded"), `config_locked_row_degrades_without_failing_the_dispatch_that_observed_it`.

Signals:

- `--test-threads=1` passes the same subset consistently, so this is cross-test interference under parallelism, not per-test logic.
- The full `cargo test -p khive-runtime --lib` run passes consistently — the failure depends on which tests get packed onto threads together, so filtered runs hit it far more often.
- Each failing test passes in isolation.

The count-of-2 shape suggests two concurrently running tests share an audit event sink (or one test's audit batch lands in another's store); the degradation shape suggests the same sharing masks the injected failure. Worth auditing the pack test fixtures for a process-shared resource (event plane sidecar path, shared writer, or global) that two tempdir-isolated tests can still collide on.

### LINKED ISSUE #2250: Six non-serial tests mutate the process-global SPAWN_COUNT that serial tests assert on
## Summary

`crates/khive-mcp/src/daemon.rs` has six tests that call
`spawn_daemon_with_exe_and_config` directly and are declared as plain `#[test]`,
with no `#[serial]`. That function contains the only increment site for the
process-global `SPAWN_COUNT` counter (`daemon.rs:743`). Several `#[serial]`
tests in the same file assert on the absolute value of that counter, so those
assertions can observe spawns that were performed by an unrelated test running
concurrently.

## Why `#[serial]` on the asserting test is not enough

`#[serial]` establishes mutual exclusion only against other `#[serial]` tests.
A bare `#[test]` is not part of that group and is free to run at the same time.
So a `#[serial]` test that resets the counters and then asserts
`SPAWN_COUNT == 0` is protected from the 62 other `#[serial]` tests in the file
and from none of these six.

## The six call sites

All in `crates/khive-mcp/src/daemon.rs`, each `#[test]` with no `#[serial]`:

| line | test |
| --- | --- |
| 3344 | `spawn_daemon_with_exe_and_config_appends_config_flag_to_command_line` |
| 3373 | `spawn_daemon_with_exe_and_config_appends_pack_flags_to_command_line` |
| 3401 | `spawn_daemon_with_exe_and_config_omits_pack_flags_when_none` |
| 3428 | `spawn_daemon_with_exe_and_config_appends_memory_db_flag_to_command_line` |
| 3467 | `spawn_daemon_with_exe_and_config_forwards_concrete_db_flag_to_command_line` |
| 3494 | `spawn_daemon_retries_a_transient_executable_file_busy_error` |

For contrast, the same file uses `#[serial]` on 62 tests, so the attribute is
established practice here and these six are the exceptions rather than the norm.

## Observed symptom

`daemon::tests::parallel_parse_failure_is_terminal_and_never_recovers` failed
once under a full-parallelism `cargo test -p khive-mcp --features
channel-email,channel-telegram` run, on an assertion that the spawn counter is
zero:

```
left: 4, right: 0
```

The test is itself `#[tokio::test(flavor = "multi_thread", worker_threads = 12)]`
plus `#[serial]`, and it calls the shared counter reset before asserting. An
observed value of 4 against six candidate bumpers is consistent with a subset of
them overlapping the window between the reset and the assertion.

## Status of the diagnosis

The structural facts above are read from source and are not in doubt: one
increment site, six non-`#[serial]` callers, an asserting test whose exclusion
group does not include them.

The causal claim — that this specific interleaving produced this specific
failure — is **not proven**. It has not been reproduced: the same test passed
8/8 under full parallelism and 25/25 in isolation on the same binary, so the
rate is low and 0/8 bounds it no tighter than roughly 31% at 95% confidence.
Filing the structural defect on its own merits; the reproduction is a separate
question and may not be worth chasing.

## Suggested fix

Add `#[serial]` to the six tests listed above, so that every test which mutates
the shared counter is in the same exclusion group as every test which asserts on
it. An alternative that removes the shared mutable state instead of scheduling
around it would be to make the counter per-test rather than process-global, but
that is a larger change to the harness.

ohdearquant:
## Summary

- serialize all six direct daemon spawn-seam tests with the existing process-global counter group
- serialize every event-store-backed runtime pack test with the config_ledger group
- add source-census regression tests so future seam callers cannot omit the required lock

## Reproduction

Before this patch, a 16-thread runtime pack run failed with cross-test config-ledger rows, and the second 16-thread daemon run left SPAWN_COUNT at 4 in permission_denied_socket_fails_without_lifecycle_or_local_fallback.

## Test plan

- [x] khive-runtime pack::tests with 16 threads, 10 consecutive repetitions
- [x] khive-mcp daemon::tests with email and Telegram features and 16 threads, 3 consecutive repetitions
- [x] khive-runtime library suite: 1,416 passed, 7 ignored
- [x] khive-mcp library suite with email and Telegram features: 499 passed
- [x] cargo fmt --all -- --check
- [x] cargo check --workspace --all-targets with email and Telegram features
- [x] cargo clippy --workspace --all-targets with warnings denied and email and Telegram features
- [x] cargo test --workspace regular test targets, followed by cargo test --workspace --doc with the explicit Rust toolchain path

Closes #2250
Closes #2224


ohdearquant:
The serialization work itself checks out: the six daemon tests take their guard before the first spawn, the retry test joins its release thread, and nothing in the diff adds a lock or static to a release build — both censuses and every `#[serial]` are inside `cfg(test)` modules.

The census tests are where the problems are, and they are the part of this change meant to keep the property true over time.

**1. The event-store census cannot see most of its own population.**

`pack.rs:6277` is `include_str!("pack.rs")`, so the scan's universe is that one file. It matches a single-line literal `with_event_store(`, and finds 49 of them there.

Outside that file there are **18 more call sites**, 11 of them in `tests/` integration binaries: `khive-pack-knowledge/tests/fixes.rs`, `khive-runtime/tests/read_verb_admission_exhaustion.rs` (four), `khive-pack-comm/tests/adr133_read_flags.rs`, `khive-pack-kg/tests/db_diagnostics_admission_obligations.rs`, `khive-pack-kg/tests/integration.rs` (two), `khive-pack-git/tests/acceptance.rs` (two). And `khive-pack-kg/tests/integration.rs:58` wraps the call in a `pack_with_events()` helper that the same file uses **59 times** — every one of those is a fixture the census is structurally unable to observe, because it is neither in `pack.rs` nor spelled with the matched literal.

So the property the test is named for — every event-backed fixture joins the ledger group — is enforced over roughly the minority of callers that happen to live in the scanned file.

**2. Both censuses report PASS when their input is empty.**

The only assertion in each is `offenders.is_empty()` (`pack.rs:6322`, `daemon.rs:3550`). Neither asserts that any candidate was found. If the call spelling changes, the callers move, or rustfmt splits the call across lines, the scan matches nothing, `offenders` stays empty, and the test passes — precisely at the moment it has stopped checking anything. A guard whose failure mode is silent success is worth less than no guard, because it also stops anyone looking. Each census needs a lower bound on its candidate count, asserted in the same test.

**3. The daemon census matches an inner name that a wrapper bypasses — latent today, so flagged as a property rather than a live defect.**

`daemon.rs:3535` matches `spawn_daemon_with_exe_and_config(`, but `spawn_daemon_with_exe` at `:732-734` forwards to it, and the counter increments in the callee at `:743`. A test calling the wrapper mutates `SPAWN_COUNT` without being counted. I grepped the callers: today the wrapper is reached only from `spawn_daemon` at `:729` and the closure at `:2193`, both production paths, so **no current test takes this route**. Worth fixing as part of the same change, not as an outage. The failure text is also only attribute line numbers — a name and file would survive contact with whoever trips it.

**4. Serialization excludes concurrency; it does not remove the shared state.**

`SPAWN_COUNT` is incremented at `daemon.rs:743` and reset only by a separate harness operation, so a serialized test still leaves the value behind for the next one. Same on the ledger side: no newly grouped test clears it on entry or guarantees cleanup on panic, and dispatch drains pending rows only on the event-store path. The reported symptom was `SPAWN_COUNT` left at 4, which is an ordering symptom, so under `--shuffle`, a newly added group member, or a single test run by name it can return with the lock fully in place. Per-test reset covering the panic path is what closes it.

Worth adding on the test plan: ten clean repetitions do not establish the absence of a race, and here they are weaker still, since the change serializes the most visible participants while the callers in item 1 stay unguarded.

The standing question for this batch applies to items 1 and 2 in particular — what changes for the caller. A census that passes on an empty scan, or over a minority of the population, produces the same green as one that checked.

oceanwaves630 (approved):

PR #2288, opened 2026-08-30
fix(config): make declared runtime surfaces truthful

### LINKED ISSUE #2271: reindex: --db bypasses the declared-backends override guard
`kkernel mcp` and `kkernel exec` validate a `--db`/`KHIVE_DB` override against declared `[[backends]]` and refuse a conflicting override (`validate_db_override_against_backends`, `DatabaseOverrideConflict`). `kkernel reindex` never calls that guard: `run_reindex` resolves a single `RuntimeConfig` from `--db`/`KHIVE_DB`/default and proceeds against whatever one file that names, even when the config declares a multi-backend topology.

Why it matters: in a split topology (e.g. a dedicated knowledge backend), reindex must be run once per backend file with an explicit `--db`. A mistyped or defaulted path silently reindexes the wrong file — including re-embedding rows on a backend whose pack is configured `no_embed` — and the ANN invalidation it performs touches only that one file's snapshots, so the intended backend's index state is left stale with no error anywhere.

Suggested fix: apply the same override validation `mcp`/`exec` use, and when `[[backends]]` is declared require the target to be one of the declared paths (or grow an explicit `--backend <name>` form like `kkernel db migrate` has).

### LINKED ISSUE #2276: Declared packs that register zero verbs start silently
## Summary

A `--pack` value is accepted at startup even when that pack registers no verbs. The declared pack set and the served verb set can therefore diverge silently: the daemon starts, answers requests, and the only symptom is that verbs an operator expected are simply absent.

## Observed

A daemon started with twelve `--pack` values serves eleven of them. `workspace` registers zero verbs. There is no error, no warning, and no non-zero exit — the flag is accepted and the process comes up healthy.

The absence is only discoverable by calling a verb from the missing pack and reading the available-verb list in the resulting refusal:

```
invalid input: unknown verb "workspace.list"; available: create, get, list, stats, ...
```

Nothing in startup output, process state, or a successful request distinguishes "this pack is loaded and serving" from "this pack contributed nothing".

## Why this matters

The pack list is how an operator declares what a deployment serves, so it is also how they and their tooling reason about what is available. A supervisor configuration or deployment manifest that reproduces the argv reads as a statement that those packs are served. When a pack silently contributes nothing, that statement is wrong, and the failure surfaces later as a missing verb at the call site rather than at the point of misconfiguration.

This is the general shape rather than one pack's problem: any future pack that fails to register — through a build feature flag, a registration regression, or a rename — degrades the same way, quietly and at a distance from the cause.

## Proposal

At startup, when a declared pack contributes zero verbs, say so loudly. Either:

- **Refuse to start**, treating it as a configuration error, or
- **Emit a warning naming the pack**, if starting anyway is preferable for operability.

Some packs are legitimately declaration-only today — `formal` is ontology with no verbs. That case should be handled explicitly rather than by leaving the check out: either mark such packs as intentionally verb-less so the rule can exempt them by declaration, or bring them under the same rule and let the warning stand as accurate.

The distinction worth preserving is between *intentionally* contributing no verbs and *accidentally* contributing none. Today those are indistinguishable from outside the process.

## Notes

`workspace` registering zero verbs appears to be pre-existing rather than a recent regression; this issue is about the missing signal, not about that pack specifically.

ohdearquant:
## Summary
- require topology-backed reindex runs to name a declared persistent SQLite backend path, including valid secondary targets
- reject declared packs that expose no public verbs unless their factory explicitly marks them as vocabulary/ontology-only
- mark workspace and formal as intentionally verb-less and document both contracts

## Verification
- khive-runtime zero-verb loader tests: 2 passed
- kkernel reindex target tests: 4 passed
- default khive-mcp server boot test: passed
- cargo check for affected packages: passed
- cargo clippy for affected packages/all targets with warnings denied: passed
- cargo fmt and git diff checks: passed

Closes #2271
Closes #2276


ohdearquant:
Two things worth saying first, because they are the load-bearing design calls and both look right.

The strict default on `intentionally_verbless()` (`khive-runtime/src/pack.rs:3239-3241`) is the correct direction. A pack that contributes nothing and boots anyway is exactly the silent condition this change exists to end, and defaulting to `false` is what makes the opt-in meaningful. And the compatibility cost lands where it should: walking the cases, the new reindex refusals are reachable only from `run_reindex` and fire before any database is opened, so ordinary boot and a first-run install with no config are unaffected. That is a maintenance-command break, not a boot break.

Two gaps.

**1. The reindex guard proves path membership, not that the target is usable — and the docs promise the latter.**

`serve.rs:2482-2492` walks the declared backends, keeps `kind = Sqlite`, canonicalizes, and returns `Ok(())` on a path match. It never consults `read_only`, existence, writability, whether the path is a directory, or whether the file is SQLite at all. There is no `read_only` reference anywhere in that region.

`read_only` is a real declared field (`khive-runtime/src/engine_config.rs:268`), so this is reachable: declare `path = "/data/snapshot.db", read_only = true`, run reindex against exactly that path, and the guard passes it through to a constructor that opens a normal writable backend. A missing file, a directory, or a non-SQLite file passes the same way and surfaces later as a generic open/schema error rather than a refusal naming the backend and the fix.

This matters more than a missing edge case because the surrounding prose calls the argument a *persistent SQLite target*, which claims usability, while the code establishes ownership. Either check the target class and `read_only` here with an explicit refusal, or narrow the documented contract to say the guard proves the path is declared and nothing further. The first is better — the whole point of the change is that a declared surface should be truthful about itself.

**2. The one message a downstream author will actually hit is the one that omits the fix.**

Since the default is strict, any out-of-tree or downstream-built zero-verb pack that compiles fine today fails at registration after this lands. The compiler does not force the update, so the error text is the entire remediation channel for that population.

The runtime error carries the answer (`khive-runtime/src/pack.rs:3311-3313`): *"its factory must declare `intentionally_verbless() = true`"*. The public server path does not (`khive-mcp/src/server.rs:972-973`): *"registers no public verbs and is not marked as intentionally vocabulary- or ontology-only"* — it names the condition and withholds the method. A downstream binary boots through the second one. Worth copying the method name into it; it costs a few words and it is the difference between a five-second fix and a source dive.

I also checked the in-tree population rather than sampling it: every pack crate except `formal` and `workspace` exposes at least one `Visibility::Verb` handler, and both of those are marked in this diff. `agent` has handlers but is not inventory-registered, so it is not a registry member. So nothing in-tree is left unmarked — the exposure is entirely out-of-tree, which is also the part no in-tree test can cover. Worth stating that limit explicitly rather than letting the added tests imply the case is covered.

Docs and code agree everywhere I checked, with the single exception in item 1.

**CI is red at this head, and it is one character.**

Two checks report failure but only one is a root — `CI gate` is the aggregator over the others, so the real failure is `Docs lint`. Reproduced locally with the repo's own formatter:

```
from docs/operations.md:454
-| `--db` / `KHIVE_DB` | ... | Target database; with `[[backends]]`, required and must match one declared SQLite path   |
+| `--db` / `KHIVE_DB` | ... | Target database; with `[[backends]]`, required and must match one declared SQLite path  |
error: Found 1 not formatted file in 206 files
```

One trailing space too many before the closing pipe on the table row at `docs/operations.md:454`. The same command on `origin/main` checks all 206 files clean, so it comes from this diff. `deno fmt` (no `--check`) from the repo root fixes it in place.

ohdearquant:
The only thing red here is `Docs lint`, and it is a one-command fix.

`deno fmt` reports `Found 1 not formatted file in 206 files`: **`docs/operations.md`**, line 454.
The table row for `--db` / `KHIVE_DB` has one space too many before its closing pipe, so the
column width does not match the rest of the table.

`deno fmt docs/operations.md` (or `make fmt`) resolves it. The other red entries on this PR are
sibling matrix jobs cancelled by fail-fast when this one failed — they are collateral, not
separate problems, so this single fix should clear the whole set.

ohdearquant (commented):

Verdict on head 9b4ca19e132578223de96bcf90410fd9ec21500e: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving at this exact commit. Verified before approving: the head matches the commit named in the request, every check run at this commit has completed with no failures, no change requests are outstanding, and the approving account is not the authoring account.

PR #2289, opened 2026-08-30
docs(adr): correct production graph site count

### LINKED ISSUE #2257: ADR-009 and ADR-014 still say six production-graph sites where both enumerate eight
Three ADRs were widened from six production-graph sites to eight. Two sentences still say six, and each sits in the same file as the enumerated list of eight that contradicts it.

Read at `52364b96763e79803c7f1393285a0a403184face`:

| File | Line | Text | Section |
| --- | --- | --- | --- |
| `docs/adr/ADR-009-backend-architecture.md` | 409 | "and the six-site boundary above is" | Rejected alternatives |
| `docs/adr/ADR-014-curation-operations.md` | 624 | "For the six production-graph sites in scope, ..." | **Decision** |

Both files carry the correct enumeration nearby: ADR-009:364 and ADR-014:674 each read "This change covers these eight production-graph sites:". `ADR-099-bulk-apply-atomic-units.md:852` has the same "eight" sentence and no stale counterpart, so it is clean.

The ADR-014 instance is the one that matters. It is in the Decision section rather than in prose, so a reader taking the Decision at its word scopes the change to six sites and excludes sites 7 and 8 — `comm.heartbeat` persistence and scheduled-event finalization — which are exactly the two the Scope section argues hardest to include. Site 8's argument in particular (a claim token is an ownership fence, not a content fence) does not follow from the other seven, so a reader who drops it loses the part that is not self-evident.

Not to be confused with a different and correct "six" in the same files: ADR-009:383, ADR-014:693 and ADR-099:871 each read "Site 8 covers all six finalization branches", which is a count of branches, not of sites. A sweep for this should not touch those.

Verified there are no other instances: a search for the `six ... site` form across `docs/` returns exactly these two, with the `eight ... site` form returning three files as a control that the search works.

ohdearquant:
## Summary
- correct ADR-009s stale six-site boundary reference to eight sites
- correct ADR-014s Decision text to cover all eight enumerated production-graph sites
- preserve the separate accurate count of six finalization branches

## Verification
- repository search finds no remaining stale six-site or six production-graph-site phrase in the synchronized ADRs
- all three ADRs retain the eight-site control enumeration and six-branch wording
- git diff check passed
- deno formatting was not runnable locally because deno is not installed; the two edits only replace number words without changing layout

Closes #2257

### DIFF
--- a/docs/adr/ADR-009-backend-architecture.md
+++ b/docs/adr/ADR-009-backend-architecture.md
@@ -406,5 +406,5 @@ read-to-replacement race without changing the ordinary upsert contract.
   still permits a stale document to overwrite concurrent changes.
 
 No open design question remains within this storage amendment: the conflict is a typed storage
-outcome, ordinary upserts retain their existing semantics, and the six-site boundary above is
+outcome, ordinary upserts retain their existing semantics, and the eight-site boundary above is
 explicit.
--- a/docs/adr/ADR-014-curation-operations.md
+++ b/docs/adr/ADR-014-curation-operations.md
@@ -621,7 +621,7 @@ relation validation, and index-maintenance behavior. The persistence failure mod
 caller whose patch was derived from a stale full-row snapshot: the caller receives a distinct
 optimistic-concurrency conflict instead of silently overwriting the newer row.
 
-For the six production-graph sites in scope, a full-row replacement carries the revision read
+For the eight production-graph sites in scope, a full-row replacement carries the revision read
 before patch application into the shared guarded store operation. The operation succeeds only if
 the row still has that revision and deletion state and the replacement revision strictly advances
 it. A conflict performs no curation update; the caller must fetch current state and explicitly


ohdearquant:
Checked this against the enumeration rather than the diff, since the whole content of the change is a number.

Both edited sentences were stale relative to the list they summarise. `ADR-014-curation-operations.md:674` already reads "these eight production-graph sites" and the list under it has exactly eight numbered entries; `:693` refers to "Site 8". So the prose at `:624` and the matching sentence in `ADR-009` were the laggards, and eight is the number the documents already commit to elsewhere.

Also swept for other carriers of the stale count, which is the part a two-line fix usually misses. At this head there is no remaining six-site or six-production-graph reference anywhere under `docs/`. The two surviving occurrences of "six" in these files — "all six finalization branches" in each — are a different population (branches within site 8, not sites) and are correctly left alone. `ADR-099` independently says eight and needed no change.

No further changes needed from my side.

ohdearquant:
Correct, and the count checks out against the thing it counts.

`ADR-009` was internally inconsistent before this: line 364 introduces the list as "these **eight**
production-graph sites" and line 409 closed the section with "the **six**-site boundary above". The
enumeration itself is the tiebreaker, and it has exactly eight entries, indices 1-8 contiguous. Same
for the two sibling copies of that list — `ADR-014:674` and `ADR-099:852` both enumerate eight.
So eight is the true number in all three documents and the stale word was the summary line, which is
the right thing to have changed.

I also swept for the rest of the class rather than just confirming the two lines, since a count that
went stale once tends to have siblings: across `docs/` and `crates/`, the only remaining
"N production-graph sites" claims are the three above, all now reading eight. Nothing else was left
behind.

One thing worth leaving a marker on, because it is the likely reason the stale six survived review
the first time: twenty lines below the fixed line, `ADR-009` says "Site 8 covers all **six**
finalization branches". That six is correct and unrelated — it counts branches, not sites. A future
reader sweeping for "six" near this section should not tidy it into "eight".

No other issues. The change is two words and both are load-bearing summary claims that now match
their enumerations.

PR #2290, opened 2026-08-30
fix(git): classify digest remote fetch failures

### LINKED ISSUE #2012: git.digest: storage timeouts and remote-fetch failures are reported as InvalidInput, so callers correctly refuse to retry a transient condition
## Summary

`git.digest` reports storage-layer timeouts as `RuntimeError::InvalidInput`. A caller receives
`invalid input: timeout during sql_bridge.reader_open` — a transient, retryable resource
condition wearing the one error kind that means "your arguments are wrong, do not retry."

## Where

`crates/khive-pack-git/src/handlers.rs` funnels every fallible resolution step through the same
blanket conversion. In `resolve_or_create_project` alone:

```rust
let slug_matches = find_projects_by_slug(runtime, token, &identity)
    .await
    .map_err(|e| RuntimeError::InvalidInput(e.to_string()))?;
```

and the same `map_err(|e| RuntimeError::InvalidInput(e.to_string()))` on
`find_normalized_noncanonical_matches`, `find_projects_by_legacy_repo_url`, and the
`resolve_project_id` call in `handle_digest`. Each of those helpers reaches
`sql.reader().await` and returns the reader-admission timeout as an `anyhow::Error`, which the
conversion then flattens into `InvalidInput` along with genuine argument faults.

`handle_digest` does the same for remote fetch failure:

```rust
let cloned = cache::ensure_clone(canonical).map_err(|e| {
    RuntimeError::InvalidInput(format!("remote clone/fetch of {:?} failed: {e}", ...))
})?;
```

A network or remote-side failure is likewise not an input fault.

## Why it matters

The error kind is the only thing a caller can branch on, and it is currently wrong in the
direction that suppresses recovery:

- A well-behaved client does **not** retry `InvalidInput`, because retrying an argument fault is
  a bug. So the one class of failure that would clear on its own is the class clients are
  correctly refusing to retry.
- This is consistent with an observed run of 74 consecutive failing calls over ten hours with no
  recovery on a poll-loop client (see #1987): nothing in the loop treated the condition as
  transient, because nothing in the response said it was.
- It also degrades triage from the outside. "Invalid input" sends a reader looking at the
  arguments — the source URL, `max_items`, the project reference — when the arguments are fine
  and the store is saturated.

## Suggested direction

Preserve the storage error's own kind rather than flattening it. The reader-admission timeout is
already typed at the storage layer (`Timeout`, operation `sql_bridge.reader_open`, per
`crates/khive-db/docs/api/pool.md`); the information exists and is discarded at this boundary.
Argument validation in `handle_digest` — the `source` parse, the `max_items` type check, the
`include` parse — is genuinely `InvalidInput` and should keep that kind; what should stop being
`InvalidInput` is everything downstream of a successful parse.

A test that would hold the line: force a reader-admission timeout during anchor resolution and
assert the returned error kind is not `InvalidInput`. The current suite passes with the
flattening in place, so a fixture that distinguishes the two kinds is the load-bearing addition
rather than another success-path case.

## Relationship to other issues

Separate from #1987, which is about the timeout happening at all; this is about what the caller
is told when it does, and it would be worth fixing even if the underlying saturation were
resolved tomorrow. Adjacent in shape to #1888 (a success envelope masking per-id failures) and
#1645 (`done: true` over a stalled cursor): in all three the response's shape misdescribes what
happened, in the direction that makes the caller do nothing.

ohdearquant:
## Summary

- return the existing RemoteFetchError after a valid remote source fails clone/fetch setup
- redact credentials and query material from the structured remote field
- keep malformed sources as InvalidInput and leave storage error typing unchanged

## Test plan

- focused remote failure classification regression test (red before production change, green after)
- cargo test --offline --locked --manifest-path crates/Cargo.toml -p khive-pack-git --lib
- cargo check --offline --locked --manifest-path crates/Cargo.toml --workspace
- cargo clippy --locked --manifest-path crates/Cargo.toml --workspace --all-targets --all-features -- -D warnings
- cargo test --offline --locked --manifest-path crates/Cargo.toml --workspace

Closes #2012

### DIFF
--- a/crates/khive-pack-git/src/handlers.rs
+++ b/crates/khive-pack-git/src/handlers.rs
@@ -156,12 +156,11 @@ impl GitPack {
         let (repo_path, expected_github_repo) = match &source {
             DigestSource::Local(p) => (p.clone(), None),
             DigestSource::Remote { canonical, gh_slug } => {
-                let cloned = cache::ensure_clone(canonical).map_err(|e| {
-                    RuntimeError::InvalidInput(format!(
-                        "remote clone/fetch of {:?} failed: {e}",
-                        redact_repo_url(canonical)
-                    ))
-                })?;
+                let cloned =
+                    cache::ensure_clone(canonical).map_err(|e| RuntimeError::RemoteFetchError {
+                        remote: redact_repo_url(canonical),
+                        message: e.to_string(),
+                    })?;
                 (
                     cloned,
                     gh_slug
@@ -2108,6 +2107,45 @@ mod tests {
                 "arbitrary malformed value must not become identity evidence: {malformed}"
             );
         }
+    }
+
+    #[tokio::test]
+    async fn remote_setup_failure_is_typed_without_reclassifying_bad_input() {
+        let _guard = crate::cache::ENV_MUTEX.lock().await;
+        let scratch = tempfile::tempdir().expect("tempdir");
+        std::env::set_var("KHIVE_GIT_DIGEST_SCRATCH_ROOT", scratch.path());
+        let (_rt, _token, registry) = fixture().await;
+
+        let remote_result = registry
+            .dispatch(
+                "git.digest",
+                json!({
+                    "source": "https://user:tok3n@127.0.0.1:1/org/repo?token=SECRET"
+                }),
+            )
+            .await;
+        let malformed_result = registry
+            .dispatch("git.digest", json!({"source": "relative/repo"}))
+            .await;
+        std::env::remove_var("KHIVE_GIT_DIGEST_SCRATCH_ROOT");
+
+        let remote_error = remote_result.expect_err("closed-port clone must fail");
+        let rendered = remote_error.to_string();
+        match remote_error {
+            RuntimeError::RemoteFetchError { remote, message } => {
+                assert_eq!(remote, "https://127.0.0.1:1/org/repo");
+                assert!(message.contains("git clone"), "{message}");
+            }
+            other => panic!("remote clone failure must stay typed, got {other:?}"),
+        }
+        assert!(
+            !rendered.contains("tok3n") && !rendered.contains("SECRET"),
+            "remote fetch error must redact credentials and query tokens: {rendered}"
+        );
+        assert!(
+            matches!(malformed_result, Err(RuntimeError::InvalidInput(ref message)) if message.contains("absolute")),
+            "malformed caller source must stay InvalidInput, got {malformed_result:?}"
+        );
     }
 
     #[test]


ohdearquant:
The redaction added here works for the shapes it recognizes: userinfo with or without a password is removed from the authority, and query and fragment material is dropped, for `https://`, `http://`, `git://`, `ssh://` and scp shorthand. The comment at `source.rs:342-347` about matching the normalizer's whitespace handling exactly is the right instinct — that is a real way these two functions could drift apart.

Two things stop this from closing the leak it targets.

**1. The rejected-source path is not redacted at all, and it is the likelier leak.**

The new redaction sits on the *accepted* branch: `handlers.rs:160-163` calls `redact_repo_url(canonical)` when a clone fails. The *rejected* branch is untouched. `source.rs:69` builds its error as `format!("source {trimmed:?}: {hint}")` — the full trimmed input — and `handlers.rs:131` forwards it verbatim with `map_err(|e| RuntimeError::InvalidInput(e.to_string()))`. Nothing on that path redacts anything.

So `http://USER:TOKEN@example.invalid/org/repo` is rejected by the scheme check and the credential is echoed whole into `InvalidInput`, which ordinary callers receive as a rendered string. The same holds for `relative/path?access_token=SECRET`, rejected as a relative source with the query intact.

That is the wrong way round. `http://` with inline credentials is exactly the spelling someone pastes by accident, and it is the one this change leaves fully exposed while masking the accepted case. `source.rs` is not in this diff, so the condition predates it — but the diff's stated purpose is to stop credentials reaching the structured remote field, and a caller cannot tell that the protection only applies once the URL was good enough to attempt.

**2. `redact_repo_url` returns unrecognized input verbatim, and never looks at the path.**

`source.rs:353-355`: if the value does not start with one of the four schemes and is not scp shorthand, the function returns `url.to_string()` unchanged. And for values it does recognize, `:359-363` splits the remainder at the first `/`, then strips before the last `@` **in the authority only**. A credential in a path segment — `https://example.invalid/org/TOKEN/repo` — survives untouched, as does anything in a `file://` URL or a string that does not parse.

Both behaviours fail open. For a function whose job is to make a value safe to print, the default should be to withhold rather than to pass through. Returning the input unchanged when the parser did not understand it is the single most dangerous line in the helper, because it produces a masked-looking value for the cases you tested and a raw one for the cases you did not.

The unit test at `source.rs:696-700` covers the userinfo/query/fragment case and passes. It cannot see either gap above, since both live in shapes it does not construct — worth adding the path-token, `file://`, and unparseable cases as explicit must-not-contain assertions.

**On the reclassification itself — worth being clear that it currently changes nothing for anyone.**

`RemoteFetchError` has four references in the whole tree: the variant declaration (`khive-runtime/src/error.rs:436`), the name mapping (`:533`), the construction site added here (`handlers.rs:160`), and the new test (`handlers.rs:2135`). There is no production consumer matching on it, and MCP callers receive the rendered string either way. So the new type is a better-shaped error that no caller yet reads differently — useful groundwork, but the caller-visible improvement in this diff is the redaction, which makes the two gaps above the substance of the change rather than a detail of it.

Nothing here argues against the direction. Making the remote a redacted-by-construction value rather than a `String` the call site remembers to sanitize would close both items at once, and would also cover `Display` and the derived `Debug`.

ohdearquant (commented):

Verdict on current head (291545ddad8fbfc1d168fa6d50be1276fb3293e4): APPROVE, zero blocking findings. This is a comment, not an approval. A human reviewer decides whether to approve and merge.

One non-blocking note. `git.digest` now returns `RemoteFetchError { remote, message }` for remote clone or fetch setup failures, with credentials and query strings redacted from `remote`, while source-parse failures stay `InvalidInput` and storage failures keep their existing types. ADR-088 Amendment 1 and the `git.digest` API reference do not yet state that contract; a follow-up documentation amendment would close the gap.

oceanwaves630 (approved):
Approved at 291545ddad8f.

PR #2291, opened 2026-08-30
fix(release): publish npm CLI alias

### LINKED ISSUE #2264: npm/cli-alias (@khive-ai/cli) is version-bumped by every release and published by no path
`npm/cli-alias` (`@khive-ai/cli`) is version-bumped by every release and published by
nothing. It has drifted three releases behind as a result, and the drift is invisible
because the bump always succeeds.

## Measured at the 0.8.0 release commit

| | value |
| --- | --- |
| version in `npm/cli-alias/package.json` | `0.8.0` |
| `dist-tags.latest` on the npm registry | `0.2.5` |
| references to `cli-alias` anywhere in the repo | none |

The last value is the substance of this issue. A repo-wide search for `cli-alias`,
excluding `.git` and `node_modules`, returns nothing at all: not in
`.github/workflows/release.yml`, not in `scripts/npm-publish.sh`, not in any build
step or doc. As a control in the same search, `kernel-darwin-arm64` appears in 6
files, so the search is working.

Both publish paths enumerate their targets explicitly and neither includes this one:

- `release.yml`'s `publish-all` job publishes the six `@khive-ai/kernel-*`
  subpackages by name, then the umbrella `khive`.
- `scripts/npm-publish.sh` iterates a hard-coded list of the same six platforms,
  then the umbrella.

## How it got here

```
52364b967  chore(release): bump workspace and npm package versions to 0.8.0
a7ab8ac3a  chore: bump all versions to 0.2.8
568dcb67f  js init commit
```

The version bump is repo-wide, so this manifest is carried along by every release
while no publisher ever ships it. The registry sitting at `0.2.5` shows it also
missed the `0.2.8` release, so this is not new to `0.8.0`.

## Why it is worth a decision rather than a quiet fix

The package is not inert. It declares `"khive": "<same version>"` as a dependency
and exposes `khive` and `khive-mcp` bin shims, so anyone who installed
`@khive-ai/cli` gets a working but three-releases-stale CLI, pinned to `khive@0.2.5`,
with no signal that a newer one exists. Publishing it and abandoning it are both
reasonable; carrying a manifest that is bumped but never shipped is the one option
that keeps producing this silently.

Two coherent resolutions:

1. **Ship it.** Add it to `publish-all` after the umbrella (it depends on the
   umbrella at the same version, so ordering matters) and to `npm-publish.sh`.
2. **Retire it.** Delete `npm/cli-alias/`, and deprecate the registry entry so
   existing installs are told where to go.

Either way the release bump should stop touching a manifest that no path publishes,
because that is the mechanism that hid the drift for three releases.

ohdearquant:
## Summary
- choose the ship-it resolution for the existing @khive-ai/cli compatibility package
- validate that alias package and exact khive dependency versions match the release version
- publish the alias after the umbrella in both GitHub Actions and the local release script
- add executable dry-run and workflow-ordering contract coverage

## Verification
- scripts/ci.sh lint: passed, including 5 CI workflow contract tests
- network-free npm-publish.sh dry-run fixture: passed
- bash syntax and git diff checks: passed
- npm pack --dry-run with an isolated cache: @khive-ai/cli@0.8.0 packaged successfully

Closes #2264

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,8 @@ name: Release
 # workflow today.
 #
 # Builds Rust binaries per-platform, publishes each khive-kernel-{platform}
-# npm subpackage, then publishes the umbrella khive package.
+# npm subpackage, publishes the umbrella khive package, then publishes its
+# exact-version @khive-ai/cli compatibility alias.
 #
 # Per ADR-026 §"Atomic release semantics": all subpackage publishes must
 # succeed before the umbrella publishes. This is enforced by separating the
@@ -21,8 +22,8 @@ name: Release
 #     artifact. No `npm publish` happens here.
 #
 #   Phase 2 (publish-all, single job): downloads all six platform artifacts,
-#     updates versions, publishes the six subpackages, then publishes the
-#     umbrella — in a single sequential job. If any subpackage publish fails,
+#     updates versions, publishes the six subpackages, the umbrella, then the
+#     dependent CLI alias — in a single sequential job. If any subpackage publish fails,
 #     the umbrella is never published. Because publishing is serialised in one
 #     job, partial failure cannot leave mismatched versions permanently on the
 #     registry (any already-published subpackage can be unpublished within 72h
@@ -181,7 +182,7 @@ jobs:
           exclude: ${{ steps.semver-excludes.outputs.list }}
 
   # ─────────────────────────────────────────────────────────────────────────
-  # Phase 2: Publish all subpackages + umbrella atomically in a single job.
+  # Phase 2: Publish all subpackages + umbrella + CLI alias in a single job.
   # Runs only after ALL Phase 1 jobs succeed (needs: build-platform with
   # fail-fast: true propagates any single failure here) and the release SemVer
   # gate passes.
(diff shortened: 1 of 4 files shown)


ohdearquant:
The ordering rationale is right and the comment at `release.yml:340-341` states it correctly: the alias carries an exact-version dependency, so it has to follow the umbrella. The source-side validation at `npm-publish.sh:28-38` is also the right check to have — it compares both the alias version and its pinned `dependencies.khive` against the umbrella version and exits non-zero on either mismatch, with no pipeline in the way.

Three things I would want fixed before this runs a release.

**1. The two release paths disagree on what "already published" means.**

`npm-publish.sh:108-118` treats an existing version as success — it prints `already on npm — skipping` and continues to `Done`. `release.yml:354-356` is a bare `npm publish`, which fails on an existing version. So the same registry state is a clean release on one path and a failed job on the other. Whichever is intended, they should not both be shipped.

**2. The skip is decided without ever looking at what is actually published.**

`npm-publish.sh:108` asks only for the alias's *version*. It never reads `dependencies.khive` from the registry — the only dependency check in the script is against the local checkout at `:30`. So if `@khive-ai/cli@VERSION` already exists pinned to the wrong umbrella version, the script skips it, reports Done, and leaves every installer resolving the wrong pair. The check that would catch it is one `npm view @khive-ai/cli@VERSION dependencies.khive` away.

There is a second problem in the same line: `2>/dev/null` sends the failure to the void, so a registry outage, an auth failure, and "package not published" are one branch. That direction happens to fall toward attempting a publish, so it is not the dangerous polarity here — but the operator is not told which of the three happened, and the same collapse in the reverse position would silently skip.

**3. The idempotency test is an unanchored regex match, in three places — latent, not currently reachable.**

```
59:  if npm view "${name}@${ver}" version 2>/dev/null | grep -q "${ver}"; then
94:  if npm view "khive@${VERSION}" version 2>/dev/null | grep -q "${VERSION}"; then
108: if npm view "@khive-ai/cli@${VERSION}" version 2>/dev/null | grep -q "${VERSION}"; then
```

`grep -q "$VERSION"` succeeds on any line *containing* the version rather than equalling it, and the pattern is a regex, so the dots are wildcards. Both are demonstrable: with pattern `0.8.0`, the string `10.8.0` matches on the substring and `0x8x0` matches on the dots.

To be clear about severity, since it would be easy to overstate: I could not construct a reachable failure. Each query pins an exact version (`pkg@VERSION`), so `npm view` returns either that exact version or nothing, and neither loose match can be triggered by that output. It is fragility rather than a live defect — but the fix is one character class (`grep -qxF`, or better `[[ "$out" == "$VERSION" ]]` on captured output), and it removes a decision about whether to publish from a pattern that only accidentally behaves. All three because they are one shape and will be fixed once or not at all.

**On partial failure.** The umbrella and the alias are two public packages published in sequence with no reconciliation between them. If the umbrella lands and the alias does not — rejection, cancellation, runner death — the registry holds `khive@VERSION` with no matching `@khive-ai/cli@VERSION`, and a rerun does not repair it: the rerun starts at the kernel publishes and fails there on already-published versions before ever reaching the alias step. Nothing in the job detects the split or tells the operator what to do about it. A separately runnable alias-only path would make the state recoverable without hand-publishing.

**On the dry run.** Worth stating plainly what it establishes: `npm-publish.sh` in dry-run mode still calls `npm view` for real but replaces every `npm publish` with an echo, so it exercises the script's control flow and none of authentication, registry acceptance, tag handling, or partial-failure behaviour. A passing fixture says the script runs, not that the release is correct — the failure modes above are all outside what it can reach.

oceanwaves630 (approved):

PR #2292, opened 2026-08-30
fix: enforce backend registry invariants

### LINKED ISSUE #2026: Filter coordinator search fan-out by declared backend kinds (ADR-165 Slice 4)
Coordinator `search` fans out to every registered backend regardless of the requested `kind`: a `kind=note` search also dispatches to the session backend, a database two orders of magnitude larger than any result it could contribute, paying its FTS and vector cost on every note search. Registration knows which kinds each backend serves; the dispatch loop discards that information.

Scope per ADR-165 Slice 4: backend registration declares served substrate kinds; the dispatch loop skips backends whose declaration cannot match the request. Declaration-based only (no per-call heuristics); a backend declaring nothing is conservatively included. Merged-result semantics unchanged for remaining backends. This is a prerequisite for any physical store split: without it each split adds one backend to every search's fan-out. Lands with ADR-166 G4.

Refs: ADR-165 Slice 4 (proposed, PR #2022), ADR-166 G4.

### LINKED ISSUE #2043: config: BackendId accepts an empty backend id — no validation at any construction path
`BackendId` (`crates/khive-runtime/src/config.rs:25` at `78af4e22ed3f`) is `pub struct BackendId(pub String)`, and `BackendId::new(name: impl Into<String>)` performs no validation, so the empty string is silently accepted. Because the tuple field is `pub`, validating `new` alone would not close the gap: `BackendId(String::new())` and deserialization-driven construction from `[[backends]]` config both bypass the constructor.

An empty backend id has no defined meaning: it participates in node-to-backend resolution and cross-backend edge routing as an ordinary map key, renders as an empty string in `Display` and diagnostics, and interacts confusingly with masked/truncated backend-id reporting.

Fix shape: make the field private, validate in the constructor (non-empty at minimum; consider the same character discipline applied to other identifiers), and reject at deserialization so config-declared backends get the same check as programmatic construction. The invariant belongs on every construction path, not just the convenience constructor.

ohdearquant:
## Summary

- make `BackendId` construction fallible and reject empty or whitespace-only identifiers at config and runtime boundaries
- add optional closed-set `served_kinds` backend metadata, reject explicit empty declarations, and carry declarations into coordinator registration
- filter federated entity and note searches before backend dispatch while preserving conservative routing when metadata is omitted
- include served-kind topology in daemon config identity and document the operator contract

## Testing

- `deno fmt --check crates/kkernel/docs/coordinator.md docs/configuration.md docs/khive-config-example.toml docs/multi-backend.md`
- `cargo check --workspace`
- `cargo test --workspace`
- `cargo test -p khive-runtime backend_id_tests -- --nocapture`
- `cargo test -p khive-runtime test_empty_backend_name_rejected -- --nocapture`
- `cargo test -p kkernel backend_id -- --nocapture`
- `cargo test -p khive-runtime backend_served_kind -- --nocapture`
- `cargo test -p kkernel served_kind -- --nocapture`
- `cargo test -p khive-mcp config_id_differs_when_backend_served_kinds_differ -- --nocapture`
- `cargo test -p kkernel coordinator_registry_copies_backend_served_kind_declarations -- --nocapture`

Closes #2043
Closes #2026


ohdearquant:
Two blocking issues, one in each commit.

## 1. An all-filtered search is indistinguishable from a genuine no-match

`fan_out_search_with_visibility` filters the registry by the requested substrate and, when nothing survives, returns three empty vectors:

```rust
// crates/kkernel/src/coordinator/dispatch.rs:585-593
let entries: Vec<(BackendId, Arc<KhiveRuntime>)> = self
    .registry
    .iter()
    .filter(|entry| entry.serves(requested_substrate))
    ...
if entries.is_empty() {
    return (vec![], vec![], vec![]);
}
```

The third vector is `per_backend`. Because it is *empty* rather than *error-bearing*, the partial flag computed one layer up is false:

```rust
// crates/kkernel/src/coordinator/service.rs:112
let partial = per_backend.iter().any(|r| r.error.is_some());
```

`any()` over an empty iterator is `false`. The MCP layer only raises `search_incomplete` when the result is both partial and empty (`crates/khive-mcp/src/server.rs:2348`), so that guard never fires here.

Net effect: with a backend declaring `served_kinds = ["event"]` and an entity stored in it, `search(kind="entity", query="known")` returns `ok=true`, `status="complete"`, `result=[]`. The record exists; the caller is told there is nothing to find. This is the failure mode the ADR-130 comment immediately above that guard was written to prevent — "the 'no match' reading is not established when the answer may be sitting on the backend that never responded" — and dispatch-side filtering reaches the same end state by a path the guard does not observe.

Note this is reachable with an entirely reasonable configuration: a substrate request maps only to note or entity, so a backend declaring `event` is excluded from *both* searchable requests and silently contributes nothing to every query.

Suggested direction: carry selected/filtered backend evidence through `CoordSearchResult` and surface a typed "no eligible backend" condition, or reject a topology at boot that has no eligible backend for a searchable substrate. Either way the all-filtered case needs envelope coverage.

## 2. `BackendId` loses its public constructor without a compatibility path

The base type is `pub struct BackendId(pub String)` with `pub fn new(name: impl Into<String>)`. This diff makes the field private and removes `new` (`crates/khive-runtime/src/config.rs:24-45`).

`BackendId` is re-exported from the crate root at `crates/khive-runtime/src/lib.rs:142` — that export predates this change, so the break lands on already-public API. `khive-runtime` is published on crates.io (currently 0.8.0), so this is a source-breaking change for any external consumer calling `BackendId::new("archive")` or constructing the tuple directly, including for identifiers the new validator would accept. The workspace version is unchanged.

The three pack-handler edits in this diff (`khive-pack-knowledge/src/handlers.rs:590`, `khive-pack-moodboard/src/handlers.rs:691-696`, `khive-pack-session/src/handlers/resume.rs:64`) adapt in-repository call sites only; they cannot help downstream users.

Suggested direction: keep a supported public constructor that routes through the fallible validator, or treat this as a deliberate breaking change with the corresponding version and migration treatment.

## Validation scope

The identifier validation itself reads correctly: it rejects the empty string and whitespace-only strings and names both the offending value and the rule. It does **not** reject leading/trailing whitespace, Unicode, very long names, or identifiers colliding only by case — worth confirming those are intended to remain accepted, since each is plausible in an existing config.

## Test coverage gaps

The added tests cover the filter's inclusion and exclusion directions at the registry level, and config-load rejection of blank and unknown values. Not covered: the all-filtered MCP result and its `ok=true`/`complete` classification; padded-whitespace, Unicode, length and case-collision identifiers; and the config → registry → MCP wiring itself — a regression passing `None` at `crates/kkernel/src/cli.rs:596-603` would leave the new filter tests green.

ohdearquant:
Adding the release consequence of the `BackendId` change, since it determines *when* this can land rather than just how it should be written.

`khive-runtime` is published on crates.io and its current release is 0.8.0. `BackendId` is re-exported from the crate root at `crates/khive-runtime/src/lib.rs:142`, and that export predates this PR — so making the field private and removing `pub fn new` is a source-breaking change to an already-public API of a published crate.

Under semver that cannot ship in any 0.8.x. For a 0.x crate the minor position is the breaking one, so the two paths are:

1. **Ride the 0.9.0 boundary** with the rest of the breaking set, or
2. **Rework to non-breaking** — keep a supported public constructor and route config-derived names through the fallible validator internally, so external callers keep compiling while the validation still runs where it matters.

Option 2 is worth a serious look, because the validation goal and the API removal are separable: nothing about validating identifiers at config load requires removing the constructor. If `new` is kept as an infallible constructor for already-valid inputs, or reintroduced as a `try_new` returning the validator's error while `new` is deprecated rather than deleted, the boot-time check in this PR still does its job.

Either way the workspace version in this diff is unchanged at 0.8.0, so as it stands the PR would publish a breaking change under a compatible version number. That needs resolving before merge regardless of which path is chosen.

This does not affect the other issue raised above — the all-filtered search returning `ok=true`/`complete`/`[]` is independent of the version question and stands on its own.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at 4d95e56a.

PR #2293, opened 2026-08-30
fix(kg-editor): preserve showcase navigation state

### LINKED ISSUE #2034: kg-editor: view switches lose scroll and panel state; unknown view params coerce silently

ohdearquant:
## Summary

- preserve structure-graph subtree and selected-node state across analysis view switches
- move focus and scroll position when users choose a view from the analysis navigation
- restore a URL-selected module when returning to the structure graph
- port the completed work from closed draft #2169 onto the current baseline while retaining the newer exact module-ID URL contract

## Tests

- `npx vitest run src/components/showcase/repo-showcase.test.tsx` (51 passed)
- `npm run typecheck`
- `npx eslint src/components/showcase/repo-showcase.tsx src/components/showcase/repo-showcase.test.tsx`
- `git diff --check`

Closes #2034


ohdearquant:
The state-preservation work is sound — lifting `structureGraphSelection` to parent state so it survives the keyed `ActiveView` remount is the right mechanism, and the focus/scroll additions are triggered by direct navigation rather than by a render effect, so they cannot steal focus while a user is typing. Two issues, one of which is the classic hazard for a port that carries a newer identifier contract.

## 1. A supplied-but-missing `module_id` falls through to a looser match

```ts
const requestedModule = requestedModuleId ? moduleById.get(requestedModuleId) : null;
if (requestedModule) {
  ...
} else if (requestedPath) {
  const matches = modulesBySourcePath.get(requestedPath) ?? [];
  if (matches.length === 1) {
    nextModuleId = matches[0].id;
    resolvedModuleRestore = true;
  }
  ...
}
```

The `else if` is entered whenever the exact lookup misses — and that includes the case where an exact ID *was* supplied and simply is not in this snapshot. The absent-ID case and the no-ID-at-all case are handled identically, which is exactly what the exact-ID contract exists to prevent.

Two concrete outcomes:

- `module_id=A` (absent from the snapshot) together with `module=src/B.ts`, where B is a unique path match: the code selects **B**, sets `resolvedModuleRestore = true`, and then canonicalizes the URL to B's path. The user asked for A, is shown B, and the link that recorded the request is rewritten to point at B — so the original request is no longer recoverable even by reloading.
- `module_id=A` alone with no path: neither branch fires, `nextModuleId` remains `defaultModuleId`, and the user silently lands on the default module.

A message is pushed when `requestedModuleId && !requestedModule`, which helps, but it sits alongside a selection the code has already marked as resolved — so the UI is in a "restored successfully" state showing the wrong module. A message is not a substitute for not selecting the wrong thing.

Suggested shape: when `module_id` is present and unresolved, do not fall through. Set an unresolved state for that ID, keep the path only as display context, and use path resolution *only* when no `module_id` was supplied. That also keeps the canonicalization from overwriting the requested ID.

## 2. Preserved selection is not validated against the preserved subtree

The refresh-time validity check accepts a selection when the subtree ID exists in the package page and the selected ID exists as either a package or a module. It never checks that the selected module still belongs to the preserved subtree — i.e. `moduleById.get(current.selectedId)?.package_id === current.subtreeId` for a package subtree.

Failure sequence: select module M inside package P; switch views; the same `RepoShowcase` instance receives a refreshed snapshot in which M has moved to package Q. P and M both still exist, so the guard returns `{subtreeId: P, selectedId: M}` as valid. Back on the graph, M is filtered out of P's module list — the selected node is not there, and the inspector falls back to the repository overview while stale selection state persists.

This is the invalidation gap the preservation change creates: state that used to be discarded by the remount now outlives the data it refers to, so it needs a relationship check rather than two independent existence checks.

## Tests

The two added tests pin the focus/scroll path and the parent-state alignment across a keyed view switch, and would redden if either regressed. What is missing is coverage of the boundary this PR actually moves:

- a URL carrying a `module_id` that is **absent** from the snapshot, with and without an accompanying path — the case where the looser fallback fires;
- a URL restore that genuinely goes through `module_id` (the existing test aligns selection from an in-app module inspection, not from a URL);
- a rerender across snapshots that moves or removes the selected module or its package while the graph is mounted behind another view.

The existing missing-selection test covers a missing *source path* with `module_id: null`, which is the complement of the failing case rather than the case itself — so the suite currently passes precisely because it does not exercise the ID path.

ohdearquant:
## `KG Studio (Next.js)` failure: one layout-input change settles the layout twice

```
FAIL  src/components/showcase/repo-showcase-layout.test.tsx > repository showcase graph layout > reuses settled positions for local interaction and invalidates on layout input changes
AssertionError: expected "vi.fn()" to be called 3 times, but got 4 times
```

The assertion is the last step of that test: after replacing `graph.modules.items` with a new
array, the layout should settle exactly once more.

```tsx
rerender(<RepoShowcase bundle={replacedModulePage} />);
expect(settleGraphLayoutSpy).toHaveBeenCalledTimes(3);
```

Four means a single layout-input change reaches the memo twice. Everything earlier in the test
passes — the local-interaction case holds at 1, the unrelated-relabel case holds at 2 — so
memoization is doing its job for the cases where nothing should recompute. Only the genuine
invalidation double-fires, which is the one case that decides whether the change delivers what
it claims.

One candidate as a starting point rather than a diagnosis; this was not reproduced locally. The
selection-reconciliation effect lists `moduleById` and `modulesBySourcePath` among its
dependencies, and both are derived from `graph.modules.items` — exactly the reference this last
rerender replaces. The effect then writes several pieces of state on every run: the
`setStructureGraphSelection` updater is guarded to return `current` when the selection is still
valid, but the surrounding `setSelectedModuleId` / `setUnresolvedModule` / `setActiveView` /
`setLocationNotice` calls are not, so a second render pass can follow the same input change.
Whether that second pass is what re-enters the layout memo is worth confirming before changing
anything — the guard on the selection updater suggests the intent was already to avoid this.

The test itself reads correct as written; it is asserting the property the branch is for.

ohdearquant (commented):

Verdict on head d2b56b62778aff04a867128d74f9e335475df4d8: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 47b727138879b596fe6ec0c67ce954e7e02a3bb6: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at c472d805.

PR #2295, opened 2026-08-30
fix(mcp): deduplicate agent json records

### LINKED ISSUE #2135: Agent-mode json responses duplicate lifted fields inside properties
## Problem

Record-shaped responses under \`presentation=agent\`, \`format=json\` (the MCP default shape) carry the same key/value pairs twice: once lifted to the top level and again inside \`properties\`.

Measured on a scratch database:

\`\`\`
kkernel exec --db :memory: --presentation agent 'gtd.assign(title="probe", assignee="x", priority="p2")'
# duplicated top-level + properties keys: ['assignee', 'priority', 'status']
\`\`\`

The same duplication appears on every row of \`gtd.tasks\` / \`gtd.next\` listings, so it scales with result count.

## Existing mechanism

ADR-078 §7 already defines the redundancy-reduction pre-pass (\`apply_redundancy_drop\` in \`crates/khive-runtime/src/presentation.rs\`) which drops exactly these duplicates, plus \`full_id\` and \`namespace="local"\`. It is currently applied only when \`format ∈ {auto, table}\`.

## Ask

Apply the §7 pre-pass in \`agent\` presentation under \`json\` format as well (ADR-078 amendment). \`verbose\` keeps the full canonical shape unchanged, so scripted/CI consumers are unaffected. Chain safety is not at risk: \`$prev\` substitution reads the canonical pre-presentation result (\`server.rs\` chain arm), verified by executing \`create | create($prev.id) | link(...)\` under \`--output-format auto\` — the substituted value was the full UUID.

## Acceptance

- \`gtd.assign\` echo under agent/json carries each of assignee/priority/status exactly once.
- A verbose/json control run shows the unchanged canonical duplication.
- Chain regression: \`create | link($prev.id, ...)\` passes under agent/json and agent/auto.

ohdearquant:
## Summary

- apply the ADR-078 redundant-record elision to `agent` presentation with JSON output
- retain JSON typing for compounded server responses while removing duplicated record content
- preserve verbose output and canonical `$prev` chaining behavior
- document the contract in ADR-078 Amendment 3

## Tests

- runtime format tests (14 passed)
- focused integration coverage for agent/verbose/json/auto chaining
- affected clippy and formatting checks
- `git diff --check`

Closes #2135


ohdearquant:
One blocking issue plus a documentation inconsistency this change makes load-bearing.

## 1. Arbitrary property metadata is erased by an equality-only key

The reduction removes a `properties` entry `(k, v)` whenever the containing object has a top-level field `k` whose value is exactly equal to `v` (`crates/khive-runtime/src/presentation.rs:147-157`). The comparison is exact — no case folding, trimming, Unicode normalization, or absent/null equivalence — which is good, but the key is still *only* equality of key and value, and that is too coarse for what `properties` is allowed to hold.

`name` is a dedicated display label while `properties` is documented as arbitrary structured metadata (`crates/khive-types/src/entity.rs:128-133`), and the create schema accepts arbitrary JSON there (`crates/khive-pack-kg/src/handler_defs.rs:98-102`). So:

```
create(kind="entity", entity_kind="concept", name="Ada", properties={"name":"Ada"})
```

produces a record where the two values are equal but not the same fact — one is the display label, the other caller-authored metadata that may carry its own provenance. Under Agent JSON the property is dropped, and if it was the only property the whole `properties` object disappears. Nothing marks the removal, so a consumer reading `properties.name` sees a successful response that is indistinguishable from the property never having been set.

Suggested direction: restrict elision to an explicit handler-owned mirror allowlist rather than generic value equality, or preserve arbitrary properties by default. A regression fixture with same-valued but semantically distinct `name` / `properties.name` would pin it — the current GTD fixture supplies genuine mirrors whose semantics are intentionally shared, so it cannot detect a key that is too coarse.

## 2. The accepted ADR and the public tool description still call `json` lossless

The amendment states the Agent JSON reductions at `docs/adr/ADR-078-output-format-shape-aware-rendering.md:601-609`, but the same accepted ADR still says elsewhere that `json` is the lossless machine contract (`:183-186`), that the reductions apply only to auto/table and that JSON retains `full_id`/`namespace` (`:233-278`), and that callers always recover the canonical value with `format=json` (`:297-301`). The public request documentation likewise says `"json" (default): compact, lossless JSON` (`crates/khive-mcp/src/tools/request.rs:73-81`).

A caller following the still-published text selects `format=json` expecting canonical `full_id`, `namespace` and a complete `properties` object, and now gets a reduced one. The code is self-consistent — Verbose JSON is the lossless escape hatch — but the document set does not say so, and the older sections actively contradict it. Those sections and the tool descriptions need amending in the same change that introduces the reduction; otherwise the contract that callers actually read is wrong.

## What is right

The scope is tight and I could not find leakage: the changed helpers have four production call sites, all on the named MCP rendering path (`crates/khive-mcp/src/server.rs:3537-3546`, `:3702`, `:3757-3759`, `:3759-3765`), with `crates/khive-runtime/src/lib.rs:131` only re-exporting. Verbose and Human JSON stay unreduced, errors are bypassed, and chain substitution is unaffected because canonical results are captured before presentation (`crates/khive-mcp/src/server.rs:2044-2050`, `:2111-2121`). Records are not deduplicated against one another — two records sharing an `id` both survive in input order — so there is no ordering-dependent winner to reason about.

ohdearquant:
## The red `CI gate` is one test, in a crate this branch changes

```
FAIL khive-mcp pending_events::tests::fired_reminder_delivers_to_creator_after_daemon_actor_changes
Summary: 1068/4675 tests run: 1067 passed, 1 failed
```

`Coverage ratchet` also fails at its `Compute coverage` step, which is downstream of the same
test run rather than an independent signal.

Unlike several other currently-red branches, this one is not ambient: the failing test lives in
`khive-mcp` and this branch edits two files there. The test name is about a reminder still
reaching its creator after the daemon actor changes, which sits close to record identity, so it
is worth checking directly against the dedup change rather than re-triggering.

ohdearquant (commented):

Verdict on head 5a122c574eddb9f7779a19933aced7dc1d63cffe: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at 39a266cc.

PR #2296, opened 2026-08-30
fix(runtime): authorize extra-namespace reads consistently

### LINKED ISSUE #2223: authorize_with_visibility checks extra namespaces against Write instead of Read
ADR-129 (Amendment 1, clause 2) specifies that `authorize_with_visibility` requires **Write on the primary namespace and Read on each extra visible namespace**. The implementation sends the same `authorize` pseudo-verb for the primary check and for every extra-namespace check (`crates/khive-runtime/src/runtime.rs`, extra-visibility loop). Since `GateRequest` carries no separate right discriminator, a right-aware gate classifies the extra checks at the pseudo-verb's full (Write) authority.

Effect: a principal holding Write on its primary namespace and Read-only rights on an extra namespace is refused the mint that ADR-129 says it should be granted. Per clause 1 this errs on the fail-closed side (over-checking refuses callers the grant would have served), so it is a functionality gap, not an authority escalation.

Possible direction: introduce a distinct read-classified pseudo-verb for the extra-namespace checks (with its classification-table entry, since an unclassified pseudo-verb is denied by default), and add tests for the Write-primary/Read-extra allow and the missing-Read denial.

ohdearquant:
## Summary

- keep `authorize` as the Write-classified primary-namespace mint check
- use `authorize.visible` for each Read-classified extra-namespace check required by ADR-129
- cover both Write-primary/Read-extra admission and missing-extra-Read refusal with verb-sensitive regression tests

## Testing

- `cargo clippy -p khive-runtime --all-targets --all-features -- -D warnings`
- `cargo test -p khive-runtime authorize_with_visibility_ -- --nocapture`
- `cargo test -p khive-runtime`

Closes #2223

### DIFF
--- a/crates/khive-runtime/src/runtime.rs
+++ b/crates/khive-runtime/src/runtime.rs
@@ -1115,15 +1115,15 @@ impl KhiveRuntime {
                 }
                 // The primary check authorizes writes to `primary` only. Each
                 // extra namespace grants read visibility, so each one takes
-                // its own gate check before it may enter the minted set — a
-                // token must never carry visibility the gate was not asked
-                // about. Any deny or gate error refuses the whole mint,
-                // naming the offending namespace (fail-closed).
+                // its own Read-classified gate check before it may enter the
+                // minted set — a token must never carry visibility the gate
+                // was not asked about. Any deny or gate error refuses the
+                // whole mint, naming the offending namespace (fail-closed).
                 for extra in &extra_visible {
                     let extra_req = GateRequest::new(
                         actor.clone(),
                         extra.clone(),
-                        "authorize",
+                        "authorize.visible",
                         serde_json::Value::Null,
                     );
                     match self.config.gate.check(&extra_req) {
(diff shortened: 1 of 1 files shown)


ohdearquant:
The direct method is implemented correctly and fails closed — every extra is checked before the token is minted, a denial returns `PermissionDenied` naming the namespace rather than dropping it, and the mint happens only after all checks pass. The problem is reach: the main dispatch path never calls it.

## The ordinary dispatch path does not perform the new check

`VerbRegistry::dispatch_with_identity` gate-checks only the requested operation (`crates/khive-runtime/src/pack.rs:2031-2040`), then mints the widened token directly:

```rust
// crates/khive-runtime/src/pack.rs:2198-2220
let primary = Namespace::local();
let mut extra_visible: Vec<Namespace> = ... // from identity or registry config
extra_visible.push(Namespace::local());
NamespaceToken::mint_with_visibility(primary, extra_visible, resolved_actor)
```

There is no `authorize_with_visibility` call and no `authorize.visible` request on this path, and `dispatch` / `dispatch_as` inherit it (`pack.rs:2596-2614`).

Enumerating rather than sampling: `authorize.visible` occurs in exactly two places in the tree — its construction at `crates/khive-runtime/src/runtime.rs:1126` and a test gate at `:2351`. Production callers of `authorize_with_visibility` are three: `crates/kkernel/src/coordinator/dispatch.rs:598`, `:861`, and `crates/khive-pack-moodboard/src/handlers.rs:928`. Every other occurrence is a test or the definition itself.

Failure scenario: configure a registry or `RequestIdentity` with `visible_namespaces=["restricted"]`. Grant the gate permission for the ordinary `list` operation in `local`, and no Read permission for `restricted` under the new contract. A caller invokes `dispatch_with_identity("list", ...)` with no explicit namespace. The registry checks only the `list` request and mints a token whose visible set is `local ∪ restricted`. `list_entities` (`operations.rs:1861-1897`), `hybrid_search_inner` (`retrieval.rs:693-785`) and `query_with_metadata` (`operations.rs:4953-4975`) all scope off that token, so records from `restricted` come back.

This is the specific shape worth avoiding in a consistency change: the coordinator search path is now consistent because it separately calls `authorize_with_visibility`, so the inconsistency looks resolved while the higher-traffic path still mints unchecked. Either route dispatch through the new predicate, or state explicitly that registry-configured `visible_namespaces` is a trusted boot-time input not subject to per-extra authorization — but the two paths should not disagree silently.

## A duplicate primary can now be refused

A caller may pass the primary itself in the extra list. Before this change the duplicate was covered by the already-allowed primary check and the token deduplicated to `[P]`. After it, `authorize.visible(P)` is evaluated independently and can refuse the whole call — even though the resulting token would grant nothing beyond `P`, which the caller was already authorized for. `NamespaceToken::mint_with_visibility` dedupes and always includes the primary (`crates/khive-runtime/src/config.rs:88-107`), so this refusal protects nothing. Worth either skipping extras equal to the primary before the check, or confirming the refusal is intended.

## Malformed visibility entries are dropped, not refused

`pack.rs:2200-2215` uses `filter_map` over `Namespace::parse` and skips unparseable entries with a warning. The result is a narrower token and a log line rather than a configuration error. That is defensible, but it means a typo in a visibility entry degrades silently to less access, which at the consumer is indistinguishable from the records not existing.

## Test coverage against the decision table

The added positive test (`runtime.rs:2368-2388`) covers the intended more-permissive row — Read-only extra now admitted — and would redden if the check reverted to `authorize`. The added negative test (`:2414-2444`) uses a gate that denies **both** verbs, so it does not isolate the case this change is actually about: allow on `authorize`, deny on `authorize.visible`. A gate distinguishing the two verbs is what pins the new behaviour; without it the suite would still pass if the extra check consulted the old verb.

Also uncovered: several extras where exactly one lacks Read (the whole mint is refused — worth asserting deliberately); the duplicate-primary boundary above; a primary denial; and gate-error paths. None of the existing integration tests help here, since they use permissive gates.

A useful check on this suite: would any added test fail if the per-extra check always returned "permitted"? Right now the negative test would, but only because its gate denies the ordinary verb too — not because the visible-verb check is exercised in isolation.

ohdearquant (commented):

Verdict on current head (27c7cd4c1bdef40a678af22488d8d6b70ecbd3b2): APPROVE, zero blocking findings. This is a comment, not an approval. A human reviewer decides whether to approve and merge.

One non-blocking note. `authorize_with_visibility` (crates/khive-runtime/src/runtime.rs:1126) applies the Read-classified `authorize.visible` check to every raw `extra_visible` entry, while `NamespaceToken::mint_with_visibility` de-duplicates and the coordinator appends `local`. When the primary namespace also appears among the extras, a request that is valid as Write-on-primary plus Read-on-distinct-extras gets an extra `authorize.visible` check on the primary, which a rights-aware gate could deny. Normalising `extra_visible` against the primary and prior entries before the checks, and passing that same set to token minting, would close it; a regression test with a duplicated primary under a read-aware gate would pin it.

oceanwaves630 (approved):
Approved at 27c7cd4c1bde.

PR #2297, opened 2026-08-30
docs(adr): budget timezone database footprint

### LINKED ISSUE #2129: Decide and record the bundled timezone-database footprint
Date-only due values resolve against a configured display timezone, which makes
`chrono-tz` and its bundled zone database a required part of the shipped binary.

## What is currently true

- `chrono-tz` is an unconditional dependency: declared at `crates/Cargo.toml`
  and consumed by `crates/khive-runtime` and `crates/khive-pack-gtd`. Nothing
  marks it `optional`, so there is no build that omits the zone database.
- The resolved version is 0.10.4.
- The footprint was reported at roughly 1.2 MB, about 3% of the shipped
  `kkernel` binary, when the dependency was introduced. **That figure is
  carried over rather than re-derived here and should be re-measured before
  anyone acts on it.**

## Why the obvious lever is probably the wrong one

`chrono-tz` 0.10 offers a `filter-by-regex` feature that trims the compiled
zone set at build time through the `CHRONO_TZ_TIMEZONE_FILTER` environment
variable (upstream README). It is tempting and, as far as I can tell, unusable
here: the display timezone is chosen at runtime by configuration, so a
compile-time zone filter converts a valid configuration into a zone the binary
cannot resolve. The failure would appear as a configuration error on a
perfectly reasonable setting, and only on builds that had been trimmed.

## The actual choice

1. Accept the footprint and record it as a deliberate budget, so the next
   person measuring binary size finds a decision rather than a surprise; or
2. Put the timezone-dependent surface behind a cargo feature, so a deployment
   that does not use date-only due values can build without the zone database.
   This costs a feature seam through the runtime config and the gtd pack, which
   may well not be worth 3%.

Either is fine. What is not fine is leaving it implicit, which is the state
today. The purpose of this issue is to get the decision written down and the
number re-derived.

ohdearquant:
## Summary

- record an A/B release-binary measurement for the full IANA timezone table versus a representative five-zone filter
- quantify the measured full-table cost as 1,170,816 bytes (1.117 MiB, 2.771%) on the documented toolchain
- retain arbitrary runtime IANA timezone support with an explicit approximately 1.2 MiB / 3% binary budget
- document the measurement procedure and toolchain limitation encountered

## Tests

- same-settings stripped release binary A/B measurement
- `git diff --check`

Closes #2129

### DIFF
--- a/docs/adr/ADR-169-timezone-correct-timestamps.md
+++ b/docs/adr/ADR-169-timezone-correct-timestamps.md
@@ -162,6 +162,34 @@ reconcile them by inference.
 The practical consequence is that a stored value echoed back verbatim already reads as the date
 the caller wrote, so D1 closes the observed defect on its own, without waiting for D2.
 
+**D6. Ship the complete IANA timezone table as an unconditional dependency and account for it as
+a deliberate binary-size budget.** Runtime configuration accepts any IANA timezone name, and
+date-only anchoring uses that configuration in the runtime, GTD, and brain packs. A compile-time
+zone filter would therefore make the validity of the same configuration depend on how a binary
+was built. A cargo feature that removes timezone-aware behavior would add a second configuration
+contract through those crates and is not justified by the measured saving.
+
+The footprint was re-measured on 2026-08-29 from source revision
+`1c23c96a5d6fa306740cfcc8108d6a114e98890c`, using `chrono-tz 0.10.4`, Rust 1.98.0,
+and an arm64 macOS 27.0 host. Both `kkernel` binaries used the workspace release profile
+(`opt-level=3`, thin LTO, one codegen unit). The control enabled `chrono-tz`'s
+`filter-by-regex` feature and selected only `UTC`, `America/New_York`, `Asia/Tokyo`,
+`America/Havana`, and `America/Santiago`; all other source and build settings were identical.
+Because the toolchain's `rust-objcopy` could not load its `libLLVM.dylib`, both artifacts were
+symmetrically stripped with `/usr/bin/strip -S` before comparison.
+
+| Build                                 | Bytes      | Difference from full |
+| ------------------------------------- | ---------- | -------------------- |
+| Complete IANA table                   | 42,248,960 | —                    |
+| Five-zone representative filtered set | 41,078,144 | -1,170,816           |
+
+The complete-table premium over that representative subset is 1.117 MiB, or 2.771% of the
+complete-table binary. This is a conservative estimate of the table's removable footprint
+because the control still contains five zones and their linked aliases. It validates the earlier
+rough statement of “about 1.2 MB / 3%.” The accepted budget is therefore approximately 1.2 MB
+and 3% on a comparable optimized build; it is a decision record, not a cross-platform byte-exact
+CI limit. Re-measure when `chrono-tz` or the release-link profile changes materially.
+
 ## Rationale
 
 D1 before D2 is the whole point of the ordering. The defect the caller actually observes is
@@ -197,8 +225,8 @@ that wiring one function is sufficient.
 
 ### Negative
 
-- Resolving IANA zone names requires a zone database in the binary, which is a new dependency
-  and a size cost that should be measured before this lands.
+- Resolving arbitrary configured IANA zone names requires the complete zone database in the
+  binary. D6 accepts its measured 1.117 MiB / 2.771% premium on the reference release build.
 - The anchoring change is observable: the same date-only string written under two different
   configured zones produces two different instants. That is the intended reading, but it is a
   behaviour change and must be stated in release notes. The scope is worth stating exactly,


ohdearquant:
The measurement is well-specified and I could reproduce its arithmetic and corroborate its inputs. One citation problem, and a note on the red check.

## Checks that pass

- The arithmetic is right: 42,248,960 − 41,078,144 = 1,170,816 bytes, which is 1.1166 MiB (stated 1.117) and 2.7713% of the complete-table binary (stated 2.771%).
- `chrono-tz 0.10.4` matches `crates/Cargo.lock`, and it is an unconditional workspace dependency consumed by `khive-pack-brain` and `khive-pack-gtd` — which is what makes the D6 argument about a compile-time filter changing the meaning of a runtime configuration hold up.
- Recording the symmetric `/usr/bin/strip -S` and the reason for it (`rust-objcopy` could not load `libLLVM.dylib`) is the right instinct — an asymmetric strip would have made the delta meaningless, and stating the workaround lets a re-measurement reproduce the conditions.
- Declaring the control's own residual — five zones plus linked aliases still present, so the figure is a floor on the removable footprint rather than the whole table — keeps the number honest, and calling this "a decision record, not a cross-platform byte-exact CI limit" forecloses someone turning it into a brittle gate.

## The one substantive issue: an unlocatable citation

> It validates the earlier rough statement of "about 1.2 MB / 3%."

I could not find that earlier statement. It is not in the text this diff amends — the Negative bullet being replaced says only that the size cost "should be measured before this lands", with no figure — and a case-insensitive search for `1.2 MB` / `1.2 MiB` / `about 3%` / `~3%` across `docs/` and `crates/` at `origin/main` returns nothing, with a positive control confirming the search reaches those trees.

So as written the sentence claims corroboration a reader cannot locate. If the rough figure lives in an issue or an earlier PR description, cite it explicitly; if it was only ever said informally, the honest form is to drop the "validates" clause and let the measurement stand on its own — it does not need the corroboration, and an unresolvable back-reference is the kind of thing that later gets treated as a second independent source when it is not one.

Stated as provisional rather than absolute: not found by that method in those two trees at that revision.

## The red check is not yours

This PR changes one markdown file, and the failure is a latency assertion in an unrelated crate:

```
khive-db::writer_timeout_sink_stalled_writer
  sink_never_adds_measurable_latency_when_its_writer_is_genuinely_slow
  pool construction took 511.665366ms against a 5000ms-per-write sink
```

Counting roots rather than red boxes: there is exactly one — `CI (ubuntu-latest, shard 1/2)`. The two other red shards are `CANCELLED` fail-fast siblings and `CI gate` is the aggregator, so four red boxes represent one failure.

Worth raising beyond this PR: nextest cancelled on that first failure, and the run reports `3869/4757 tests were not run due to test failure`. A wall-clock assertion on a shared runner will redden unrelated changes periodically, and each time it does it also suppresses 81% of that shard — so a real regression landing in the same window would be invisible behind a timing flake. That is a CI-health item rather than anything this PR should fix.

ohdearquant:
One real failure here; the rest of the red is fail-fast collateral from sibling matrix shards.

The failing test is `sink_never_adds_measurable_latency_when_its_writer_is_genuinely_slow`. That
is a latency assertion, so it is timing-sensitive by construction and a shared runner under load
is exactly where it will be least stable. Before treating it as a defect in this change, worth
establishing which it is — a re-run that passes indicates flakiness rather than a regression, and
a re-run that fails the same way makes it real.

This branch is also 10 commits behind `main`. Merging up first is worth doing regardless, so the
question is asked against current code.

ohdearquant:
## The red `CI gate` is not this change

`CI (ubuntu-latest, shard 1/2)` fails one test:

```
FAIL khive-db::writer_timeout_sink_stalled_writer
     sink_never_adds_measurable_latency_when_its_writer_is_genuinely_slow

panicked at crates/khive-db/tests/writer_timeout_sink_stalled_writer.rs:58:5:
pool construction took 511.665366ms against a 5000ms-per-write sink
  — the sink must never add filesystem-bound latency to pool boot
```

This branch changes exactly one file, `docs/adr/ADR-169-timezone-correct-timestamps.md`, and
touches nothing under `crates/khive-db/`. A single-markdown change cannot alter pool
construction latency, so the failure is ambient.

It is a wall-clock threshold on a shared runner, and it overshot by about two percent. Its two
sibling tests passed in the same run at comparable durations — `..._when_its_directory_is_unwritable`
at 0.546s and `..._when_its_file_is_a_blocked_fifo` at 0.303s — which is the signature of
runner contention rather than a latency regression: the assertion that tripped is the one whose
budget sat closest to normal scheduling noise.

Worth knowing before anyone re-runs this: PR 2315 modifies
`crates/khive-db/tests/writer_timeout_sink_stalled_writer.rs` along with its two siblings, to
derive these bounds from fixtures instead of fixed wall-clock numbers. That is the systematic
fix for this exact assertion. Once it lands, rebasing this branch picks it up; until then a
fresh trigger is the only thing available and it will pass or fail on runner load, not on
anything in this diff.

oceanwaves630 (approved):
Approved at c6535b65. ADR-169 D6 spec-gate signed at this head; docs-only diff; required checks green.

PR #2298, opened 2026-08-30
fix(kg-editor): skip unauthenticated showcase probe

### LINKED ISSUE #2036: kg-editor: static fallback emits 404 console noise (analysis probe, favicon)
Observed in the repository showcase running on the static fallback.

**Symptom.** Every load emits console noise:
- `GET /api/showcase/analyses/<id>` → 404 (the probe for the server-first source before falling back to the static bundle), logged as an error on each load.
- `GET /favicon.ico` → 404 (no favicon is shipped).

**Expected.** The static-fallback probe handles the expected-absent route quietly (e.g. HEAD/fetch with explicit 404 handling that does not log an error), and the app ships a favicon so the request resolves.

ohdearquant:
## Summary

- load the curated static showcase bundle directly when no nonblank operator token is available
- preserve authenticated DB-snapshot reads and their 404-only static fallback
- preserve API probing for dynamic-only entries that have no curated asset

## Tests

- `npx vitest run src/lib/adapters/preferred-showcase-source.test.ts`
- `npm run typecheck`
- `npx eslint src/lib/adapters/preferred-showcase-source.ts src/lib/adapters/preferred-showcase-source.test.ts`
- `git diff --check`

Closes #2036

### DIFF
--- a/apps/kg-editor/src/lib/adapters/preferred-showcase-source.test.ts
+++ b/apps/kg-editor/src/lib/adapters/preferred-showcase-source.test.ts
@@ -52,6 +52,7 @@ describe("preferred showcase source", () => {
     const result = await loadPreferredShowcaseBundle(
       configuredStaticEntry,
       fetchBundle,
+      { accessToken: "operator-secret" },
     );
 
     expect(result.source).toBe("khive-db-snapshot");
@@ -62,6 +63,7 @@ describe("preferred showcase source", () => {
       credentials: "same-origin",
       redirect: "error",
       signal: expect.any(AbortSignal),
+      headers: { authorization: "Bearer operator-secret" },
     });
   });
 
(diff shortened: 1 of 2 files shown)


ohdearquant:
The design holds up. I went in expecting the main risk to be silent staleness — curated content served where live data was expected, with nothing to tell them apart — and that turns out not to apply here, which is worth saying explicitly rather than leaving as an unstated pass.

## The fallback is labeled, end to end

`LoadedShowcaseBundle` carries an explicit `source` discriminant (`"khive-db-snapshot" | "curated-static-fallback"`, `preferred-showcase-source.ts:14-20`), the early return sets `source: "curated-static-fallback"`, and the UI renders it: `showcase.tsx:99` maps the discriminant to the visible string "curated static fallback", with `repo-showcase.tsx:3011` defaulting to the same. So a user with a blank or whitespace-padded token is not shown live-looking data — the provenance travels with the bundle and reaches the screen.

The caveat worth stating: that safety lives in the *caller*, not the adapter. A future consumer that reads `bundle` and discards `source` would reintroduce exactly the problem. Since `source` is the only thing standing between this change and silent staleness, it is worth a comment on the type saying so.

## The predicate and its boundary

```ts
if (!options.accessToken?.trim() && entry.assetPath) {
```

Two things follow, and both are correct:

- **A present-but-unusable token is not blank.** An expired, malformed, or revoked token is non-empty, so it takes the API branch and its failure surfaces — 401/403 propagates as an error rather than being papered over with curated content. That is the right call, and it is the row I most expected to be wrong.
- **Dynamic-only entries still probe.** `entry.assetPath` is falsy for them, so the early return cannot fire and the public-deployment path is preserved.

The 404-only fallback is intact at `:136-143`, with every other non-OK status still throwing at `:145-148`. The change does convert one case: a blank token against a curated entry previously probed unauthenticated and would throw on a 401/403, and now returns the curated bundle without asking. That is the intended behaviour and it is labeled, so it reads as deliberate.

## The one gap: no test for a non-blank unusable token

The suite covers `undefined` (via omitted options, `:128-142`), whitespace-only (`:94-113`), and valid tokens including a 503 (`:164-173`) — but nothing exercises a non-blank malformed or expired token against a 401/403.

That is precisely the boundary this PR moves. The predicate's whole correctness rests on "non-empty implies usable, so let the server decide", and a later refactor toward a *usable*-token check — or any accidental broadening of the early return — would reclassify an expired token as blank and serve cached curated data instead of surfacing the auth failure. The current tests would stay green through that change, because none of them distinguishes "blank" from "present but rejected".

Worth adding: an obvious non-secret placeholder token with a 401 or 403 response, asserting the API URL was called, the trimmed `Authorization` header was sent, the promise rejects, and the static asset was *not* fetched. Keep the whitespace test alongside it — the pair is what pins the boundary, since each alone admits the other's failure.

ohdearquant (commented):

Verdict on current head (bbdf69631ff3bdba8e1297ac00f7d14a75e811c4): APPROVE, zero blocking findings. This is a comment, not an approval. A human reviewer decides whether to approve and merge.

Two non-blocking notes, both documentation.

1. ADR-147 (`docs/adr/ADR-147-repo-showcase-bundle.md:396`) still says a configured entry requests the opaque ID route first. With this change (`apps/kg-editor/src/lib/adapters/preferred-showcase-source.ts:84`), an absent or blank operator token loads the approved static asset directly and only a non-blank token keeps route-first behaviour with 404-only fallback. An amendment stating that order would bring the ADR back in line with the code.

2. `apps/kg-editor/README.md:145` and `apps/kg-editor/DEMO.md:23` describe the bearer token as sent on every snapshot request. With no non-blank token there is now no snapshot request at all; the two documents should say that a non-blank token enables the authenticated request and that a blank one loads the curated asset directly.

oceanwaves630 (approved):
Approved at bbdf69631ff3.

PR #2299, opened 2026-08-30
fix(brain): report zero-match unbind counts

### LINKED ISSUE #2139: brain.unbind returns ok:true when zero bindings matched
## Defect

`brain.unbind` reports success when the given selector matches no rows. A mistyped
`profile_id` (or a binding already removed) returns `ok: true`, indistinguishable from a
real unbind, so a caller cannot tell whether state changed.

Reproduce:

1. `brain.unbind(profile_id="<any well-formed id with no binding>")`
2. Response is `ok: true` with no matched/removed count.

## Why it matters

Unbind is a state-clearing operation; its success signal is the only confirmation the
caller gets. A zero-match success turns typos into silent no-ops that read as completed
cleanup. Same class as delete-by-id verbs, which do report not-found.

## Acceptance

- Response carries a matched/removed count, and zero matches is distinguishable from a
  successful removal (either `ok: false` with a not-found error, or an explicit
  `removed: 0` the caller can gate on — consistent with how sibling verbs treat missing
  subjects).
- Test covering the zero-match path.

ohdearquant:
## Summary

- make zero-match `brain.unbind` responses report an explicit numeric removal count
- expose canonical `removed` while retaining the compatibility `unbound` field
- document and test the zero-removal response contract

## Tests

- focused brain unbind tests (3 passed)
- affected clippy and formatting checks
- `git diff --check`

Closes #2139

### DIFF
--- a/crates/khive-pack-brain/src/handlers.rs
+++ b/crates/khive-pack-brain/src/handlers.rs
@@ -2495,7 +2495,12 @@ impl BrainPack {
                 });
 
                 let removed = before - state.bindings.len();
-                Ok(json!({ "unbound": removed }))
+                // `removed` is the canonical mutation count. Keep `unbound`
+                // as a compatibility alias for existing callers and smoke
+                // tests. In particular, both fields remain present at zero so
+                // a successful no-op cannot be mistaken for a confirmed
+                // removal.
+                Ok(json!({ "removed": removed, "unbound": removed }))
             },
         )
         .await
--- a/crates/khive-pack-brain/src/tests.rs
+++ b/crates/khive-pack-brain/src/tests.rs
@@ -516,7 +516,28 @@ async fn dispatch_unbind_removes_binding() {
         )
         .await
         .unwrap();
+    assert_eq!(result["removed"], json!(1u64));
     assert_eq!(result["unbound"], json!(1u64));
+}
+
+#[tokio::test]
+async fn dispatch_unbind_reports_zero_when_selector_matches_nothing() {
+    let (pack, rt) = make_pack();
+    let registry = empty_registry();
+    let token = rt.authorize(Namespace::local()).unwrap();
+
+    let result = pack
+        .dispatch(
+            "brain.unbind",
+            json!({"profile_id": "well-formed-but-unbound"}),
+            &registry,
+            &token,
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(result["removed"], json!(0u64));
+    assert_eq!(result["unbound"], json!(0u64));
 }
 
 // ── UE5-H1: brain.profiles lifecycle filter public API ───────────────────
--- a/docs/guide/api-reference.md
+++ b/docs/guide/api-reference.md
@@ -1308,6 +1308,10 @@ Remove rows from the profile resolution table. At least one filter is required.
 request(ops="brain.unbind(actor=\"role:implementer\")")
 ```
 
+The result includes `removed`, the number of matching bindings deleted. A successful
+request that matched nothing returns `removed: 0`. The legacy `unbound` field carries
+the same count for compatibility.
+
 ### `brain.bindings` — Assertive
 
 List rows in the profile resolution table, optionally filtered.


ohdearquant:
The new test is worth having. The production change does not do what the code comment says it does.

## The stated justification describes a property that already held

The added comment says:

> In particular, both fields remain present at zero so a successful no-op cannot be mistaken for a confirmed removal.

At `origin/main` the handler already returned:

```rust
let removed = before - state.bindings.len();
Ok(json!({ "unbound": removed }))
```

`json!` always emits the key, and a selector matching nothing yields `before == state.bindings.len()`, so `removed` is `0` and the response was already `{"unbound": 0}` — an `Ok`, with the count present, at zero. A no-op was already distinguishable from a confirmed removal by exactly the mechanism the comment credits to this change.

So the caller-visible effect of the production diff is that the response now carries **two keys with the same value** instead of one. No count changes, no status changes, no previously-absent field becomes present. If the intent is to make `removed` the canonical name and keep `unbound` as an alias, that is a reasonable rename — but it should be described as a naming change, not as closing a reporting gap, and the comment as written will mislead the next reader into thinking the zero case was previously unreportable.

If there *is* a real defect here — a path where `brain.unbind` errors or omits the count instead of returning zero — it is not the one in this diff, and pointing at it would change my read entirely.

## The test is the real contribution, and it is a good one

`dispatch_unbind_reports_zero_when_selector_matches_nothing` pins behaviour that genuinely had no coverage: a well-formed selector matching nothing returns `Ok` with a zero count rather than an error. That property was load-bearing and untested, and a future change that made a zero-match request fail — a plausible "nothing to do, so report a problem" refactor — would previously have gone unnoticed. Keep it regardless of what happens to the production hunk.

Note it currently asserts both keys. If `unbound` is meant to be a deprecated alias, asserting both is right for now; if `removed` is meant to replace it, the test should be the thing that fails when the alias is finally dropped, which it will be.

## Field-consumer note

The only readers of `unbound` in the tree are three assertions in `crates/khive-pack-brain/src/tests.rs` (`:519`, `:779`, `:4609`) — there is no in-repository production consumer. The `removed` occurrences at `docs/schemas/khive-review-v1.schema.json` belong to an unrelated schema and are not this field. That does not argue against keeping the alias, since this is a public verb response and external callers are not visible from here — but it does mean the "existing callers" the comment cites cannot be confirmed from the repository, and the compatibility argument rests on out-of-tree consumers rather than in-tree ones.

## Documentation

The `docs/guide/api-reference.md` addition is accurate and reads well. One consequence worth making explicit there: since both fields carry the same count, a caller should not treat their agreement as corroboration — they are one value emitted twice.

ohdearquant:
This is correct, and the compatibility half of it is load-bearing rather than defensive — worth
saying because a duplicated field usually is not.

`unbound` has two live consumers outside this crate: `tests/smoke_test.py:1134` asserts on it
(`unbound.get("unbound", 0) >= 1`) and prints it at :1137, and `tests/smoke_brain.py:335` reads
it. Dropping it in favour of `removed` would have broken both, and neither is in this crate's test
suite, so the local tests would have stayed green while the smoke run went red. Keeping the alias
was the right call and the comment's stated reason checks out.

The count itself is sound. `before` is captured before `retain`, and `retain` only removes, so
`before - state.bindings.len()` cannot underflow — worth confirming explicitly since an unsigned
subtraction on a length is where that bug usually lives.

The new zero-match test earns its place. It is not merely a second example: before this change the
no-op path emitted `unbound: 0` and nothing asserted it, so the field that distinguishes "matched
nothing" from "removed something" was untested on exactly the branch where the distinction matters.
The `removed` assertion in it also fails without the production change (the key would be absent),
so the test is pinned to the fix rather than passing either way.

One advisory, not a blocker: two fields now carry the same value with `unbound` documented as
"legacy", but there is no deprecation marker in the code or a stated removal condition. The next
person to see the duplication has nothing telling them what has to become true before it can go. A
one-line note naming the two smoke-test call sites as the blocking consumers would make that
self-answering.

PR #2300, opened 2026-08-30
fix(mcp): fail oversized daemon responses

### LINKED ISSUE #2219: list verbs: frame-budget overflow returns ok:true with result omitted — silent end-of-list to pagers
When a \`knowledge.list\` response exceeds the response frame budget, the operation returns:

\`\`\`json
{"ok":true, "result_omitted":"operation succeeded; result omitted because the response frame budget was exceeded", "tool":"knowledge.list"}
\`\`\`

with top-level \`status: success\`, \`succeeded: 1\`, and no error field (241 bytes total).

Measured at \`limit=500\`, \`offset=0\` against a large store: 500 → result omitted / 0 rows visible; 200 → 200 rows / 4.2MB; 100 → 100 rows / 2.0MB.

**Why this is a defect and not a presentation choice:** the natural pagination loop terminates on an empty page. A caller walking the store at \`limit=500\` receives what is indistinguishable from an empty first page, stops at offset 0, and reports a complete walk over zero rows. The failure is silent and fails toward looking finished — the worst direction for a caller computing a set difference or completeness check.

**Suggested fix:** the frame-budget-exceeded path should return a retryable error (e.g. \`retryable: true\`, code naming the frame budget) so callers know to reduce \`limit\` and retry, instead of \`ok:true\` with the payload dropped. At minimum the summary should not count the op as succeeded when the result was discarded.

Applies to any verb routed through the same frame-budget guard, not just \`knowledge.list\`.

ohdearquant:
## Summary

- replace silently dropped oversized daemon results with a typed, non-retryable frame-budget error whose `recoverable` field tells the caller whether to reduce the result size and reissue (`reduce_result_size`) or read the committed outcome (`read_outcome`)
- recompute batch status and summary after applying the response budget
- document the fail-safe response behavior

## Tests

- frame-omission regression tests (4 passed)
- real daemon oversized-result integration test
- affected check, clippy, and formatting checks
- `git diff --check`

Closes #2219


ohdearquant:
The core of this is right, and it is a real fix rather than a relabeling. Before this diff an oversized daemon result came back as `ok: true` with no `result` and a `result_omitted` string, counted as a success, under a top-level `status: "success"` — a pager could read the missing payload as an empty terminal page. After it, the entry is `ok: false` with a typed error and the aggregate is recomputed to `partial`. The caller outcome genuinely changes, and `refresh_frame_budget_outcome` keeps `results.len()` and `summary.total` consistent so a refused entry can no longer be counted as succeeded. Refusal rather than truncation, with no partial `result` emitted beside the error, is the right shape.

Two problems, one of which can duplicate a mutation.

## 1. In a chain, the envelope contradicts the execution contract — and invites a replay

The documented contract is explicit:

> A failed chain operation prevents subsequent operations from running; completed operations are not rolled back.
> — `docs/guide/request-and-dsl.md:84-85`

But frame fitting happens *after* the whole chain has been dispatched. So for `large_result(...) | create(...)`:

- both operations execute;
- the first entry is then rewritten to `ok: false` with `response_frame_budget_exceeded`;
- the second entry is returned as `ok: true`.

The returned envelope therefore shows a failed chain operation *followed by a successful one*, which the contract says cannot happen. A caller reasoning from the documented semantics concludes the `create` never ran. It did.

Now combine that with `retryable: true`: a client that retries the chain on a retryable error re-executes `create`, and the mutation happens twice. That is the part that makes this more than a cosmetic inconsistency — the two behaviours are individually defensible and jointly produce a duplicate write.

The distinction the envelope is missing is between *"this operation did not happen"* and *"this operation happened but its result could not be delivered"*. Those need different outcomes, because only the first is safe to replay. At minimum the post-render refusal should be distinguishable from an ordinary chain failure, and it should not be presented as safely retryable while later operations in the same chain have already committed.

## 2. `retryable: true` is unconditional for a deterministic refusal

```rust
"message": "operation result exceeded the daemon response frame budget; reduce limit or result size and retry",
"retryable": true,
```

The message and the flag disagree about who is meant to act. The message tells a human to *change the request*; the flag tells a machine it may reissue *this* request. Reissuing it identically produces the identical oversized result — nothing transient is being cleared — so an automatic retry loop makes no progress. The test itself constructs a deterministic `large_result(bytes = MAX_FRAME_BYTES + 1024)`, which is exactly the case that never succeeds on replay.

Either classify it non-retryable for an identical request, or define a distinct "retry after reducing the request" contract and say plainly that automatic identical retries are prohibited. The ADR and guide repeat `retryable: true` (`docs/adr/ADR-016-request-dsl.md:456-460`, `docs/guide/request-and-dsl.md:121-125`) without either caveat, so whichever way this resolves, all three need to say the same thing.

Worth noting this repository has the inverse defect on record — a genuinely transient condition marked `retryable: false` — so the direction matters and should be decided from the failure's determinism, not from which mistake feels more familiar.

## Budget mechanics — checked, no issue found

`MAX_FRAME_BYTES` is 8,388,608, a compile-time constant with no runtime knob. The fitter measures the serialized `DaemonResponseFrame` including JSON escaping, and uses `request_id = Some(u64::MAX)`, so it can conservatively *over*-estimate by a few bytes for a shorter id but never underestimates the real transport frame. The comparison is `<=`, so the exact boundary is accepted. That all lines up with `write_frame`'s own check.

The separate `BATCH_RESPONSE_BUDGET_BYTES` (`MAX_FRAME_BYTES / 2`) measures pre-presentation entry bytes, a smaller and differently-scoped unit — conservative in the safe direction, so it can refuse queued work that would have fit rather than the reverse. Not changed here, just noting the two budgets measure different things.

## Test gaps

The three changed tests pin the single-entry case well and would redden on a revert to `ok: true`/`result_omitted`, on a dropped typed error, on lost search-degradation metadata, or on an over-cap frame. What none of them covers:

- a refused entry **inside a multi-entry batch** beside a successful sibling — so nothing asserts the recomputed multi-entry `summary` counts or their relationship to `results.len()`, which is precisely what `refresh_frame_budget_outcome` exists to get right;
- **any chain case**, including the duplicate-side-effect scenario above;
- identical-request retry behaviour, despite `retryable: true` being asserted;
- the exact-at-cap boundary, as opposed to an over-cap value.

The batch and chain cases are the ones I would want before this lands, since they are where both findings live.

oceanwaves630 (approved):
Approving. CI is green on this head and the remaining delta is a documentation link fix.

PR #2301, opened 2026-08-30
fix(git): bound remote ingest work

### LINKED ISSUE #2072: git.digest: include=["issues"] pays a full git clone although issues ingest reads the GitHub API
An issues-only digest (`git.digest(source=<url>, include=["issues"])`) clones the full repository before any issue is read. Issues and pull requests ingest through the `gh` CLI, and the repository probe resolves owner/repo from the cloned checkout (`crates/khive-pack-git/src/ingest.rs`, gh probe path), so the clone exists only to derive a slug that is already present in the source URL.

Consequence: a tracker whose git history exceeds the per-clone size cap cannot be digested for issues at all today — the clone fails and the request errors before the API read starts, even though the API read itself would succeed. Measured against a ~1.7 GiB-history repository: 150-230 s of transfer, then failure, zero items ingested.

Proposed fix: when no requested source requires a worktree (issues/PRs only), resolve owner/repo from the source URL and skip the clone entirely; commits-inclusive digests keep the current path.

### LINKED ISSUE #2073: git.digest: per-clone size cap is enforced only after the full transfer completes
The per-clone size cap (`KHIVE_GIT_DIGEST_CLONE_MAX_BYTES`, default 1 GiB) is measured after `git clone` finishes and the error path then deletes the checkout (`crates/khive-pack-git/src/cache.rs`, `CloneTooLarge`). An over-cap repository therefore costs the full transfer before failing: measured 150-230 s and ~1.7 GiB of traffic per attempt, repeated on every retry by any caller with a polling cadence.

Proposed fix, in preference order: probe the advertised repository size before transfer where the host exposes it; otherwise enforce the cap during transfer (e.g. monitor the staging directory and abort the child process at the cap) so an oversized clone fails in seconds. The error message should keep its current remediation text.

ohdearquant:
## Summary

- avoid cloning for API-only remote digests while retaining the supplied GitHub identity
- monitor fresh clones and terminate/reap work that exceeds the configured cap
- document the remote-ingest work-bound contract

## Tests

- `khive-pack-git` library suite (263 passed)
- acceptance suite (94 passed)
- workspace all-target/all-feature check and clippy
- formatting check

Closes #2072
Closes #2073


ohdearquant:
The conditional-clone design holds up: I traced `repo_path` through both branches and the non-clone path never reaches a consumer that needs a real repository. One blocking gap in the bounds, one stale doc, and a coordination note.

## 1. The size cap does not bound an existing-slot fetch

The new monitoring is only on the fresh-clone path. `clone` samples the child every 25 ms and kills it on cap excess (`crates/khive-pack-git/src/cache.rs:922-950`), returning `CloneTooLarge` — a refusal, with staging cleaned up. Good shape.

The existing-slot path has no equivalent. It waits for `git fetch` to complete (`crates/khive-pack-git/src/cache.rs:1161-1185`) and only then measures and evicts an oversized slot (`:306-318`, `:380-389`). So a remote digest that hits an already-owned cache slot can pull an arbitrarily large pack before the 1 GiB default cap is observed. If the transfer fills the filesystem, the failure arrives as host-wide resource pressure or a failed fetch, before cleanup begins. The caller does get a hard error rather than a partial report — but the process was never stopped at the cap, which is the property being claimed.

That matters because the amended ADR states this cap bounds single-repository disk consumption before LRU eviction applies (`docs/adr/ADR-088-amendment-1-git-digest.md:169-178`). That is true for fresh clones and not for fetches into an existing slot.

Suggested direction: route existing-slot fetch/refetch through the same monitored child/process-group path with a final measurement — or narrow the ADR and docs to persistent-slot enforcement and say plainly that transfer-time usage is unbounded for that path. Either is defensible; the current pairing of code and claim is not.

## 2. `docs/api/cache.md` still shows the pre-`--no-checkout` command

`crates/khive-pack-git/docs/api/cache.md:94-99` shows `git clone --filter=blob:none`; the implementation runs `git clone --filter=blob:none --no-checkout` (`crates/khive-pack-git/src/cache.rs:904-906`). The omitted flag is the material one — the source comment at `:880-887` notes that a checkout would materialize all reachable blobs, which is most of the point of the change. An operator reading that section cannot tell the worktree is deliberately suppressed.

## 3. Coordination: #2290 edits the same region of `handlers.rs`

This PR rewrites `handlers.rs` lines 154-180 (conditional `repo_path`) and 219-233 (remote dispatch). #2290 edits an overlapping range in the same file. The two changes are complementary rather than contradictory — they fix different things — but they will not merge cleanly, and the resolution needs checking rather than trusting.

Specifically: this PR's remote-clone error is `RuntimeError::InvalidInput` with message `remote clone/fetch of {:?} failed: {e}`, where the source slot is `redact_repo_url(canonical)` and `{e}` is a `CacheError` rendered as-is (`handlers.rs:161-165`). Whichever change lands second should be verified to still produce exactly that, since a clean automatic merge can preserve both sides' edits and still yield a wrong result.

Related, and worth stating because this PR's own error paths are otherwise careful: `handlers.rs:131` forwards `parse_source` errors unchanged, and the parser embeds the trimmed user-supplied source without redaction (`source.rs:49-88`). An invalid source spelling carrying credentials therefore reaches that error message. This predates this PR and is not introduced by it — it is what #2290 addresses — but it is the one remaining unredacted source path in this file after this change lands.

## Bounds, as read

- `max_items` (default 500, clamped 1..2000, `handlers.rs:115-147`) is a resumable stop, not a refusal: `done:false` plus `stopped_early` on the source expose the condition, and effective/requested values with a clamp warning come back at `:249-258`. It counts create attempts, not skips or page fetches.
- `PAGE_LIMIT` (1000, `ingest.rs:2369-2405`) — a stalled full page refuses to claim exhaustion rather than silently reporting completeness.
- Per-clone cap: 1 GiB default, `KHIVE_GIT_DIGEST_CLONE_MAX_BYTES`, refusal not truncation. The 25 ms figure is a sampling interval, not a wall-clock bound; there is no new timeout or depth bound.

## Test gaps

`digest_remote_issues_only_never_invokes_git_clone` (`tests/acceptance.rs:6491-6566`) pins the no-clone branch well and would redden on a reintroduced clone or a probe that stopped being source-bound. The cache unit test at `cache.rs:1879-1947` covers the fresh-clone kill path. Not covered: PR-only and mixed API-only sources, an unsluggable remote, and — most relevant to finding 1 — any test asserting a *bounded fetch* outcome. The suite asserts a bounded `CloneTooLarge` and nothing equivalent for the fetch path, which is precisely where the bound is missing.

## What is right

The capability comment in `source.rs` checks out: both branches carry `expected_github_repo`, and `probe_gh_repository` validates it before invoking `gh repo view <slug>` (`ingest.rs:863-895`), including on the path that no longer clones. The no-slug case reports the source unavailable instead of falling back to the daemon cwd's `origin` — fail-closed, which is the right call. The no-clone branch also disables origin derivation, clears `GH_REPO`/`GH_HOST`, and passes explicit `--repo` to listings.

ohdearquant (commented):

Verdict on head 6c9adb25b862ec32c1d22340732f7dc2d89c1d9d: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving at this exact commit. Verified before approving: the head matches the commit named in the request, every check run at this commit has completed with no failures, no change requests are outstanding, and the approving account is not the authoring account.

PR #2302, opened 2026-08-30
fix(runtime): allow UUIDs near generic token prose

### LINKED ISSUE #1963: comm.send: credential-proximity guard false-positives on document titles near ids, and its suggested remediation does not clear the refusal
Reproduced twice on the same payload: `comm.send` refused a message where the phrase "design language and token contract" (a real document title; the trigger word is `token`) sat in the line above a document UUID. No credential or secret was present anywhere in the message.

Two defects:
1. The proximity window treats a generic vocabulary word near a UUID as a credential signature. Titles containing words like "token" adjacent to record ids are routine in this corpus, so the false-positive rate on legitimate cross-reference messages is structural, not incidental.
2. The refusal's own suggested remediation (move the id to its own line) did not clear the guard on retry; only deleting the word did. Guidance that does not work trains senders to strip content rather than restructure it.

Proposal: tighten the trigger to credential-shaped values (entropy/format of the VALUE side, not vocabulary proximity), or at minimum make the refusal's remediation advice something that actually passes.

ohdearquant:
## Summary

- distinguish generic standalone `token` prose from explicit credential labels when classifying UUID-shaped values
- preserve high-entropy detection for `service token <opaque>` and `token=<value>`
- preserve UUID blocking for assignments and explicit labels such as `api_key`

## Tests

- `cargo test --manifest-path crates/Cargo.toml -p khive-runtime secret_gate::tests:: --lib` (242 passed)
- `cargo check --manifest-path crates/Cargo.toml -p khive-runtime`
- `cargo clippy --manifest-path crates/Cargo.toml -p khive-runtime --all-targets -- -D warnings`
- `git diff --check`

Closes #1963

### DIFF
--- a/crates/khive-runtime/src/secret_gate.rs
+++ b/crates/khive-runtime/src/secret_gate.rs
@@ -15,8 +15,10 @@
 //!    secret keys, URL userinfo (`scheme://user:pass@`).
 //! 2. **High-entropy token heuristic** — base64/hex/base64url runs ≥ 24 chars
 //!    near a trigger word (key, secret, password, credential, bearer, auth,
-//!    apikey, api_key, access_key, private_key). The word `token` alone is
-//!    NOT a trigger, to avoid blocking `tokenizer_*`, `token_count`, etc.
+//!    apikey, api_key, access_key, private_key). A standalone `token` still
+//!    triggers opaque entropy detection, but does not by itself label a UUID
+//!    as a credential; compound identifiers such as `tokenizer_*` and
+//!    `token_count` remain excluded.
 //!
 //! Credential-shaped labels and assignments dominate the allowlist below.
 //! Public VCS revisions and plausible file paths remain exempt in ordinary
@@ -953,7 +955,8 @@ fn check_entropy_heuristic<'a>(
         // `shannon_entropy` over its bytes is a true per-character entropy.
 
         // Compute the trigger window before any shape-based allowlist decision.
-        // UUID and base64 content-hash exemptions remain trigger-sensitive.
+        // UUIDs require credential-label context rather than a generic mention
+        // of `token`; base64 content-hash exemptions remain trigger-sensitive.
         // VCS revisions and file paths use narrower syntactic context below.
         let window_start = floor_char_boundary(text, tok_offset.saturating_sub(TRIGGER_WINDOW));
         let window_end = floor_char_boundary(text, tok_offset + raw_token.len() + TRIGGER_WINDOW);
@@ -967,17 +970,26 @@ fn check_entropy_heuristic<'a>(
         let near_trigger = contains_trigger(&window[..raw_start])
             || contains_trigger(&window[raw_end..])
             || has_inline_credential_trigger(raw_token);
+        let uuid_near_credential_label = contains_credential_label_trigger(&window[..raw_start])
+            || contains_credential_label_trigger(&window[raw_end..])
+            || has_inline_credential_trigger(raw_token);
 
-        // Step 1 (see doc: per-token flagging sequence). UUID/content-hash checks fall
-        // through to detection near a trigger rather than being silently passed —
-        // hex-shaped entropy alone (<=4.0 bits/char) can never reach ENTROPY_THRESHOLD.
-        if near_trigger && value_candidates(token).any(is_uuid_canonical) {
+        // Step 1 (see doc: per-token flagging sequence). UUIDs fall through only
+        // beside an explicit credential label; the generic word `token` remains
+        // trigger context for opaque values but is common in design prose. Content
+        // hashes retain the broader trigger rule. Hex-shaped entropy alone (<=4.0
+        // bits/char) can never reach ENTROPY_THRESHOLD.
+        let has_uuid_candidate = value_candidates(token).any(is_uuid_canonical);
+        if uuid_near_credential_label && has_uuid_candidate {
             return Some((token, "uuid-near-trigger"));
         }
         if near_trigger && value_candidates(token).any(is_base64_content_hash) {
             return Some((token, "content-hash-near-trigger"));
         }
-        if !near_trigger && (is_uuid_canonical(token) || is_base64_content_hash(token)) {
+        if !uuid_near_credential_label && is_uuid_canonical(token) {
+            continue;
+        }
+        if !near_trigger && is_base64_content_hash(token) {
             continue;
         }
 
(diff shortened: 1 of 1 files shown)


ohdearquant:
The intent is right — a generic `token` in design prose should not turn a UUID into a credential — but the new exemption and the existing detection normalize the token differently, so the change only takes effect for some spellings of the same value. I measured before/after against the built crate rather than reasoning from the diff; table below.

## The exemption is bypassed by adjacent punctuation

The flagging branch asks whether any *value candidate* is a UUID:

```rust
let has_uuid_candidate = value_candidates(token).any(is_uuid_canonical);
if uuid_near_credential_label && has_uuid_candidate { ... }
```

`value_candidates` normalizes through `wrapper_strip_repeated` → `strip_wrappers`, which trims `{ } ( ) [ ] " ' \` . , ;`.

The new exemption branch asks a different question:

```rust
if !uuid_near_credential_label && is_uuid_canonical(token) {
    continue;
}
```

Here `token` is `strip_delimiters(raw_token)`, and `strip_delimiters` trims only `" ' \` : = , ;` — it does **not** trim `.` `(` `)` `[` `]` `{` `}`. `is_uuid_canonical` requires length exactly 36, so for those characters the value is still recognized well enough to be *flagged*, but no longer well enough to be *exempted*.

When the exemption is skipped, control reaches the hex-credential step. A canonical UUID is five pure-hex runs of 8+4+4+4+12, and `contains_normalized_hex_credential` concatenates consecutive hex runs across separators: 32, which is in `HEX_CREDENTIAL_LENGTHS`. So the value blocks again, now reported as `hex-credential-token`.

## Measured behaviour, before and after

Same content in both columns, content prefix `token contract\n` unless noted, value `550e8400-e29b-41d4-a716-446655440000`:

| token as written | before | after |
| --- | --- | --- |
| `api_key <uuid>` (control) | `uuid-near-trigger` | `uuid-near-trigger` |
| `token=<uuid>` (control) | `uuid-near-trigger` | `uuid-near-trigger` |
| `<uuid>` | `uuid-near-trigger` | allowed |
| `<uuid>,` | `uuid-near-trigger` | allowed |
| `<uuid>;` | `uuid-near-trigger` | allowed |
| `"<uuid>"` | `uuid-near-trigger` | allowed |
| `` `<uuid>` `` | `uuid-near-trigger` | allowed |
| `<uuid>.` | `uuid-near-trigger` | **`hex-credential-token`** |
| `(<uuid>)` | `uuid-near-trigger` | **`hex-credential-token`** |
| `[<uuid>]` | `uuid-near-trigger` | **`hex-credential-token`** |
| `Your token is <uuid>` | `uuid-near-trigger` | allowed |
| `Your token is <uuid>.` | `uuid-near-trigger` | **`hex-credential-token`** |
| no trigger word, `<uuid>` and `<uuid>.` | allowed | allowed |

The practical consequence: `Your token is <uuid>` is now allowed, and the same sentence ending in a period is not. A UUID inside parentheses or brackets — ordinary in Markdown — is also unchanged. So the false positive this PR targets is still live for a large share of real prose, and which spelling works is not predictable from the rule as documented.

The asymmetry predates this PR: the old third branch also tested `is_uuid_canonical(token)`. It was masked, because whenever a trigger was near, the first branch fired and blocked every spelling uniformly. This change makes it caller-visible.

Suggested direction: have the exemption ask the same question as the detection —

```rust
if !uuid_near_credential_label && has_uuid_candidate {
    continue;
}
```

`has_uuid_candidate` is already computed one line above, so the two branches would then agree on what counts as a UUID, and the punctuated forms would behave like the bare one.

## Test coverage

`allows_uuid_on_line_after_benign_token_contract_title` uses a UUID at the very end of the content with no trailing punctuation — the one shape that takes the new path. Adding the same assertion with a trailing period would fail today. `generic_token_uuid_exemption_keeps_strong_credential_controls` is a good must-block control, but all four of its cases carry an explicit label or assignment, so none of them exercises the boundary this change actually moves.

Worth adding in both directions: a bare-`token` case with punctuation (currently blocked, presumably should not be), and a must-block case whose only context is bare `token`, to pin the relaxation's intended limit.

## On the relaxation itself

Naming the class plainly, since it is a deliberate security trade: after this change a canonical UUID is emitted unmasked whenever the only credential context within the trigger window is the standalone word `token` — not `token=`, not `token:`, not an `api_key`-style label, all of which still block via `has_assignment_credential_trigger`. UUID-shaped bearer and session tokens are a real credential shape, so that class is not empty. That may well be the right call given how often `token` appears in design prose, but it is worth stating in the module doc as the rule's actual boundary.

Minor, from reading rather than measurement: within `contains_credential_label_trigger`, `has_token_assignment` looks fully subsumed by `has_assignment_credential_trigger` — the latter trims punctuation back to the label and matches `label == "token"`, covering `token=` and `token:` plus the spaced and quoted forms the former misses. The doc comment credits `has_token_assignment` for keeping `token=<uuid>` authoritative; the guarantee appears to come from the other predicate.

ohdearquant:
Approving.

The behavioural change is a single clause. `contains_credential_label_trigger` is
`contains_trigger` minus `has_standalone_token`, and it is consulted only on the
`is_uuid_canonical` path — content hashes keep the broader `near_trigger` rule.

I measured the disagreement set between the two predicates over a matrix of
UUID-in-context cases rather than reasoning about it. They diverge on exactly one
class: a bare `token` word with no assignment syntax anywhere in the window.

Previously flagged, now allowed — this is the class the change is for:

- `token is <uuid>`, `service token <uuid>`, `refresh token <uuid>`,
  `token contract` on the line above a `<uuid>`.

Still flagged, via `has_token_assignment` / `has_assignment_credential_trigger`,
both of which survive in the narrowed predicate:

- `token=<uuid>`, `token: <uuid>`, `{"token": "<uuid>"}`, `X-Auth-Token: <uuid>`,
  `Authorization: Bearer <uuid>`, `api_key <uuid>`, `secret <uuid>`.

The quoted-key form is worth calling out, because it is the one that looks at risk
on a reading. `has_token_assignment` requires `=` or `:` immediately after `token`,
so it does not match `"token": ...` by itself. `has_assignment_credential_trigger`
is what covers it — it trims the closing quote before taking the label — so the JSON
shape is retained. That is measured, not inferred.

Two structural points I checked:

- Splitting the combined `!near_trigger && (is_uuid_canonical || is_base64_content_hash)`
  skip into two separate skips is behaviour-preserving for content hashes, and the new
  early `continue` for UUIDs costs nothing: hex-and-dash content caps near 4.0 bits/char
  and cannot reach the 4.5 threshold, so the explicit UUID rule was always the only
  thing that could flag a canonical UUID. Falling through to the entropy arm was never
  reachable for this shape.
- The flag path keeps `value_candidates(...).any(is_uuid_canonical)` rather than
  whole-token matching, so punctuated values such as `token=<uuid>` are still inspected.

`generic_token_uuid_exemption_keeps_strong_credential_controls` is the right test to
have added. A loosening carrying only a must-not-flag test cannot show what it still
catches, and that arm is what makes this reviewable.

PR #2303, opened 2026-08-30
ci(coverage): separate measurement from ratchet

### LINKED ISSUE #2278: Coverage ratchet reports failure when it never measured coverage
## Summary

The `Coverage ratchet` job runs two steps in sequence: `Compute coverage`, then `Check coverage does not regress`. When the compute step fails, the check step is **skipped** and the job reports `failure`.

The result is a red check named for a judgment it never rendered. "Coverage ratchet: failure" is indistinguishable, from outside, between:

- coverage was measured and it regressed — a real finding the author must act on, and
- coverage could not be measured at all — an infrastructure problem the author cannot act on.

Those need opposite responses, and the current signal cannot tell them apart.

## Observed

Two open pull requests currently show the identical step pattern:

```
job: Coverage ratchet   conclusion=failure
  step 7: failure   Compute coverage
  step 8: skipped   Check coverage does not regress
```

In both cases every other job is green — all CI shards pass — and the only other red is `CI gate`, whose sole failing step is `Check aggregated job results`. So `CI gate` is derivative: one underlying cause presents as two red checks, and neither names it.

The job is not universally broken. A third pull request's `Coverage ratchet` concluded `success`, so the compute step fails conditionally rather than always. Whatever that condition is, the reporting problem is independent of it.

## Why this matters beyond the immediate cause

A gate in this shape **can never report PASS when its compute step is unhealthy**, and it blocks merges while doing so. Worse, it fails in the direction that looks like the author's fault: the natural reading of a red "Coverage ratchet" is "my change reduced coverage," which sends the author looking for a regression that was never measured.

This is a general property of compute-then-judge gates, not a property of coverage specifically. Any gate that computes an input and then judges it needs to distinguish "the judgment failed" from "the input could not be produced."

## Proposal

1. **Separate the signals.** Report an unavailable measurement distinctly from a failed judgment — a separate check name, or a conclusion that does not read as a coverage result. `neutral` exists for exactly this and does not block.
2. **Do not let a skipped judgment present as a rendered one.** If `Check coverage does not regress` never ran, no coverage result exists, and the surfaced status should say so in those words.
3. **Consider whether an unmeasurable ratchet should block at all.** A gate that cannot measure has no evidence of regression. Blocking on absence of evidence is a policy choice worth making deliberately rather than inheriting from step ordering.
4. Separately, the compute-step failure itself needs a root cause; this issue is about the reporting, and fixing the reporting is useful even after the compute failure is resolved, because it will recur.

## Notes

`CI gate` correctly aggregates and is not itself defective — but when counting problems from a red build, it is worth knowing that an aggregate and a fail-fast cancellation are not separate defects.

ohdearquant:
## Summary

- report coverage computation under a distinct advisory measurement job
- preserve the failed compute outcome for an explicit unavailable warning and job summary
- run Coverage ratchet only when a measurement exists, while real measured regressions still fail
- make CI gate track both the measurement signal and the conditional judgment

## Behavior

When cargo llvm-cov fails, Coverage measurement records the failure as advisory and Coverage ratchet is shown as skipped. No regression result is claimed and CI gate accepts the legitimate skip.

When measurement succeeds, Coverage ratchet receives the measured value and preserves the existing baseline and epsilon comparison.

## Test plan

- [x] RED: the two contract tests for the split failed against the original combined job; the report-step test fails once `continue-on-error` is removed from that step
- [x] full CI workflow contract suite: 6 passed
- [x] scripts/ci.sh lint
- [x] workflow YAML syntax parse
- [x] Prettier workflow format check
- [x] Ruff check and format check for the Python contract test

## Coordination

Open PRs #2283, #2284, and #2291 also touch one or both files. Their changes are in disjoint workflow/test sections; the new test class is placed away from #2284s insertion point to reduce conflict risk.

Closes #2278


ohdearquant:
The problem this targets is real — a coverage job that goes red without having measured anything reports "coverage regressed" when it means "coverage is unknown", and separating measurement from judgment is the right decomposition. But as implemented the split converts a fail-closed gate into a fail-open one.

## When coverage cannot be measured, CI now goes green with no coverage judgment

Trace the failure path:

1. `compute_coverage` carries `continue-on-error: true`, so when `cargo llvm-cov` fails the **step** records `failure` but the **job** `coverage-measurement` concludes `success`.
2. `available` is `${{ steps.compute_coverage.outcome == 'success' }}` → `'false'`.
3. `coverage-ratchet` is guarded by `if: needs.coverage-measurement.outputs.available == 'true'` → the job is **skipped**.
4. `ci-gate` aggregates with:

```bash
bad=$(echo "$NEEDS" | jq -r 'to_entries[] | select(.value.result != "success" and .value.result != "skipped") | .key')
```

`coverage-measurement` is `success` and `coverage-ratchet` is `skipped`. Neither is `bad`. **The gate passes.**

Compare `origin/main`: there, `Compute coverage` has no `continue-on-error`, so a failing measurement fails the job, `ci-gate` sees `failure`, and the gate blocks. Today a broken measurement stops the merge. After this change it does not.

That is a strictly worse position than the defect being fixed. The current behaviour is confusing — a red check whose label implies a regression that was never evaluated — but it is loud, and someone investigates. The proposed behaviour is quiet: an unmeasurable-coverage PR merges under a fully green wall, and the only trace is a warning annotation and a step-summary block that nobody is required to read. Adding `coverage-measurement` to the gate's `needs` does not help, because that job is engineered to succeed regardless.

Note the general shape, since it recurs: `continue-on-error: true` plus a downstream `if:` guard plus a gate that treats `skipped` as passing is a three-part construction where each part is individually reasonable and the combination silently removes enforcement.

## What would keep both properties

The goal — distinguish "regressed" from "unmeasurable" — does not require dropping enforcement. Options, in rough order of preference:

- Have the ratchet job run unconditionally and fail with a distinct message when `available != 'true'`. The check then stays red for an unmeasurable build, but says the honest thing, which was the actual complaint.
- Keep the skip, and add a third small job that runs when `available != 'true'` and fails. Same result, keeps the ratchet job's own logic clean.
- Keep the skip and make it visible to the gate by having `ci-gate` special-case `coverage-ratchet`: treat `skipped` as failure for that one need. Least attractive, since it puts the policy far from the job it governs.

If the intent is genuinely that a broken coverage tool should not block merges, that is a defensible policy — but it is a policy change, and it should be stated in the job comment and in the PR description rather than arriving as a side effect of `continue-on-error`.

## The added tests cannot see this

`scripts/tests/test_ci_workflows.py` asserts the presence of literal strings in the YAML — `assertIn("continue-on-error: true", measurement)`, `assertIn("if: needs.coverage-measurement.outputs.available == 'true'", ratchet)`, `assertIn("- coverage-measurement", gate)`. Every one of those passes on the current diff, and every one would still pass if the gate let an unmeasured build through, which it does.

These tests confirm that the author built the structure the author intended. They assert nothing about what the workflow *decides*. The property worth testing is the outcome one: given a failed measurement, does the aggregate gate block? That is expressible against the same YAML — evaluate the gate's `jq` selector over a synthetic `needs` object with `coverage-measurement: success, coverage-ratchet: skipped` and assert the result is a failure. As written, the suite is a spelling checker for the diff that accompanies it.

Also worth fixing while here: `test_ratchet_requires_available_measurement_and_gate_tracks_both_jobs` asserts `assertNotIn("cargo llvm-cov", ratchet)`. That is a reasonable structural check, but paired with an outcome test it would be redundant; on its own it is the kind of assertion that keeps passing after the behaviour it was protecting has gone.

ohdearquant:
Adding the acceptance condition, so the shape of a mergeable fix is unambiguous rather than left as "address the comment".

**The wall must stay fail-closed on unmeasurable coverage.** Separating "regressed" from "unknown" is a labeling improvement worth having — that part of the motivation is right. But `unknown` has to block. Trading a loud-but-mislabeled red for a silent green is strictly worse than the current state, because the current state at least stops the merge and gets investigated.

Concretely, the ratchet should be able to report **three** states — `PASS`, `FAIL`, `UNMEASURABLE` — with **only `PASS` admitting a merge**. `UNMEASURABLE` may carry a distinct message, a distinct annotation, and a distinct name in the checks list; it may not carry a green wall. If there is a reason to let an unmeasurable build merge, that is a policy decision and belongs in the job comment and the description as one, not as an emergent consequence of `continue-on-error` plus a skip-tolerant aggregator.

If a quarantined state is preferred over an outright block, it needs a forced consumer — something that already runs and that stops the merge when the state is set. A warning annotation and a step-summary block are not forced consumers; nothing fails if nobody reads them.

**On the tests.** The question to write them against is: *what changes for the caller?* Here the caller is the merge wall, and the answer under this diff is "an unmeasurable build now merges" — which no assertion in the added suite can observe, because every assertion checks for a literal string in the YAML. A test that evaluates the gate's own selector over a synthetic `needs` object (`coverage-measurement: success`, `coverage-ratchet: skipped`) and requires a failure would have caught this, and would keep catching it after someone edits the YAML into a different but equally permissive shape.

This is worth stating as a general point because it keeps recurring in this repository: a check that asserts the presence of the code you just wrote confirms authorship, not behaviour. The assertion has to be able to distinguish the intended outcome from the outcome you are trying to prevent.

ohdearquant:
Requesting changes. Splitting measurement from judgment is the right idea, but as
written this converts a fail-closed gate into a fail-open one.

The chain:

1. `cargo llvm-cov --workspace` fails. This is not hypothetical — the job's own
   comments call it the disk-heaviest command in CI and discuss the writer-checkout
   pressure it creates.
2. `continue-on-error: true` means the step fails but the `coverage-measurement`
   **job** succeeds, with `available` set to `false`.
3. `coverage-ratchet` carries `if: needs.coverage-measurement.outputs.available == 'true'`,
   so it is **skipped**.
4. `ci-gate` selects jobs with
   `.result != "success" and .result != "skipped"`. A skipped job is green there, so
   the gate prints "All gated jobs green (or legitimately skipped)" and passes.

Net effect: whenever coverage measurement fails, the coverage-regression gate silently
disappears and a PR that regresses coverage below the baseline merges with no judgment
having been made. On `main` today the same llvm-cov failure fails `coverage-ratchet`,
`ci-gate` sees a failure, and the PR is blocked — I checked, there is no
`continue-on-error` on that job today, so the change in behaviour is real and is in the
unsafe direction.

The `::warning` and step-summary notice do not close this. Nothing consumes them; no
check goes red and nothing blocks, so the signal is only visible to someone who already
went looking.

Suggested shape, keeping the split you want: don't make the ratchet's *existence*
conditional on the measurement. Let `coverage-ratchet` always run
(`if: always()` alongside `needs: coverage-measurement`) and let it decide:

- `available == 'true'` → compare against the baseline exactly as now;
- otherwise → fail with the "measurement unavailable" message.

That still gets the benefit the PR is after — a flaky llvm-cov produces an honest
"could not measure" failure instead of a misleading "coverage regressed" one — without
removing the wall. If the intent really is that infrastructure flakiness should not
block a merge, that is a policy call that should be explicit and owned rather than
implemented as a skip that the aggregate gate reads as green.

One knock-on: the two new tests assert the current structure, including
`if: needs.coverage-measurement.outputs.available == 'true'`, so they pin the fail-open
shape in place and will need updating. Worth adding an assertion that the ratchet cannot
be skipped, since that is the property actually being protected.

Not a blocker, just noting it verified fine: `coverage-baseline.txt` is at the repo root,
so the new job reading it without a `working-directory` is correct.

ohdearquant:
## An unmeasurable coverage run now passes the merge wall

The split itself is right — measurement and ratchet are genuinely different jobs and separating them
is an improvement. The problem is what the unavailable state does at the aggregate.

`Compute coverage` gains `continue-on-error: true`:

```
.github/workflows/ci.yml:703-705
  - name: Compute coverage
    id: compute_coverage
    continue-on-error: true
```

That is new. `origin/main` has no `continue-on-error` anywhere in this workflow — the whole-file
count is 0 on the base and 1 at this head — so on the base a failed `cargo llvm-cov` failed the job.

Downstream, the aggregate accepts a skipped result:

```
.github/workflows/ci.yml:~800
  bad=$(echo "$NEEDS" | jq -r 'to_entries[]
        | select(.value.result != "success" and .value.result != "skipped") | .key')
```

So when the measurement step fails, the sequence is: the step is tolerated, the reporting step
still succeeds, the job exposes `available == 'false'`, the ratchet is skipped by its own
condition, and the aggregate sees `{measurement: success, ratchet: skipped}` — no `bad` entry, gate
green. A run that could not measure coverage at all is indistinguishable at the merge wall from one
that measured and passed.

The asymmetry is worth stating plainly, because it is what makes this blocking rather than a nit.
Failing a PR because coverage could not be measured is loud, bounded, and fixed by a retry. Letting
an unmeasured PR through is silent and unbounded, and it is silent precisely on the runs where
something was already wrong with the toolchain. The warning annotation and the step-summary entry
are real, but no required check consumes either, so nothing in the merge path reads them.

The new tests assert the literal structures — that the steps exist and are shaped as written — but
none of them evaluates the failure-state outcome, so they pass while this holds.

**Suggested direction, keeping the split:** make the unavailable state fail closed. Either run the
decision job unconditionally and have it exit non-zero with an explicit `UNMEASURABLE` result, or
add a required consumer that fails when `available != 'true'`. Worth adding a contract test that
feeds the synthetic needs-state `{coverage-measurement: success, coverage-ratchet: skipped}` and
asserts the aggregate decision fails — that test is what would have caught this, and it is the one
shape the current tests do not cover.

For transparency on how this was produced: an automated pass over this diff was run on a model from
the same family as the one that authored it, so its output was treated as input evidence only. The
three load-bearing facts above — the `continue-on-error` addition, the base's whole-file count of
zero, and the aggregate's `!= "skipped"` predicate — were each re-derived directly from the two refs
before this was posted.

oceanwaves630 (approved):
Approved at 785f7a5717b6.

PR #2304, opened 2026-08-30
fix(knowledge): support explicit-null atom patches

### LINKED ISSUE #1999: knowledge: atom upsert cannot express an explicit clear of source_uri/source_type/finalized
## Context

The atom upsert path preserves `source_uri`, `source_type`, and `finalized` when the caller omits them (`COALESCE(?N, column)` in `crates/khive-pack-knowledge/src/knowledge/crud.rs`), which fixed re-ingest runs clobbering attribution and review status with NULL.

## Limitation

The wire types are `Option<T>`, so an explicitly supplied JSON `null` and an omitted field both deserialize to `None` and bind as SQL `NULL` — `COALESCE(NULL, column)` then preserves the old value. There is now no way to deliberately clear these fields through the upsert surface.

This is not a regression: before the preservation fix, both the omitted and the explicit-null form clobbered the column, so "clear" only ever worked as an accident of the data-loss bug. But the operation is genuinely inexpressible today.

## Proposed direction

Distinguish field-absent from field-null at the wire (double-option pattern or an explicit sentinel), or add a curation-layer verb for clearing attribution/review state, keeping the upsert surface preserve-on-omit. Whichever shape lands should add a test pair where omitted and explicit-null produce different outcomes — the current conflation is exactly what a round-trip test would catch.

ohdearquant:
## Summary

- distinguish omitted, explicit-null, and concrete values for atom source/finalization patches
- clear `source_uri` / `source_type` on null and define `finalized: null` as resetting the flag to false without demoting lifecycle status
- document the accepted tri-state contract in ADR-047 and live verb help

## Tests

- `cargo test -p khive-pack-knowledge`
- `cargo check --workspace`
- `git diff --check`

Closes #1999

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/crud.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/crud.rs
@@ -114,10 +114,10 @@ impl KnowledgeHandlers {
             khive_runtime::secret_gate::reject_reserved_secret_gate_property(
                 atom_in.properties.as_ref(),
             )?;
-            if let Some(ref uri) = atom_in.source_uri {
+            if let Some(Some(uri)) = &atom_in.source_uri {
                 khive_runtime::secret_gate::check(uri)?;
             }
-            if let Some(ref st) = atom_in.source_type {
+            if let Some(Some(st)) = &atom_in.source_type {
                 khive_runtime::secret_gate::check(st)?;
             }
         }
@@ -199,13 +199,23 @@ impl KnowledgeHandlers {
             let source_uri = atom_in
                 .source_uri
                 .as_ref()
+                .and_then(Option::as_ref)
                 .map(|s| s.trim())
                 .filter(|s| !s.is_empty());
+            // Blank strings keep their established update behavior (preserve),
+            // while JSON null is now an explicit clear.
+            let source_uri_present =
+                matches!(&atom_in.source_uri, Some(None)) || source_uri.is_some();
             let source_type = atom_in
                 .source_type
                 .as_ref()
+                .and_then(Option::as_ref)
                 .map(|s| s.trim())
                 .filter(|s| !s.is_empty());
+            let source_type_present =
+                matches!(&atom_in.source_type, Some(None)) || source_type.is_some();
+            let finalized_present = atom_in.finalized.is_some();
+            let finalized = atom_in.finalized.flatten().unwrap_or(false);
 
             if insert {
                 statements.push(SqlStatement {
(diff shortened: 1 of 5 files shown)


ohdearquant:
This one is correct. I went looking for the defect that usually hides in a "tri-state" patch — omitted and explicit-null collapsing to the same Rust value, so the third state exists only in the documentation — and it is genuinely not here.

## Verified

**The three states are distinct at the wire.** `#[serde(default, deserialize_with = "tri_state")]` over `Option<Option<T>>`, with:

```rust
fn tri_state<'de, T, D>(deserializer: D) -> Result<Option<Option<T>>, D::Error> {
    Ok(Some(Option::deserialize(deserializer)?))
}
```

Absent key takes the field default (`None`); a present key invokes the deserializer, so `null` becomes `Some(None)` and a value becomes `Some(Some(v))`. That is the canonical shape and it applies to all three patch fields.

**It survives to storage.** `AtomInput` derives only `Deserialize` and is never re-serialized, so there is no round trip that could rewrite an explicit null back into an omission. `crud.rs:199-218` splits each field into a presence flag plus payload, and the `CASE WHEN presence` branches at `:249-264` bind SQL NULL for the clear, so the distinction reaches the storage boundary intact rather than being flattened by a `COALESCE`.

**The reachable state matrix gains nothing new.** `finalized: null` produces the same false payload as an explicit `finalized: false`, and `reviewed` + `finalized=false` was already reachable through the existing explicit-false path — so no combination becomes reachable *only* via the new route. Consumers were enumerated rather than sampled: `atom_from_row`/`atom_to_json`, `knowledge.stats` (`eval_coverage` counts `finalized=1`), `knowledge.search` (eligibility and scoring key on `status`, not `finalized`), `knowledge.suggest`, auto-`compose`, and `knowledge.index` (`atom_embed_text` reads neither field). The only behavioural consequence is the intended mutation and its effect on `eval_coverage`, which is correct — resetting a finalization flag should lower a finalization-coverage metric.

## One migration hazard worth documenting

Since `null` is now the explicit clear, a client that models these fields as an ordinary `Option<T>` and serializes `None` as JSON null will clear `source_uri`/`source_type` on a patch that never intended to mention them. The server cannot defend against this, because that payload is indistinguishable from a deliberate clear — which is the correct design, not a defect.

The in-repository caller (`sections.rs:939-951`, sending concrete strings and `finalized: true`) is unaffected, and no checked-in client serializes an optional atom patch. But this is a public verb, so out-of-tree callers are the exposed population. Worth a sentence in the ADR and the verb help saying explicitly that clients must omit fields they do not intend to change — `skip_serializing_if = "Option::is_none"` for Rust clients, and the equivalent elsewhere. The tri-state contract is only safe if callers know that "I left it unset" and "I sent null" are now different requests.

Blank strings retaining the old preserve-on-update behaviour (`:199-216`) rather than becoming a second clear spelling is the right call — one clear operation, not two.

PR #2305, opened 2026-08-30
test(memory): stabilize keyword widening fixture

### LINKED ISSUE #2235: memory: keyword_only widening test is non-deterministic — same commit fails then passes
`keyword_only_widening_ignores_vector_leg_candidates` is non-deterministic: the same commit failed once and passed on re-run with no code change.

## Observation

Test: `khive-pack-memory`, `handlers::recall::tests::keyword_only_widening_ignores_vector_leg_candidates`
(`crates/khive-pack-memory/src/handlers/recall.rs`, assertion at the end of the test body).

Both runs are the identical tree:

| run | job | result |
|-----|-----|--------|
| first | CI (ubuntu-latest, shard 1/2) | FAILED |
| re-run, same commit, no push between | CI (ubuntu-latest, shard 1/2) | success |

The other three shards (ubuntu 2/2, macos 1/2, macos 2/2) passed in both.

Failure output was an empty left side against a one-element right side:

```
assertion `left == right` failed: keyword_only widening must not stop on vector-leg-only candidates
  left: []
 right: ["<target uuid>"]
```

So the recall returned no rows at all, rather than returning the wrong rows.

## What has been ruled out

**Feature unification is not the cause.** `test-internals` is an empty feature
(`test-internals = []`) whose only two `cfg` sites are an additive `mod test_internals`
and a `pub use` re-export. Neither can change recall behaviour, so a workspace build
unifying that feature cannot explain the difference.

**Boundary exclusion on the time window is not the cause.** The test pins the target's
`created_at` to exactly the same value it later passes as the window's lower bound, which
made an exclusive comparison the obvious suspect. It is inclusive: the note store builds
`created_at >= ?` (`crates/khive-db/src/stores/note.rs:834`), so a target sitting exactly
on the boundary is inside the window, not outside it.

## What is not established

The mechanism. The above are two hypotheses checked and refuted, not a diagnosis. Recording
them so the next person does not spend the same two checks.

Worth noting about the shape of the test, as a starting point rather than a conclusion: it
depends on a 160-filler ranking arrangement chosen so the target falls outside a
150-candidate first fetch and inside a widened 200. That makes the assertion sensitive to
any tie-breaking or ordering that is not fully determined, and an empty result is consistent
with the target landing outside the widened window rather than with widening being skipped.

## Why it is worth fixing rather than re-running

A test that fails once per N runs on an unchanged tree spends a CI cycle each time it
trips, and it trains readers to re-run rather than read. It also sits on the recall
widening path, where a genuine intermittent defect and a flaky assertion would look
exactly alike from the CI summary.

ohdearquant:
## Summary

- replace the fixture’s mutable 150/200 production-default boundary with an explicit request-local 4→8 candidate window
- remove zero-similarity vector tie ordering from the target by assigning it a deterministic score below the decoys but above the raw relevance floor
- retain the vector-only, unwindowed crowd-out, and exact windowed widening assertions

## Tests

- focused regression test: 50/50 repeated passes
- `cargo test --manifest-path crates/Cargo.toml -p khive-pack-memory --lib` (324 passed, 1 ignored)
- `cargo check --manifest-path crates/Cargo.toml -p khive-pack-memory --all-targets --all-features`
- `cargo clippy --manifest-path crates/Cargo.toml -p khive-pack-memory --all-targets --all-features -- -D warnings`
- `git diff --check`

Closes #2235

### DIFF
--- a/crates/khive-pack-memory/src/handlers/recall.rs
+++ b/crates/khive-pack-memory/src/handlers/recall.rs
@@ -6776,9 +6776,27 @@ mod tests {
     }
 
     const STRATEGY_PROBE_QUERY: &str = "widening strategy probe query";
+    const STRATEGY_PROBE_TARGET: &str = "widening strategy probe query sentinel";
+    const STRATEGY_PROBE_INITIAL_CANDIDATE_LIMIT: u32 = 4;
+    const STRATEGY_PROBE_MAX_RECALL_CANDIDATES: usize = 8;
+    const STRATEGY_PROBE_KEYWORD_FILLERS: usize = 5;
+    const STRATEGY_PROBE_VECTOR_DECOYS: usize = 2;
 
-    /// Query and vector decoys share one direction; everything else is
-    /// orthogonal, so decoys are top vector hits while never matching FTS.
+    fn strategy_probe_recall_config() -> Value {
+        json!({
+            "candidate_limit": STRATEGY_PROBE_INITIAL_CANDIDATE_LIMIT,
+            "scoring": {
+                "max_recall_candidates": STRATEGY_PROBE_MAX_RECALL_CANDIDATES,
+            },
+            // One re-gather round is the behavior this fixture exercises.
+            "ann_overfetch_max_rounds": 2,
+        })
+    }
+
+    /// Query and vector decoys share one direction. The keyword target has a
+    /// distinct, lower vector score that remains above `min_raw_relevance`,
+    /// while fillers are orthogonal. This keeps the decoys first without
+    /// letting zero-score vector tie ordering intermittently filter the target.
     struct StrategyProbeVecService;
 
     #[async_trait]
@@ -6793,6 +6811,10 @@ mod tests {
                 .map(|t| {
                     if t == STRATEGY_PROBE_QUERY || t.starts_with("vector decoy") {
                         vec![1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+                    } else if t == STRATEGY_PROBE_TARGET {
+                        // cosine(query, target) ~= 0.316: below the decoys,
+                        // above the default raw-vector floor of 0.10.
+                        vec![1.0, 3.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
                     } else {
                         vec![0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
                     }
@@ -6842,11 +6864,13 @@ mod tests {
         let ns = Namespace::parse("local").expect("local namespace");
         let token = rt.authorize(ns).expect("authorize local");
 
-        // 160 out-of-window keyword fillers that out-rank the target on FTS,
-        // so the target sits outside the 150-candidate first fetch and inside
-        // the widened 200 (default candidate_limit 150, max 200).
+        // One more out-of-window keyword filler than the request-local initial
+        // cap. All fillers out-rank the target on FTS, so the target sits
+        // outside the 4-candidate first fetch and inside the explicitly capped
+        // widened fetch of 8. Keeping both bounds request-local avoids coupling
+        // this fixture to production defaults or a 150-row rank boundary.
         let mut filler_ids = Vec::new();
-        for i in 0..160 {
+        for i in 0..STRATEGY_PROBE_KEYWORD_FILLERS {
             let filler = rt
                 .create_note(
                     &token,
@@ -6870,7 +6894,7 @@ mod tests {
                 &token,
                 "memory",
                 None,
-                "widening strategy probe query sentinel",
+                STRATEGY_PROBE_TARGET,
                 Some(0.7),
                 None,
                 vec![],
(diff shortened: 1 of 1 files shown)


ohdearquant:
This holds up. I checked the things that would make a stabilization like this pass for the wrong reason, and they come back clean.

## Verified

- **The injected config is genuinely honored, and a typo could not silently restore defaults.** `RecallParams` is `#[serde(deny_unknown_fields)]` (`crates/khive-pack-memory/src/handlers/common.rs:307-308`), and all three keys exist: `candidate_limit` and `ann_overfetch_max_rounds` at `crates/khive-pack-memory/src/config.rs:68,96`, `max_recall_candidates` at `crates/khive-pack-memory/src/scoring.rs:487`. This was the failure mode worth ruling out — an ignored config block would leave the fixture running at the 150/200 defaults where 5 fillers put the target *inside* the first fetch, so the test would exercise no widening at all and still pass.
- **The arithmetic is consistent.** `STRATEGY_PROBE_KEYWORD_FILLERS = 5` against an initial cap of 4 is the "one more than the initial cap" the comment claims, so the target lands at FTS rank 6 — outside the first fetch of 4, inside the widened 8.
- **The vector score claim is right.** cosine([1,0,…], [1,3,0,…]) = 1/√10 = 0.3162, matching the stated ~0.316, and above the 0.10 raw floor while below the decoys' 1.0. Giving the target a distinct non-zero score is the correct fix for tie-order flakiness — a zero-score tie is resolved by whatever order the store returns, which is exactly the kind of thing that flips between runs.
- **The new count assertion is not vacuous.** `assert_eq!(vector_probe_ids.len(), STRATEGY_PROBE_VECTOR_DECOYS)` pins the population rather than only checking membership, so a probe returning extra rows now reddens instead of passing the `contains` loop.

## One trade worth stating explicitly

The fixture previously ran against production defaults (150 first fetch, 200 widened) and now runs against request-local values (4 and 8). That is the right call for stability — 160 rows to reach a rank boundary is a slow and brittle way to test a mechanism — but it does mean this test no longer exercises the default `candidate_limit` or the default widening ceiling. A regression that changed those defaults, or that broke the path where they are *derived* rather than supplied, would not be caught here.

That is probably fine, since the test is named for the widening strategy rather than for the defaults. Worth confirming that some other test covers the defaults, and if none does, that gap is now unowned rather than incidentally covered. The comment already says the decoupling is deliberate, which is the right instinct; naming what it gives up would make it complete.

## Minor

Replacing the literal decoy names with `format!("vector decoy {i}")` keeps the `t.starts_with("vector decoy")` branch in the fake embedding service working, and the loop bound is now the same constant the assertion uses — so the count and the fixture cannot drift apart. Good.

ohdearquant:
This is the right fix for the flake class, and the arithmetic holds.

The vector geometry checks out. Query is `[1,0,0,…]`, the new target is `[1,3,0,…]`, so
cosine = 1/√10 = 0.316 — the comment's number is exact. That sits below the decoys (identical to
the query, so 1.0) and above the raw-vector floor, and the floor really is 0.10: `ScoringConfig`'s
default `min_raw_relevance` is `0.10` at `crates/khive-pack-memory/src/scoring.rs:524`. Giving the
target a distinct non-zero score is the actual de-flake — a zero-score target was ordered against
the other zero-score rows by whatever the tie broke to that run.

The filler arithmetic is consistent too: 5 fillers against a request-local initial cap of 4 puts the
target at FTS rank 6, outside the first fetch and inside the capped widened fetch of 8. All three
config keys are real fields on the live path (`candidate_limit`, `scoring.max_recall_candidates`
consumed at `recall.rs:212,374` and `sub_handlers.rs:103,219`, `ann_overfetch_max_rounds`), and 8 is
far under the `MAX_RECALL_CANDIDATES` DoS clamp, so nothing is silently reshaped.

**The part worth calling out, because it is what makes the rest safe.** `ScoringConfig` is
`#[serde(default)]` and is *not* `deny_unknown_fields` — unlike `RecallParams`. So a misspelled key
inside the nested `scoring` object would be silently dropped and the fixture would quietly fall back
to the production defaults it is trying to stop depending on. That failure would normally be
invisible: with only 5 fillers and a default `candidate_limit` of 150, nothing is crowded out, no
widening happens, and the reproduction arm passes without reproducing anything.

What stops that is the precondition assertion already in the test — "the target must be crowded out
without a window". If the config silently failed to apply, the target *would* come back unwindowed
and that assert fires. So the fixture cannot pass vacuously. Please keep that assertion; it is not a
redundant sanity check, it is the only thing standing between this test and a silent no-op.

One observation, not blocking. The sibling fixture
`recall_widens_when_out_of_window_candidates_crowd_out_eligible_ones` (`recall.rs:6657`) still
carries exactly the coupling this PR removes from its neighbour — 160 fillers at `:6671` and a
comment pinning `candidate_limit` 150 / `max_recall_candidates` 200. It has the same precondition
guard, so it fails loudly rather than silently, but it will go red the day a production default
moves, for a reason that has nothing to do with what it tests. The same request-local treatment
would apply cleanly, and doing it while this pattern is fresh is cheaper than rediscovering it.

PR #2306, opened 2026-08-30
test: clean up temporary database fixtures

### LINKED ISSUE #2263: Test tmp_db() leaks SQLite -wal/-shm and the blob GC lock into $TMPDIR
`tmp_db()` in the `khive-mcp` test modules creates a database at a
`NamedTempFile` path. SQLite then creates `-wal` and `-shm` beside it, and the
blob GC creates `<db>.khive-blob-gc.lock`. `NamedTempFile`'s drop removes only
the file it created, so three siblings survive every run that touches those
paths.

## Measured

On one developer machine, `$TMPDIR` held 19.9 GiB across 39,685 entries:

| shape | count |
| --- | --- |
| `.tmpXXXXXX-wal` | 10,952 |
| `.tmpXXXXXX-shm` | 10,952 |
| `.tmpXXXXXX.khive-blob-gc.lock` | 10,928 |
| bare `.tmpXXXXXX` | 6,570 |

The three leaked families are matched sets, one per run. `file(1)` on a sample
`-wal` reports `SQLite Write-Ahead Log, version 3007000`.

Age: none older than 7 days, and 25,590 created within the last day, so this
accumulates at roughly 20 GiB/week on a machine that runs the suite regularly
rather than being a historical artifact.

## Where

- `crates/khive-mcp/src/components.rs` — `tmp_db()` in `#[cfg(test)] mod tests`
- `crates/khive-mcp/src/pending_events.rs` — same helper

The lock file itself is not the bug. `database_gc_lock_path` in
`crates/khive-db/src/stores/blob.rs` opens `<database_path>.khive-blob-gc.lock`
and holds it, which is correct; a held lock is not unlinked. The bug is that the
database path is a tempfile whose cleanup guard does not know the sidecars
exist.

Worth noting that `components.rs` already has a `freeze_snapshot_sidecars`
helper that enumerates `["-wal", "-shm"]` for the same paths, so the sidecars
are known in that module; only the cleanup path is unaware of them.

The 6,570 bare `.tmpXXXXXX` are a second case with a different cause. They are
**directories**, leaked by a deliberate `std::mem::forget` of the tempdir guard,
not surviving base files from a guard that failed to run. Corrected, with
measurements and the 12 leak sites, in
https://github.com/ohdearquant/khive/issues/2263#issuecomment-5447057129

## Suggested fix

Have `tmp_db()` return a `TempDir` and place the database inside it. Removing
the directory takes the database, both sidecars, and the lock together, with no
list of suffixes to keep in sync as new sidecar kinds appear.

Test-only, so no effect on a served deployment. The cost is developer and CI
disk.

ohdearquant:
## Summary

- place both khive-mcp scratch databases inside owned TempDir directories so SQLite WAL/SHM and blob-GC lock sidecars are removed together
- replace all 12 deliberately forgotten test TempDir guards with owned runtime, pool, or test-scope guards
- add focused coverage for recursive database-and-sidecar cleanup

## Verification

- cargo test --manifest-path crates/Cargo.toml -p khive-mcp -p khive-runtime -p khive-pack-memory -p kkernel -p khive-db --no-run
- focused tests: MCP cleanup (1), stalled writer (1), atomic runner (10), atomic prepare (40), atomic apply (18), memory feedback (10), recall lifetime cases (4), ANN lifetime cases (2)
- cargo check --manifest-path crates/Cargo.toml --workspace --all-targets --all-features
- cargo clippy --manifest-path crates/Cargo.toml --workspace --all-targets --all-features -- -D warnings
- git diff --check

Closes #2263

### DIFF
--- a/crates/khive-db/tests/writer_timeout_sink_stalled_writer.rs
+++ b/crates/khive-db/tests/writer_timeout_sink_stalled_writer.rs
@@ -23,22 +23,19 @@ use khive_db::{ConnectionPool, PoolConfig};
 
 const WRITE_DELAY_MS: u64 = 5_000;
 
-fn point_sink_at_slow_writer() {
+fn point_sink_at_slow_writer() -> tempfile::TempDir {
     let dir = tempfile::tempdir().expect("tempdir");
     std::env::set_var("KHIVE_WRITER_TIMEOUT_SINK_DIR", dir.path());
     std::env::set_var(
         "KHIVE_WRITER_TIMEOUT_SINK_WRITE_DELAY_MS",
         WRITE_DELAY_MS.to_string(),
     );
-    // Leak the tempdir so it stays alive for the rest of the process — the
-    // writer thread keeps writing (slowly) into it for as long as the
-    // process runs.
-    std::mem::forget(dir);
+    dir
 }
 
 #[test]
 fn sink_never_adds_measurable_latency_when_its_writer_is_genuinely_slow() {
-    point_sink_at_slow_writer();
+    let _sink_dir = point_sink_at_slow_writer();
 
     let dir = tempfile::tempdir().unwrap();
     let db_path = dir.path().join("stalled_writer_sink_test.db");
--- a/crates/khive-mcp/src/components.rs
+++ b/crates/khive-mcp/src/components.rs
@@ -624,12 +624,13 @@ mod tests {
     use super::*;
     use khive_runtime::{KhiveRuntime, Namespace, RuntimeConfig};
     use std::sync::atomic::{AtomicU32, Ordering};
-    use tempfile::NamedTempFile;
+    use tempfile::TempDir;
 
-    fn tmp_db() -> (NamedTempFile, String) {
-        let f = NamedTempFile::new().expect("tempfile");
-        let path = f.path().to_str().expect("utf8 path").to_string();
-        (f, path)
+    fn tmp_db() -> (TempDir, String) {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let path = dir.path().join("khive-test.db");
+        let path = path.to_str().expect("utf8 path").to_string();
+        (dir, path)
     }
 
     #[cfg(unix)]
@@ -708,6 +709,22 @@ mod tests {
             restart_delay_ms(u64::MAX, u64::MAX),
             u64::MAX,
             "saturating jitter arithmetic must remain overflow-safe"
+        );
+    }
+
+    #[test]
+    fn tmp_db_guard_removes_database_and_all_sidecars() {
+        let (dir, db_path) = tmp_db();
+        let dir_path = dir.path().to_path_buf();
+        for suffix in ["", "-wal", "-shm", ".khive-blob-gc.lock"] {
+            std::fs::write(format!("{db_path}{suffix}"), b"fixture").expect("write fixture");
+        }
+
+        drop(dir);
+
+        assert!(
+            !dir_path.exists(),
+            "dropping the directory guard must remove the database and every sibling sidecar"
         );
     }
 
--- a/crates/khive-mcp/src/pending_events.rs
+++ b/crates/khive-mcp/src/pending_events.rs
@@ -3328,7 +3328,7 @@ mod tests {
     use khive_storage::event::EventFilter;
     use khive_storage::types::PageRequest;
     use khive_types::{Details, HandlerDef, KhiveError, VerbCategory, Visibility};
-    use tempfile::NamedTempFile;
+    use tempfile::TempDir;
     use tokio_util::sync::CancellationToken;
 
     #[derive(Debug)]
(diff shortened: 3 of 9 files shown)


ohdearquant:
Most of this is a real improvement, and the sidecar reasoning behind it checks out: the fixtures
build the database as `dir.path().join(...)` (`crates/khive-mcp/src/components.rs:629-633`, and the
matching helper in `pending_events.rs:3553-3557`), WAL and SHM siblings are derived beside that path,
and the blob-GC lock is derived directly from the DB path at
`crates/khive-db/src/stores/blob.rs:1050-1069`. I looked for a sidecar whose path is computed
independently of the database and found only one — which is the problem below.

**The one independently-derived path is exactly the guard that must not be scoped to a test.**
`KHIVE_WRITER_TIMEOUT_SINK_DIR` is chosen separately at
`crates/khive-db/src/timeout_sink.rs:227`, and the sink is process-wide with a writer thread retained
by the pool (`timeout_sink.rs:653-659`, `741-790`, wired from `pool.rs:867-873`). The removed code
said so itself:

```rust
// Leak the tempdir so it stays alive for the rest of the process — the
// writer thread keeps writing (slowly) into it for as long as the
// process runs.
std::mem::forget(dir);
```

That reason has not changed. After this diff the directory is bound to `_sink_dir` in the test body
(`crates/khive-db/tests/writer_timeout_sink_stalled_writer.rs:26-38`), so it is removed when the test
returns while the sink thread keeps running with a 5s write delay still configured. A later delayed
open, heartbeat, retry, rename or append then hits a deleted path, and any other test in that binary
can observe a sink I/O error or a missing sink file. It fails under load rather than deterministically,
which is the worst shape: it reddens changes that have nothing to do with it. The DB tempdir at
`:40-53` is a different directory and does not keep the sink alive. Either retain the sink guard for
process lifetime as before, or add an explicit sink shutdown/join and remove the directory only after
it returns.

**Two more sites hand a directory to work that outlives the guard.** In both, the field order in the
wrapper is locally correct — `runtime` declared before `_temp_dir`, so the runtime drops first — but
field order only governs the wrapper's own copy, not a clone captured by a detached task.

`memory.recall` spawns a tracked task with cloned runtime and registry state
(`crates/khive-pack-memory/src/handlers/recall.rs:1059-1082`) that writes the ledger and then appends
a second event at `:1086-1126`. The four converted sites (`:2030-2048`, `:2185-2191`, `:2640-2648`,
`:2784-2791`) all return after dispatch; the actor test polls only the ledger row (`:2743-2775`) and
the projection test returns straight after its assertions (`:2859-2891`). Nothing waits for the event
append. `#[serial(background_tasks)]` serializes test *entry*, it does not join outstanding tasks. So
the directory can be removed while the cloned runtime is still writing, producing an intermittent
`SQLITE_CANTOPEN` or sidecar-cleanup failure *after* the visible assertion has already passed.

The ANN helper (`crates/khive-pack-memory/src/ann.rs:3923-3943`) has the same shape for its background
callers. `spawn_rebuild_task_inner` captures a runtime clone in a detached tracked task (`:724`), does
DB work through `:751`, releases the warming guard, and can enqueue a debounced follow-up at
`:781-789`. Tests such as
`ensure_ann_background_releases_warming_guard_after_success_and_allows_later_rebuild` (`:3946-3950`)
wait on warming and generation conditions, which is not the same as proving the chain has stopped. The
helper is sound for its synchronous callers; it is the background ones that need a drain or a
transfer of ownership into the task.

The remaining conversions look right, and it is worth saying which, because the per-site distinction
is the whole point here rather than a pattern applied uniformly: `ann.rs:4394-4408` (locals drop
`rt2`, `rt1`, then `tmp`), both `feedback.rs` helpers, `atomic_prepare.rs:1769-1837`, and
`atomic_apply.rs:1520-1548` have correct order and no detached reader. `atomic_runner.rs:404-438`
joins its explicit tasks at `:1060-1077`; the residual question there is only the pool's own writer
task, whose join handle is detached on pool drop (`crates/khive-db/src/pool.rs:507-519`), so a
close-time operation could in principle race removal even though no queued test operation remains.

Production impact is nil and I confirmed it rather than assuming: the three `src/` files touched
(`atomic_prepare.rs:1756`, `atomic_runner.rs:394`, `atomic_apply.rs:1514`) have their changes inside
`#[cfg(test)]`, so no non-test build reaches the new owners. Before this diff a caller in situation X
got the same production behaviour; after it, the same caller gets the same production behaviour.

On evidence for the cleanup claim itself: the only new assertion is
`crates/khive-mcp/src/components.rs:715-729`, which writes the main DB, `-wal`, `-shm` and blob-lock
names by hand, drops the guard, and asserts the directory is gone. That would catch a regression in
`TempDir` cleanup, but it does not exercise SQLite actually creating or checkpointing a WAL, real
blob-GC lock acquisition, or any background teardown — so the sidecars it names are the ones it
assumes rather than the ones it observed. The three lifetime issues above have no coverage at all,
and by their nature would not be caught by a passing run.

One coordination note: `writer_timeout_sink_stalled_writer.rs` is also the subject of #2310, so
whichever lands second will need a rebase on that file.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at 8ffff658.

PR #2307, opened 2026-08-30
perf(knowledge): reuse query work per request

### LINKED ISSUE #2218: knowledge.stats fails consistently at ~153K atoms: 'stats finalized: timeout during query_row'
Live deployment, reproduced on demand (was intermittent, became consistent after the corpus grew ~4.7K rows to ~153K atoms):

```
{"error":"internal: stats finalized: timeout during query_row","ok":false,"tool":"knowledge.stats"}
```

The failing query is the third of five sequential scalar counts in `stats` (`crates/khive-pack-knowledge/src/knowledge/crud.rs:823-830`):

```sql
SELECT COUNT(*) FROM knowledge_atoms
WHERE namespace = ?1 AND finalized = 1 AND deleted_at IS NULL
  AND tags NOT LIKE '%type:domain%'
```

Observations:

- Every count in the verb is a full-scan shape at this corpus size (`tags NOT LIKE` defeats any index), and they run sequentially on one reader. The error consistently names the **third** query, which suggests a shared per-request deadline being consumed by the first two scans rather than the finalized predicate itself being uniquely slow — but that attribution is unverified.
- Downstream cost: any monitoring that reads corpus totals from `knowledge.stats` loses its measurement while this fails (one such consumer correctly fails closed and reports instrument-fault rather than a number).

Remedy directions (pick after measuring): indexed counts (partial index or a maintained counter), a single-pass `SUM(CASE...)` replacing the two atom scans, or a per-statement rather than per-request read deadline for diagnostic verbs.

### LINKED ISSUE #2232: perf(knowledge): one request embeds the same query up to seven times, against ADR-051
The same request's query text is embedded more than once on several public retrieval paths. This contradicts an accepted design decision rather than merely costing extra work: `docs/adr/ADR-051-section-embeddings-hybrid-compose.md:78-95,90` states the query is embedded once and reused across scoring stages.

## The seam

`embed_cosine_scores` (`crates/khive-pack-knowledge/src/knowledge/search.rs:864-879`) takes the query as a string and prepends it to the candidate batch:

```rust
let mut texts = Vec::with_capacity(candidate_texts.len() + 1);
texts.push(query.to_string());
texts.extend_from_slice(candidate_texts);
```

So every caller that has already embedded that query pays for it a second time. `crates/khive-runtime/src/retrieval.rs:636-646` already accepts an optional precomputed query vector, so the reuse path exists and is simply not threaded through.

## Sites

1. **`knowledge.search`** — ANN embeds `raw_query` at `search.rs:1849-1855`, then reranking calls the helper at `:1910-1916`, which embeds the identical query again. **2x** query representation on the ANN-plus-rerank path.

2. **`knowledge.suggest`** — same shape: ANN embed at `search.rs:2009-2015`, rerank re-embed at `:2065-2073`. **2x** on the normal non-timeout path. This one also feeds auto-compose, so it multiplies downstream.

3. **KG blending in compose** — `search_kg_entities` (`search.rs:1511-1514,1554-1562`, retrieval leg `crates/khive-runtime/src/retrieval.rs:738-746,553-575`) embeds the query for the concept hybrid search, again for the document hybrid search, and again for the final entity-candidate cosine batch. **3x** when both kind loops run and candidates survive. The candidate vectors themselves are distinct needed artifacts; only the query repeats.

4. **Atom and section scoring in compose** — atom reranking embeds the query through `rerank_text_items` (`search.rs:2406-2407`), then section scoring calls `runtime.embed_query(&raw_query)` separately at `:2446-2451`. **2x** when atoms and sections are both present. ADR-051:90 specifically says these two stages share one query embedding.

Aggregated, a single `knowledge.compose` request on the normal path with sections, both KG kinds, surviving candidates, and suggest reranking can compute **up to 7 representations of one query**. That figure is an aggregate of the sites above, not an additional count.

## How the multipliers were established

Derived from source, not measured end to end, and the reason is worth recording for whoever picks this up: live probes of `knowledge.search` and `knowledge.suggest` returned `degraded.lexical_timeout=true`, and the degraded branch at `search.rs:1910-1916` and `:2065-2073` skips reranking entirely. So the observed counters (`embed_calls: 30` and `embed_calls: 1` respectively) describe the degraded path and are **not** measurements of the duplicate. Anyone confirming this should first get a non-degraded run.

## Suggested fix

Compute one query vector per request and thread it through. Concretely: give `embed_cosine_scores` an optional precomputed query vector, retain the ANN `query_emb` where one was already produced, and batch only candidate texts. For compose, carry a request-scoped query representation from auto-suggest through the atom, section, and KG stages. Nothing here needs new machinery — the optional-vector parameter already exists one layer down.

## Not in scope

The per-embedding-model fan-out is intentional and sanctioned (`docs/adr/ADR-031-multi-engine-retrieval.md:28-30,59-64`); identical sibling content is already grouped so one computation serves each model (`crates/khive-runtime/src/atomic_message.rs:303-370`, with the expectation asserted at `:815-864`). Separate FTS and per-model vector projections of a record are likewise required by the accepted consistency contract. This issue is only about one query being embedded repeatedly inside a single request.

ohdearquant:
## Summary

- combine atom and finalized totals into one knowledge.stats aggregate scan
- cache one successful query embedding across ANN, rerank, compose section, and KG blend stages
- preserve lexical-only degradation behavior and add request-level recording coverage

## Verification

- cargo test -p khive-pack-knowledge --lib (157 passed)
- focused integration: stats_reflects_current_corpus
- focused integration: auto_compose_embeds_request_query_exactly_once
- cargo clippy -p khive-pack-knowledge --all-targets -- -D warnings
- cargo fmt --all -- --check
- git diff --check

Closes #2218
Closes #2232


ohdearquant:
The stats consolidation is correct, and I checked it against the failure it usually has. The two
original queries counted `knowledge_atoms` with `namespace = ?1 AND deleted_at IS NULL AND tags NOT
LIKE '%type:domain%'`, the second adding `finalized = 1`. The replacement keeps that one WHERE clause
and computes `COUNT(*)` alongside `SUM(CASE WHEN finalized = 1 THEN 1 ELSE 0 END)`
(`crates/khive-pack-knowledge/src/knowledge/crud.rs:846-860`). The populations coincide exactly: NULL
`tags` stays excluded because `NOT LIKE NULL` is not true, and NULL `finalized` falls to the CASE ELSE
branch just as it failed the old `= 1` predicate. `eval_coverage` cannot move for an unchanged
database, and the single scan gives a more coherent pair under concurrent writes. Result-identical
and only faster, which is the intended answer.

The embedding reuse is a different story, because the slot being reused is not keyed.

It is a bare `&mut Option<Vec<f32>>` threaded through the handlers — no query text, no model, no
revision, no dimension, no normalization, and critically no role. That would still be safe if only
one thing ever produced the vector. Two things do, and they call different APIs:

- When ANN or suggest runs first, the slot is filled via `embed_query`, which for anything that is
  not a BGE variant dispatches to `service.embed_query(&texts, emb_model)`
  (`crates/khive-runtime/src/retrieval.rs:320-326`).
- When it is not already populated, `embed_cosine_scores` pushes the query into the *same batch* as
  the candidate texts (`search.rs:876-885`) and embeds the lot through `embed_batch`, which always
  calls generic `service.embed(texts, emb_model)` (`retrieval.rs:405-411`).

For instruction-tuned models those are not the same vector — E5 and Qwen prefix a query instruction
on the query-role path and nothing on the generic path. The BGE carve-out in `embed_query` is the
tell: someone already knew these two APIs are not interchangeable for every model, and encoded the
exception.

So the ordering consequence is concrete. Atom and entity rerank previously embedded
`[raw_query, candidates]` together through generic `embed_batch`, scoring a generic query vector
against generic candidate vectors. After this change, when ANN populated the slot first, rerank scores
those same generic candidate vectors against a query-role vector. Cosine similarity between an
instruction-prefixed query embedding and unprefixed candidate embeddings is a different number, so
scores and result order shift — silently, with no error and no degradation marker. The same path
reaches compose atom rerank and KG entity rerank. The reverse case exists too: if the slot was filled
by the generic batch, compose section scoring then consumes a generic vector where it previously made
its own `embed_query` call.

For symmetric models — AllMiniLM, and BGE through the compatibility branch — this genuinely is
result-identical and only faster. That is why the suite does not catch it:
`query_embedding_cache_reuses_query_vector_across_reranks` (`search.rs:3301-3422`) uses AllMiniLM with
a provider that ignores role, so it would pass unchanged even if the key omitted every discriminator
listed above. It proves the vector is reused, which is the intended behaviour, not that reuse is
sound. A test that separates the two needs a provider returning distinguishable vectors for
`embed` versus `embed_query`.

The fix is either a real key — query text plus model, revision, dimension, normalization and role —
or two slots, one query-role and one generic, with each stage drawing from the one matching what it
used before.

Second issue, independent of the key. `search_kg_entities` now returns early:

```rust
let Some(query_vector) = query_embedding.clone() else { return Ok(Vec::new()); };
```

(`search.rs:1620-1622`). Previously it called `hybrid_search(token, query, None, candidate_k, ...)`
and the core runtime embedded the query itself when handed `None`. That difference matters because a
knowledge pack can be configured `no_embed` while `core()` is wired to the main runtime's embedders.
Under that configuration compose can no longer populate the slot through the pack runtime, so it now
returns an atom-only briefing with entities omitted, where before the core hybrid search discovered
them. A transient embedding failure has the same shape: the old path could still embed on its own,
the new one never attempts it. This is lost results rather than slower ones, and the caller cannot
distinguish it from a genuine no-match — the existing degradation marker keys on ANN and candidate
availability, not on an empty query vector.

Failure caching is handled correctly, for what it is worth: `embed_cosine_scores` returns before
assigning the slot on an embedding error or cardinality mismatch (`search.rs:886-906`), so one blip
is not persisted for the rest of the request. The "one successful embedding" wording matches the
code.

Coverage gaps beyond the role case: no test exercises an initial query-embedding failure with reuse
enabled, the `no_embed` pack with a configured core runtime, or transient recovery. On the stats side
the existing coverage checks true/false finalized and the 0.5 calculation but not NULL `finalized`,
NULL `tags`, or the domain-tag exclusion — those are the predicates the consolidation had to preserve,
so they are the ones worth pinning now that both numbers come from one scan.

ohdearquant:
## The red `CI gate` is one test, in the crate this branch changes

```
FAIL khive-pack-knowledge::integration kg_blend::kg_blend_failure_degrades_to_atom_only_response
Summary: 2398/4675 tests run: 2397 passed, 1 failed
```

`Coverage ratchet` also fails at `Compute coverage`, downstream of the same run.

Every file in this branch is under `crates/khive-pack-knowledge/`, and the failing test is a
`khive-pack-knowledge` integration test, so this is not the ambient-flake pattern showing up on
other branches today. The assertion is about a blend failure degrading to an atom-only
response, which is a fallback path — exactly the kind of behaviour that changes shape when
query work is reused across a request. Worth confirming the degraded path still runs when the
cached work is present.

ohdearquant (commented):

Verdict on head 701c3c5214eba5a6007a50b2878d6733fb64ef70: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approved. CI is green at this head.

The three-state role-specific embedding closes the masking path cleanly. Because
`as_deref()` returns `None` for both `NotAttempted` and `Failed`, read sites are
unchanged, while `is_not_attempted()` is true only for `NotAttempted` and is what
gates the fill (`search.rs:2290`, and again at `:2714`). So a slot whose provider
call already failed is neither retried nor overwritten by the generic batch, which
is the behaviour the type was introduced for. Failure paths set `Failed` rather
than leaving the default, so the distinction is actually established and not just
representable.

PR #2308, opened 2026-08-30
feat(knowledge): add projected cursor traversal

### LINKED ISSUE #2220: knowledge.list: key-only projection and a declared stable order (or cursor) for full-store walks
Two gaps in \`knowledge.list\` make a full-store key walk (e.g. computing a set difference against an external inventory) impractical and unsound:

**1. No projection.** Every row carries its full \`content\`. Measured against a large store: 200 rows = 4.2MB, so a full walk at the largest workable page size is ~769 calls, ~5.3 hours, ~3.3GB transferred — for a question that needs only \`slug\`. Content is ~99% of the bytes. Latency is also flat (~25s per call from limit=50 through 500), so the cost is not row-count-driven and smaller pages buy nothing.

**2. No declared order.** The live schema exposes only \`type/limit/offset/status/exclude_status\` — no \`order_by\`, no cursor. Offset pagination without a declared stable order over a store receiving concurrent writes can skip rows and double-count rows, and neither is visible in the results. For completeness-sensitive consumers (anti-joins, mirror parity checks) that makes the walk unsound, not merely slow.

**Ask:** a fields/projection parameter (at minimum a keys-only mode returning \`slug\`/\`id\`), plus a declared stable order (\`created_at,id\` ascending) or an opaque cursor. Either alone helps; together they turn a 5-hour unsound walk into a cheap sound one.

Related: #2219 (frame-budget overflow silently truncates the same walk).

ohdearquant:
## Summary

- add strict per-type `fields` projection to `knowledge.list` and push it into the SQL select list, retaining only hidden `id`/`created_at` pagination keys
- add `after=""` / `next_after` keyset traversal ordered by `created_at ASC, id ASC`, with full-UUID validation, namespace/type checks, machine-readable order metadata, and mutually exclusive offset mode
- preserve legacy offset behavior under its now-declared `created_at DESC, id DESC` order
- document live traversal semantics: inserts behind an issued boundary wait for a fresh walk, inserts ahead may extend the current walk, and `next_after`—not the changing `total`—is the completion signal
- update strict params, vocabulary parity, ADR/API docs, and pack guidance

## Tests

- `cargo test --manifest-path crates/Cargo.toml -p khive-pack-knowledge --lib` (157 passed)
- `cargo test --manifest-path crates/Cargo.toml -p khive-pack-knowledge --test integration` (118 passed)
- cursor concurrent-insert regression repeated 20/20 passes
- focused schema-routing and parameter-strictness tests passed
- `cargo check --manifest-path crates/Cargo.toml -p khive-pack-knowledge --all-targets --all-features`
- `cargo clippy --manifest-path crates/Cargo.toml -p khive-pack-knowledge --all-targets --all-features -- -D warnings`
- `git diff --check`

Closes #2220


ohdearquant:
The keyset predicate is correct, and that is the part worth stating first, because it is where this
feature class usually fails. Both branches emit the lexicographic tuple comparison

```sql
(created_at > ?cursor_created_at OR (created_at = ?cursor_created_at AND id > ?cursor_id))
```

with `ORDER BY created_at ASC, id ASC` (`crates/khive-pack-knowledge/src/knowledge/crud.rs:879-899`,
`:994-1017`). For rows A<B<C<D all sharing timestamp T at `limit=2`, the first page returns A,B with
`next_after=B` and the next returns C,D — neither skipped nor re-emitted. A `created_at >` predicate
would have dropped C,D and `created_at >=` would have repeated A,B. Equal timestamps are the normal
case here because bulk import writes many rows in one instant, so this matters more than it looks.

Two things to fix.

**An explicit `after: null` silently restarts at page one.** `after` is declared
`#[serde(default)] pub after: Option<String>` (`crates/khive-pack-knowledge/src/knowledge/schema.rs:163`),
so serde maps an explicit JSON `null` to `None` — the same value an omitted key produces. Cursor mode
is then selected by presence alone, before the value is parsed:

```rust
let cursor_mode = p.after.is_some();   // crud.rs:810
```

A caller that round-trips a terminal cursor back into the next request — which is the natural thing
to do with a field whose terminal value *is* `null` — therefore gets a successful `created_at DESC`
first page instead of an error. The failure is a silent restart, so a client looping on `next_after`
rescans from the top without ever seeing a signal. Every other malformed cursor is rejected
(`deadbeef` → `InvalidInput` at `:79-88`; well-formed but absent UUID → `NotFound` at `:856-861`),
which makes `null` the one input that fails open.

The remedy is already in this crate. `tri_state` on `main`
(`crates/khive-pack-knowledge/src/knowledge/schema.rs:113`) exists to preserve exactly the
distinction serde's plain `Option<T>` collapses, and is already applied to `finalized`, `source_uri`
and `source_type` in the same file. Declaring `after: Option<Option<String>>` through it lets the
handler reject `Some(None)` as `InvalidInput` while leaving omission alone. An integration assertion
for `{"after": null}` would pin it.

**The API reference omits the projection vocabulary.** `validate_list_fields` rejects anything
outside two exhaustive per-type allowlists (`crud.rs:24-52`, `:91-113`), and ADR-047 lists those
names (`docs/adr/ADR-047-knowledge-pack.md:242-248`), but the caller-facing table at
`docs/guide/api-reference.md:1721-1729` only says "strict non-empty projection" and shows `id,slug`
as an example. Against a fail-closed endpoint that leaves a caller guessing. Either list both field
sets there or link the normative allowlist.

Two notes that are not defects:

- SQL construction is safe. `list_select_columns` does interpolate field names into the SELECT list
  (`:124-135`), but its only caller runs after `validate_list_fields` (`:816`), and the allowlist is
  the fixed per-type constant set. Column identifiers cannot be bound parameters, so an exhaustive
  allowlist with a rejecting non-matching branch is the correct mechanism, and that is what is here.
  Cursor, namespace, status, limit and offset all remain bound. Projections that omit `id` or
  `created_at` do not break the walk: both are added internally and stripped after rendering.
- `next_after` is not opaque — it is the boundary row's UUID, unsigned, and any existing UUID in the
  same namespace and type is accepted as a deliberate jump. That is a defensible design, but it is a
  contract worth stating in the reference rather than leaving implicit.

Backward compatibility holds for the offset path: same rows, same `created_at DESC, id DESC`
ordering, same envelope plus an additive `order` field.

Two test gaps beyond the null case. There is no domain-branch equal-timestamp cursor test — the
existing equal-timestamp domain coverage is offset-only, so a mutation in the domain cursor SQL would
escape the suite even though the atom branch is well covered. Cursors pointing at soft-deleted and
hard-deleted boundary rows are also untested, and they behave differently: the cursor lookup omits
`deleted_at`, so a soft-deleted boundary still resolves while a hard-deleted one returns `NotFound`.

One mechanical note: this branch changes `crud.rs`, `schema.rs`, `vocab.rs` and ADR-047, all four of
which moved on `main` a few minutes ago, so it needs a rebase regardless. Picking up `tri_state` in
the same pass is the cheap route.

ohdearquant:
The only thing red here is `Docs lint`, and it is a one-command fix.

`deno fmt` reports `Found 1 not formatted file in 206 files`: **`docs/guide/api-reference.md`**,
around lines 1728-1729. The `status` and `exclude_status` rows are padded to a different column
width than the rest of the table.

`deno fmt docs/guide/api-reference.md` (or `make fmt`) resolves it. The other red entries on this
PR are sibling matrix jobs cancelled by fail-fast when this one failed — collateral rather than
separate problems, so this single fix should clear the whole set.

ohdearquant (commented):

ohdearquant (commented):

Verdict on head bedb8d37a0653582fd5a726b30e6e139cafe8d4e: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at 13fa2a25.

PR #2311, opened 2026-08-30
fix(runtime): keep cross-pack reads available under audit pressure

### LINKED ISSUE #2217: Read verbs fail on audit-lane admission exhaustion (QueueAdmissionExhausted on gtd.tasks)
Observed on a live deployment, 2026-08-25 ~07:35 UTC, twice ~30s apart, verbatim:

```
{"error":"internal: audit obligation commit failed for verb \"gtd.tasks\": QueueAdmissionExhausted","ok":false,"tool":"gtd.tasks"}
```

Shape of the failure:

- The refusing op is a **read** (`gtd.tasks(status=..., assignee=...)`); what exhausted is the read's **audit obligation commit** — the write-side obligation attached to a read op — not the read itself.
- Not connection- or batch-wide: two `memory.remember` ops in the same batch committed with ids while the `gtd.tasks` op in that batch refused; `comm.*` verbs worked throughout on the same session.
- Transient: the same call succeeded on a later retry (~40 min after), so this is admission-queue contention, not a persistent fault.

Why it matters: under writer-queue pressure, audit-obligated reads fail even though the domain read itself needs no writer slot. A read surface that goes dark exactly when the write plane is congested inverts the degradation callers expect.

Structural fix in flight: the events-daemon split (ADR-170, #2190/#2195/#2202/#2203) moves the idempotent audit-batch lane off the domain writer, which removes the contention source for exactly this class. Until it lands, worth considering whether an audit obligation refused on admission should degrade (queue the obligation, or classify as retryable with the read still served) rather than fail the read outright — the retryable classification exists (`QueueAdmissionExhausted` is transient by definition) but the read's result is discarded with it.

ohdearquant:
## Summary

- expand the explicit audit-admission degradation allowlist from the original KG-only set to 39 reviewed, side-effect-free Assertive verbs across agent, blob, brain, comm, GTD, KG, knowledge, moodboard, schedule, and session packs
- keep all five Assertive handlers with durable or accounting-bearing incidental effects fail-closed: `memory.recall`, `db_diagnostics`, `knowledge.search`, `knowledge.suggest`, and `knowledge.compose`
- replace the KG-only vocabulary check with a dynamic cross-pack source census that discovers every `khive-pack-*` source tree and requires all 44 public Assertive verbs to be classified exactly once
- add a forced `QueueAdmissionExhausted` regression proving the live-reported `gtd.tasks`, `gtd.next`, and `comm.inbox` reads still return their results while the five incidental-effect probes remain strict
- update ADR-103 Amendment 3, ADR-133 Amendment 1, and the audit-batch guide for the expanded named set and unchanged counted-undercount contract

## Verification

- `cargo test --manifest-path crates/Cargo.toml -p khive-runtime admission_degrade_safe_assertive_census_matches_live_pack_sources --lib`
- `cargo test --manifest-path crates/Cargo.toml -p khive-runtime --features fault-injection,test-internals --test read_verb_admission_exhaustion`
- `cargo check --manifest-path crates/Cargo.toml --workspace --all-targets --all-features`
- `cargo clippy --manifest-path crates/Cargo.toml --workspace --all-targets --all-features -- -D warnings`

Closes #2217


ohdearquant:
To answer the shape question first: this sheds selected audit work under audit-lane admission
pressure. It does not add a separate capacity lane and it does not let a read skip the gate,
namespace, or authorization path — `gate_request_with_identity`, namespace resolution and token
minting are untouched and still run before dispatch. The containment work is genuinely careful:
`ADMISSION_DEGRADE_SAFE_VERBS` is an explicit fail-closed allowlist rather than a category proxy,
relaxation additionally requires `DispatchSucceeded` (`pack.rs:3833-3835`), gate-denied,
config-locked, unknown-verb and failed-dispatch producers all pass `false`, and the cross-pack
census makes a newly added Assertive handler strict until someone reviews it. The queue is bounded
(4096 pending, 256 per generation) and refuses at the bound rather than growing.

**The defect is that the exemption's trigger is broader than the exemption's stated bound.** Both
ADRs are explicit that this must not cover durability loss:

> It applies to two specific terminal reasons, not persistent store failure. A persistently failing
> store still fails these dispatches exactly as D2 requires — the exception exists only for the audit
> lane's own transient admission pressure, not for durability loss. (ADR-133)

> **Bounded to transient audit-lane admission pressure**, not persistent store failure. A persistent
> commit failure for one of these rows is unaffected by this amendment. (ADR-103)

But `AdmissionDeadlineExpired` is not a statement about admission pressure. It is produced by a bare
timeout elapsing, whatever the reason:

```rust
match tokio::time::timeout(self.config.admission_deadline, rx).await {
    Ok(Ok(result)) => result,
    Ok(Err(_recv_error)) => Err(AuditTerminalReason::DriverJoinLost),
    Err(_elapsed) => Err(AuditTerminalReason::AdmissionDeadlineExpired),
}
```

(`crates/khive-runtime/src/audit_batch.rs:693-702`). A store that is *persistently* failing but slow
— still inside `append_events_idempotent` retries when the five-second deadline elapses — produces
exactly this reason, and `pack.rs:3872-3883` then takes the relaxed arm, increments
`AUDIT_ADMISSION_UNRESOLVED_OBLIGATIONS`, and returns `Ok(())`. The caller receives success for a
read whose accounting row later exhausts its retries and is never committed. That is the case both
documents promise is excluded.

The code is honest about the ambiguity, which is what makes it findable: the source comment says the
row "is left in place for the driver to drain", and the warning text says "it may still commit". A
terminal reason that cannot say whether the work survived is not a safe predicate for deciding that
losing it is acceptable. Relatedly, "counted, not silent" holds for the *event* but not for the
*outcome* — the unresolved counter is an upper bound that never resolves, so no operator surface ever
distinguishes an eventually-committed row from a lost one.

Worth noting `audit_batch.rs` is not in this diff. The mismatch is between a new consumer and a
pre-existing terminal reason, so the fix has two plausible homes: give the batch a reason that
separates "deadline elapsed, driver healthy and still working" from "deadline elapsed, driver
terminally failing", or keep the relaxed arm but resolve the obligation afterwards so a later
`StoreFailure` still surfaces. Either way this needs a regression test injecting a store that is slow
past the deadline and then terminates with `StoreFailure` — there is currently no such test, and it
is the only case that separates the implemented behaviour from the documented one.

Two notes that are not blocking.

Read classification is an exact-name allowlist plus a live `VerbCategory::Assertive` check. That is
fail-closed for *new* names, which is the important direction. The residual exposure runs the other
way: `Assertive` is a speech-act declaration rather than an effect capability, and the census parser
is deliberately textual, so an already-allowlisted verb that later *acquires* a durable or
accounting-bearing side effect stays eligible without anything flagging it. The source comment
acknowledges this. A periodic re-review of the allowlist, rather than only the drift check on
additions, would close the remaining direction.

On the feature gate — I checked this rather than leaving it open, because it is easy to assume wrongly
in either direction. `read_verb_admission_exhaustion.rs` is gated
`#![cfg(all(feature = "fault-injection", feature = "test-internals"))]`, neither feature is default,
and `scripts/ci.sh` runs `cargo nextest run --workspace --partition ...` with no `--features` flag.
That looks like a dead test, but it is not: `khive-pack-kg`'s dev-dependency on `khive-runtime`
enables both features, and under resolver 2 that unifies across a workspace test build. Confirmed
against CI logs from a green run rather than by reasoning — both shards show
`khive-runtime::read_verb_admission_exhaustion` cases executing and passing, including
`read_verb_dispatch_survives_audit_lane_admission_deadline_expiry`. The centrepiece test is live.

ohdearquant (commented):

Verdict on head 6c0dbc0355d17a9af4f0c3886ab78560c484c533: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 8979659159e29bf210f14c67ff740dac0804c23b: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approved. CI is green at this head.

The per-registration trust bit is the right shape, and it is consulted rather
than merely recorded: an untrusted pack short-circuits before its handlers are
examined at all (`if !trusted { continue; }` in the eligibility loop), so it
cannot enter the degrade-safe set by any path, whatever it reports as its name.
Eligibility then still requires an `Assertive` verb handler AND membership in
the `(pack, verb)` allowlist, so the bit widens nothing on its own.

Deciding it once at build into a flat verb set also removes the per-dispatch
pack/handler scan rather than adding a second check beside the old one, which is
what keeps the guard from drifting away from the thing it guards.

PR #2312, opened 2026-08-30
perf(knowledge): overlap search candidate stages

### LINKED ISSUE #1996: knowledge: run ANN and lexical candidate stages concurrently to stop serializing the read deadline
Follow-up to #1993. The search and suggest handlers now run query embedding + ANN candidate retrieval to completion before starting the lexical/FTS stage. The ordering is deliberate — ANN results must already be in hand when the deadline-prone lexical stage spends the request read deadline, so a lexical timeout degrades to ANN-backed results instead of erroring — but strict sequencing means embedder/ANN latency depletes the shared deadline before FTS begins, making premature `degraded.lexical_timeout` responses more likely under embedder load.

The two stages are independent until fusion; they can run concurrently while preserving the invariant that matters (ANN results exist at the moment a lexical timeout is observed — a concurrently-completed ANN task satisfies it the same way). Expected effect: fewer degraded responses under load, no contract change.

Sites: `crates/khive-pack-knowledge/src/knowledge/search.rs` search and suggest handlers (ANN block preceding `search_core`/`search_decomposed`).

ohdearquant:
## Summary

- run the independent ANN query/hydration and lexical candidate stages concurrently for both `knowledge.search` and `knowledge.suggest`
- keep both futures attached to the caller task so the existing absolute read deadline, cancellation, timeout fail-open behavior, eligibility filters, hydration diagnostics, fusion, and rerank guards remain unchanged
- add controlled paused-clock regressions that prove lexical work starts while ANN is blocked and that a completed ANN result survives a lexical deadline expiry
- keep query-vector reuse and request caching from #2232 out of this change

## Validation

- `cargo test -p khive-pack-knowledge` — 402 passed, 1 ignored; doctests passed separately
- `cargo check -p khive-pack-knowledge --all-targets --all-features`
- `cargo clippy -p khive-pack-knowledge --all-targets --all-features -- -D warnings`
- `cargo fmt --manifest-path crates/Cargo.toml --check --all`

Closes #1996


ohdearquant:
The dependency audit comes back clean, which is the thing that had to be true: the lexical leg never
read an ANN-produced value. Hits, availability and hydration counts are consumed only after the join
(`crates/khive-pack-knowledge/src/knowledge/search.rs:2128-2141`), candidate vectors and dedup and
score maps are local to their own futures, and the query embedding is a local inside the ANN future
rather than a slot both legs touch.

Two design choices deserve credit rather than silence. `join_candidate_stages` uses `tokio::join!`
and keeps both futures **attached** to the caller task (`search.rs:1411-1419`) — spawning them would
have detached them from the task-local read deadline and from cancellation, which is the usual way
this refactor goes wrong. And determinism holds: `join!` preserves source association, lexical
ranking has a slug tie-break, and RRF sorts by descending score then id, so completion order does not
reach the result. That is worth stating explicitly because "results may now vary run to run" is the
default outcome of overlapping two producers, and here it does not happen.

**The one real behaviour change is contention that did not previously exist inside a single
request.** Both legs acquire SQL readers, and they now hold them at the same time. Before this diff
only one candidate stage was ever active per request, so the two could not compete with each other.
What makes it worth fixing rather than noting is that the two legs classify the resulting failure
differently:

```rust
fn is_read_timeout(e: &RuntimeError) -> bool {
    matches!(e, RuntimeError::Storage(khive_storage::StorageError::Timeout { .. }))
}
```

(`search.rs:523-528`). That matches storage read timeouts only, so a pool **admission** timeout is
not fail-open — on the lexical path it propagates as a hard error. On the ANN path the equivalent
loss converts to an empty `AnnCandidateOutcome` and, on the generic fallthrough, carries no
`ann_unavailable` marker.

So under `max_readers=1` or a globally saturated pool, the same contention produces two different
caller experiences depending on which leg loses: a verb-level error, or a silently lexical-only
result that is indistinguishable from a query with no vector matches. The second is the worse one,
and this diff is what makes it reachable from within one request. Either keep sequential scheduling
when the reader budget cannot support two legs, or give the ANN-lost-to-admission case its own
caller-visible degradation classification so the quiet path stops being quiet.

**Tests cannot distinguish the two completion orders.** `search_and_suggest_candidate_stages_overlap`
(`ann_degrade_tests.rs:274`) blocks ANN and proves only that lexical starts before ANN is released;
`controlled_lexical_timeout_keeps_concurrent_ann_result` (`:338`) explicitly waits for `ann_finished`
before releasing lexical. Neither arranges lexical-first completion, and none compares the two orders.
The implementation is order-independent today, so this is not a live defect — but it means the suite
would stay green through a future change that accumulates in completion order or leaves score ties
unresolved, and pagination and cached ordering are what break when that happens. Running the same
query twice under deterministic gates forcing each order, with a tie case, and asserting identical
result JSON would close it.

One visibility note. The rename to `wall_clock_lexical_timeout_degrades_suggest_without_verb_error`
also deletes two assertions from that test — `result["total"] > 0` ("ANN candidates fetched before
the lexical stage must still produce results") and the check that the top result is the seeded
vector-backed domain — leaving only `degraded.lexical_timeout`. In fairness the coverage is not lost:
`controlled_lexical_timeout_keeps_concurrent_ann_result` asserts ANN survival on the real handler
legs under paused time with `rebuild_ann: false`, which is deterministic where the old test was
wall-clock. So this is fine on the merits. It is worth calling out only because a rename is the
easiest place for an assertion deletion to pass unread, and the assertions removed were the ones
covering the exact property this diff puts at risk.

ohdearquant (commented):

Verdict on head 8306c79641cba6c41dab0341c9b4609662642a30: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

ohdearquant (commented):

Verdict on head 6c9e66d34023f837f620680651b1f2a4a45e1a36: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head e3db0d346fa37b9286ff45dbb0bac876f20423a2: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 979b7455ade0d5dcbf493781d315dcecbb9d4a46: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630:
Closing on age. 174 commits behind main with a large diff on a search path that has changed substantially since, so landing it would mean resolving a conflict rather than reviewing an optimization.

If a measurement still wants this overlap, it should come back as a fresh change against current main with the benchmark that motivates it.

Issue #2273, opened 2026-08-28
knowledge FTS5 external-content indexes desync; hard DELETE raises SQLITE_CORRUPT


oceanwaves630:
## Symptom

On a long-lived store, `DELETE FROM knowledge_atoms` / `DELETE FROM knowledge_sections` fails with `database disk image is malformed` (SQLITE_CORRUPT), while `PRAGMA integrity_check` reports `ok`. The FTS5 integrity-check command fails instantly on both indexes:

```sql
INSERT INTO fts_knowledge(fts_knowledge, rank) VALUES('integrity-check', 1);
-- => database disk image is malformed
INSERT INTO fts_sections(fts_sections, rank) VALUES('integrity-check', 1);
-- => database disk image is malformed
```

`fts_knowledge` and `fts_sections` are external-content FTS5 tables (`content=knowledge_atoms` / `content=knowledge_sections`), so `integrity_check`/`quick_check` cannot see the desync — FTS5's own integrity-check command is the only detector.

## Mechanism (one proven source)

The trigger pair on `knowledge_atoms` is asymmetric:

- `fts_knowledge_ai` (AFTER INSERT) is guarded: `WHEN new.deleted_at IS NULL` — soft-deleted rows never enter the index.
- `fts_knowledge_ad` (AFTER DELETE) is unconditional — it issues the FTS5 `'delete'` command for every row, including rows whose index entry was never created (inserted soft-deleted) or already removed (`fts_knowledge_au` removes the entry when a row is soft-deleted and does not re-insert it).

An FTS5 `'delete'` against a missing entry is undefined for external-content tables and surfaces as SQLITE_CORRUPT. Any store containing soft-deleted atoms (the measured store has 518) is guaranteed to hit this on hard delete. `fts_sections` has no `deleted_at` column and symmetric triggers, yet also fails integrity-check on the measured store, so at least one additional desync source exists (likely a historical write path that bypassed the triggers).

Beyond the hard-delete crash, a desynced external-content index silently returns wrong/missing rows for FTS-backed knowledge search.

## Suggested fix

1. Guard the delete side to match the insert side: `fts_knowledge_ad` and the delete half of `fts_knowledge_au` should fire only `WHEN old.deleted_at IS NULL`.
2. Ship a repair path for existing stores: `INSERT INTO fts_knowledge(fts_knowledge) VALUES('rebuild')` (and same for `fts_sections`), e.g. as part of `reindex` or a dedicated maintenance verb.
[...]

PR #2313, opened 2026-08-30
fix(knowledge): repair external-content FTS indexes

ohdearquant:
## Summary\n\n- add migration V23 to rebuild both knowledge FTS5 indexes and move atom external content to a live-row view\n- make atom insert, update, soft-delete, resurrection, and hard-delete trigger transitions symmetric\n- expose an atomic rebuild plus rank-1 integrity check through knowledge.index(rebuild_fts=true) and the existing kkernel reindex operator path\n- document the repaired invariants and migration ledger\n\n## Diagnosis\n\nfts_knowledge named the full atom table as external content while intentionally omitting soft-deleted rows from the index. That made FTS5 rank-1 integrity compare different row sets, and a later hard delete could issue a second FTS delete for an already-removed row. The section update trigger was narrowed in V2 to stop embedding-only churn, but that migration did not rebuild fts_sections, so stores that had already drifted stayed desynchronized.\n\nV23 fixes the atom content/index contract and repairs both existing indexes. The operator path rebuilds both indexes and verifies each against its external content object in the same writer batch.\n\n## Verification\n\n- cargo test --manifest-path crates/Cargo.toml -p khive-db --lib (839 passed)\n- focused V23 lifecycle/integrity migration regression\n- focused knowledge FTS drift/rebuild regression\n- existing knowledge hard-delete regressions (3 passed)\n- cargo test --manifest-path crates/Cargo.toml -p kkernel reindex::tests (37 passed)\n- affected cargo check and cargo clippy --all-targets -- -D warnings\n- cargo fmt --check, git diff --check, and SQL lint\n\nCloses #2273


ohdearquant:
The structural properties that matter most for a migration against live data all hold, and they are
worth stating first because they are what makes the rest reviewable rather than alarming.

Every removal in V23 is derived state: three atom FTS triggers, the `fts_knowledge` virtual table,
and the repair view (`crates/khive-db/sql/023-knowledge-fts-repair.sql:14-18`). No base
`knowledge_atoms` or `knowledge_sections` row is dropped, deleted, truncated or updated, and both
indexes are reconstructed from surviving content. Atomicity is right too — the runner opens an
immediate transaction (`migrations.rs:1262`), runs the script, and inserts the ledger row **inside
the same transaction** (`:1394-1403`) before committing, so there is no window producing "data
migrated but not recorded" or the reverse. A kill mid-flight rolls back both.

The rowid question also resolves cleanly, and for a better reason than the one I expected.
`knowledge_atoms` is `id TEXT PRIMARY KEY`, so it keeps an implicit rowid, and the new view projects
that same underlying rowid with `content_rowid=rowid`. But the decisive point is that V23 does not
migrate the index in place at all: it drops `fts_knowledge` and rebuilds it (`:17`, `:25-34`, `:58`),
so no stale entry survives to be repointed at the wrong document. That is the failure mode this
change class usually has, and dropping rather than converting avoids it entirely.

The trigger matrix is complete. Insert guards on live rows, delete guards on old-live, and update
guarded independently on both sides with `deleted_at` included in the `AFTER UPDATE OF` column list.
Resurrection (soft-deleted to live) indexes exactly once because the old-live branch is suppressed
and the new-live branch fires; embedding-only section updates stay silent under the V2 narrowed
trigger without suppressing a state transition. Those are the two shapes this bug class hides in and
both are handled.

**The finding is that the repair is globally scoped but reachable from an ordinary namespaced
caller.** The helper takes no token at all:

```rust
async fn rebuild_fts_indexes(runtime: &KhiveRuntime) -> Result<(), RuntimeError> {
```

(`crates/khive-pack-knowledge/src/knowledge/index_handler.rs:239`), and it issues
`INSERT INTO fts_knowledge(fts_knowledge) VALUES('rebuild')` plus the same for `fts_sections` —
whole-store rebuilds of both indexes. In the handler it runs *before* any namespace-scoped work and
independently of it:

```rust
let ns = token.namespace().as_str().to_owned();
if rebuild_fts {
    rebuild_fts_indexes(runtime).await?;
}
```

(`:27-31`). So a caller with ordinary access to `knowledge.index` can pass `ids=[]` and
`rebuild_fts=true` and rebuild the entire corpus, holding the writer transaction for the duration.
Cost scales with the whole store rather than the caller's namespace, concurrent writers queue behind
it, and on non-WAL configurations readers can block or take a busy error. There is no operator-only
capability, size guard, rate limit, or confirmation on the path. That is a materially wider blast
radius than the namespace-scoped embedding work sitting in the same handler, and the verb's
advertised contract does not carry the distinction. Either gate global repair behind the operator
reindex capability, or scope it to the caller's namespace with a separate privileged global mode.

**The operator flags do not mean what the operations guide says.** `docs/operations.md:462-465`
describes `--sections-only` as running only sections and `--no-sections` as skipping them, but
`crates/kkernel/src/reindex.rs:719-731` sets `rebuild_fts: true` whenever a knowledge pass runs, and
the shared handler always rebuilds both indexes. Those flags govern embedding work only. An operator
reaching for `--sections-only` specifically to bound the work will still incur a full atom-and-section
FTS rebuild — which is exactly the situation where someone chose the narrow flag to stay safe.

**ADR-047 points at the wrong migration, and I checked it is not just squashed history.**
`docs/adr/ADR-047-knowledge-pack.md:152-166` states the corpus tier "introduces two tables via
versioned migration V19 (`knowledge_atoms_and_domains`)". In the current tree V1 is the complete
baseline (`migrations.rs:107`, "all schema in one migration"), `knowledge_atoms` is created at
`crates/khive-db/sql/schema.sql:135` and `fts_knowledge` at `:186`, and the V19 slot is occupied by
an unrelated migration, `019-list-cursor-backfill-repair.sql`. So the reference is not merely stale:
the number resolves to a real but different migration, which is worse, because an operator diagnosing
which databases need V23 will land somewhere plausible and wrong. Correct it to name V1 as the
baseline, or mark the passage historical and link the current ledger.

One more, lower stakes but worth deciding deliberately rather than by default: V23 rebuilds a drifted
store and a clean store into the same state and reports the same success, and the runtime helper
returns only `fts_rebuilt: true`. Nothing records that a given store *was* desynchronized. Since the
consequence of the original defect was wrong search results over some window, an operator who wants
to know whether their deployment was affected has no way to find out after the repair runs. A
repaired-row count or a drift flag persisted at migration time would preserve that.

On tests: the V23 migration test is good — it builds a V20 database, hand-constructs the V22 state,
deliberately desynchronizes both indexes, applies V23, and exercises soft-delete, hard-delete,
resurrection and a live content update afterwards, so a regression in the rebuild, the view
definition, or either guard would redden it. What is missing is idempotence (V23 applied twice), a
fresh-versus-upgraded schema equivalence assertion, and the empty-index and missing-index arms of the
integrity check. Four of the transitions above are not independently exercised post-V23
(absent to live, embedding-only section update, live-to-live section content, live to hard-delete),
and the three `rebuild_fts: false` additions in `tests/fixes.rs` only keep existing struct literals
compiling — they add no runtime coverage. Given the finding above, a scope and authorization test for
the public rebuild entry point belongs on that list too.

ohdearquant:

`Docs lint` fails at `Check markdown formatting`: `docs/guide/api-reference.md` are not `deno fmt` clean.
wrapping, no content change.

ohdearquant:
## The FTS rebuild is global, and it is reachable from the ordinary public verb

The rebuild helper takes the runtime and nothing else:

```rust
// crates/khive-pack-knowledge/src/knowledge/index_handler.rs:239-247
async fn rebuild_fts_indexes(runtime: &KhiveRuntime) -> Result<(), RuntimeError> {
    let mut writer = runtime.sql().writer().await...;
    writer.execute_batch(vec![
        SqlStatement { sql: "INSERT INTO fts_knowledge(fts_knowledge) VALUES('rebuild')".into(), ... },
```

No caller token, no namespace. It acquires one writer and rebuilds the complete indexes.

The reach is what makes it worth raising: `rebuild_fts` is accepted by the ordinary public
`knowledge.index` verb (`index_handler.rs:24-31`, advertised at `vocab.rs:205-208`) and dispatched
with user parameters at `pack.rs:241-243`, alongside the operator path at
`kkernel/src/reindex.rs:719-733`. On `origin/main` neither the option nor the helper exists — the
base options are `{ atoms, sections }` only — so both the global capability and its public exposure
arrive together in this change.

`execute_batch` being atomic protects against partial index state, which is good, but atomicity is
not a bound: the writer stays occupied for the whole corpus rebuild regardless. So any caller
admitted to `knowledge.index` can occupy the global writer for a corpus-sized amount of time, and
the cost does not scale with anything about that caller's own namespace.

To be clear about scope, this is not a claim that the runtime gate is bypassed — the gate runs
before pack dispatch. It is about what an *admitted* call is now able to do.

**Suggested direction:** keep ordinary `knowledge.index` namespace-scoped and move the global repair
to an explicitly privileged maintenance verb. If it stays on the public verb, it wants an explicit
authorization check plus a resource or lock policy, and the API contract should say plainly that
this option is global rather than namespace-scoped.

## `--sections-only` and `--no-sections` no longer bound the reindex

Those flags previously bounded what the reindex touched. With the FTS rebuild running unconditionally
alongside them, a caller asking for a deliberately narrow reindex still pays for the global rebuild,
so the flags no longer mean what their names say. Worth either gating the rebuild behind the same
scoping or renaming the expectation in the help text.

## The changed API reference fails the docs-lint gate

`Docs lint` is failing at this head on the changed reference page. Mechanical, but it blocks the
gate.

Separately and **not** blocking: ADR-047 still attributes a migration to V19 incorrectly. That is
pre-existing rather than introduced here, and worth a follow-up rather than a fix in this change.

authored it, so its output was treated as input evidence only. The helper signature, the public verb
acceptance path, and the base's option set were each re-derived from the refs before posting.

ohdearquant (commented):

Verdict on head f387dd79d1cc9a630027864c38e28007321da77d: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 66ba95fbc1195e037139da3b82420ff24931ea44: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head b15200e6d5a66c7a25a5f9ebf8011caf4dad0731: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 22db13e901c105ce9027bee824925b273950d06b: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 146757b15bd7687db801a64f338a812693c55d63: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approved at 146757b1.

PR #2314, opened 2026-08-30
fix(runtime): freeze redaction surfaces mask-only

### LINKED ISSUE #2059: Git redaction surface: define stored-target contract or permanent mask-only status
ADR-115 Amendment 1 §3 named obligation. Define whether the git redaction surface remains permanently mask-only or gains a final stored target with the same stamp and atomic success-event guarantees as the runtime finalization boundary. No exemption is allowed on this surface until this is specified and implemented.

### LINKED ISSUE #2060: Session redaction surface: define stored-target contract or permanent mask-only status
ADR-115 Amendment 1 §3 named obligation. Define whether the session redaction surface remains permanently mask-only or gains a final stored target with the same stamp and atomic success-event guarantees as the runtime finalization boundary. No exemption is allowed on this surface until this is specified and implemented.

### LINKED ISSUE #2061: MCP redaction surface: specify final stored target, stamp location, and atomic success-event boundary
ADR-115 Amendment 1 §3 named obligation. Preserve current masking until a final stored target, stamp location, and atomic success-event boundary are specified and implemented for the MCP redaction surface. No exemption is allowed on this surface until then.

ohdearquant:
## Summary

- Declare Git ingest, session mirror, and MCP diagnostics as closed, permanent mask-only redaction surfaces.
- Add a typed `RedactionSurface` contract and route every named call site through the canonical masker without any manifest, stamp, or exemption-event capability.
- Resolve ADR-115 obligations #2059/#2060/#2061 with explicit final-target, stamp-location, atomic-event, and readback contracts.
- Document the contract in runtime, Git, session, MCP, and API references.

## Verification

- RED: `cargo test --manifest-path crates/Cargo.toml -p khive-runtime --lib named_redaction_surfaces -- --nocapture` failed to compile before the typed contract existed.
- GREEN: runtime library — 1,417 passed, 7 ignored.
- Session parser — 34 passed.
- Git ingest-focused library tests — 41 passed.
- MCP backend diagnostic masking regression — 1 passed.
- `cargo check --manifest-path crates/Cargo.toml -p khive-pack-session -p khive-pack-git -p khive-mcp --all-targets`
- `cargo clippy --manifest-path crates/Cargo.toml -p khive-runtime -p khive-pack-session -p khive-pack-git -p khive-mcp --all-targets -- -D warnings`
- `git diff --check`

## Contract

Git and session persist only masked projections and never write `khive:secret_gate` or an exemption-success event. MCP diagnostics have no durable stored target. Caller identity, verb, namespace, path, and request arguments cannot select another mode.

Closes #2059
Closes #2060
Closes #2061

## Bounded masking trade

- On truncated input the trailing fragment-drop walk runs regardless of trigger context inside the window: the trigger word that would decide it may lie past the boundary, and the walk cannot see there. A fragment-shaped token chain at the tail of any truncated window is dropped, trigger or not; the cost is at most `MAX_BRIDGE_FRAGMENTS - 1` tokens of an already-truncated tail, and `truncated` is reported. The unbounded surfaces are unaffected.


ohdearquant:
The typed contract is a real improvement over three independent direct calls, and the ADR amendment is specific enough to check against. Two ordering defects block it, and the census turns up a surface the set does not name.

## Masking runs after truncation on two of the three surfaces

The canonical URL-userinfo detector only fires when the terminating `@` is inside the text it is handed. `find_url_userinfo` (`crates/khive-runtime/src/secret_gate.rs:845-861`) locates `://`, bounds the authority at the first `/ ? # space newline`, and then looks for `@` *within that bound*. No `@` in scope, no detection — by construction, not by accident.

Two of the three named surfaces cut the input before that call.

**MCP diagnostics.** `bounded_backend_error_message` (`crates/khive-mcp/src/server.rs:212-229`) does:

```rust
let bounded_input: String = message.chars().take(MAX_BACKEND_ERROR_INPUT_CHARS).collect();
let masked = mask_mcp_diagnostic(&bounded_input);
```

`MAX_BACKEND_ERROR_INPUT_CHARS` is 4096 (`:53`), and the output bound is 1024 (`:52`). So for input shaped `https://user:<4096+ chars of password>@host`, the `@` is cut before masking, nothing matches, and the first 1024 characters — which contain the scheme, the username and the start of the password — are returned on the wire. `bounded_backend_error_key` (`:232-258`) has the same ordering.

**Session tool previews.** `extract_block` (`crates/khive-pack-session/src/mirror/parse.rs:642-657`) truncates to 500 characters while building the preview:

```rust
let input_str = truncate(&serde_json::to_string(&input).unwrap_or_default(), 500);
```

and the enclosing parser masks afterwards (`:153-154`, `:264-269`). A credential whose `@` falls past character 500 is invisible to the detector, and the `text` projection can retain the raw prefix while the full `raw` event is masked — the two projections disagree about the same event.

What makes this actionable rather than a design argument is that **the repository already contains the fix, in the same file as the detector.** `bounded_masked_log_text` (`crates/khive-runtime/src/secret_gate.rs:462-482`) tracks whether it truncated and compensates:

```rust
let masked = mask_secrets(&bounded_input);
let masked = if mask_input_truncated {
    redact_crossing_boundary_url_userinfo(&masked)
} else { masked };
```

The two surfaces above take the same shape of input bound without the corresponding fallback. Either mask the complete input before bounding, or route these through a boundary-aware helper equivalent to the one above. Both need a regression whose credential terminator sits beyond the input bound — a test that keeps the whole URL inside the bound passes either way and proves nothing.

This also means ADR-115's obligations for the session and MCP surfaces are not delivered as the amendment now states them. The amendment marks both resolved and declares the surfaces permanently mask-only; for a boundary-shaped input neither guarantee holds today.

## "Closed" is a convention here, not an invariant

`RedactionSurface` is a finite enum (`:280-289`) and the exhaustive match at `:312-325` will fail to compile if a variant is added without a contract mapping. That is the whole of the compile-time closure, and it protects the *mapping*, not the *routing*. `mask_secrets` is still public at `:358`, `mask_for_redaction_surface` is public, and `RedactionSurfaceContract` has public fields (`:299-309`), so a new call site can redact without touching the enum and nothing fails: not the compiler, not a test, not a census.

Worth saying plainly in the ADR which of the two is being claimed. As written the text reads as an invariant; the code provides a convention.

## The set is missing a surface

Enumerating redaction call sites independently rather than starting from the three: `crates/kkernel/src/coordinator/dispatch.rs:22-64` defines `bounded_backend_id_for_log` and `bounded_backend_cause_for_log`, which call the canonical masker directly and feed warning sites at `:602-604`, `:685-688`, `:864-868`, `:982-1018`. That is a fourth coordinator-diagnostic display surface, outside the enum and outside the amendment. The bounded-log path (`secret_gate.rs:462-482`, used from `runtime.rs:1066-1069`, `:1150-1153`, `:1179-1182` and `pack.rs:1655-1658`, `:1944-1947`) is a fifth.

Some of these may be deliberate exclusions — the coordinator comments treat the raw backend result as pre-wire and the VCS and repo-source paths at `crates/khive-vcs/src/sync.rs:233-238` and `crates/khive-pack-git/src/source.rs:183` answer to different policies. The problem is that the exclusions are not stated anywhere, so a reader cannot tell an omission from a decision. A set claim needs its complement written down.

## Tests

The two added tests (`secret_gate.rs:4478-4514`) check that each variant maps to mask-only with no stamp or event, and that the wrapper masks a sample key. Since all three variants currently resolve to the same implementation, a mutation that reverts any named call site to a direct `mask_secrets` leaves both green — the tests cover the mapping, and the mapping is not what this change is about. Nothing covers routing, a redaction call outside the contract, or either truncation boundary.

Suggested additions: the two boundary regressions above; a test asserting the session `text` and `raw` projections agree for a boundary-shaped input; and, if the routing claim is to hold, something mechanical — a restricted module boundary, or a census test over the tree — rather than a convention.

ohdearquant:
Requesting changes. The mechanism here is good and worth saying so explicitly, because it is
stronger than the usual version of this: `redaction_surface_contract` is an exhaustive `const fn`
match over `RedactionSurface`, and `mask_for_redaction_surface` is an exhaustive match over
`RedactionSurfaceMode`. Adding a variant to either is a compile error, not a test that quietly
keeps passing. A declaration enforced by the type checker is a different class of thing from one
enforced by a list a test iterates.

## The closed set is closed to code that opts in, and two call sites have not

The summary says every named call site is routed through the canonical masker and the surface
set is closed. The enum is closed. Entry into it is voluntary, because `mask_secrets` remains
`pub` and can still be called directly — and two callers do:

```
crates/kkernel/src/coordinator/dispatch.rs:32   khive_runtime::secret_gate::mask_secrets(&bounded_input)
crates/kkernel/src/coordinator/dispatch.rs:64   khive_runtime::secret_gate::mask_secrets(&bounded_input)
```

against the three that were routed:

```
crates/khive-mcp/src/server.rs:206              mask_for_redaction_surface(...)
crates/khive-pack-git/src/ingest.rs:25          mask_for_redaction_surface(RedactionSurface::GitIngest, ...)
crates/khive-pack-session/src/mirror/parse.rs:13 mask_for_redaction_surface(RedactionSurface::SessionMirror, ...)
```

This is not a leak — both call sites do mask, with the canonical detector, which is the property
that actually matters. The problem is the claim. Those two functions are
`bounded_backend_id_for_log` and `bounded_backend_cause_for_log`, and their own doc comments
describe them as *the same surface this PR is naming*:

> the MCP boundary applies the same canonical secret masker before exposing it on the wire

> This mirrors the MCP wire boundary so the earlier coordinator diagnostic cannot leak a
> credential that the response would later redact.

So a fourth redact-not-block surface, self-described as mirroring `McpDiagnostic`, exists in the
tree today and sits outside the contract that declares the set closed. The compile-time force
only reaches code that already decided to use the enum, so a fifth surface can be added tomorrow
without ever touching `RedactionSurface` — which is the same shape as declaring a set of writers
in a list, arriving through visibility instead of through a test.

Two ways out, either is fine. Name them — a `CoordinatorDiagnostic` variant, which the exhaustive
match then defends for free. Or make entry non-optional by narrowing `mask_secrets` visibility so
the contract wrapper is the only route, which is the stronger version because it makes the
closure claim true rather than asserted.

## Cross-cutting: this and #2341 both add a section named "Amendment 2"

Not visible from inside either change. `ADR-115` on the base carries `## Amendment 1
(2026-08-19)` with subsections 1 through 5. This PR adds:

```
## Amendment 2 (2026-08-29): Permanent mask-only redaction surfaces
```

and #2341 adds:

```
### 6. Amendment 2: write-inventory, knowledge, and stamp-echo completion
```

Different content, same name. They do not textually conflict — `git merge-tree` on the two heads
returns clean — so whichever order they land in, the document ends up with a top-level Amendment
2 and, nested inside Amendment 1, a subsection also calling itself Amendment 2. A clean
auto-merge means the edits were far apart, not that they agree.

This PR has the correct form: top level, dated, matching Amendment 1's shape. The renumbering is
the other change's to do, and it is already on that thread; flagging it here so the collision is
not discovered after both are in.

ohdearquant (commented):

Verdict on head a25ac2d15f6c88143a9f10864fe38d0bf57895fc: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

ohdearquant (commented):

Verdict on head 29291754b3a650eb6e66ef3e55efc9127f00e38f: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 3e5113c91f1b399197b62415de76d8d793f4a305: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 6afb3f735b8ba0561d6f3b3cf8ae54ba6d1c72c5: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2315, opened 2026-08-30
test(timing): derive latency bounds from fixtures

### LINKED ISSUE #2310: test: wall-clock latency assertions redden unrelated changes and hide co-shard failures
## What happened

`khive-db::writer_timeout_sink_stalled_writer::sink_never_adds_measurable_latency_when_its_writer_is_genuinely_slow` failed on CI against a documentation-only change (a single `.md` file, no Rust touched):

```
pool construction took 511.665366ms against a 5000ms-per-write sink
  — the sink must never add filesystem-bound latency to pool boot
thread ... panicked at khive-db/tests/writer_timeout_sink_stalled_writer.rs:58:5
```

The threshold is 500ms. The measurement was 511ms. Nothing in the change under test could affect it.

## Why this shape fails

The property being protected is real and worth a test: pool construction must not block on the sink's writer thread, so a writer that takes 5000ms per write must not slow `ConnectionPool::new`.

But look at the numbers the assertion has to separate:

- construction **does not** block on the writer → a few milliseconds
- construction **does** block on the writer → at least 5000ms

That is a 5000ms-wide gap, and the threshold was placed at 500ms — roughly 2% above the noise floor of a busy shared runner, and 10× below the failure it is meant to detect. All of the margin was spent on the side where nothing is being discriminated. A runner under load crosses 500ms routinely; it will never accidentally cross 5000ms.

So the test is not measuring the code. It is measuring how busy the runner was, and the failure it produces is indistinguishable from the real regression.

## Amplifier: one timing flake hides most of the shard

`nextest` cancels the run on first failure. In this run:

```
Summary [8.296s] 888/4757 tests run: 887 passed, 1 failed
warning: 3869/4757 tests were not run due to test failure
```

So a wall-clock flake does not merely produce a spurious red — it stops 81% of that shard from executing. Any genuine regression landing in the same window is invisible behind it. That is what makes this worth fixing as its own change rather than tolerating.

## The class, enumerated

This is not a single site. Wall-clock comparisons against a `Duration` bound, workspace-wide:

| Location | Bound |
| --- | --- |
| `khive-pack-knowledge/src/knowledge/vamana.rs:3797` | `< 200ms` |
| `khive-pack-knowledge/src/knowledge/vamana.rs:5765` | `< 200ms` |
| `khive-db/tests/writer_timeout_sink_stalled_writer.rs:82` | `< 250ms` |
| `khive-db/tests/writer_timeout_sink_stalled.rs:131` | `< 250ms` |
| `khive-db/tests/writer_timeout_sink_stalled_fifo.rs:100` | `< 250ms` |
| `khive-db/tests/writer_timeout_sink_stalled_writer.rs:59` | `< 500ms` ← the one that fired |
| `khive-db/tests/writer_timeout_sink_stalled_fifo.rs:83` | `< 500ms` |
| `khive-db/src/checkpoint.rs:4522` | `< 500ms` |
| `khive-pack-memory/src/handlers/recall.rs:1450` | `< 3s` |
| `khive-pack-memory/src/handlers/recall.rs:5783` | `< 5s` |
| `khive-runtime/src/operations.rs:15303` | `< 5s` |
| `kkernel/src/coordinator/tests.rs:697` | `>= 5s && < 6s` |

The 200ms and 250ms bounds carry the least headroom and should be expected to fire next. The last one is a two-sided 1-second band around a 5-second operation — its lower bound (`>= 5s`, asserting something took *at least* a duration) is the robust direction and is not part of this problem; its 6s upper bound is.

## Proposed rule

**A wall-clock bound in a CI-run test must be derived from the hypothesis it discriminates, not from the expected value.** Place the threshold in the middle of the gap between "correct" and "the defect", and express it in terms of the fixture's own constants so it cannot drift.

For the sink family, `WRITE_DELAY_MS` is already in scope, so:

```rust
// blocking on the writer would cost >= WRITE_DELAY_MS; anything well below
// that proves the caller path never touched it
assert!(construct_elapsed < Duration::from_millis(WRITE_DELAY_MS / 2), ...);
```

This keeps the full discriminating power — a genuine regression costs at least 5000ms and still fails loudly — while raising the flake margin from ~2% to 5×. The same substitution applies to the `250ms` checkout assertions, whose stated claim is "not anywhere near `WRITE_DELAY_MS`".

Where a bound cannot be tied to a fixture constant, the alternatives in order of preference are: assert a logical condition instead (did the operation observably complete before the slow work could have finished), or log the duration and drop the assertion. A timing number that is recorded but not asserted still supports investigation without voting on the build.

## Acceptance

- Each converted assertion still fails when the property it protects is actually broken — demonstrate by making the caller path block and showing the assertion fires.
- No bound is a bare literal where a fixture constant expresses the same value.
- The `>= 5s` lower-bound assertion is preserved; only upper bounds near the expected value are in scope.

ohdearquant:
## Summary

Several test assertions bounded latency with a wall-clock literal that had no derivation from the
fixture it was meant to discriminate. A bound like `elapsed < 500ms` guarding a claim about a
5,000ms-per-write sink has ten times more headroom than the hypothesis needs, and its failure margin
is host scheduling noise rather than the behaviour under test — so it reddens changes that never
touched the code path.

That happened: pool construction measured 511.665366ms against the 500ms bound while the claim it
protects concerns a 5,000ms write, on a change that touched only documentation. Because the suite
runs sharded with fail-fast, one such failure cancelled its siblings and left most of the suite
unrun, so the cost is much larger than one red test.

## Changes

Each bound is now derived from the fixture constant that makes it discriminating:

- `crates/khive-db/tests/writer_timeout_sink_stalled_writer.rs:59,82` — both bounds become
  `Duration::from_millis(WRITE_DELAY_MS / 2)`. The 2,500ms midpoint sits far below the 5,000ms write
  the assertion exists to rule out, while leaving scheduler noise no way to reach it.
- `crates/khive-db/src/checkpoint.rs:4524` — bound becomes half of
  `checkpoint_config.truncate_busy_timeout`, the midpoint between normal PASSIVE completion and the
  TRUNCATE regression it separates.
- `crates/khive-pack-knowledge/src/knowledge/vamana.rs:3797,5765` — terminal-unavailable
  short-circuit bounds become `ANN_WARM_WAIT_TIMEOUT_MS / 2`, with the same constant passed as the
  wait timeout so the two cannot drift apart.

Two assertions are removed rather than rebased, because they duplicated a guard already present:

- `crates/khive-db/tests/writer_timeout_sink_stalled_fifo.rs:78,85` — these calls already run through
  `bounded()`, which waits on a channel with `recv_timeout(BOUND_TIMEOUT)` and fails the test if the
  operation does not complete. A blocked caller path therefore still fails; the wall-clock ceilings
  added a second, noisier way to say the same thing. The `timed_out` assertions remain.
- `crates/khive-db/tests/writer_timeout_sink_stalled.rs:124` — the elapsed ceiling is replaced by a
  `sync_channel` plus `recv_timeout(HANG_GUARD_TIMEOUT)`. This fixture models a path with no finite
  write duration, so completion is the discriminating observation and a duration ceiling was never
  measuring the right thing.

Four sites are deliberately left alone: two recall paths and one runtime operation whose ceilings are
broad secondary guards around assertions that already state the real condition, and the coordinator
test, which runs under `start_paused = true` and therefore measures deterministic virtual time rather
than host latency. Its lower `>= 5s` bound is load-bearing and unchanged.

## Verification

The two rewritten bounds were shown to still fail when they should. Temporarily making `emit_timeout`
fill the bounded queue synchronously forces the caller to wait behind the delayed writer:

```text
pool construction took 5.016870167s against a 5000ms-per-write sink — the sink must never add
filesystem-bound latency to pool boot

checkout_timeout was 50ms but writer() took 5.015753834s against a 5000ms-per-write sink —
emit_timeout must be a non-blocking enqueue, never blocking on the writer thread's own I/O latency
```

Both exit 101. With the mutation reverted the test passes in 0.07s. A bound that cannot be shown to
fail is not evidence that it still guards anything, which is the whole point of widening these.

Closes #2310

### DIFF
--- a/crates/khive-db/src/checkpoint.rs
+++ b/crates/khive-db/src/checkpoint.rs
@@ -4458,9 +4458,9 @@ mod tests {
         let dir = tempfile::tempdir().unwrap();
         let path = dir.path().join("high_water_test.db");
 
-        // busy_timeout = 2000ms: a TRUNCATE regression blocks ~2s (clearly caught by
-        // the <500ms assertion below), but PASSIVE returns well within 500ms even on
-        // a heavily loaded CI runner. 4x margin on both sides vs. the old 200ms/50ms.
+        // busy_timeout = 2000ms: a TRUNCATE regression blocks ~2s, while the
+        // assertion below uses the midpoint of that gap instead of a noise-floor
+        // threshold. The checkpoint config carries the same 2000ms default.
         let pool = Arc::new(
             ConnectionPool::new(PoolConfig {
                 path: Some(path.clone()),
@@ -4500,12 +4500,13 @@ mod tests {
                 .unwrap();
         }
 
+        let checkpoint_config = CheckpointConfig::default();
         let conn = checkpoint_conn(&pool);
         let start = std::time::Instant::now();
         checkpoint_once(
             &pool,
             &conn,
-            &CheckpointConfig::default(),
+            &checkpoint_config,
             &mut TruncateState::default(),
         )
         .expect("checkpoint_once must succeed against a healthy dedicated connection");
@@ -4515,15 +4516,18 @@ mod tests {
         reader.execute_batch("COMMIT;").ok();
         drop(reader);
 
-        // PASSIVE returns in <1ms even with an open reader snapshot.
-        // A TRUNCATE regression would block ~busy_timeout (2000ms) and fail here.
-        // 500ms threshold is generous for CI jitter while staying well below 2000ms.
+        // PASSIVE returns without waiting for the reader snapshot. A TRUNCATE
+        // regression would block for the configured busy timeout (2000ms), so
+        // use the midpoint of that gap rather than a noise-floor threshold.
+        let max_elapsed = checkpoint_config.truncate_busy_timeout / 2;
         assert!(
-            elapsed < std::time::Duration::from_millis(500),
-            "checkpoint_once with active reader snapshot took {:?}; \
-             expected <500ms (PASSIVE must not block on readers; \
-             a TRUNCATE regression would block ~2000ms)",
-            elapsed
+            elapsed < max_elapsed,
+            "checkpoint_once with active reader snapshot took {:?}; expected <{:?} \
+             (PASSIVE must not block on readers; a TRUNCATE regression would block \
+             for the configured {:?})",
+            elapsed,
+            max_elapsed,
+            checkpoint_config.truncate_busy_timeout
         );
     }
 
--- a/crates/khive-db/tests/writer_timeout_sink_stalled.rs
+++ b/crates/khive-db/tests/writer_timeout_sink_stalled.rs
@@ -116,20 +116,17 @@ fn sink_never_adds_measurable_latency_when_its_directory_is_unwritable() {
     // write anywhere in this test.
     let held = pool.writer().expect("first checkout should succeed");
     let pool_for_thread = Arc::clone(&pool);
-    let start = Instant::now();
-    let timed_out = std::thread::spawn(move || pool_for_thread.writer().is_err())
-        .join()
-        .unwrap();
-    let elapsed = start.elapsed();
+    let (timed_out_tx, timed_out_rx) = std::sync::mpsc::sync_channel(1);
+    std::thread::spawn(move || {
+        let _ = timed_out_tx.send(pool_for_thread.writer().is_err());
+    });
+    let timed_out = timed_out_rx
+        .recv_timeout(HANG_GUARD_TIMEOUT)
+        .expect("writer admission blocked on the unwritable sink directory");
     drop(held);
 
     assert!(
         timed_out,
         "a second writer checkout while the first is held must time out"
-    );
-    assert!(
-        elapsed < Duration::from_millis(250),
-        "checkout_timeout was 50ms but writer() took {elapsed:?} against an unwritable sink \
-         directory — emit_timeout must be a non-blocking enqueue, never blocking I/O"
     );
 }
--- a/crates/khive-db/tests/writer_timeout_sink_stalled_fifo.rs
+++ b/crates/khive-db/tests/writer_timeout_sink_stalled_fifo.rs
@@ -18,7 +18,7 @@
 use std::process::Command;
 use std::sync::mpsc;
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use khive_db::{ConnectionPool, PoolConfig};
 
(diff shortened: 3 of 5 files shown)


ohdearquant:
The core idea here is right, and three of the five files apply it well. Replacing a noise-floor
threshold with a midpoint derived from the thing that would actually regress turns a flaky budget
assertion into a robust two-outcome discriminator, which is what these tests were always trying to be.

Where it applies cleanly:

- `crates/khive-db/src/checkpoint.rs` — `checkpoint_config.truncate_busy_timeout / 2` sits at 1000ms
  between "PASSIVE returns immediately" and "a TRUNCATE regression blocks ~2000ms". Strictly better
  than the old 500ms: same catch, more headroom against CI jitter, and the bound now moves with the
  config instead of drifting away from it.
- `crates/khive-db/tests/writer_timeout_sink_stalled_writer.rs` — `WRITE_DELAY_MS / 2` ties the bound
  to the fixture's own injected delay. This is the pattern the title promises.
- `crates/khive-pack-knowledge/src/knowledge/vamana.rs` — the `5_000` literal now reads
  `ANN_WARM_WAIT_TIMEOUT_MS`, which I confirmed is the production constant (`vamana.rs:615`), so the
  test can no longer silently disagree with production.

## Two tests had their latency assertions deleted rather than derived

`crates/khive-db/tests/writer_timeout_sink_stalled.rs` drops
`assert!(elapsed < Duration::from_millis(250), …)` and replaces the thread join with
`recv_timeout(HANG_GUARD_TIMEOUT)`, where `HANG_GUARD_TIMEOUT` is 10 seconds (`:16`).

That is not the same assertion loosened — it is a different assertion. The old one measured latency
against a fixture whose `checkout_timeout` is 50ms (`:92`). The new one only establishes that the
call returns within ten seconds. The discrimination window went from 250ms to 10s, so the regression
the test exists to catch — blocking sink I/O adding latency to writer admission — now has to block
for over ten seconds to be noticed. A sink that blocks for two seconds, which is the exact class the
sibling `…_stalled_writer` test models with `WRITE_DELAY_MS`, passes silently here.

`crates/khive-db/tests/writer_timeout_sink_stalled_fifo.rs` goes further: both the pool-construction
and the checkout latency assertions are removed, and the file now contains no `Instant` or `elapsed`
reference at all. What remains asserts that the second checkout timed out and that the FIFO was not
rotated away — both worth keeping, neither about latency.

The part that makes this worth blocking on rather than noting: **both functions are still named
`sink_never_adds_measurable_latency_…`**. A test whose name asserts a property it no longer measures
is worse than a deleted test, because the suite goes on reporting coverage for it and the next person
reading the name has no reason to look. If the intent is that these two no longer check latency, the
names have to say so.

## The fix is the PR's own idea, applied consistently

The deleted bounds were already fixture-derived in spirit — the old failure message literally reads
"checkout_timeout was 50ms but writer() took {elapsed}", so 250ms was `checkout_timeout × 5` written
as a literal. Writing it as `checkout_timeout * 5` keeps the property, removes the magic number, and
matches exactly what the other three files now do. Same for the FIFO construction bound.

The property under test in both cases is binary — non-blocking enqueue versus blocking I/O — which is
the shape the midpoint technique handles best. There is no reason these two needed a different
treatment from `…_stalled_writer.rs`, which kept its assertion and derived the bound.

## One sensitivity note, not a blocker

The `vamana.rs` bounds move from 200ms to `ANN_WARM_WAIT_TIMEOUT_MS / 2` = 2500ms. That still
separates "short-circuits" from "polls out the full 5s timeout", so the stated property holds. But it
is a 12.5× loosening, and a partial regression that short-circuits at, say, 1000ms instead of
immediately would now pass. Fine if the property is deliberately binary — worth a word in the comment
saying so, since the next reader will wonder whether 2500ms was chosen or inherited.

No builds or tests were run for this pass; every finding is from source reading.

ohdearquant (commented):

oceanwaves630 (approved):
Approving at 3976754d. Required checks are green at this head (Secret scan (gitleaks), CI gate); the remaining UNSTABLE signal is a cancelled optional bench job, not a required context. The automated round at this head reported no blocking findings, and I read the diff: latency bounds now derive from each fixture's own configured timeout instead of hand-picked noise-floor constants, which removes the sensitivity that made these tests flake under load. Approving as the second reader on that basis.

Issue #2242, opened 2026-08-27
writer-task BEGIN busy is proven safe to retry and nothing retries it, so a recoverable condition reaches callers as a failed request


ohdearquant:
### Summary

A writer-task `BEGIN` refusal is proven safe to retry, is reported to the caller as
`retryable: true`, and is retried by nothing. There is no bounded retry at the seam where the
safety proof is local, and the wire response deliberately carries no backoff hint, so the only
behaviour available to a caller is an immediate blind retry — or, more commonly, surfacing a
designed recoverable condition to a human as a failed request.

### What the code does today

`StorageError::WriterTaskBusy` is constructed when the writer task's `BEGIN IMMEDIATE` is
refused busy or locked after its full busy timeout (`crates/khive-db/src/writer_task.rs:568`).
`RuntimeError::retryable_failure_context` classifies it as the `writer_task_begin_busy` stage
with `operation: "writer_task_begin"` (`crates/khive-runtime/src/error.rs:620-628`), and it
reaches the wire as `retryable: true`.

The safety proof behind that classification is strong and specific. ADR-067 Amendment 4 states
that the queue accepted the request but the operation closure never ran, so retrying that one
failed operation cannot duplicate a write, and the writer task remains live and serves the next
request normally. `contended_begin_increments_its_own_failure_counter`
(`crates/khive-db/src/writer_task.rs:1390-1476`) demonstrates exactly that: it forces a real
refusal against a held write lock, then sends a second request that succeeds against the same
live task.

So the process knows the operation did not execute, knows a retry is safe, knows the task is
still healthy — and returns the failure to the caller anyway.

### Why the hint is absent, and why that is not the bug

`retry_after_ms` is `None` for this stage (`error.rs:627`) while the adjacent
`WriteQueueFull` admission failure carries `Some(timeout_ms)` (`error.rs:594`). That asymmetry
is deliberate and documented: the ADR-131 admission hint means the queue never accepted the
request, which is the opposite of what happened here, so reusing the field would tell the caller
something false.

The point is not that the hint should be reused. It is that withholding the hint and providing
no retry together leave the caller with nothing: it is told the condition is retryable and given
no basis for when. The safest thing a well-behaved caller can then do is retry immediately,
which is also the thing most likely to hit the same contended lock.

### Proposed direction

Add a bounded retry at the writer-task `BEGIN` seam, where the proof that the closure did not
run is local and does not have to be re-established by every caller. Bounded in attempts and in
[...]

PR #2316, opened 2026-08-30
fix(db): retry contended writer BEGIN

ohdearquant:
## Summary\n\n- retry only busy or locked BEGIN IMMEDIATE refusals, before request closure execution, with three total attempts and fixed 5 ms / 10 ms backoffs\n- keep writer_task_begin_busy as the surfaced-refusal counter and add writer_task_begin_busy_absorbed for refusals followed by another internal attempt\n- amend ADR-067 and operator/API documentation with the retry, latency, exactly-once, and diagnostics contract\n\n## Safety boundary\n\nThe retry loop owns only BEGIN IMMEDIATE. It never replays the FnOnce request closure, DML, COMMIT, ROLLBACK, top-level operations, or ambiguous terminal paths. Exhaustion still returns StorageError::WriterTaskBusy and preserves the existing runtime/MCP contract: stage writer_task_begin_busy, retryable true, operation writer_task_begin, and no admission retry hint.\n\n## Verification\n\n- deterministic real-contention retry-success test: one absorbed refusal, closure executes exactly once, one committed row\n- deterministic sustained-contention test: two absorbed refusals, one surfaced refusal, closure executes zero times\n- cargo test --manifest-path crates/Cargo.toml -p khive-db --lib (839 passed)\n- db_diagnostics wire projection test passed\n- runtime typed-classification test passed\n- MCP writer_task_begin_busy envelope test passed\n- affected cargo check and cargo clippy --all-targets -- -D warnings passed\n- cargo fmt --check and git diff --check passed\n\nCloses #2242


ohdearquant:
The property the whole design rests on holds, and it holds for a checkable reason rather than by convention. One documentation/observability issue is worth settling before merge; one doc line is now false.

## The retry is genuinely upstream of every side effect

At the moment a retry decision is made, the request still owns its `FnOnce` closure and no DML has run. `run_writer_task` dequeues the request, enters the `BEGIN IMMEDIATE` loop (`writer_task.rs:960-980`), and the closure is only transferred after `Ok(())` breaks the loop. So a retried `BEGIN` cannot re-execute caller work, and there is no path where the closure is invoked and then has its transaction fail into a retry. The source comment at the absorbed-counter increment states the same invariant, and the code matches it.

The backoff holds the dedicated writer connection and its `spawn_blocking` thread, but no pool mutex and no open SQL transaction — so a retry does not wait while holding the resource other callers are contending for, which is the failure this pattern usually has.

## The refusal classification is variant-based, which is the right answer

`is_busy_or_locked` (`timeout_sink.rs:1113-1122`) matches on the structure, not the text:

```rust
rusqlite::Error::SqliteFailure(inner, _)
    if matches!(inner.code,
        rusqlite::ErrorCode::DatabaseBusy | rusqlite::ErrorCode::DatabaseLocked)
```

So a permanent error whose message merely contains busy-like text cannot enter the retry loop. That is the specific way this class of change converts a fast failure into a slow one, and it is closed here by construction. Everything else falls through to the existing generic path unchanged.

## The counter contract needs one sentence changed

ADR-067:846-852 says observability "distinguishes every refusal **without changing the existing counter's meaning**," then immediately defines the new behaviour: `writer_task_begin_busy` "continues to count only the final refusal surfaced to the caller."

Both halves are individually defensible and the pairing is what misleads. The counter's **definition** genuinely has not changed — it always meant refusals surfaced to the caller. What changed is that far fewer refusals are now surfaced, because absorbed ones are not. So for an identical workload the counter's **value drops**, and an operator reading "meaning unchanged" will reasonably conclude an existing alert or dashboard keyed on it is unaffected. It is not: a briefly contended write that previously produced 1 now produces 0.

That is a metric-definition change riding along with a behaviour change, and it needs saying in the words an operator will act on. The mitigating fact is worth stating in the same breath, because it makes this a documentation fix rather than a code one: the two counters are genuinely disjoint — a refusal increments `absorbed` when a delay remains and `busy` when the loop exits with a surfaced error, never both — so `absorbed + busy` reconstructs the pre-change total exactly. No information is lost; it is split, and consumers need to be told to sum them.

Suggested: drop "without changing the existing counter's meaning", say plainly that the value of `writer_task_begin_busy` decreases for an unchanged workload because absorbed refusals no longer reach it, and state that the pre-change quantity is `writer_task_begin_busy_absorbed + writer_task_begin_busy`.

Related, and the reason this matters beyond bookkeeping: contention that previously surfaced now partially disappears. If nothing reads the absorbed counter, the system loses its early warning and only becomes visible once contention is bad enough to exhaust all three attempts. Worth confirming something actually consumes `db_diagnostics.writer_contention`.

## `writer-task.md:6` is now false

> "...drains a bounded channel of typed write requests, issuing `BEGIN IMMEDIATE` once per request."

`WRITER_BEGIN_RETRY_DELAYS` is `[Duration; 2]` (`writer_task.rs:65`), so it is up to three attempts per request. Since this is the function-specific technical reference, it is the document a reader consults precisely when reasoning about attempt counts.

## Tests

The classifier unit test (`timeout_sink.rs:1524`) covers the mapping for busy and locked codes, and the success-after-retry and exhaustion tests cover the busy paths. What is missing is the arm that separates this change from a broader one: **no test asserts that a non-retryable `BEGIN` failure is not retried.** A constraint, read-only or auth failure should execute once, take no backoff, and increment only `writer_task_begin_errors`. Today a mutation widening the predicate — matching on the message string, or adding a third error code — leaves the whole suite green, and that mutation is exactly the defect the variant-based classifier is there to prevent.

Also worth adding for item 4: an arithmetic assertion that `absorbed + busy` equals the refusal count for a workload driven through a known number of contended attempts. That is the property the documentation fix above will be asking consumers to rely on, and nothing currently pins it.

ohdearquant:
Two items to add to my earlier comment. Both verified against `6bcac1a`.

## The success-after-retry test can fail intermittently

`wait_for_absorbed_begin_refusals` breaks on `absorbed >= expected`
(`crates/khive-db/src/writer_task.rs:1124-1138`), but the assertions that follow demand exact
values — `absorbed == 1`, and `busy == 0` with the message "a refusal still inside the retry loop
must not look surfaced" (`writer_task.rs:1485-1491`).

Nothing holds the writer still between the wait returning and the snapshot being taken. If the
remaining retries land in that window, the same run observes `absorbed == 2` and `busy == 1` and
fails on an assertion about an intermediate state it was never guaranteed to catch. A `>=` wait
cannot pin an `==` assertion: the wait establishes a lower bound, and the assertion claims an exact
instant, which is a strictly stronger statement than the synchronization provides.

This is the kind of flake that surfaces months later and reads as a real regression. Either
synchronize deterministically so the intermediate state is stable while it is read, or assert the
invariant instead of the instant — `absorbed + busy` is the quantity that is well-defined at every
point here, which is the same identity the counter documentation should be pointing consumers at.

## `WriterTaskBusy` presents a per-attempt timeout as the total window

`crates/khive-storage/src/error.rs:210-213`:

> `"writer task could not begin within {timeout_ms}ms because SQLite remained busy; request was not executed"`

With `WRITER_BEGIN_RETRY_DELAYS = [5ms, 10ms]` (`writer_task.rs:65-66`), an exhausted request makes
three `BEGIN` attempts, so the elapsed window before this error surfaces is up to three busy-timeout
periods plus 15ms of backoff — not `timeout_ms`. An operator sizing a client timeout or an alert
threshold from this sentence would take a number close to a third of the real worst case.

Same shape as the counter item in the earlier comment: the value is still correct for what it names,
while the sentence around it now describes something else. Worth either reporting the actual elapsed
window or wording it explicitly as the per-attempt busy timeout.

ohdearquant:
This holds up. The parts that usually go wrong in a retry change are all handled deliberately,
and two of them are handled better than asked:

- **The retry cannot reach the closure.** The loop body is `conn.execute_batch("BEGIN IMMEDIATE")`
  and nothing else; the request still owns its `FnOnce` at every `continue`. Exactly-once is a
  structural property here, not a discipline.
- **Classification is not duplicated.** The retry predicate is
  `crate::timeout_sink::is_busy_or_locked(&error)`, and the failure path counts by matching the
  value the shared classifier returned rather than re-testing the error. The comment says why —
  a second copy of the busy rule could drift from the error actually replied. That is the right
  instinct and it is the one most often skipped.
- **The counters are genuinely disjoint**, not just described as such. The final refusal breaks
  out of the loop *before* `record_writer_task_begin_busy_absorbed`, so it is counted once, at
  the failure path. A success on attempt two leaves one absorbed and no surfaced refusal, which
  is what the documented meaning says.

Two things I checked that are not problems, noted so they do not get re-raised:
`std::thread::sleep` is correct here because the whole drain body runs inside
`tokio::task::spawn_blocking`, not on a runtime worker. And the worst-case latency is disclosed
accurately in the amendment — "three busy-timeout windows plus 15 ms of explicit backoff" — which
is the number that matters, since the backoff is the small term and the re-entered busy handler
is the large one.

## One observation: the bound is stated per request, but the cost is queue-wide

The amendment bounds a *transaction-wrapped request* to three busy-timeout windows. The drain
loop is strictly serial — `rx.recv().await`, then `spawn_blocking(...).await`, then the next
`recv` — so while one request is retrying, nothing else in the queue is being drained.

That makes the operationally interesting number a head-of-line stall rather than a per-request
latency: with the default `busy_timeout` of 30s, a single sustained-contention request can now
hold the queue for up to roughly three windows instead of one, and every request behind it waits
that out. The change is correct and the per-request bound is stated correctly; it is the
queue-wide consequence that the amendment does not name, and it is the one an operator would
feel first.

Not asking for a code change for it. Worth a sentence in Amendment 5 saying the bound is
per-request and that the drain is serial, so the worst case is inherited by queued work — that
is the kind of thing that is obvious while writing it and expensive to re-derive from a latency
graph six months later.

ohdearquant (commented):

Verdict on head 57e59631bbe76092f77da02d6ecc650c3ae7d7df: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head a12153da610e1a3eee4413ad7f8151ef9bf5dec1: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Approved at 89f26782.

Issue #2243, opened 2026-08-27
every filtered note page runs an unbounded COUNT(*) to answer has_more, so a one-row page costs work proportional to the whole result set


ohdearquant:
### Summary

Every filtered note page pays an unbounded `SELECT COUNT(*)` over the whole matching set before
it returns its bounded page, so a request for one row costs work proportional to the entire
result set. The count exists to populate `has_more`, which needs one bit, not a total.

### Where

`query_notes_filtered` builds `SELECT COUNT(*) FROM notes{where}`
(`crates/khive-db/src/stores/note.rs:1367`) and `query_note_page_snapshot` executes that count
before the limited data query (`crates/khive-db/src/stores/note.rs:527-555`). Any caller taking
a bounded page through this method pays it — `comm.inbox` with a small `limit` is the case that
surfaced it, but nothing about the problem is specific to comm.

### Why it matters

The cost is invisible at small mailbox sizes and grows with the store, which is the shape of
problem that is cheap now and expensive exactly when the system is busiest. It also means the
paging API's cost is a function of total matches rather than page size, so a caller cannot
reduce its cost by asking for less — the usual remedy does not work here.

### Suggested direction

`has_more` can be answered by over-fetching a single row: request `limit + 1`, return `limit`,
and report `has_more` from whether the extra row appeared. That makes page cost proportional to
page size and removes the count entirely.

The complication is that some callers may want the true total rather than the bit. Those should
ask for it explicitly, so the expensive number is requested rather than paid for by default. A
bounded variant that saturates at a cap — the shape used for the unread badge — is the honest
middle option for callers that want a number but can accept "at least N".

### Note on scope

Found while reviewing #2166, which does not touch this code: `query_notes_filtered` and
`query_note_page_snapshot` are byte-identical on `main`. Filing separately so the paging
contract gets its own decision rather than being changed as a side effect of a feature branch.

ohdearquant:
Closing as an exact duplicate of #2214. The same filtered-note COUNT path and paging contract are tracked there.

PR #2318, opened 2026-08-30
perf(storage): add count-free note and graph pagination

### LINKED ISSUE #2204: events-split follow-ups: per-page COUNT, per-read socket dial, duplicated freeze loop

### LINKED ISSUE #2214: storage: filtered note queries always execute an unbounded COUNT(*)
\`query_notes_filtered\` (and the unfiltered \`query_notes\` page path) unconditionally executes \`SELECT COUNT(*)\` over the full filtered candidate set to populate \`Page::total\`, before the LIMIT'd data query (crates/khive-db/src/stores/note.rs, \`query_note_page_snapshot\`). Every filtered list therefore performs work proportional to the matching-row population regardless of the requested page size — including \`comm.inbox\`'s page query and its mailbox-wide \`unread_count\` (#2166 reduced that path from ceil(N/200) counts to one, but one unbounded count per call is the storage-layer floor today).

Design ask, two candidate shapes (not mutually exclusive):

1. **Count-free page projection** — a variant (or a \`PageRequest\` flag) that skips the COUNT and returns \`total: None\`, for callers that only need the page.
2. **Bounded count projection** — \`SELECT COUNT(*) FROM (SELECT 1 FROM notes WHERE <filter> LIMIT :cap)\`, returning a saturating count with an explicit cap, for badge/summary consumers where "cap+" semantics are acceptable.

Either bounds per-request work at the storage layer instead of asking every caller to reason about backlog-proportional cost. Callers to migrate afterwards: the comm inbox/unread paths first.

ohdearquant:
## Summary

- add fail-closed count-free note-page contracts and SQLite implementations for both simple and filtered queries, preserving deterministic ordering while returning `total: null`
- migrate row-only note consumers in runtime list operations plus comm, GTD, and schedule handlers; retain the exact filtered page for `session.list`, whose response genuinely exposes a snapshot-consistent total
- remove implicit `COUNT(*)` work from single- and multi-namespace graph pages; explicit `count_edges*` APIs remain exact for callers that request counts
- migrate `comm.inbox` scans to count-free overfetch pagination while keeping the addressed-plus-legacy recipient partition on one snapshot
- expose bounded unread projection metadata (`unread_count_cap` / `unread_count_saturated` and `count_cap` / `count_saturated`)

## #2204 disposition

This PR completes the remaining live page-count scope after re-deriving the current-main caller census:

- event page count removal landed separately in #2268
- graph page totals had no consumer, but the page methods do have production callers in `KhiveRuntime::list_edges` and portability export; both discarded the total, so these paths now execute row SQL only
- note page totals are preserved only for `session.list`; every production row-only note caller now uses a count-free method
- the socket-pool item was measured and deliberately declined in the issue discussion
- the freeze/code-path item landed separately in #2266

The graph correction is intentionally based on current-main call sites rather than the older issue comment that described those page methods as having no production callers.

## Contract refinement

This deliberately refines the exact unread-count contract introduced by #2166 in response to the unbounded-work problem reported in #2214 (and its closed duplicate #2243):

- counts below the cap remain exact
- a count equal to the cap remains exact and reports `*_saturated: false`
- a count above the cap returns the cap (`1000`) and reports `*_saturated: true`
- the addressed and legacy-recipient partitions are still counted in one SQLite snapshot, then combined with saturating arithmetic

The existing exact note-page storage APIs remain available for callers that genuinely require a total. Count-free trait defaults fail closed rather than silently falling back to exact pages.

## Verification

- `cargo test --workspace`
- focused regressions deny SQLite `COUNT` and prove filtered/unfiltered note pages plus single-/multi-namespace graph pages still succeed
- focused regressions cover concurrent note-page snapshots, deterministic ordering, below-cap exactness, above-cap saturation, and addressed/legacy partition snapshot consistency

Closes #2204
Closes #2214


ohdearquant:
## `unread_count` changes meaning above 1,000 without changing its name

`crates/khive-pack-comm/src/handlers.rs:940-944` clamps the published value with
`count: observed.min(cap)`, where `UNREAD_COUNT_CAP = 1_000` (`handlers.rs:864`). The value is
still published under the pre-existing key `unread_count` at `handlers.rs:571` and
`handlers.rs:834`, now beside two new siblings `unread_count_cap` and `unread_count_saturated`.

A consumer that reads only `unread_count` — which is every consumer written before this change —
previously received 5,000 for a 5,000-message mailbox and now receives 1,000, with nothing in the
field it actually reads to indicate the value was clamped. A field that keeps its name and type
while changing its meaning above a threshold is a breaking change for everyone who does not learn
about the new flag.

This is the one item I would not merge as-is. Any of these resolves it: keep the exact value under
the existing key and publish the bounded one under a new name; gate the bounded semantics behind an
opt-in; or state explicitly that the break is intended and accepted.

## Verified and sound

The caller census was re-derived per store rather than carried from one to the other, since the
populations differ:

- `query_edges` and `query_edges_in_namespaces` consumers (runtime `list_edges`, its per-namespace
  fallback, portability export) read only `.items`.
- Count-free note callers in `khive-runtime`, `khive-pack-comm`, `khive-pack-gtd` and
  `khive-pack-schedule` likewise read only `.items`.
- `session.list` remains the only production consumer of a total, and still uses the exact path
  (`crates/khive-pack-session/src/handlers/list.rs:89-101` calls `query_notes_filtered` and reads
  `page.total`).
- Count-free note defaults fail closed with `Unsupported` rather than delegating to the exact path
  (`crates/khive-storage/src/note.rs:704-712`); `SqlNoteStore` overrides them and
  `PolicyEnforcingNoteStore` forwards.
- Ordering is deterministic with UUID tie-breaks (`crates/khive-db/src/stores/note.rs:719-738`,
  `crates/khive-db/src/stores/graph.rs:904-929`), so a row at a page boundary is neither skipped nor
  duplicated absent concurrent mutation.
- Both unread partitions are counted in one storage snapshot using cap+1 probes
  (`handlers.rs:914-944`).

## Non-blocking notes

- Graph has no separate count-free method. The singleton-namespace path delegates straight to
  `query_edges` (`crates/khive-storage/src/graph.rs:136-149`), so a downstream graph backend can
  still compute totals unless it adopts the convention by hand.
- The merged fold at `crates/khive-runtime/src/events_split.rs:1829-1832` is
  `(Some(a), Some(b)) => Some(a + b), _ => None`, so one count-free side erases the merged total.
  Both current SQL event stores already return `None`, which makes this adjacent pre-existing
  behaviour rather than a regression introduced here — but the decision to skip a count is being
  made per store, where caller intent is not known, rather than where it is.
- `crates/khive-pack-comm/src/vocab.rs:267` and the README describe exactness only *below* the cap
  and omit that `saturated=false` is still exact *at* the cap.
- The SQLite authorizer rejects the `COUNT` function; that asserts a statement shape, not that every
  production caller uses the count-free method, and not that the work is bounded.
- Coverage gaps: no test covers a total-consuming caller against a count-free path, the merged-total
  fold with `None` on one side, or a page boundary under a non-unique custom JSON ordering key.

ohdearquant:
The count-free page migrations look sound — caller coverage is complete and the ordering is
deterministic. The problem is narrower than the change and sits on the wire contract.

## The existing unread-count fields now carry a capped lower bound under their old names

On `origin/main` these two fields are exact:

```rust
"unread_count": unread_count,           // inbox
Ok(json!({ "count": count, "actor": caller_actor }))   // unread
```

At this head both publish the clamped value, under the same names:

```rust
// crates/khive-pack-comm/src/handlers.rs:834
"unread_count": unread.count,
"unread_count_cap": unread.cap,
"unread_count_saturated": unread.saturated,

// :856
"count": unread.count,
"count_cap": unread.cap,
"count_saturated": unread.saturated,
```

and the value is capped at 1,000:

```rust
// :864, :941-945
const UNREAD_COUNT_CAP: u32 = 1_000;
let observed = exact.count.saturating_add(legacy.count);
Ok(UnreadCount { count: observed.min(cap), cap,
                 saturated: exact.saturated || legacy.saturated || observed > cap })
```

So a mailbox with 5,000 unread messages used to return `unread_count: 5000` and now returns
`unread_count: 1000` alongside `unread_count_saturated: true`.

The new flags are additive and correct, and that is exactly what makes this easy to miss: a client
that reads only the field it has always read gets `1000` — a number that is well-formed, plausible,
and wrong, with nothing in that field to signal it. The failure is silent on the read path that
every existing consumer already uses, which is the property that separates this from a cost
optimization.

Worth being precise about the direction, since the cap itself is a reasonable idea: refusing to
count an enormous mailbox is bounded and visible; returning a confident smaller number for it is
neither.

**Suggested fix:** keep the existing field exact and introduce a distinctly named bounded field, or
make bounded semantics an explicit opt-in or versioned contract. Repurposing the established field
in place is the part to avoid.

## Smaller: the boundary rule is missing from two of the four surfaces

`crates/khive-pack-comm/src/vocab.rs:267` and the pack README describe the count as exact below
1,000 and explain saturation above it, but neither states that `saturated=false` is exact *at*
1,000. The API reference (`docs/guide/api-reference.md:1417-1422`, `:1470-1472`) and the lifecycle
document (`crates/khive-pack-comm/docs/api/message-lifecycle.md:180-219`) both state it correctly.
Worth aligning all four when the contract above is settled, since discovery metadata is the surface
a caller reads first.

authored it, so its output was treated as input evidence only. The base and head field shapes, the
cap constant, and the `observed.min(cap)` clamp were each re-derived from the two refs before
posting.

ohdearquant (commented):

ohdearquant (commented):

oceanwaves630 (approved):
Approved at aba0b9cc.

PR #2319, opened 2026-08-30
fix(db): preserve proven rollback finality

### LINKED ISSUE #2213: writer task drops proven-rollback finality on ordinary error paths
## Observation (at main 97b44ad4, crates/khive-db/src/writer_task.rs:339-433)

\`execute_wrapped_transaction_profiled\` carries a finality channel (\`terminal_state: Option<WriterTaskRequestState>\`, with \`NotStarted\` / \`TransactionRolledBack\` / \`SideEffectsUnknown\`), but two ordinary paths discard proven finality:

1. **Operation error + verified rollback** (\`Ok(Err(_))\` → \`RollbackDisposition::RolledBack\`): returns the original operation error with \`terminal_state: None\`. The one case where the writer can PROVE no effect persisted is reported identically to cases where it cannot. Only the panic path stamps \`TransactionRolledBack\`.
2. **Commit error + verified rollback**: returns \`StorageError::Pool\` with \`terminal_state: None\` — untagged, and \`Pool\` conflates a commit failure with pool exhaustion.

## Why it matters

Retry classifiers (e.g. the audit-batch \`classify_store_error\` path) key on error class as a proxy for effect-finality. Retryability and effect-finality are different questions: a caller deciding whether replay is safe needs "did the transaction roll back", which the writer task already knows on these paths and drops.

Structural constraint on any fix: a store closure cannot claim committed success — COMMIT happens after it returns — so finality must be carried at the writer-task boundary (where \`terminal_state\` already lives), not in the closure's return type.

## Open questions before a contract change

- Whether \`(updated_at, deleted_at)\` is an adequate revision signal for callers that retry conditionally (\`replace_note_if_unchanged\` and the property-patch paths return a bare \`changed > 0\`, conflating row-disappeared / row-moved / non-advancing replacement).
- Receipt retention semantics if the boundary starts reporting proven-rollback.
- Magnitude of the change across callers.

Related: the wire-level \`side_effects_unknown\` flag added on the events socket (#2190 series) reconstructs \`SideEffectsUnknown\` across the boundary; a proven-rollback tag would want the same treatment.

ohdearquant:
## Summary
- carry verified rollback finality on ordinary writer operation and COMMIT errors while retaining the original source classifiers
- keep request closures exactly-once and the writer live after proven rollback; retain terminal SideEffectsUnknown when rollback cannot be verified
- propagate typed request finality through runtime, MCP, and events protocol v3, and document the public contract

## Validation
- cargo test -p khive-db --lib (838 passed)
- cargo test -p khive-storage -p khive-runtime -p khive-mcp (passed)
- cargo test -p khive-runtime --features fault-injection,test-internals --test adr133_audit_batch d1_retry_table_uses_typed_request_state (1 passed)
- cargo check -p khive-storage -p khive-db -p khive-runtime -p khive-mcp --all-targets
- cargo clippy -p khive-storage -p khive-db -p khive-runtime -p khive-mcp --all-targets -- -D warnings
- cargo fmt --all --manifest-path crates/Cargo.toml -- --check
- git diff --check

Closes #2213

### DIFF
--- a/crates/khive-db/src/stores/blob.rs
+++ b/crates/khive-db/src/stores/blob.rs
@@ -5054,8 +5054,15 @@ mod tests {
         .await
         .expect_err("the probe must refuse when an id it would delete already names a row");
         assert!(
-            matches!(error, StorageError::Unsupported { .. }),
-            "expected StorageError::Unsupported, got {error:?}"
+            matches!(
+                &error,
+                StorageError::WriterTaskRequestFailed {
+                    request_state:
+                        khive_storage::WriterTaskRequestState::TransactionRolledBack,
+                    source,
+                } if matches!(source.as_ref(), StorageError::Unsupported { .. })
+            ),
+            "expected a proven-rollback wrapper retaining StorageError::Unsupported, got {error:?}"
         );
 
         let reader = backend.pool().reader().unwrap();
--- a/crates/khive-db/src/stores/note_tests.rs
+++ b/crates/khive-db/src/stores/note_tests.rs
@@ -965,8 +965,14 @@ async fn atomic_note_property_patch_rolls_back_when_one_target_is_ineligible() {
         .await
         .expect_err("an ineligible target must abort the atomic patch");
     assert!(
-        matches!(&error, StorageError::Conflict { message, .. }
-            if message.contains(&ineligible_id.to_string())),
+        matches!(
+            &error,
+            StorageError::WriterTaskRequestFailed {
+                request_state: WriterTaskRequestState::TransactionRolledBack,
+                source,
+            } if matches!(source.as_ref(), StorageError::Conflict { message, .. }
+                if message.contains(&ineligible_id.to_string()))
+        ),
         "the conflict must name the first failing id {ineligible_id}; got {error:?}"
     );
     assert!(!error.is_retryable(), "a precondition conflict is terminal");
@@ -1071,8 +1077,14 @@ async fn atomic_note_property_patch_writer_task_commits_and_rolls_back() {
         .await
         .expect_err("a later ineligible row must abort the writer-task transaction");
     assert!(
-        matches!(&error, StorageError::Conflict { message, .. }
-            if message.contains(&ineligible_id.to_string())),
+        matches!(
+            &error,
+            StorageError::WriterTaskRequestFailed {
+                request_state: WriterTaskRequestState::TransactionRolledBack,
+                source,
+            } if matches!(source.as_ref(), StorageError::Conflict { message, .. }
+                if message.contains(&ineligible_id.to_string()))
+        ),
         "the conflict must name the first failing id {ineligible_id}; got {error:?}"
     );
     assert_eq!(
@@ -1703,7 +1715,11 @@ async fn pooled_transaction_commit_failure_with_verified_rollback_keeps_writer_u
     assert!(
         matches!(
             &result,
-            Err(StorageError::Pool { operation, .. }) if operation == "test_pooled_commit"
+            Err(StorageError::WriterTaskRequestFailed {
+                request_state: WriterTaskRequestState::TransactionRolledBack,
+                source,
+            }) if matches!(source.as_ref(), StorageError::Pool { operation, .. }
+                if operation == "test_pooled_commit")
         ),
         "a denied COMMIT followed by a verified rollback must report the commit error: {result:?}"
     );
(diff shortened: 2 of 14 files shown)


ohdearquant:
## Verified rollback finality rests on an invariant nothing enforces

The predicate that decides a request failed cleanly is `crates/khive-db/src/writer_task.rs:295`:

```rust
Ok(()) if conn.is_autocommit() => RollbackDisposition::RolledBack
```

That establishes the connection left *a* transaction. It does not establish that everything the
closure wrote was undone. The closure type is
`Box<dyn FnOnce(&Connection) -> Result<R, StorageError> + Send>` (`writer_task.rs:70`), so it
receives a raw connection, and the "DML only — never a bare `BEGIN` / `COMMIT` / `ROLLBACK`" rule
lives in the doc comment at `writer_task.rs:62-69` with nothing enforcing it. A closure that
committed and then opened a fresh transaction before returning an error would satisfy this predicate
with its first write already durable, and the caller would be told the request rolled back cleanly.

On severity, so this is not overstated: I enumerated all 41 writer-task submission sites
(`send_bounded` / `send_top_level`, non-test, closure bodies extracted by brace matching rather than
a line window) and **none** of them issues transaction control. Every one is DML-only. The
`BEGIN`/`COMMIT` pairs live in the separate `with_writer` fallback path, whose closures own their
own transaction and never reach this predicate — `crates/khive-db/src/stores/graph.rs:1521` versus
`:1533` is the pattern throughout.

So this is not a live defect; it is a stated guarantee resting on an unenforced convention, where
the failure is silent and runs in the worst direction — reporting a clean failure for a durable
write. Worth closing while the invariant still holds everywhere rather than after some future
closure breaks it: either enforce the restriction at the boundary (hand the closure a wrapper that
cannot issue transaction control) or conservatively classify what the predicate cannot prove as
`SideEffectsUnknown`.

## Verified and sound

- The unknown branch is genuinely reachable, not decoration: rollback errors, a successful rollback
  with autocommit still false, and poisoned connections all reach
  `WriterTaskTerminated(SideEffectsUnknown)` and retire the writer (`writer_task.rs:412`, `:1013`).
- Exactly-once holds across a surviving writer: the operation is `FnOnce`, invoked once
  (`writer_task.rs:348`), replied once (`:471`), and never replayed anywhere in this change.
- `AdmissionDeadlineExpired`, `DriverCancelled` and `DriverJoinLost` can all be produced while work
  is still in flight, and the code treats them as unresolved rather than as clean failure — the
  conservative direction.
- Protocol version skew is rejected before dispatch, so no cross-version write proceeds, and a
  missing `writer_task_failure` defaults to `None`, which declines to claim rollback finality rather
  than asserting it.

## Non-blocking notes

- The events bridge flattens the nested source type. `storage_error_response` sends
  `source.to_string()` plus retryability (`crates/khive-runtime/src/events_split.rs:1168`) and the
  client reconstructs a `Pool` or `InvalidInput` (`:1604`), so
  `Unsupported("append_events_idempotent")` arrives as a generic error. That is narrower than the
  documented promise that source classification stays inspectable.
- Session state (authorizers, PRAGMAs) is not reset on a writer that survives a failure, so the same
  unenforced raw-connection contract can carry settings into the next request.
- No test covers a message with a missing finality field combined with `retryable: true`.

ohdearquant:
The retry classification is sound. Extending the request-state match so `WriterTaskRequestFailed`
is handled alongside `WriterTaskTerminated` is right: both carry the same `WriterTaskRequestState`,
and a given state means the same thing on either variant.

One point is worth recording explicitly, because it looks wrong until you check it.
`SideEffectsUnknown` is documented as a state that "must not be blindly retried"
(`crates/khive-storage/src/error.rs:28-31`), and the classifier maps it to `Retry`. That is safe
here, and the reason is the retry loop itself: it re-issues
`store.append_events_idempotent(events.clone())` (`crates/khive-runtime/src/audit_batch.rs:440`),
so the retry is idempotent by contract rather than blind, and it is bounded by
`max_commit_attempts`. The mapping is also unchanged from the base — the same
`SideEffectsUnknown => RetryDecision::Retry` arm is present in `origin/main` — so this change
widens which error variant reaches that arm, not what happens once it does.

The remaining item is the low, non-blocking one: verified rollback is still an unenforced
raw-connection convention. That is a follow-up, not a reason to hold this.

No blocking findings.

oceanwaves630 (approved):
Approved at c9d0ff53. Required checks green (Secret scan, CI gate); reviewed at head.

PR #2320, opened 2026-08-30
fix(db): reconcile WAL pins and report storage composition

### LINKED ISSUE #2042: walpin: orphaned .pid.beacon.tmp sidecar temps are enumerator-invisible and accumulate unboundedly
## Summary

The walpin sidecar's write-temp-then-rename discipline can orphan `.{pid}.beacon.tmp` files, and the sidecar enumerator (`list_names` in `crates/khive-db/src/walpin.rs`) skips all dot-names by first raw byte, by design. An orphaned beacon temp is therefore invisible to every enumeration-based sweep and is only ever cleaned if the same PID later re-registers against the same sidecar directory (`write_atomic`'s best-effort `unlink_tolerant(tmp_name)` keyed on the caller's own PID). A once-per-process beacon whose process dies between the temp write and the rename orphans its temp effectively forever.

## Measurements (live sidecar, macOS)

- `khive.db.walpin/` holds 7 visible `*.beacon` entries and **1130 hidden `.{pid}.beacon.tmp` orphans** (1137 dirents total).
- Oldest orphan mtime 2026-07-20; newest 2026-08-18 (the day of measurement) — accumulation is ongoing, roughly 35-40/day on this workload.
- Sampled orphans (oldest, mid-range, newest) all contain **complete, valid beacon JSON**, so `write_all` + `sync_all` completed and the loss window is at/around the final `renameat`. `rename_over` itself has no refusal path, which points at process termination inside the sync-to-rename window; on macOS `sync_all` is `F_FULLFSYNC` and can block for seconds under heavy WAL/checkpoint IO — the same contention conditions this sidecar exists to diagnose. That termination-during-fsync mechanism is consistent with the evidence but has not been reproduced under instrumentation.

## Why it matters

Disk cost is trivial (87 bytes each), but the enumerator's doc comment states skipped dot-entries **still count toward the raw scan bound (`RAW_SCAN_FACTOR * max`)**. Unbounded temp accumulation therefore steadily consumes the scan budget of every `list_names` pass and will eventually force truncated enumerations of the *real* beacon/heartbeat population — degrading exactly the sidecar-health attribution the beacons implement.

## Suggested fix shape

Age-based reaping of dot-temp entries as a dedicated arm (not via the dot-skipping enumerator): a `.{pid}.*.tmp` entry older than a small multiple of the producer sweep cadence is orphaned by definition (renames are immediate), so it can be unlinked safely; keep the existing symlink refusal discipline. Alternatively (or additionally), `ensure_sidecar_dir` could sweep stale temps at initialization. The current live population (1130 orphans on one machine) is a ready validation corpus for whichever arm lands.

### LINKED ISSUE #2133: db_diagnostics: no size composition, and wal_pin census is degraded by construction
Two gaps found while trying to answer "where are the bytes" for a 23 GB store using only the verb surface. Measured against `khive` 0.7.0, build `59076981b3508bc1083e3a6bebc947f2e4eae337`.

## 1. No size composition anywhere on the surface

`db_diagnostics()` returns writer contention, checkpoint counters, edge integrity, the WAL sidecar size, and a holder census. `stats()` returns entity, note and edge counts. Neither returns a per-table or per-index byte breakdown.

For a store at 23 GB with 199,868 notes, that works out to roughly 121 KB per note. Whether that is expected depends entirely on how the bytes split between row data, FTS5 trigram indexes, and embedding spaces, and the surface cannot say. So the question that decides whether a retention policy is even needed is not answerable without going around the verbs to the file, which defeats the point of having a diagnostic verb.

What would close it: a per-table and per-index size report, with the vector/embedding storage broken out separately, since that is the component most likely to dominate and the one whose growth has a different remedy from row retention.

## 2. The WAL-pin census cannot complete without mutating

`db_diagnostics().wal_pin` returns `status: "degraded"` and `fully_attributed: false` on a healthy system. Its own stated reason:

> sidecar-to-holder reconciliation not available: this tree's khive-db has no non-destructive sidecar reconciliation primitive; `walpin::enumerate_live` and `walpin::housekeep_live` both perform mutating cleanup, so diagnostics called neither and collected only the read-only OS holder census below

The verb is correct to refuse: a diagnostic must not mutate to report. But the consequence is that this field is degraded by construction on every call, for everyone, forever, rather than degraded because something is wrong. A field that can never reach `ok` carries no signal, and a reader who learns to skip it will also skip it on the day it matters.

The census that does run is separately incomplete here (250 PIDs uninspectable, so additional holders cannot be ruled out), which is an ordinary permissions limit and is reported honestly.

What would close it: a read-only `walpin` enumeration primitive that reconciles sidecars to holders without cleanup, so `wal_pin` can report `ok` on a healthy store and `degraded` only when it is actually degraded.

## Context

Both were hit while measuring, not while debugging a failure. The store itself reads healthy in the same call: 1 writer acquisition timeout across 165,398 acquisitions, no audit append or flush failures, `audit_degraded` false, 82 checkpoint pressure episodes started and 82 recovered with 0 consecutive skips, and a 64 MB WAL file holding 24 pages.

ohdearquant:
## Summary

- reap bounded stale walpin producer temps only for dead or PID-reused producers while retaining fresh, live, malformed, and untrusted evidence
- add a bounded non-mutating sidecar inspection path and reconcile it with the OS holder census so healthy evidence can report complete
- add deterministic per-object SQLite page accounting with separate row, index, FTS, vector, mixed-embedding, internal, freelist, and unaccounted totals
- document the operator contract and explicit precision/cap boundaries

## Verification

- `cargo test --manifest-path crates/Cargo.toml -p khive-db --lib` (844 passed)
- `cargo test --manifest-path crates/Cargo.toml -p khive-pack-kg --lib` (186 passed)
- `cargo clippy --manifest-path crates/Cargo.toml -p khive-db -p khive-pack-kg --all-targets -- -D warnings`
- `git diff --check`

Closes #2042
Closes #2133


ohdearquant:
Two problems here should block, and both are in the reaper/reporting half rather than the accounting
math. The page-composition work is largely sound; the parts that decide *what to delete* and *when to
claim completeness* are not yet safe.

## A failed directory read is reported as a successful one

`crates/khive-db/src/walpin.rs:1050-1055`:

```rust
let entry = unsafe { libc::readdir(dirp) };
if entry.is_null() {
    break;
}
```

`readdir` returns NULL for end-of-directory **and** for an error; the only way to tell them apart is
to set `errno` to zero before the call and read it after. This loop does neither. An I/O error
mid-walk therefore ends the enumeration silently, `truncated` stays `false`, and the partial listing
is indistinguishable from a complete one.

That flows straight into the completeness status. `crates/khive-db/src/diagnostics.rs:566-570`
computes `fully_attributed` as `census_is_complete && !sidecar_listing_truncated && …`, so a failed
sidecar read can yield `WalPinAttributionStatus::Complete` whenever the entries that *did* get read
happen to cover the OS-confirmed holders. The report then asserts complete attribution on evidence it
could not finish gathering — which is the exact outcome the reconciliation exists to prevent, and it
fails in the permissive direction.

The convincing part is that this file already knows the rule. The OS holder census at
`walpin.rs:2534-2541` has an explicit iterator-error branch that sets `truncated = true` and carries
the comment "the walk is no longer provably a complete enumeration of live PIDs." The sidecar walk
needs the same treatment. Note the new sentinel does not cover this: it guards the explicit
count/raw-scan bounds, not a `readdir` failure.

## A malformed producer temp is deleted without ever being parsed

`walpin.rs:3116-3118` returns the candidate as reapable on a dead PID:

```rust
if !is_process_alive(pid) {
    return Ok(Some(entry));
}
```

Parsing happens afterwards, at `:3122-3131`, and the malformed branch (`return Ok(None)` when
identity does not deserialize) sits below it — so that retention branch is reachable **only while the
process is alive**. Once the PID dies, a corrupt or non-JSON `.<pid>.json.tmp` / `.<pid>.beacon.tmp`
is unlinked purely because its filename PID is gone, its contents never inspected.

This contradicts two statements in the change itself. The function's own doc comment
(`walpin.rs:3024-3027`) says malformed and uninspectable records "remain on disk so a later
TRUNCATE-no-progress attribution pass can consume their `Unknown` evidence instead of observing a
falsely clean directory." ADR Amendment 13 (`docs/adr/ADR-091-wal-snapshot-lifetime.md:1521-1524`)
lists "malformed identity" among what is retained.

Worth noticing that the retention is not just wrong, it is *unstable*: the same malformed temp is
retained while its PID lives and becomes reapable on a later pass once that PID dies. A retention
guarantee that expires on a timer is the kind that looks correct in every test that runs promptly.
The body and its embedded PID/start identity need validating before any dead-PID fast path, with any
parse, stat or ownership uncertainty classified non-reapable.

## The start-time comparison is not process identity, and can authorize deleting a live producer

The reuse test compares a persisted whole-second `started_at` against a freshly queried OS start
time, treating a difference above `START_TIME_EPSILON_SECS = 2` as PID reuse and returning the entry
for unlink (`walpin.rs:3136-3137`).

The comment at `walpin.rs:61-65` is candid that these two numbers come from different clocks — the
writer's own `SystemTime::now()` versus `proc_pidinfo` / `/proc/<pid>/stat` — and then concludes this
is "rounding slack, not a real identity ambiguity window." That conclusion is where it breaks.
`SystemTime` is wall-clock and can step; the OS start time is derived from boot time and monotonic
ticks and does not. An NTP correction or manual clock change larger than two seconds makes them
disagree by more than the epsilon, and the code reads that disagreement as PID reuse and deletes a
live producer's still-load-bearing pin. A wall-clock adjustment must never become deletion authority.

The inverse also holds and shows the comparison is not identity at all: a genuinely different process
occupying the same PID whose start time lands inside the two-second window compares as "same". That
direction is safe — residue is retained — but it demonstrates the predicate is a heuristic wearing an
identity name.

Credit where due: the surrounding liveness handling is careful. `is_process_alive`
(`walpin.rs:1096-1109`) treats `EPERM` from `kill(pid, 0)` as alive rather than dead, and an
unavailable `process_start_time_secs` returns `None`, which retains the temp (`:3119-3120`). Both
fail in the safe direction. The gap is specifically the start-token comparison, which wants an
OS-native token from the same clock the reaper reads.

## Composition is measured across several statements, not one snapshot

`diagnostics.rs:954-975` issues separate `PRAGMA page_size`, `page_count` and `freelist_count`
statements and then a separate `dbstat` query. The enclosing `scope.run` (`:1313-1322`) installs
cancellation and progress handling but does not open a read transaction.

A commit landing between those statements means the object and class totals describe one database
state while `page_count`/`freelist_count` describe another. The residual is then arithmetic over two
different databases: a page can appear to belong to none, or the accounted total can exceed the
available non-freelist pages. `ORDER BY d.name` makes the object ordering deterministic, which is not
the same property as the measurement being consistent — determinism of order does not give a
consistent snapshot. All reconciliation inputs want to come from one explicit read transaction.

## The residual is clamped, so over-accounting is silently invisible

`diagnostics.rs:1020-1024` computes `unaccounted_bytes` with two `saturating_sub` operations. Under
normal operation the class buckets are mutually assigned per `dbstat` row and freelist is subtracted
separately, so double-counting is not expected — but if an inconsistent read (see above) or a
`dbstat` discrepancy pushes accounted bytes past `database_bytes - freelist_bytes`, the negative
residual clamps to zero and the report looks perfectly reconciled.

So `unaccounted` is a residual only on one side. The case where the number would actually reveal a
classification bug is precisely the case it hides, which inverts the value of publishing it. A
signed or checked delta with an explicit over-accounted state would make the bucket mean what it
says. `docs/guide/api-reference.md:863-871` documents the bucket but not the clamping.

No builds or tests were run for this pass; every finding above is from source reading. The
verification list in the description is an unverified external claim and was not relied on.

ohdearquant:
## The red `CI gate` is one timing-attribution test, and the numbers are close

`CI (ubuntu-latest, shard 2/2)` fails one test:

```
FAIL khive-db writer_task::tests::writer_stage_sample_attributes_a_slow_body
panicked at crates/khive-db/src/writer_task.rs:2590:9:
a fast COMMIT must not receive the body's delay: WriterStageObservation {
  queue_wait_micros: 83, transaction_acquire_micros: 14,
  body_micros: 60168, commit_micros: 72725, total_micros: 133225, ... }
```

The test injects a slow body and asserts the commit stage is not credited with that delay.
It failed because the commit measured 72.7ms against a body of 60.2ms — the commit really was
slower than the body, so stage attribution was arguably correct and the wall-clock premise is
what broke.

Two readings, and the diff does not settle it. Against ambient: this is a wall-clock assertion
about one stage being fast, which is exactly the shape that tips over on a contended runner —
several branches failed timing assertions today. Against ambient: this branch does edit two
files under `crates/khive-db/`, and reconciling WAL pins is plausibly able to move commit
latency, which would make a genuinely slow commit the finding rather than the noise.

The discriminator is cheap: re-run and see whether `commit_micros` lands well under
`body_micros` or hovers near it again. A single pass either side is not evidence — the ratio
across a few runs is. If commit stays in the same band as body, this is a real attribution
change and not a flake.

ohdearquant:
Three things here, and the first one is mechanical.

## The required gate is failing on an introduced lint error

`CI gate` is `FAILURE` at this head. The ubuntu shard reports
`khive-db/src/walpin.rs:1003:34`, *useless conversion to the same type: u64*, under `-D warnings`,
followed by `could not compile khive-db`. The line is:

```rust
let current_device = u64::try_from(current.st_dev).unwrap_or(u64::MAX);
```

This is introduced rather than inherited — `remove_if_same` does not exist on `origin/main` (0
occurrences there, 2 at this head), so the whole surrounding function is new. Worth fixing first
since nothing else can be judged against a red gate.

## The new cleanup lane deletes malformed temps before reading them, which contradicts the ADR in this same change

This is the one I would hold on. `stale_orphan_temp` returns a deletion candidate as soon as the
filename PID is not alive:

```rust
// crates/khive-db/src/walpin.rs:3116-3117
if !is_process_alive(pid) {
    return Ok(Some(entry));
}
```

The body is only parsed *after* that, at `:3122-3131`. So a stale `.{dead_pid}.json.tmp` or
`.{dead_pid}.beacon.tmp` whose contents are malformed, corrupted, or deliberately shaped never
reaches the parse at all — it is handed to `remove_if_same` at `:3190-3195` and unlinked.

The reason this matters more than an ordinary ordering nit is that the ADR added by this same PR
promises the opposite:

> Fresh temps, live matching producers, **malformed identity**, uninspectable processes, non-owned
> entries, symlinks, and all unrecognized dot-names are retained.

(`docs/adr/ADR-091-wal-snapshot-lifetime.md`, producer-temp residue section.)

`read_checked_entry` does reject symlinks, non-regular files, and oversized files at `:920-974`, but
that evidence survives only because the error returns before the delete path; it does not cover the
dead-PID-with-invalid-body case. And a malformed temp whose PID is currently alive is retained now,
then reclassified as reapable on the next pass once that PID exits — so the retention is a function
of timing rather than of the record.

The direction is the bad one: this is the path that consumes exactly the evidence that would explain
a corruption or an attribution question later, and it does so silently, leaving a record that looks
clean.

Suggested fix: parse and validate body, filename PID, and producer identity *before* returning a
deletion candidate. Anything unparseable or uninspectable returns retain; only a validated record
with a positively dead producer, or a validated start-time mismatch proving PID reuse, may reach
`remove_if_same`.

## The size composition is assembled from reads that are not one snapshot

`database_size_composition` issues `PRAGMA page_size`, `PRAGMA page_count` and
`PRAGMA freelist_count` as three independent `query_row` calls (`crates/khive-db/src/diagnostics.rs:955-966`),
then prepares and runs a separate `dbstat` statement (`:968-975`), with no transaction spanning
them. `ORDER BY d.name` stabilizes the object list within the one `dbstat` statement, but it does
not put the pragmas and the walk in the same view.

A concurrent commit, checkpoint, or freelist change can therefore mix values from different database
states — an object present in the `dbstat` walk while `page_count` predates its allocation, or the
reverse. The API and ADR describe a file-wide decomposition, so this should hold under writes rather
than only when the database is quiet. Taking all four inputs under one read transaction would close
it.

## Smaller, same area

`unaccounted_bytes` is `database_bytes.saturating_sub(freelist_bytes).saturating_sub(accounted_bytes)`
(`:1020-1024`). If the named buckets ever exceed the measured used bytes, the residual clamps to
zero with no flag, log, or signed value — which turns the single strongest signal that the
decomposition is wrong into a result indistinguishable from a healthy one. The test at `:2295-2345`
asserts the identity using that same saturating formula, so it cannot catch it. Checked or signed
arithmetic with an explicit over-accounting field would make `unaccounted` do the job it exists for.

Separately and **not** blocking this change: ordinary housekeeping still deletes a malformed beacon
whose filename PID is dead (`:3323-3333`, heartbeat equivalent at `:3250-3260`), which is
pre-existing — `origin/main` has the same branch. It is now in tension with the new ADR text and is
worth a follow-up, but it is not something this PR introduced.

authored it, so its output was treated as input evidence only. The load-bearing facts above — the
failing required gate and the lint site, `remove_if_same`'s absence on the base, the early return
ahead of the parse, the ADR's retention sentence, and the four unsynchronized reads — were each
re-derived from the refs before posting.

oceanwaves630 (approved):

PR #2322, opened 2026-08-30
feat(storage): classify batch write refusals

### LINKED ISSUE #2079: Batch write surfaces report only first_error: bulk ingest callers cannot classify per-item refusals
## Problem

`BatchWriteSummary` (crates/khive-storage/src/types/mod.rs:42) is the return shape for batch write operations:

```rust
pub struct BatchWriteSummary {
    pub attempted: u64,
    pub affected: u64,
    pub failed: u64,
    pub first_error: String,
}
```

A bulk ingest caller whose run accumulates hundreds of refusals gets a count and one error string. There is no per-item refusal record on the response and no durable refusal ledger in the store, so after the fact the caller cannot distinguish deterministic refusals (schema validation, duplicate content, invalid references) from transient ones (contention, timeouts). The only classification method available today is re-submitting the refused items and reading fresh responses.

Observed in practice: a large backfill recorded 738 store refusals as a bare counter; nothing durable can say why.

## Ask

Per-item error reporting on batch write surfaces, in one of these shapes (in preference order):

1. An `errors: Vec<(index_or_id, error_class)>` field, possibly capped with an explicit truncation marker.
2. A sampled `errors` vector plus `counts_by_class` so large failure sets stay bounded but classifiable.
3. At minimum, counts grouped by error class alongside `first_error`.

A refusal is a decision the store made; the caller should be able to read that decision without re-submitting the workload.

ohdearquant:
## Summary

- extend `BatchWriteSummary` additively with bounded per-item refusal details, complete counts by stable class and retryability, and explicit truncation metadata
- populate the contract across SQLite entity, note, text, sparse, vector, and guarded graph batch decisions while preserving all-or-nothing error semantics
- retain the legacy four-field successful JSON shape and `first_error` behavior; cap detail samples at 128 and sampled messages at 512 characters
- document the storage and writer-task contracts and update compatibility initializers across the workspace

## Verification

- `cargo test -p khive-storage` (93 unit + 48 integration tests)
- `cargo test -p khive-db --lib` (838 passed)
- focused vector, guarded-graph, serde/cap, and SQLite classification regressions
- affected-crate `cargo clippy --all-targets -- -D warnings`
- `git diff --check`

Closes #2079


ohdearquant:
The atomicity story here is carefully done and the per-store behaviour is more considered than the
summary suggests. Four things to settle, two of which should block.

First, a correction to the description's own framing: the legacy *successful* JSON was three fields,
not four. `first_error` carries `skip_serializing_if = "String::is_empty"`, so it was already absent
from a successful summary. The new successful shape is still exactly `attempted`, `affected`,
`failed`, so the compatibility claim holds — it just holds for a different reason than stated.

## Message truncation is invisible to a consumer

`bounded_batch_error_message` (`crates/khive-storage/src/types/mod.rs:173-183`) cuts a message to
511 characters and appends `…`. `BatchWriteError` (`:74-85`) has exactly `index`, `item_id`, `class`,
`retryability`, `message` — no truncation flag. The only truncation metadata on the summary,
`errors_truncated` and `errors_omitted`, describes the *list*, not any individual message.

So a genuine 512-character message, or any original message that legitimately ends in `…`, is
indistinguishable from one that was cut. A consumer cannot tell whether `errors[].message` is
complete, which is the same defect the list-level metadata was added to prevent — solved for one cap
and left open for the other. The existing test (`:233-250`) checks length and suffix, so it cannot
detect this either.

Worth an explicit per-entry `message_truncated` flag serialized with the detail, tested in both
states.

## Adding public fields to `BatchWriteSummary` is a source-breaking change, and it reaches users

`BatchWriteSummary` (`crates/khive-storage/src/types/mod.rs:96-113`) is `pub` and derives
`Clone, Debug, Default, Serialize, Deserialize` — it is **not** `#[non_exhaustive]`. Four public
fields are added, so any struct literal that constructed it exhaustively stops compiling. The diff
demonstrates this itself: initializers across the workspace had to be rewritten onto
`..BatchWriteSummary::default()` (`crates/khive-db/src/checkpoint.rs:5802`,
`crates/khive-mcp/src/serve.rs:11355`, and six others).

The reach is what makes this High rather than a note. `khive-storage` carries full publish metadata
— description, license, repository, keywords, categories — with no `publish = false`, and its
sibling-crate dependencies are declared with explicit versions beside their paths
(`khive-score = { version = "0.8.0", path = ... }`), which is the shape of a crate that is actually
released rather than workspace-only. So this is not contained to this repository, and the wire
compatibility argument does not cover it: JSON compatibility and Rust source compatibility are two
different promises, and "additively" is only true of the first.

Either expose the detail through a new type and leave the existing struct alone, add
`#[non_exhaustive]` now and accept that this release is the breaking one, or state plainly that this
is a semver-breaking change to a published API and handle it in the release contract.

## `cancelled` is advertised as safe to retry without knowing why it was cancelled

`classify_batch_sqlite_error` (`crates/khive-db/src/stores/mod.rs:27-47`) maps every
`ErrorCode::OperationInterrupted` to `Cancelled` + `Transient`. Its only input is the `rusqlite`
error, so it has no way to distinguish a deadline or transport interruption from an explicit caller
cancellation — and it labels both as transient.

`retryability` exists precisely so automation can act on it without reading messages. An operation a
caller deliberately interrupted is the one case where retrying is wrong, and this is the value that
tells a future consumer to retry it. Nothing in this repository reads the field yet, so this is a
latent contract defect rather than a live bug — but it is the decision every future consumer
inherits, and it is cheaper to fix before anything depends on it. Preserve the cancellation cause
through classification and mark only deadline/transport interruptions transient; explicit
cancellation should be `Unknown` or non-retryable.

## The guarded graph path drops the new detail before it reaches callers

`crates/khive-runtime/src/operations.rs:6152-6164`: when `outcome.refused` is present, the code
returns `RuntimeError::GuardedWriteFailed` carrying only `entry_index` and the two missing-endpoint
values. `outcome.summary` — with the freshly populated `errors`, `error_counts` and truncation
fields, including the collateral `batch_aborted` entries the guard just classified — is discarded.

Storage-level callers see the per-item evidence; `link_many` callers do not. On the one path where
the new classification is most informative, the contract stops at the runtime boundary. Either carry
the bounded projection into `GuardedWriteFailure`, or document and test that this operation
deliberately surfaces only the single guard failure.

## Verified and sound

- **Atomicity is not uniform, and the diff is right about which store is which.** Entity, note, text,
  sparse and vector are partial-success by construction (statement-level refusals, or per-item
  savepoints at `stores/text.rs:828-852` and `stores/vectors.rs:1028-1069`), so a refused row is
  isolated and successful rows commit. Ordinary `upsert_edges` stays all-or-nothing
  (`stores/graph.rs:1097-1116` returns `Err` on first failure and both wrappers roll back). The
  guarded batch pre-checks every endpoint inside the transaction and writes nothing on refusal, so it
  changes the *evidence*, not the atomicity. No path reports a refused row as committed.
- **The counts are complete, not sampled.** `record_failure` updates `error_counts` before the cap
  branch (`types/mod.rs:124-169`), so aggregates survive truncation of the detail list — the failure
  mode where a count silently describes only the sample is closed here.
- **`first_error` is preserved exactly**, copied before the message cap applies, and the graph guard
  pre-sets the legacy message so a collateral entry cannot displace it.
- **No driver-message string matching.** Classification is by `sqlite_error_code()` variant, with
  unrecognized codes falling to `driver/unknown` and non-SQLite errors to `unknown/unknown`.
- **The compatibility initializers are honest.** All but one are test fakes defaulting to zero/empty
  on successful writes; `crates/khive-runtime/src/events_split.rs:1654` is the one real forwarder and
  supplies a measured `affected`. None fabricates a class count or a sampled refusal.

## Coverage gaps

`conflict` has no producer anywhere in this change, and neither ADR amendment enumerates the nine
serialized class names or notes that one is currently unemitted — worth doing, since the value of a
closed vocabulary is that consumers can match on it exhaustively.

No test drives a real store past 128 refusals (only the unit-level 131-item case at
`types/mod.rs:190-229`); there is no refusal-detail regression for entity, note, text or sparse; no
per-store SQL-error classification test; and nothing pins the runtime/MCP guarded refusal projection,
which is the fourth finding above.

No builds or tests were run for this pass; every finding is from source reading. The verification
list in the description is an unverified external claim and was not relied on.

ohdearquant (commented):

ohdearquant (commented):

ohdearquant (commented):

ohdearquant (commented):

oceanwaves630 (approved):

PR #2325, opened 2026-08-30
feat(identity): disclose resolved actor at host boundaries

### LINKED ISSUE #2277: A shared daemon inherits actor identity from its working directory's owner
## Summary

Actor identity resolves from the khive config file discovered at a process's current working directory, with no upward walk. That rule is clear and predictable. The problem is what it implies for a long-lived shared daemon, which must have *some* working directory: if that directory belongs to one of the daemon's users, then the daemon's own configuration identity is that user's, and — more seriously — **any client invoked from that directory authenticates as that user**.

There is a second, more general asymmetry underneath it: identity resolution is loud when it *fails* and silent when it *succeeds as somebody*.

## Observed

A shared daemon was configured with a working directory that happened to contain a user-owned config. Consequences:

- The daemon's configuration identity was derived from that user's configuration.
- Routine verification commands run from that directory — the natural choice, since it is the daemon's own working directory — executed under that user's identity and returned that user's scoped data: task lists, schedule entries, stored memories, profile bindings, mailbox.

Nothing in the output indicated whose identity was in force.

When identity resolution *fails*, a clear warning is emitted saying the actor resolved to the unattributed default, that sends will be stamped unattributed, that mailbox reads will be unscoped, and that an explicit actor should be configured.

When it *succeeds as an unexpected identity*, there is no output at all. The safe-looking case is the silent one, and it is the more dangerous of the two: acting as the wrong principal is worse than acting as nobody.

## Why the obvious workaround is not enough

"Run verification from a config-bearing directory" is the natural instruction, and it is exactly the instruction that produces this outcome — it names a *property* (a config exists there) when the thing that matters is a *value* (which actor it names). Documentation can be corrected, but the failure mode is silent and the check is easy to skip, so a mechanism is preferable.

## Proposals

Any one of these closes the daemon half; the third closes the general case.

1. **A daemon-owned configuration.** Give the daemon its own configuration directory and a dedicated non-user actor, so no user's configuration is load-bearing for a shared service. This is the narrowest fix and probably the right default.

2. **Refuse to inherit a user identity in daemon mode.** When starting as a daemon, require identity to come from an explicit flag or a service-kind actor, and refuse — or warn loudly — if the resolved actor is an ordinary user actor.

3. **Disclose the resolved identity symmetrically.** Report the resolved actor on invocation, not only when resolution degrades to unattributed. An operator should never have to infer which principal a command ran as by inspecting the addressee field of returned rows. This is the smallest change and has the broadest reach, because it makes every future instance of this class self-revealing rather than silent.

A related hardening, if cheap: let a command *assert* its expected identity and fail if it does not match, so scripts and automation can be explicit rather than positional.

## Notes

The working-directory rule itself is not in question; predictable cwd-scoped resolution is a reasonable design. The gap is that a shared daemon has no identity of its own to fall back on, and that success is quieter than failure.

ohdearquant:
## Summary

- disclose the resolved actor on successful exec dispatch before daemon forwarding or local execution
- emit the same attributed-or-local actor line at both MCP host startup boundaries
- keep JSON and MCP stdout untouched by using stderr and the forced khive.boot target
- document the invariant and the existing fail-closed expect-actor assertion

## Verification

- cargo check --manifest-path crates/Cargo.toml --workspace --all-targets
- cargo test -p khive-mcp resolved_actor_disclosure
- git diff --check

Closes #2277


ohdearquant:
The design here is right in the way that matters most for an identity diagnostic: it does not
re-derive the answer. `resolved_actor_disclosure` calls `khive_runtime::resolve_actor`
(`crates/khive-mcp/src/serve.rs:3869`), which is the same pure function the attribution paths use —
token minting and gate checks at `crates/khive-runtime/src/runtime.rs:1035` and `:1098`, pack
dispatch at `crates/khive-runtime/src/pack.rs:1643` and `:1928`. A disclosure that resolved identity
independently could print one actor while a different one signed the records, which would be worse
than printing nothing; that failure is closed by construction here.

The inputs match too, which is the other half of that argument. `serve` passes `server.actor_id()`
→ `registry.actor_id()` → the same field `pack.rs:1643` resolves for attribution, and `exec` passes
`cfg.actor_id`, the same config the runtime is built from. Since `resolve_actor` reads no ambient
state, identical input gives identical output.

Three other claims verified:

- **Visibility under the default log level holds.** `crates/kkernel/src/cli.rs:887` force-enables
  `khive.boot=info` in the filter, so the line survives the default `--log warn`. This was the claim
  most likely to be wrong — an `info!` that nothing enables would make the feature invisible in
  exactly the configuration operators run.
- **stdout is untouched at both surfaces.** `exec` writes via `writeln!(std::io::stderr(), ...)`
  (`crates/kkernel/src/exec.rs:2106-2110`), and `serve` goes through the tracing subscriber.
- **The `exec` coverage is complete, not just the two edited sites.** `run_exec_inline`
  (`exec.rs:2132`) delegates to `run_exec_inline_with_forward`, so the two disclosure calls at
  `:2260` and `:2556` sit on every dispatch path, each one paired with the existing database
  disclosure rather than added beside it.
- **`--expect-actor` is genuinely fail-closed** (`exec.rs:1876-1889`): a mismatch returns a refusal
  rather than warning, and it resolves through the same `resolve_actor`, so the assertion and the
  disclosure cannot disagree about what "the actor" means.

Using `{:?}` for the id is a good choice worth keeping deliberately. `resolve_actor` trims only to
decide emptiness and then stores the *untrimmed* string, so an id like `" svc-indexer "` is a real
possibility; the debug formatting makes that visible instead of rendering it as an ordinary name.

## The gap: nothing tests that the line is ever emitted

The only tests are the two pure-string cases at `serve.rs:4355` and `:4361`. They pin the format of
the returned `String`, but no test observes the line at any of the four call sites. Delete the
`tracing::info!` at `serve.rs:161` or `:2008`, or the `disclose_resolved_actor` call at `exec.rs:2260`
or `:2556`, and the suite stays green — which means the change's entire user-visible behaviour is
unpinned. A test asserting a formatter's output cannot detect a formatter nobody calls.

The repo already has the pattern for this and it is a close fit: `crates/kkernel/tests/mcp_closed_stderr.rs`
launches the binary, reads the first startup diagnostic off stderr, and asserts it contains
`"database:"` (`:72-75`). The actor line goes to the same stream at the same forced target, so the
same test could assert `actor:` alongside it. Worth adding at least the startup-boundary case, and
ideally one `exec` case, before this lands.

## Smaller note

`actor_is_unattributed` is `actor.id == "local"` (`crates/khive-runtime/src/actor_identity.rs:27-29`),
so an *explicitly* configured `--actor local` is disclosed as `(resolved; unattributed local
fallback)`. The classification is correct — that identity really is the unattributed pool however it
was set — but "fallback" names a cause that did not occur, and this line's whole job is to tell an
operator how the identity came about. "unattributed" alone would say the true thing in both cases.

ohdearquant (commented):

oceanwaves630 (approved):

PR #2327, opened 2026-08-30
perf(kg-editor): bound large graph layout work

### LINKED ISSUE #1952: kg-editor: layout solver runs O(n²) force passes synchronously, multi-second stall on 200-node graphs
The kg-editor layout solver performs 160 O(n²) force passes plus up to 200 O(n²) overlap-cleanup passes synchronously on the main thread (`apps/kg-editor/src/lib/graph-layout.ts:101-119,166-190`, invoked from `apps/kg-editor/src/components/studio.tsx:459-462`). A valid 200-node graph measured ~3.0 s per layout call locally (environment-dependent), blocking the UI for the whole solve.

A compounding symptom: `StructureGraph` recomputes the unchanged layout on every zoom or selection render because `settleGraphLayout` is called directly in the render body (`apps/kg-editor/src/components/showcase/repo-showcase.tsx:216-218,257-259,301,381-398`); the 51-node showcase averaged ~96 ms per call.

Candidate directions:
- Memoize the settled layout keyed on graph identity so zoom/selection renders reuse it.
- Cap iteration counts adaptively by node count, or move the solve off the main thread (worker) with an interim coarse layout.
- Early-exit the cleanup passes when displacement falls below a threshold.

Introduced surface predates #1923; filing so the cost is tracked against the schema's allowed 200-node graphs.

ohdearquant:
## Summary

- bound synchronous all-pairs force and overlap-cleanup work with deterministic node-count-based iteration budgets above 64 nodes
- preserve the established full iteration schedule, exact golden layout, bounds, ordering, seeded placement, and cleanup early exit for review/showcase-sized graphs
- add a deterministic 200-node bench-style regression that counts distance evaluations instead of relying on wall-clock timing

## Closure proof

Issue #1952 identified two compounding costs. Input-stable memoization across zoom and selection already landed in #2101 and its regression remains green here. This PR addresses the remaining first-layout cost at the supported 200-node maximum:

- RED: 7,164,201 synchronous distance evaluations
- GREEN: no more than 400,000 evaluations (12 force passes and 5 cleanup passes at 200 nodes)
- focused run: 5 tests passed, including the existing zoom/selection memoization regression

Closes #1952

## Verification

- npx vitest run src/lib/graph-layout.test.ts src/components/showcase/repo-showcase-layout.test.tsx --reporter=verbose
- npm run lint
- npm run typecheck
- npm test (20 token-contract tests and 331 Vitest tests)
- npm run check:showcase
- npm run build
- cargo fmt --all -- --check
- cargo check --workspace
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test --workspace

Playwright e2e was also attempted after installing the pinned Chromium binary into a task-local temporary cache. The managed macOS sandbox denied Chromium Mach port registration (bootstrap_check_in: Permission denied) before any application assertion could run; the production build prerequisite passed.


ohdearquant:
The budget arithmetic is exactly as documented: at 200 nodes there are 19,900 pairs, so
`250_000 / 19_900` floors to 12 force passes and `100_000 / 19_900` floors to 5 separation passes,
matching the comment. And the temperature schedule was rebased correctly —
`0.9 - iteration / forceIterations * 0.72` keeps the full annealing sweep at any cap. Leaving the
original `ITERATIONS` in that expression would have frozen large graphs near the starting
temperature and quietly undone the layout, so that substitution is load-bearing and right.

Three things worth settling before this lands.

## Capping the separation pass converts a guarantee into best effort

The separation loop is a convergence loop, not a fixed-work loop: it breaks early when a pass moves
nothing (`if (!movedAny) break;`, `graph-layout.ts:217`). Before this change it ran until no pair was
closer than `MIN_SEPARATION`, with 200 as a safety ceiling. After it, a graph above
`FULL_QUALITY_NODE_LIMIT` gets a computed cap instead — five passes at 200 nodes — and can exit with
pairs still overlapping.

That is a change in what the function promises, and it bites hardest where overlap is most likely: a
dense 200-node graph is both the case that needs the most separation passes and the case now
allowed the fewest. The force loop has no such property — it is fixed-iteration and capping it only
costs quality — so the two loops are not the same kind of thing and capping them the same way is
what hides this.

Nothing currently pins the invariant at any size. The two existing tests cover order-independence
against a five-node snapshot and stage-bound clamping at twenty nodes; neither asserts
`MIN_SEPARATION`. Worth stating the post-change contract explicitly, and adding a test that asserts
separation holds below the limit and documents what actually holds above it.

## "The supported 200-node maximum" is not enforced

The comment reads as though 200 were a ceiling something guarantees. Nothing does. The only
occurrences of 200 in this area are that comment, `SEPARATION_ITERATIONS`, the new test's array
length, and unrelated HTTP status codes; both production call sites pass the graph straight through
without a length bound (`components/studio.tsx:504`,
`components/showcase/repo-showcase.tsx:574`).

Past 200 the divisor keeps growing, so the budgets keep shrinking: by roughly 700 nodes both
`boundedPairIterations` results floor to the `Math.max(1, …)` clamp and the output is essentially the
initial golden-angle seeding with one settling pass. Bounding the work is the goal and that part
succeeds — the main thread is protected. But the resulting behaviour is "returns an unlaid-out
graph", not "returns a slower layout", and a reader of that comment would not expect it. Either
enforce the maximum at the callers or reword it to describe the degradation.

## The new test counts a proxy that a plausible optimization zeroes

`bounds pair-distance work for a schema-maximum graph` counts calls to `Math.sqrt` and `Math.hypot`.
Those are genuinely the per-pair distance primitives today (`:137` in the force loop, `:199` in the
separation loop), so the count tracks pair work as written.

It stops tracking it under a change someone would plausibly make. Replacing `Math.sqrt(distanceSquared)`
at `:137` with a squared-distance comparison is the standard optimization for exactly this loop, and
it would drive the counter toward zero while leaving the O(n²) iteration structure completely
unchanged. The test then passes a change that does not reduce pair work at all — it is measuring the
distance function, not the loop.

To be fair to it, the test is not vacuous: raising `FULL_QUALITY_NODE_LIMIT` above 200 would put
this case back on the full path at 160 × 19,900 ≈ 3.18M calls, far past the 400,000 bound, and it
catches that. Asserting the computed iteration counts directly — `boundedPairIterations(200, …)`
returning 12 and 5 — would pin what the change is actually about, and would survive a change of
distance primitive. Also note the bound has about 18% headroom over the 350,000 the budgets permit,
so it does not pin the budgets themselves.

## Smaller note

There is a visible step at the threshold: 64 nodes gets 160 force and up-to-200 separation passes,
65 nodes gets 120 and 48. The separation drop is the sharp one, though its early exit softens it in
practice for sparse graphs. Worth a sentence saying the discontinuity is intended.

ohdearquant (commented):

Verdict on head 711d5878fd34615cd318b447f8457a5f40c037aa: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving on the reviewer report at this head, with CI green (20 passing, 7 skipped, 0 failing, 0 pending).

PR #2328, opened 2026-08-30
perf(memory): scope FTS gather to memory corpus

### LINKED ISSUE #1906: perf: FTS5 segment scans dominate daemon CPU; short recall queries take 17-19s on a 173k-note store
A CPU sample of the serving daemon (5s, macOS `sample`) shows the hottest userland frames are all FTS5 query internals — `fts5MultiIterNext`, `fts5SegIterNext`, `fts5SegIterSeekInit`, `fts5ExprNodeTest_STRING` — with `walFindFrame` close behind. A freshly restarted daemon sustains ~85-97% of a core with no interactive load; under load it climbs to several cores.

Correlated daemon-log evidence from the same window:

- `memory.recall exceeded slow-request threshold total_ms=17744..19192` for queries of 60-105 bytes returning 5-8 rows (store: ~173k notes).
- `backend search task timed out backend=main timeout_ms=5000` recurring — the dispatch deadline abandons the search, but the underlying read runs to completion (the cancellation gap tracked separately in #1895), so each slow query burns a core well past its deadline.
- `WAL TRUNCATE attempt made no progress; wal_pages_before=20015 wal_pages_after=20015` — long-running abandoned readers pin the WAL, and a 20k-page WAL makes every read pay `walFindFrame` traversal, which further slows queries. The three mechanisms are self-reinforcing.

The cancellation half is #1895 and the WAL-pin visibility half is #1903. What this issue asks: why does a sub-100-byte match query cost 17s of FTS5 iteration on a 173k-row corpus? Working hypothesis is segment accumulation from years of small write transactions (the multi-iterator is merging across many small b-trees), which would make this an index-maintenance problem:

1. Measure FTS5 segment structure on a representative store (`fts5_structure` / integrity-check tooling).
2. If confirmed, add periodic `'optimize'` (or bounded `'merge'`) maintenance under the checkpoint/maintenance task, and consider `automerge`/`crisismerge` tuning for the write path.
3. Add the segment count to `db_diagnostics` output so regression is observable.

A recall p95 target for acceptance should come from the bench lane rather than this issue, but the current 17s observed against a 10s slow-threshold is far outside any defensible bound.

### LINKED ISSUE #1907: perf: memory recall's FTS gather runs MATCH over the entire note corpus, not the memory corpus
`crates/khive-pack-memory/src/text_gather.rs` builds its `TextSearchRequest` with `filter.kinds = [SubstrateKind::Note]` — substrate-level only. The MATCH expression therefore executes against the single shared FTS index covering every note in the store (~173k rows today: edge-annotation notes (~58k), message rows, task rows, kg notes), and rows outside the memory corpus are discarded only by the post-MATCH WHERE filter. The memory rows are a small fraction of the corpus, but every recall pays FTS5 segment iteration over all of it. With `TextQueryMode::AnyTerm` and common terms, the candidate set before filtering is enormous.

This compounds #1906 (segment accumulation): the shared index's segments contain everything, so even a perfectly merged index makes memory recall scan non-memory content.

Fix directions, smallest first:
1. Push a kind-class predicate into the indexed columns so the FTS query itself can prune (FTS5 column filter or an indexed auxiliary column), rather than post-filtering.
2. Partition the FTS index per corpus class (memory / kg / channel / task) so each search pays only for its own corpus. Aligns with pack modularity; requires a migration and trigger updates, but the `TextSearch` trait boundary already isolates callers.

Acceptance: recall FTS gather latency measured before/after on a production-scale store, with the same query set; the memory-scoped index should be iterating segments proportional to the memory corpus only.

ohdearquant:
## Summary

- add schema V23 with an indexed granular `record_kind` classifier on the two shared FTS5 tables; backfill live base-record kinds while preserving unmatched/stale FTS rows unscoped
- stamp the classifier on normal, batch, merge, namespace-rename, and reindex writes, and scope every memory ranked/unranked/rank-within-cap/count/term-statistics path to `memory`
- keep user lexical terms title/body-only, retain an exact SQL equality predicate for trigram correctness, and use a zero-weight custom hidden-rank configuration so classifier text cannot affect relevance or defeat FTS5's optimized top-K path
- document the storage, migration, and memory-recall contract in ADR-015, ADR-021, ADR-062, and the memory text-retrieval API notes

## Production-shaped evidence

Read-only source snapshot (the live database was never mutated):

- 212,801 `fts_notes` rows
- 93,695 live notes
- 27,832 live memory notes; 27,829 corresponding FTS rows

Five fixed AnyTerm queries were run before and after against copies of the same snapshot: `memory|recall`, `agent|task`, `khive|runtime`, `database|query`, and `performance|issue`, each at top-K 150.

- post-MATCH baseline candidate visits: 368-1,171
- indexed memory MATCH candidate visits: exactly 150 for every query
- median query latency: 85.52-360.09 ms before, 37.61-109.29 ms after (about 1.65x-3.29x faster)
- plan changed from shared lexical scan plus note lookup (`SCAN f VIRTUAL TABLE INDEX 32:M8`) to the classifier/rank-constrained FTS scan (`SCAN fts_notes VIRTUAL TABLE INDEX 32:rM9`)

The focused RED fixture held 50 memories constant and increased matching non-memory notes from 200 to 2,000. Before pushdown, visits rose from 220 to 2,020; after pushdown, both runs visit only the requested 20 memory rows.

## Verification

- `cargo test -p khive-db memory_kind_match_work_is_bounded_by_memory_corpus`
- `cargo test -p khive-db record_kind_`
- `cargo test -p khive-db v23_backfills_indexed_record_kinds_and_preserves_unmatched_fts_rows`
- `cargo test -p khive-db --lib`: all 841 non-flaking tests passed; unrelated concurrent note-page seam test flaked once and passed immediately in isolation (the same full suite passed all 842 on the prior run)
- `cargo test -p khive-pack-memory`: 324 unit + 89 integration passed; 1 documented model-loading test ignored
- `cargo test -p khive-runtime --lib`: 1,415 passed; 7 documented heavyweight/audit tests ignored
- `cargo test -p khive-storage -p khive-retrieval`: 310 passed
- `cargo clippy -p khive-storage -p khive-db -p khive-runtime -p khive-retrieval -p khive-pack-memory -p kkernel --all-targets -- -D warnings`
- `scripts/lint-sql.sh`: 32 SQL files OK
- `git diff --check`

## Related scope

This does not close #1906. V23 performs a one-time FTS rebuild, but #1906 still requires periodic optimize/merge maintenance, segment-count diagnostics, and per-stage recall timings.

Closes #1907

### DIFF
--- a/crates/khive-db/benches/db_hot_path.rs
+++ b/crates/khive-db/benches/db_hot_path.rs
@@ -85,6 +85,7 @@ fn build_text_corpus() -> Vec<TextDocument> {
         .map(|i| TextDocument {
             subject_id: Uuid::new_v4(),
             kind: SubstrateKind::Note,
+            record_kind: None,
             namespace: NAMESPACE.to_string(),
             title: Some(gen_title(i)),
             body: gen_body(&mut rng, i),
@@ -449,6 +450,7 @@ fn bench_fts_upsert_batch(c: &mut Criterion) {
                             .map(|i| TextDocument {
                                 subject_id: Uuid::new_v4(),
                                 kind: SubstrateKind::Note,
+                                record_kind: None,
                                 namespace: NAMESPACE.to_string(),
                                 title: Some(gen_title(i)),
                                 body: gen_body(&mut rng, i),
--- a/crates/khive-db/sql/023-fts-record-kind.sql
+++ b/crates/khive-db/sql/023-fts-record-kind.sql
@@ -1,0 +1,88 @@
+-- V23: index the granular source-record kind inside the shared FTS tables.
+-- The copy preserves every indexed row, including stale rows whose base record
+-- no longer exists; those rows receive an empty classifier and remain visible
+-- to unscoped text searches exactly as before.
+
+CREATE VIRTUAL TABLE fts_entities_v23 USING fts5(
+    subject_id UNINDEXED,
+    kind UNINDEXED,
+    title,
+    body,
+    tags UNINDEXED,
+    namespace UNINDEXED,
+    metadata UNINDEXED,
+    updated_at UNINDEXED,
+    record_kind,
+    tokenize = 'trigram'
+);
+
+INSERT INTO fts_entities_v23(
+    subject_id,
+    kind,
+    title,
+    body,
+    tags,
+    namespace,
+    metadata,
+    updated_at,
+    record_kind
+)
+SELECT
+    f.subject_id,
+    f.kind,
+    f.title,
+    f.body,
+    f.tags,
+    f.namespace,
+    f.metadata,
+    f.updated_at,
+    coalesce(e.kind, '')
+FROM fts_entities AS f
+LEFT JOIN entities AS e
+    ON e.id = f.subject_id
+   AND e.deleted_at IS NULL;
+
+DROP TABLE fts_entities;
+ALTER TABLE fts_entities_v23 RENAME TO fts_entities;
+
+CREATE VIRTUAL TABLE fts_notes_v23 USING fts5(
+    subject_id UNINDEXED,
+    kind UNINDEXED,
+    title,
+    body,
+    tags UNINDEXED,
+    namespace UNINDEXED,
+    metadata UNINDEXED,
+    updated_at UNINDEXED,
+    record_kind,
+    tokenize = 'trigram'
+);
+
+INSERT INTO fts_notes_v23(
+    subject_id,
+    kind,
+    title,
+    body,
+    tags,
+    namespace,
+    metadata,
+    updated_at,
+    record_kind
+)
+SELECT
+    f.subject_id,
+    f.kind,
+    f.title,
+    f.body,
+    f.tags,
+    f.namespace,
+    f.metadata,
+    f.updated_at,
+    coalesce(n.kind, '')
+FROM fts_notes AS f
+LEFT JOIN notes AS n
+    ON n.id = f.subject_id
+   AND n.deleted_at IS NULL;
+
+DROP TABLE fts_notes;
+ALTER TABLE fts_notes_v23 RENAME TO fts_notes;
--- a/crates/khive-db/src/backend.rs
+++ b/crates/khive-db/src/backend.rs
@@ -802,6 +802,7 @@ impl StorageBackend {
              namespace UNINDEXED, \
              metadata UNINDEXED, \
              updated_at UNINDEXED, \
+             record_kind, \
              tokenize = '{}'\
              )",
             table_key, tokenizer
@@ -1356,6 +1357,7 @@ mod tests {
         let doc = khive_storage::types::TextDocument {
             subject_id: id,
             kind: khive_types::SubstrateKind::Entity,
+            record_kind: None,
             title: Some("Test Title".to_string()),
             body: "This is a searchable document about Rust.".to_string(),
             tags: vec!["rust".to_string()],
@@ -1395,6 +1397,7 @@ mod tests {
         let doc = khive_storage::types::TextDocument {
             subject_id: id,
             kind: khive_types::SubstrateKind::Note,
+            record_kind: None,
             title: None,
             body: "Hello world.".to_string(),
             tags: vec![],
@@ -1534,6 +1537,7 @@ mod tests {
         let doc = TextDocument {
             subject_id: uuid::Uuid::new_v4(),
             kind: SubstrateKind::Entity,
+            record_kind: None,
             title: Some("Title".to_string()),
             body: "Body text.".to_string(),
             tags: vec![],
@@ -1727,6 +1731,7 @@ mod tests {
         let doc = khive_storage::types::TextDocument {
             subject_id: entity_id,
             kind: khive_types::SubstrateKind::Entity,
+            record_kind: None,
             title: Some("Issue1029Repro".to_string()),
             body: "issue 1029 repro body".to_string(),
             tags: vec![],
@@ -1820,6 +1825,7 @@ mod tests {
         let doc = khive_storage::types::TextDocument {
             subject_id: entity_id,
             kind: khive_types::SubstrateKind::Entity,
+            record_kind: None,
             title: Some("Issue1029TwoPools".to_string()),
             body: "issue 1029 two-pool repro body".to_string(),
             tags: vec![],
@@ -1914,6 +1920,7 @@ mod tests {
         let doc = khive_storage::types::TextDocument {
             subject_id: uuid::Uuid::new_v4(),
             kind: khive_types::SubstrateKind::Entity,
+            record_kind: None,
             title: Some("Issue1029Starved".to_string()),
             body: "issue 1029 starvation diagnostic body".to_string(),
             tags: vec![],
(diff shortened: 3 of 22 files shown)


ohdearquant:
Reviewed at head `e625ed5aab53a65f1334fdc6ecbee0274601c62e`. The scoping mechanism is well built and the
relevance/top-K reasoning holds up. The thing worth changing is a design direction rather than a bug:
the classifier is optional by construction and its absence fails toward *invisible*.

## The absent-classifier case fails in the dangerous direction

`TextDocument::record_kind` is `Option<String>` with `#[serde(default)]`
(`crates/khive-storage/src/types/text.rs:70`), and recall requires both `record_kind : "memory"` in the
MATCH and `record_kind IN (?)` as the exact backstop. So a row whose classifier is empty is not
returned — no error, no warning, no count that moves. For a search index that is the wrong failure
direction: an unclassified row silently stops existing, and the only symptom is a result that was
there last week and isn't now.

Compare the alternative: treating an unclassified row as *includable* preserves correctness and costs
only the pruning optimization for rows the migration could not classify. That is the same trade the
exact-equality backstop already makes for trigram correctness, applied to the one case where being
wrong is unrecoverable by the caller.

To be clear about how reachable this is today, because it changes the ask rather than the direction —
I chased each population the migration leaves unclassified and none is a live regression:

- **Soft-deleted base rows.** V23's `LEFT JOIN ... AND e.deleted_at IS NULL` gives these `''`. But
  hydration already drops deleted notes (`crates/khive-pack-memory/src/handlers/common.rs:1246` and
  `:1578`), so before this change they were retrieved and then discarded. Pruning them at the index is
  an improvement, not a loss.
- **FTS rows with no base match.** Orphans with nothing to hydrate; they could not be returned
  usefully before either.
- **Live base rows with no FTS row.** Unchanged by this diff — not findable before or after.
- **A live row whose FTS row carries an empty classifier.** This is the only case that would be a real
  regression, and it needs a writer that omits the classifier. Every in-repo production write path
  stamps it through the canonical constructors. I also checked whether a soft-deleted note could be
  restored after migration and inherit its `''` — it cannot: entity/note un-delete is *guarded
  against* rather than provided (`crates/khive-runtime/src/curation.rs:4977` asserts a CAS refusal
  precisely so a stale snapshot cannot "silently resurrect a deleted entity"), and the
  `ON CONFLICT ... DO UPDATE SET deleted_at = NULL` resurrection path is edges, which have no FTS row.

So this is a latent hazard rather than a live loss, and the ask is to close it structurally rather
than to hunt for lost rows. The cheapest closure is to stop letting the classifier be absent: make it
non-optional on `TextDocument` so every writer must supply it and a missing one is a compile error
rather than an invisible row. `#[serde(default)]` is the part that makes this durable — a document
serialized by any producer that predates V23 deserializes to `None` silently, so the optionality is
not only a local convenience, it is a forward-compatibility hole that fails toward invisible.

Failing that, a classifier-coverage check that can go red is the minimum: today nothing counts
unclassified rows, and the raw population guard counts FTS rows rather than classifier coverage
(`docs/adr/ADR-021-memory-pack.md:194-195`).

## Namespace rename copies the old classifier instead of deriving it

`crates/khive-db/src/stores/text.rs:1644-1692` selects the existing optional value and writes it back
unchanged, so a `''` survives a rename rather than being repaired. Worth noting the reach honestly:
whole-tree grep finds only the definition at `text.rs:1563` with no non-test callers, so this is not
reachable in current production and should not block. But it conflicts with ADR-062's "every FTS
write" wording, and it is the one write path that had a free opportunity to repair a row and declines
it.

## What holds up

- **The relevance claim is sound, and stronger than "zero-weighted".** All public lexical terms are
  wrapped as `{title body} : (...)` at `text.rs:630-641`, so a user term cannot reach the classifier
  column through the public path at all — including prefix and trigram forms. The
  `bm25(0.0, 0.0, 1.0, 1.0, 0.0, ...)` configuration at `:645-651` is then a second, independent
  guarantee. Two mechanisms, not one, which is the right shape for a claim like this.
- **Top-K is genuinely preserved.** Ranked SQL uses the hidden `rank MATCH` with `ORDER BY rank` and
  `LIMIT` (`text.rs:1052-1056`), and rank-within-cap uses the same (`:1485-1499`). An exposed
  `ORDER BY bm25(...)` would have defeated the optimization; none is present.
- **Migration failure is atomic.** Each migration runs as an IMMEDIATE transaction with the SQL before
  the ledger insert (`migrations.rs:1262-1274`, `:1369-1375`), so a failed V23 rolls back rather than
  leaving a half-classified table.
- **Reindex is a real repair path** — it enumerates live records through the canonical constructors
  (`crates/kkernel/src/reindex.rs:399-448`) and replaces empty rows. Worth stating in the docs that
  FTS5 `rebuild` is *not* equivalent: it rebuilds postings only (`text.rs:1315-1332`).

## Tests

The added tests are real and each names a mutation it would catch — classifier filtering across
ranked/unranked/rank-cap, count and DF leakage, the short-kind fallback, and the bounded-candidate
plan. What is missing is the case above: a live base row paired with an empty-classifier FTS row, run
through the *public* scoped query rather than raw migration rows. The migration test asserts empty
classifiers for deleted and stale fixtures (`migrations_tests.rs:1151-1159`) but never asks what a
scoped search then returns, so the silent-exclusion behaviour is currently unobserved by the suite in
either direction.

Also worth a test: re-running the migration. Present coverage is ledger idempotence only
(`migrations_tests.rs:989-1005`), which checks that V23 is skipped, not what classifier state a second
run would leave.

## Docs

ADR-015 (`:210-216`) registers V23 and describes the empty-unmatched behaviour but omits that only
*existing* FTS rows are classified and missing rows are not created. ADR-062 (`:207-225`) documents the
two-part MATCH/equality contract accurately, but its "every FTS write" wording is broader than the
code delivers given rename and the optional field. The text-retrieval notes describe the scope and term
statistics correctly and omit the unclassified-row exclusion.

No builds, tests, benchmarks or database queries were run for this pass; every claim above was checked
against source at the head named, and the production-shaped measurements in the description were not
relied on or repeated.

ohdearquant:
The migration and the scoped filter both hold up.

The load-bearing safety property is that V23 drops and renames the FTS tables
(`crates/khive-db/sql/023-fts-record-kind.sql:45-46` and `:87-88`), which is only acceptable if it
cannot half-apply. It cannot: `run_migrations_locked` executes the file with
`tx.execute_batch(migration.up)`, records the ledger row, and commits, all on the same transaction
(`crates/khive-db/src/migrations.rs:1369-1413`). A failure anywhere in the batch rolls the whole
thing back, and an already-applied version is skipped on rerun via the ledger check.

The unclassified-row observation is worth keeping as a follow-up rather than a hold. Rows that end
with an empty classifier are exactly those whose join found no live base record, so they belong to
records that are already gone; scoped retrieval excluding them is the intended behaviour rather
than a loss. What is missing is the test that pins it — a live base record paired with a same-ID
row carrying a NULL or empty classifier — and a documented line saying scoped retrieval fails
closed on that state.

No blocking findings.

oceanwaves630 (approved):
Approved at 512c37ff. Required checks green (Secret scan, CI gate); reviewed at head.

PR #2329, opened 2026-08-30
fix(code): gate direct findings ingestion

### LINKED ISSUE #2048: code-ingest and git-ingest write through storage with namespace authorization only — no gate request is constructed
## Observation (at main f0eb4d22c07c)

The bulk ingest paths authorize a namespace and then write directly through storage traits:

- `crates/kkernel/src/code_ingest.rs:216` — `.authorize(resolved_ns)` only
- `crates/kkernel/src/git_ingest.rs:60` — `.authorize(resolved_ns)` only

No `GateRequest` is constructed anywhere on these write paths; the only `GateRequest` references under `crates/kkernel/src/` are in test code (`coordinator/tests.rs:64`, `exec.rs:5755`).

## Why this matters

These writes are invisible to any policy expressed at the gate seam. A policy scoped to deny a specific bulk-ingest write cannot observe or block these paths today, because they bypass verb dispatch entirely by design. That makes ingest a third write route alongside registry dispatch and the coordinator intercept, with a strictly weaker authorization check (namespace mint only).

## Ask

Either route ingest writes through the same gate seam as verb dispatch (constructing a gate request per logical write or per batch), or document the bypass explicitly as part of the gate contract so policy authors know ingest is out of scope.

ohdearquant:
> AI-assisted contribution: OpenAI Codex

## Summary

- wrap the direct findings-ingest batch in a stable code.findings_ingest operation at the normal intercepted gate and audit seam
- expose only bounded input-kind and record-count metadata to policy while retaining namespace authorization as defense in depth
- ensure gate denial or evaluator failure occurs before runtime setup and before entity, note, or edge writes
- document that git ingestion already routes its create/link records through the registry and therefore did not need a second batch gate

## Verification

- cargo test -p kkernel code_ingest::tests (23 passed)
- cargo test -p kkernel --test code_ingest_cli (7 passed)
- cargo test --workspace (all tests and doctests passed)
- cargo check --workspace --all-targets
- cargo clippy --workspace --all-targets -- -D warnings
- cargo fmt --all -- --check
- git diff --check

Closes #2048


ohdearquant:
The gate itself is built correctly, and I checked the mechanism rather than taking the summary's word
for it. One thing should be fixed before this lands, and it is in the drain path rather than the gate.

## What I verified holds

**The interception seam really does gate before the closure runs.** `dispatch_intercepted_with_metadata_with_identity`
(`crates/khive-runtime/src/pack.rs:1708`) builds the gate request from the verb *string*, calls
`self.gate.check`, and on `GateDecision::Deny` returns `RuntimeError::PermissionDenied` before
`dispatch(...)` is ever invoked; a gate `Err` routes to `gate_unavailable_error` and also returns.
Both directions fail closed. Notably it performs no registry lookup of the verb, which is what makes
a synthetic `code.findings_ingest` work here — the "unknown verb" rejections at `pack.rs:1628` and
`:2574` are on the ordinary dispatch path, not this one.

**The denial test asserts outcomes, not absence of a crash.** It pins exactly one gate consultation
per batch, the verb name, the bounded `record_counts` args, and then opens the database and asserts
`entities`, `notes` and `graph_edges` are each empty. That last assertion is the one that would catch
a reordering regression, and it is the one most such tests omit.

**`drop(registry)` is a real fix, not noise.** `build_code_ingest_registry` hands `runtime.clone()` to
`register_packs`, so the registry holds a handle clone; without the explicit drop, `drop(runtime)`
would not close the write queue and the drain would block to its timeout. Good catch.

**The git-ingest claim in the amendment checks out independently.** I grepped `git_ingest.rs` for
direct store use (`.entities(` / `.notes(` / `.graph(` / the three `*Store` traits) and got zero hits,
with the same pattern returning 5 in `code_ingest.rs` as a live control. So "each logical create/link
write goes through `VerbRegistry` dispatch" is accurate, and the ADR text matches the code.

## The new `?` escapes this file's own writer-drain invariant

```rust
let runtime = KhiveRuntime::new(cfg).map_err(|e| anyhow::anyhow!("{e}"))?;
let registry = build_code_ingest_registry(&runtime)?;   // new
```

The comment thirty lines below states the rule this breaks: the drain must happen "on BOTH paths, so
an early error return cannot leave the file still moving." `build_code_ingest_registry` can fail —
`PackRegistry::register_packs` on a bad pack set, `builder.build()` on a registration conflict — and
when it does, the function returns with `settle_writer_drain` never called and
`take_writer_task_join_or_warn` never taken.

This is reachable through configuration, and it is specifically the state `WRITER_DRAIN_TIMEOUT` was
introduced to make loud: the runtime already exists at that point, so its pool and writer task exist,
and construction has already run migrations. Dropping the runtime on the way out closes the queue but
does not *await* the task's exit, which is the whole property the bounded timeout provides. The
process can return, and exit, with the database file still in motion.

The earlier `KhiveRuntime::new(cfg)?` is fine by contrast — no runtime exists yet, so there is nothing
to drain. The registry line is the only new early return that sits inside the guarded window.

Binding the result and letting it fall through to the existing drain, rather than `?`, keeps the
invariant intact and is roughly a one-line change.

## One precision point on the ordering claim

"A deny or gate outage returns before the runtime setup closure and before any entity, note, or edge
write" is accurate as written — `runtime_setup` is inside the closure, and the test proves the three
tables stay empty. Worth being explicit that this is narrower than it may read: `KhiveRuntime::new`
and pack registration both run *before* the gate, so a denied ingest still creates the database file
and applies migrations. The denial test demonstrates this incidentally — it can only count rows in
those three tables because the tables exist. That is a defensible design, but it is a different
guarantee from the "zero filesystem effect" the validation path above it claims, and the amendment
reads as though they were the same.

## CI

Both reds on this PR reduce to one cause: the aggregated gate reports `Gate failure — jobs not green:
secret-scan`, with `ci`, `docs`, `doc-build`, `coverage-ratchet`, `supply-chain`, `data-leak-guard`,
`wasm-parity`, `vamana-portability`, `minio-blob-compat`, `kg-editor` and `marketplace` all green.
Nothing in this diff caused it.

No builds or tests were run for this pass; every finding is from source reading, and the verification
list in the description was not relied on.

ohdearquant (commented):

Verdict on head 3599799d7e3fef1cec5f41ae41009164b2cdf034: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving at f83904b8af5059726ed25ffacb3ffec89b848d62. Checks are green at this exact commit. I confirmed at this head that the ingest registry attaches the runtime event store, so gate consultations on this path persist an audit event, and that the audit-store selection is a named variant rather than a bare boolean at both call sites.

PR #2330, opened 2026-08-30
fix(deps): close lattice 0.9.0 follow-ups

### LINKED ISSUE #1896: lattice 0.9.0 bump follow-ups: blake3 pure-feature unification, stale version metadata, dead lru advisory waiver
Follow-ups from the lattice 0.9.0 / lru 0.18.2 dependency refresh (#1894), none blocking that merge but all worth resolving before the next release cut.

## 1. lattice-embed 0.9.0 forces blake3 `pure` workspace-wide (performance)

lattice-embed 0.9.0 declares its blake3 dependency with `features = ["std", "pure"]`; 0.8.0 declared no extra features (crates.io dependency metadata for both versions). Because cargo unifies features across the graph, the workspace's single blake3 1.8.5 instance is now compiled with `pure` — the portable Rust implementation — replacing the assembly/SIMD/NEON paths for every in-tree consumer, including content hashing in the blob store and khive-db.

Resolution options:
- upstream: gate `pure` behind lattice-embed's wasm feature (where it is presumably needed) and pick up a patch release; or
- benchmark the hashing paths under `pure` and record an explicit accepted-risk if the delta is negligible.

Either way the release cut should not ride an unmeasured hashing regression silently.

## 2. khive-pack-moodboard hard-codes engine version strings

- `src/model.rs:89` (and test expectations at 1251/1263) report `version: "0.8.0"` for the inference engine
- `src/preference.rs:32` pins `FANN_CRATE_VERSION = "0.7.1"`

Both now disagree with the pinned lattice crates (=0.9.0). Version metadata should be derived from the dependency (or a single const tied to the workspace pin) rather than free-standing string literals that rot on every bump.

## 3. deny.toml carries a dead advisory waiver

`crates/deny.toml` `[advisories].ignore` still contains the RUSTSEC-2026-0002 entry with a rationale block written for lru 0.12.5. The workspace now pins lru 0.18.2, where that advisory (and RUSTSEC-2026-0253) do not apply, so the ignore entry and its comment block are dead config and should be removed. Stale waivers are worse than none: they read as if the advisory still applied and mask a future regression to an affected version.

ohdearquant:
## Summary

- centralize the moodboard lattice provenance version and pin it to every exact lattice workspace dependency with a regression test
- remove the obsolete `lru` advisory waiver after confirming the resolved 0.16.4 and 0.18.2 versions are patched
- document the workspace-wide `blake3/pure` feature-unification risk, its bounded 0.9.0 acceptance, and the merged upstream 0.10.0 exit
- add a standalone locked feature-comparison harness so the default and pure backends can be reproduced without Cargo feature unification contaminating the comparison

## BLAKE3 disposition

On Apple M2 Max, the isolated release-mode ABBA measurement found no stable 64-byte regression, but found a material 50.6-54.0% throughput reduction for 4 KiB through 64 MiB inputs (about a 2.0-2.2x CPU-hashing slowdown). This PR therefore records an explicit accepted risk only for the exact lattice 0.9.0 pins; it does not characterize the impact as negligible. The exit is the first compatible published lattice release containing upstream lattice#1403, which removes BLAKE3 from `lattice-embed`. The benchmark record also states its shared-host, single-architecture, and microbenchmark limitations.

## Test plan

- [x] `cargo test --workspace`
- [x] `cargo test -p khive-pack-moodboard --all-targets`
- [x] `cargo deny check advisories` with cargo-deny 0.19.9
- [x] default and pure harness `cargo check --release --locked`
- [x] default and pure harness `cargo clippy --release --locked --all-targets -- -D warnings`
- [x] harness `cargo fmt --all -- --check`

Closes #1896


ohdearquant:
Good PR, and the BLAKE3 section is the part worth calling out first: finding a 50-54% throughput
reduction and recording it as an explicitly accepted, pin-scoped risk with a named exit condition —
rather than rounding it to "no significant change" — is the right way to land this. Stating the
shared-host, single-architecture and microbenchmark limitations in the record itself means the next
reader can tell what the number does and does not support.

## Verified independently

**The advisory waiver removal is correct.** I did not take the description's word for it: the lockfile
at this head resolves `lru` to 0.16.4 and 0.18.2, and the removed waiver's own comment names 0.16.3 as
the fix version for RUSTSEC-2026-0002. Both resolved versions are past it, so the ignore entry is
genuinely obsolete rather than merely inconvenient, and dropping `ignore` to `[]` is right.

**The centralization actually happened.** `LATTICE_VERSION` has use sites in `lib.rs`, `model.rs`,
`preference.rs`, `preference_artifact.rs` and `preference_handlers.rs`, and the only surviving `0.9.0`
literal in the crate is prose in a `preference.rs` doc comment ("`lattice-fann` 0.9.0's
`BackpropTrainer` optimizes MSE"). That one is not provenance data so it does not affect correctness,
but it is the one instance the constant cannot keep in sync, and it will drift silently at the 0.10.0
exit — worth phrasing so it does not name a version, since that line is about behaviour rather than
about which version is pinned.

I also like that the assertion compares against `format!("={LATTICE_VERSION}")` rather than the bare
number: it enforces the exact-pin form, not just the version, which is the property that actually
matters here.

## The regression test hardcodes the population it claims to cover

```rust
for dependency in ["lattice-embed", "lattice-inference", "lattice-fann"] {
```

The summary describes this as pinning "every exact lattice workspace dependency." Today that list is
complete — I enumerated `workspace.dependencies` at this head and there are exactly three `lattice-*`
entries, matching the array. So this is not a live defect.

But it is a drift detector whose population is a literal, and the drift it cannot detect is the one
most likely to happen: a fourth lattice crate added to the workspace. That crate would get no pin
assertion, the test would stay green, and the green would read as coverage. The failure mode is
invisible precisely because the test passes.

Deriving the list instead — filter `dependencies` for keys starting with `lattice-`, assert the
result is non-empty so it cannot pass vacuously if the prefix or the table shape ever changes, then
run the existing per-crate assertion over it — costs a few lines and makes the test's population match
the sentence describing it. The non-empty assertion is the important half: without it, a derived
filter that silently matches nothing is worse than the hardcoded array.

## Minor

The `toml::Value::Table` arm reads `specification["version"]`, which panics on a lattice dependency
declared without an explicit `version` key (a path or git dependency, say). Panicking in a test is an
acceptable outcome, but the message would be an index panic rather than the `expect` text just below
it; folding that into the same `expect`-style error would make a future misconfiguration diagnose
itself.

No builds or tests were run for this pass; the lockfile and manifest facts above were read at the PR
head, and the benchmark numbers in the description were not re-measured.

ohdearquant:
CI triage on this head, since four checks are red and only one of them is yours.

- **Docs lint — yours, and trivial.** `Found 1 not formatted file in 207 files`; the diff it prints is
  column alignment in the new benchmark table (`docs/benchmarks/blake3-pure-0.9.0.md`). Running the
  repo's markdown formatter over that file clears it.
- **CI (ubuntu-latest, shard 1/2) — not yours.** The failure is
  `handlers::recall::tests::keyword_only_widening_ignores_vector_leg_candidates`, panicking at
  `crates/khive-pack-memory/src/handlers/recall.rs:6987`. Nothing in this PR touches
  `khive-pack-memory`. That test lives at `recall.rs:6837` on `main` and its body spans the panic
  line, and PR #2305 ("test(memory): stabilize keyword widening fixture") rewrites exactly that
  region — its hunks cover 6776-7012. So this is the known unstable fixture that #2305 is fixing, not
  a regression from this change.
- **Secret scan (gitleaks) — not yours.** Repo-wide, currently red across the open queue; #2323 scopes
  the scan to the change under review.
- **CI gate — aggregate.** It fails because the two above fail; it carries no independent signal.

So the only thing to act on here is the markdown formatting. Worth rerunning the shard after #2305 and
#2323 land rather than investigating the memory-pack failure from this branch.

ohdearquant:
Correcting my previous comment on one point, because it pointed you away from a failure that is
actually yours.

I wrote: "**Secret scan (gitleaks) — not yours.** Repo-wide, currently red across the open queue."
Both halves are wrong.

**It is not queue-wide.** Across the 54 currently open pull requests the secret scan is SUCCESS on 31
and FAILURE on 22. It is branch-dependent, so "everyone is red" was never a safe attribution.

**It is this branch's.** This branch's merge base is `a46914ae`, which is exactly current `main` — zero
commits behind — and its `.gitleaksignore` is byte-identical to main's. `main` at `a46914ae` passed
the secret scan on its own CI run. So the branch is main plus this PR's commits, scanned with the same
allowlist, and it fails where main passes. The finding is in the commits this PR adds.

What the log does and does not tell you: the job runs `gitleaks detect --source . --redact
--no-banner` and reports `4637 commits scanned` and `leaks found: 1`. Because of `--redact`, the log
names no file, line, or rule — which is why my first read had nothing to attribute and I filled the
gap with the wrong inference instead of saying so.

I could not identify the specific finding without running gitleaks, which I did not do. Two things
that may help you find it quickly:

- The only file this PR adds that contains long hex runs is `bench/blake3-feature-compare/Cargo.lock`
  (10 of them). That is a natural candidate for the `generic-api-key` rule, though `crates/Cargo.lock`
  carries 446 such runs on main and passes, so the path or surrounding context evidently matters.
- The repo allowlists by exact `commit:path:rule:line` fingerprint in `.gitleaksignore`, and all four
  existing entries are `generic-api-key` matches on confirmed non-credentials — two test fixtures and
  two ADR documentation lines. If your finding is the same kind of false positive, the fix is a new
  fingerprint entry rather than a code change. Note the file's own comments record that a line-number
  shift once re-attributed a finding to a new commit and required a second pair of entries, so a
  fingerprint is tied to the exact commit and line.

The rest of my earlier triage stands: Docs lint is a real formatting failure in the new benchmark
table, the shard failure is the unrelated `keyword_only_widening_ignores_vector_leg_candidates`
fixture that #2305 rewrites, and the CI gate is an aggregate with no independent signal. My review
comment above this is unaffected.

ohdearquant:
My correction above is itself wrong, and since it told you to go hunting, here is the retraction with
the measurement that settles it.

I claimed the finding had to be in this PR's own commits, reasoning that the branch is `main` plus its
commits, `main` passes, so the difference must be yours. That inference assumes the scan's population
is the branch's history. It is not.

- The `main` run at `a46914ae` reports **4610 commits scanned**, no leaks.
- This PR's run, whose merge base is that same `a46914ae`, reports **4637 commits scanned**, 1 leak.
- This PR contributes **one** commit.

27 additional commits were scanned, not 1. `gitleaks detect --source .` walks every ref present in the
checkout, and a pull-request checkout carries refs that a push build does not. So the scanned set
includes commits from branches unrelated to this PR, and the finding cannot be attributed to your
change — or ruled out — from anything in that log. The honest status is UNRESOLVED in both directions,
which is what I should have written the first time instead of guessing twice.

For what it is worth, #2323 is precisely a fix for this: it bounds the scan to the pull request's own
commit range and adds `-v`, so a finding names its file, line, commit and rule instead of reporting
only `leaks found: 1`. Its own rationale describes this exact failure — a finding on any pushed branch
turning the job red on unrelated pull requests with no way to tell which change is at fault. That
matches what I measured here.

So: nothing for you to chase on the secret scan. The two real items from my first comment stand — the
markdown formatting in the new benchmark table is yours, and the `keyword_only_widening_ignores_vector_leg_candidates`
shard failure is the unstable fixture #2305 rewrites.

Apologies for the churn on your PR.

ohdearquant:

`Docs lint` fails at `Check markdown formatting`: `docs/benchmarks/blake3-pure-0.9.0.md` are not `deno fmt` clean.
wrapping, no content change.

ohdearquant:
Two small things, both in the same direction: the change is fine and the artifacts around it are
slightly out of step.

## The docs gate is red on the new benchmark file

`Docs lint` fails at this head. `deno fmt --check` wants padded Markdown table rows at
`docs/benchmarks/blake3-pure-0.9.0.md:41-46`, and the job ends with `Found 1 not formatted file in
207 files`. The file itself is new, so this is introduced rather than inherited. Applying the
repository's Deno formatting to that table clears it.

## The advisory ignore list is now empty but the comment above it still describes entries

`crates/deny.toml:29` is now `ignore = []`, while the comment immediately above it still carries the
"ignore list ... now ACTIVE" wording and the rationale that belonged to the removed
RUSTSEC-2026-0002 entry. The base had that wording with a populated array below it, where it read
correctly; with the array emptied, the comment describes something that is no longer there.

This is maintainer-facing text rather than a runtime path — `.github/workflows/ci.yml:339-361` runs
`cargo-deny check advisories` against this file — so it is not urgent. It is worth fixing while the
context is fresh, because a stale rationale next to an empty list is the kind of thing that later
gets read as permission to re-add an entry without re-deriving why.

authored it, so its output was treated as input evidence only. The failing docs gate and the
`ignore = []` change against the base's populated array were re-derived from the ref before posting.

ohdearquant (commented):

Verdict on head 73d45c619a1ba57bdedb05a91e7be56ca7ae79e7: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head a864d3fd987d0ceb47239719d59e8e33215d1490: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):

PR #2331, opened 2026-08-30
fix(runtime): resolve committed write audit outcomes

### LINKED ISSUE #2256: Write verbs hard-fail on AdmissionDeadlineExpired after the domain write has already committed
## Summary

A write verb whose audit obligation returns `AdmissionDeadlineExpired` fails the dispatch, but the domain write has already committed. The caller is told the operation failed when it succeeded. `create` and `link` are not idempotent, so a caller that believes the error and retries produces a duplicate entity or a duplicate edge, and neither dispatch can observe the duplicate because the first one reported failure.

This is the inverse of the invariant ADR-133 D2/D3/D4 protects. That invariant is one-sided by construction, and the code says so.

## Mechanism, read at `52364b96763e79803c7f1393285a0a403184face`

The domain write runs first and its result is already in hand when the audit row is attempted:

- `crates/khive-runtime/src/pack.rs:1798-1811` — `producer` is derived from `result.is_ok()`, then `append_audit_event_best_effort` is called. The audit event itself is built from the dispatch's returned value at `:1750`.
- `crates/khive-runtime/src/pack.rs:1712-1724` — the audit outcome is folded back in with `result = fold_audit_obligation(result, audit_outcome)`.
- `crates/khive-runtime/src/pack.rs:3798-3802` — the fold: `(Ok(_), Err(audit_err)) => Err(audit_err)`.

Its own doc comment at `:3789-3793` states the intent plainly:

> A dispatch that would otherwise report success cannot claim it once the row accounting for it fails to commit (ADR-133 D2/D3/D4), so `Ok` becomes the audit's `Err`. ... the obligation is on never reporting a false success, not on replacing one error with another.

So the trade is deliberate. What is missing is the other half: nothing states an obligation against reporting a false *failure*, and for non-idempotent verbs a false failure is not benign.

The audit append is a separate call (`store.append_event` at `:3765`), not part of the domain write's transaction, so nothing rolls the domain write back.

## Why the #2228 degrade does not cover this

`#2228` (185e2859f) added `AdmissionDeadlineExpired` as a distinct variant and degraded read verbs under admission pressure. That degrade is scoped, deliberately, to a *successful* dispatch of a verb on `ADMISSION_DEGRADE_SAFE_VERBS`. `pack.rs:3657-3669`:

> every failed dispatch, every write verb, every gate-denial/unknown-verb/git.digest row stays strictly obligation-bearing.

So writes take the `:3745-3753` path and hard-fail. That is the current design, working as written.

The sharp part is what the variant means. `crates/khive-runtime/src/audit_batch.rs:110-126`:

> The row was already enqueued ... it remains enqueued and unresolved, and is committed (or terminally failed) by the generation driver independently of this caller's timeout, so the caller cannot tell from this reason alone whether the row eventually committed ... Retrying is only safe for an idempotent caller.

A landed, non-idempotent write is therefore failed on a condition that is explicitly *unresolved* rather than a known loss, and the retry the error invites is the one the variant's own doc calls unsafe.

## Observed

Against an installed release build, a batch of graph writes returned, for ten ops (two `create`, eight `link`):

```
internal: audit obligation commit failed for verb "create": AdmissionDeadlineExpired
internal: audit obligation commit failed for verb "link": AdmissionDeadlineExpired
```

All ten domain writes had committed. Verified afterwards by re-read: the created concepts resolve by id with `created_at` timestamps matching the run, and the eight edges appear in `neighbors()` with edge ids. A separate op in the same batch failed on relation validity and said so in those terms, so the store distinguishes the two cases internally while the caller cannot.

This also supplies the evidence #2248 flagged as missing. That issue records the commit behaviour as derived-from-source and explicitly untested; these ten ops are the observation, with the records read back.

## Relationship to the existing issues

- **#2248** describes the two arms as returning the same variant. That was already stale when filed: #2228 split them the previous evening. Its *consequence* section is what remains live, and is restated here in post-split terms.
- **#2117** covers the same conflation before the split.
- **#2147** / **#2217** cover the read-verb side, which #2228 addressed.

What none of them states in current terms: after the split, a write verb still hard-fails on the unresolved variant, and the domain write has already landed when it does.

## Suggested direction

The read/write asymmetry is sound, so the fix is probably not to extend the degrade to writes. Two narrower options:

1. Make the error say what is actually known. A write that committed and whose audit row is unresolved is a different fact from a write that failed, and the current message is indistinguishable from a genuine dispatch failure. Surfacing the committed domain result alongside the unresolved-accounting condition lets a caller avoid the unsafe retry.
2. Resolve the row rather than abandoning it. If the caller's bounded wait is the only thing that elapsed, and the generation driver goes on to commit or terminally fail the row regardless, the dispatch could wait on the resolved outcome for obligation-bearing writes specifically, where the cost of a false failure is highest.

Either way, a test that forces the deadline arm on a write verb and then asserts both the domain row and the audit row would pin the behaviour, which is the same test #2248 asks for.

ohdearquant:
Authored with OpenAI Codex.

## Summary

- keep the same enqueued audit receiver alive after the admission wait threshold for successful non-degrade-safe operations
- preserve bounded admission behavior for allowlisted reads, failed/denied outcomes, pure observability, and pre-enqueue queue exhaustion
- make git digest receipts use the same resolved-outcome contract
- record the distinction in ADR-133 and the audit batch API documentation

The write handler is never rerun and its audit row is never re-enqueued. A committed domain effect now reports success only after that generation commits, while genuine terminal audit failures remain failures.

## Test plan

- [x] focused TDD regression: committed non-idempotent write remains pending past the audit deadline, then succeeds once its single audit row commits
- [x] `cargo test -p khive-runtime --all-features`
- [x] `cargo test --workspace` including doctests
- [x] `cargo check --workspace --all-targets`
- [x] `git diff --check`

Closes #2256

---

## Contract change, added 2026-09-11 after merging main — read this before merging

Merging `main` exposed a contract decision this branch makes for **every verb**, not only for the
receipt verb. It is stated here rather than resolved by a test edit.

**Two deadlines, two different meanings.** `submit_until_resolved` keeps waiting past the
**admission** deadline, and that wait is itself bounded by a **resolution** deadline. Those two
expiries are not the same fact:

- **Admission expiry**: the row is enqueued and the seam keeps waiting. The generation commits it
  independently of this caller, so the dispatch reports its committed result and the row counts on
  the unresolved-obligation counter. Unchanged.
- **Resolution expiry**: the caller waited for the real outcome and never received one. The audit
  commit is **unconfirmed, not proven absent**. Reporting success there would tell the caller a
  record exists when nobody knows whether it does.

**So an ordinary verb now receives what the receipt verb already received**: `AuditObligation` with
`ResolutionDeadlineExpired`, `DomainDisposition::Committed`, the domain result retained, and **no
retryable failure context** — the domain effect committed, so a retry would write twice. The handler
is never replayed and the row is never re-enqueued; the original driver work continues on its own.

Before this branch, an ordinary committed write whose audit row was merely slow reported success and
counted as unresolved. That remains true at the admission deadline. It changes only once the caller
has also waited out the resolution deadline, which before this branch could not happen because no
such wait existed.

**The counter split is part of the contract.** An expired resolution wait propagates as the
dispatch's own error, so it lands on the propagated-obligation counter. The unresolved-obligation
counter is for dispatches that **succeeded** and left a row enqueued past the admission deadline,
and this one did not succeed. The fixture asserts that the second counter does **not** move, which
is what keeps the two facts from collapsing into one.

### Also in this range

`main` added a closed `wire_code` mapping over `AuditTerminalReason` while this branch added three
variants, so the merged tree did not compile (`E0004`). Arms added in enum order:
`resolution_deadline_expired`, `driver_append_abandoned`, `store_wedged`.

The disposition fixture set `admission_deadline` to 30ms with `..Default::default()`, leaving
`resolution_deadline` at its 30s default, so the fixture's own 1s timeout expired before dispatch
returned. It now sets both.

### Verification of this range

Pinned 1.95.0, `KHIVE_PACKS` and `KHIVE_ADDITIONAL_EMBEDDING_MODELS` unset:

- `cargo test -p khive-runtime --no-fail-fast` — 1587 + 23 + 99 + 12 + 8 + 4 + 2 + 1 + 1 + 1 + 1 passed, 0 failed
- `cargo clippy -p khive-runtime --all-targets -- -D warnings` — 0
- `cargo fmt --all -- --check` — 0

Both fixture arms now assert the obligation reason, the `Committed` disposition, the retained domain
result, the absence of retryable context, exactly one domain write, exactly one eventual audit row,
and the counter split.


ohdearquant:
The problem this fixes is real and the diagnosis is right: reporting a committed non-idempotent write
as failed, on a deadline that only describes the audit lane's own admission, is a genuine defect and
the retry it invites is genuinely unsafe. The refactor is also clean — `submit` now delegates to
`submit_with_wait_policy(row, false)`, and that path is byte-for-byte the old behaviour including the
comment explaining why `AdmissionDeadlineExpired` is distinct from a queue-full refusal.

What I want to push back on is the specific remedy, because I think it treats the wrong half of the
problem as the defect.

## The bound was not the bug; the reason was

After the deadline elapses, `submit_with_wait_policy` falls through to a bare `rx.await` with no
timeout. With `admission_deadline` defaulting to 5s, this converts a bounded 5-second wait into an
unbounded one for `DispatchSucceeded && !degrade_allowlisted` — which is the ordinary write path, not
an edge case.

The stated justification is that returning `AdmissionDeadlineExpired` "would report a false operation
failure and invite an unsafe retry." That is correct, but the harmful property there is that
`AdmissionDeadlineExpired` is *indistinguishable from a failure for which retry is safe*. Waiting
forever does not remove that ambiguity — it relocates it to the client, and makes it worse. A caller
that eventually gives up on its own transport timeout gets a cancellation carrying no
`AuditTerminalReason` at all, which is strictly less information than the reason it replaced, and it
invites exactly the same unsafe retry of the same committed write. The one place a caller could have
been told "your effect committed, do not retry" is the return value that this change removes.

A distinct terminal reason meaning "domain effect committed, audit outcome unresolved, retry is not
safe", returned on a longer but finite bound, would preserve the entire safety property this PR is
after while keeping the caller bounded. That seems strictly better than an unbounded await, and it is
a smaller change than the one already made.

Worth being precise about how narrow the hang actually is, because the surrounding code is careful:
every join outcome in `supervisor_loop` resolves its waiters — `DriverPanicked` / `DriverCancelled`
on a join error, `Err(reason)` on `GenerationResult::Failed`, per-row dispositions on `Committed`, and
`fail_driver` on a length mismatch. So a driver that *dies* still unblocks the caller. The unbounded
case is specifically a driver that stays alive and never completes its generation — `child.await`
inside the supervisor has no timeout either — which is the wedged-store case. That is the one where a
5s answer was doing real work.

(I looked for an outer per-request bound on the serve path and did not find one; that is
"not found by this grep" rather than a claim it does not exist, so please correct me if a transport
timeout bounds this elsewhere.)

## Nothing covers the arm that actually changed

`write_verb_waits_past_audit_deadline_until_row_commits` is a good test — it holds the generation,
asserts the dispatch is still pending 80ms past the deadline, then releases and asserts it resolves
with the committed `{"created": true}` and the expected counts. It would redden on a revert to
`submit`, and the pending assertion fails in the safe direction on a slow machine rather than flaking.

But it only covers the generation that *eventually* commits. The behaviour that changed for the worse
is the generation that never resolves, and there is no arm for it. A test that wedges the driver and
asserts the caller still receives some bounded, correctly-labelled terminal outcome is the one that
would have surfaced the concern above — and under the current design it would have to assert the
caller hangs, which is itself a useful thing to have to write down.

## Checked and fine

- The predicate `producer == AuditProducer::DispatchSucceeded && !admission_degrade_eligible` reduces
  to `DispatchSucceeded && !degrade_allowlisted`, which matches the summary exactly. Failed and denied
  outcomes keep the bounded wait, correctly — their caller-visible result is already fixed.
- `persist_git_digest_receipt` switching unconditionally looks asymmetric next to that predicate, but
  it early-returns `PersistenceUnavailable` unless `result` is `Ok`, so it is success-only by
  construction, and its doc comment already declares it "intentionally strict while every other
  dispatch audit remains best-effort." The ADR-133 comment above the call site stays true — arguably
  more true than before.
- Pre-enqueue refusals (`PreflightRejected`, `AdmissionClosed`, `QueueAdmissionExhausted`,
  `Lifecycle::Failed`) all still return before the wait, so the "pre-enqueue queue exhaustion"
  carve-out in the summary holds.

No builds or tests were run for this pass; every finding is from source reading, and the checked
boxes in the test plan were not relied on.

ohdearquant:
## The strict path drops the caller's bound entirely after the deadline

Following up on the earlier note — there is a sharper version of it, and it is blocking.

At this head, `wait_until_resolved` waits twice. The first wait is bounded. The second is not:

```rust
// crates/khive-runtime/src/audit_batch.rs:558-570
if wait_until_resolved {
    match tokio::time::timeout(self.config.admission_deadline, &mut rx).await {
        Ok(Ok(result)) => return result,
        Ok(Err(_recv_error)) => return Err(AuditTerminalReason::DriverJoinLost),
        Err(_elapsed) => tracing::warn!(
            ?producer,
            "strict audit obligation remains enqueued after the admission wait deadline; \
             waiting for its real terminal outcome"
        ),
    }
    return match rx.await {          // <- no timeout, no cancellation branch
        Ok(result) => result,
        Err(_recv_error) => Err(AuditTerminalReason::DriverJoinLost),
    };
}
```

The elapsed arm no longer ends the wait — it logs and then falls through to an unbounded `rx.await`.
On `origin/main` the same site is a single bounded wait whose elapsed arm terminates:

```rust
// origin/main:crates/khive-runtime/src/audit_batch.rs:693,702
match tokio::time::timeout(self.config.admission_deadline, rx).await {
    ...
    Err(_elapsed) => Err(AuditTerminalReason::AdmissionDeadlineExpired),
```

So the deadline changed meaning: it used to end the caller's wait, and now it only decides when to
emit a warning. For a `DispatchSucceeded` row outside the degrade allowlist this is the selected
path, which makes a live-but-stalled writer or a generation that never completes hold the caller
indefinitely, and take `close_and_drain()` / `quiesce()` with it.

Nothing upstream restores the bound. The request wrapper installs a task-local context but supplies
no independent abort, and this write is not inside the read-phase deadline, so the request deadline
is not an effective ceiling here.

The direction matters: waiting too long is silent and unbounded and consumes request capacity
exactly when the system is already degraded, whereas ending the wait early is loud and recoverable.
The change trades the recoverable failure for the silent one.

The added regression covers a generation that is *delayed and then commits*. It has no
never-completing driver case, so it passes either way — it cannot distinguish the bounded
implementation from this one.

**Suggested direction:** keep a finite, cancellation-aware bound after the ordinary admission
threshold, and give the caller a distinct outcome meaning *the domain effect committed, the audit
outcome is unresolved*, with explicit no-retry semantics — a generic failure here would invite a
retry of an operation that already committed. Then add a case with a driver that never completes;
that is the test that separates the two implementations.

wrote it, so its output was treated as input evidence only. The three facts above — the unbounded
second await at this head, the base's terminating elapsed arm, and the selection of this path for
ordinary successful dispatch — were each re-derived from the two refs before posting.

ohdearquant (commented):

Verdict on head ac960774c376a560cd6c5d8d17847bd1737cc308: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 08f1e6d263135f4ae212d545ff9f65ff105cb8a6: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 79f4d3fc00981da8ea93d506dde6215b4ddd2328: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

ohdearquant:
Reviewed at `0c92b8bf`. The admission-deadline half of #2256 looks right to me, and the reasoning in the doc comments on `AuditTerminalReason` and the two counters is unusually easy to follow. Two things from running the current `main` build tonight.

**The defect is live, and it does not only surface as `AdmissionDeadlineExpired`.** Under write contention on 0.8.0, two verbs dispatched in one batch both returned:

```
internal: audit obligation commit failed for verb "gtd.complete": StoreFailure
internal: audit obligation commit failed for verb "comm.inbox": StoreFailure
```

`status` was `partial` with `summary.succeeded: 0`. The write had committed: the next call on the same row returned `invalid input: task <id> is in terminal state "done"; no further transitions allowed`. A `comm.send` in the same period returned `AdmissionDeadlineExpired` and was delivered. So the same caller-visible sentence carries at least two terminal reasons, and at least one of them can follow a committed effect.

**The gap that survives this change.** `append_audit_event_best_effort` now routes a successful non-degrade-safe operation through `submit_until_resolved`, which is what removes the false failure at the admission boundary. But when the generation then resolves *terminally* — `StoreFailure`, `IdentityConflict`, `DriverPanicked`, `DriverCancelled` — the caller still gets the unchanged message:

```rust
"audit obligation commit failed for verb {verb:?}: {reason:?}"
```

That sentence does not say the domain effect may already be durable, so a caller reading `ok: false` as "did not happen" retries a committed non-idempotent write. That is the same hazard #2256 describes, moved from the deadline path to the terminal path rather than closed.

The fix is already in this diff for the neighbouring case. `GIT_DIGEST_RECEIPT_FAILURE` says exactly the right thing:

```
git.digest writes may have committed, but no durable success receipt was confirmed;
inspect ingest state before retrying
```

Suggest the general obligation failure carry the same shape whenever the operation's own result was a success, since by construction that is precisely when the effect is durable and the audit is not. Two callers cannot be expected to key retry logic on a `{reason:?}` enum they have no stable contract for.

**Test-population note.** The listed regression covers the deadline arm ("remains pending past the audit deadline, then succeeds once its single audit row commits"). An arm where the generation resolves terminally *after* a committed write would pin the sentence above, and would fail today.

Issue #940, opened 2026-07-13
bench-track: packs component exceeds the 20m job timeout on main, starving packs trend data and mislabeling every main run cancelled


ohdearquant:
## Symptom

27 of the last 30 `Bench Trend Tracking` runs on main conclude `cancelled` (2 success, 1 in progress as of 2026-07-13 morning). Inspection of a representative run (29235108059, main push) shows the actual failure:

- `Component benches / packs / Criterion quick profile`: killed at 20m15s with "The job has exceeded the maximum execution time of 20m0s"
- All other jobs succeeded in the same run: e2e 4m25s, scoring-retrieval 6m49s, storage-brain 10m2s, ann-query 5m15s, with artifacts uploaded

GitHub reports timeout-killed jobs as cancelled, so the whole run concludes `cancelled`. The workflow-level concurrency design is NOT at fault: `cancel-in-progress` is already PR-only (bench-track.yml:49-56) and behaves as documented.

## Impact

1. **packs trend data is missing from the main ledger** for nearly every main push — the packs component (khive-pack-gtd, khive-pack-kg, khive-pack-knowledge, khive-pack-memory, khive-pack-schedule) has outgrown the shared 20-minute budget in `bench-component.yml`. The largest-stable-gaps mining surface (PR #862) is blind on the packs axis.
2. **Signal masking**: every main run reading `cancelled` makes the trend pipeline look supersession-starved when it is actually one component's timeout; real cancellations would be indistinguishable.

## Fix directions (pick after measuring which crate grew)

- Measure the per-crate quick-profile time inside the packs component first; khive-pack-memory has grown substantially (ANN invalidation suites) and is the likely driver.
- Then either: give the packs component a larger `timeout-minutes` (make it a per-component input on `bench-component.yml`), or split packs into two components, or trim the quick-profile bench set for the oversized crate.
- Regardless of choice: a timeout kill of one component should not read as a cancelled run — consider `continue-on-error` semantics or a final status job so main-run conclusions carry signal.

## Evidence

- Run tally: `gh run list --workflow 'Bench Trend Tracking' --branch main --limit 30` → 27 cancelled / 2 success
- Representative durations of "cancelled" main runs: 20-41 minutes wall (not pending-collapses)
- Run 29235108059 job breakdown as above

ohdearquant:
Step-level evidence from the killed job (86767889183) sharpens the root cause — it is the **bench-target compile**, not the benches:

[...]

PR #2332, opened 2026-08-30
perf(db): bound FTS5 segment maintenance

### LINKED ISSUE #1906: perf: FTS5 segment scans dominate daemon CPU; short recall queries take 17-19s on a 173k-note store

ohdearquant:
## Summary

- add bounded, periodic FTS5 incremental-optimize maintenance to the main backend's existing dedicated checkpoint task: one table per 300-second due call, round-robin, one 500-page merge budget, zero busy wait, and no unbounded `optimize`
- resume SQLite-persisted merge work after daemon restart, skip single-segment indexes, keep application writers preferred, and expose process-lifetime maintenance outcome counters
- report both FTS indexes' level/segment structure from the documented `%_data` structure record instead of a corpus-sized `%_idx` count
- add `embed_ms`, `fts_ms`, `ann_ms`, `fresh_tail_ms`, and `hydrate_ms` to completed slow-recall warnings, with explicit max-for-concurrent/sum-for-sequential timing semantics
- document the maintenance, diagnostics, and recall telemetry contracts in ADR-091, ADR-033, and the affected API docs

## Production-shaped evidence

The live database was inspected read-only and never mutated:

- database size: 4.2 GiB
- `fts_notes`: 212,803 rows across 11 segments
- `fts_entities`: 8,848 rows across 10 segments
- `EXPLAIN QUERY PLAN` for `COUNT(DISTINCT segid)` over `%_idx` reports a virtual-table scan, so that approach would make diagnostics proportional to index size
- decoding each index's documented `%_data.id = 10` structure record reports the same segment census with one bounded row read per index

The focused maintenance regression creates fragmented FTS segments, proves MATCH results and row count are unchanged, then proves bounded merge steps converge the index to one segment. Additional tests cover persisted-cycle restart recovery, positive continuation, round-robin selection, threshold skips, and a held-writer busy skip/retry with counter assertions.

## Operational bounds

Defaults are conservative and configurable through `KHIVE_FTS_MERGE_ENABLED`, `KHIVE_FTS_MERGE_INTERVAL_SECS`, `KHIVE_FTS_MERGE_PAGES`, and `KHIVE_FTS_MERGE_MIN_SEGMENTS`. Maintenance is main-backend-only and independent of checkpoint health: errors are counted/logged without invalidating an otherwise healthy checkpoint connection.

Persistent FTS5 `automerge` and `crisismerge` settings remain unchanged deliberately. Retuning them could shift unpredictable merge work onto foreground commits; the explicit owner supplies a bounded observable budget instead. A later retune should be justified by separate production workload evidence.

## Verification

- `cargo test -p khive-db fts_maintenance -- --nocapture`: 7 passed
- focused `khive-db` diagnostics tests, including interruptible diagnostics: passed
- focused slow-warning fields and timing-aggregation tests: passed
- `cargo test -q -p khive-db -p khive-pack-memory -p khive-runtime --lib`: 888 + 326 + 1,415 passed; 8 documented heavyweight/model tests ignored
- `cargo check --workspace --all-targets --all-features --locked`
- `cargo clippy -p khive-db -p khive-pack-memory -p khive-runtime --all-targets -- -D warnings`
- `RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps -p khive-db -p khive-pack-memory -p khive-runtime`
- `scripts/ci.sh lint`: formatting, SQL, ADR references, artifact checks, writer census, and workflow checks passed
- `git diff --check`

## Scope notes

- The issue owner's p95 target remains bench-lane work; this PR supplies deterministic work/counter/structure evidence rather than a machine-specific wall-clock assertion.
- Draft #2328 / #1907 independently scopes memory MATCH work to the memory corpus. This PR is based directly on `main` and does not depend on or stack on that branch.

Closes #1906


ohdearquant:
Reviewed at head `d925d7d6acf70bcaf2fe5d61e9868bb4123246f5`. The maintenance design is sound in shape —
one table per due turn, round-robin, a parameterized `merge` rather than an unbounded `optimize`,
busy treated as a non-error, a successful checkpoint preserved when maintenance fails. The problem is
that the documented safety property is stronger than the mechanism delivers, and it is stated as a
guarantee in an ADR.

## The writer-priority claim is not what `busy_timeout = 0` buys

`crates/khive-db/docs/api/checkpoint.md:100-104` says an application writer "therefore wins
immediately", and `docs/adr/ADR-091-wal-snapshot-lifetime.md:1525-1526` says the merge "never waits
behind an application writer".

A zero busy timeout means maintenance does not *wait* when it **loses** the race for the write lock —
it fails immediately and is counted as busy. It says nothing about the case where maintenance
**wins**. There is no queue priority, no cooperative cancellation and no preemption around the
statement at `crates/khive-db/src/fts_maintenance.rs:332-341`. Once the merge holds the lock, an
application writer waits behind it for as long as the merge takes, bounded by the writer's own busy
timeout. The dedicated standalone connection (`checkpoint.rs:1845-1870`) avoids the pooled writer
mutex, but it still contends for SQLite's database-level write lock.

Two things compound it:

- **The budget is page work, not time.** `KHIVE_FTS_MERGE_PAGES` defaults to 500
  (`fts_maintenance.rs:16`), passed through as a positive count that SQLite consumes in its own merge
  loop. There is no wall-time bound and no interrupt point, and the variable accepts any positive
  `u32` (`:119-127`), so 500 is a default rather than a ceiling.
- **The merge is synchronous inside the checkpoint task** (`checkpoint.rs:2073-2079`, inside the
  successful-checkpoint branch). While it runs, no checkpoint runs. The loop ticks at 500 ms with
  `MissedTickBehavior::Skip` (`:1969-1970`), so a long merge does not queue work — it simply leaves
  the WAL uncheckpointed for the duration and then drops the missed ticks.

Also worth noting the zero timeout is scoped to the merge statement only. It is set at `:335-336` and
restored at `:339` immediately after `execute`; the preflight (`:386-392`) and postflight (`:425-431`)
structure reads run under the connection's normal timeout. So "zero busy wait" describes one statement,
not the due call.

I would take either fix, but not neither: soften the text to describe fail-fast lock admission and say
plainly that maintenance can delay a writer once it holds the lock, **or** add a real admission
mechanism plus a wall-time policy. What should not survive is an ADR asserting a property the code does
not implement — that is the claim someone will rely on later when deciding this path is safe to leave
unbounded.

## The structure decoder accepts a record SQLite considers corrupt

`parse_structure_record` validates truncation, per-level leaf ranges, and header-vs-levels total
(`fts_maintenance.rs:251-283`), but never checks `merge_input_segments <= level_segments`. A record
declaring one level with `merge_input_segments = 1` and `level_segments = 0` parses cleanly: total
segments is 0, matching the header, so it returns `FtsIndexStructure { segment_count: 0, .. }` while
`has_active_merge` (`:302-307`) reports an active merge. SQLite's own decoder rejects exactly this
(`nTotal < pLvl->nMerge`).

That collapses the one distinction `fts_segments_error` exists to make: a corrupt record is reported
as a healthy empty index. And since `nMerge` is the restart-resumption mechanism, maintenance will try
to continue a merge state SQLite would refuse to load. Mirroring the invariant is a two-line check.

Separately, there is no explicit FTS5 V2 marker handling — a V2 record is rejected as an implausibly
large level declaration. That is fail-closed, which is fine, but it is an untested compatibility path
and worth an explicit version error so the failure names itself.

## Some maintenance failures are invisible to the counter that advertises them

In `run_merge_statement`, the `PRAGMA busy_timeout` read (`:332-334`) and the zero-set (`:335-336`)
both return early on error, **before** `FTS_ATTEMPTS` is incremented at `:337` and before any
`FTS_ERRORS` increment. Merge execution errors do increment (`:364-366`), as do restore failures
(`:342-347`) and structure read failures (`:386-390`, `:425-430`) — but a setup failure lands in
neither total.

The caller logs it (`checkpoint.rs:2109-2118`), so it is not swallowed, but
`db_diagnostics.fts_maintenance.errors` is documented as the visibility surface
(`checkpoint.md:111-112`, ADR-091 `:1527-1530`). An operator can read a clean error counter while the
maintenance path fails before every merge — and a connection/setup failure then looks identical to
"no maintenance was due". Worth centralizing the accounting so every fallible step after the due check
increments, and defining whether "attempt" means work considered or statement executed.

## Timing fields cannot distinguish "skipped" from "0 ms"

`RecallStageTimings` is five plain `Duration`s with a zero default
(`crates/khive-pack-memory/src/handlers/common.rs:485-492`) and the emitter always serializes numbers
(`recall.rs:76-83`). But FTS returns before starting its timer when there are no lexical terms
(`common.rs:866-870`), the no-registered-models path returns empty while keeping defaults
(`:1020-1033`), and the exact sqlite-vec fallback writes `fresh_tail_elapsed: Duration::ZERO`
explicitly (`:1425`, `:1810`).

So "no FTS terms", "no embedding models", "fresh-tail not applicable" and "ran in under a millisecond"
all emit `0`. That is the one thing a slow-recall warning most needs to tell an operator apart.
`Option<Duration>`, or a status alongside each value, would fix it; reserve numeric zero for a stage
that actually ran.

The arithmetic itself is right and worth saying so: sequential rounds sum (`:494-500`), hydration sums
(`:502-504`), and concurrent per-model ANN/fresh-tail take maxima (`:506-509`). The issue is stage
state, not max-versus-sum.

## Smaller

- **Public docs miss the new fields.** `fts_segments` and `fts_maintenance` are absent from
  `docs/guide/api-reference.md:784-788`, `crates/khive-pack-kg/README.md:33`, ADR-023 `:179`, and the
  runtime/handler comments. Worth stating explicitly that `segment_count` is *physical FTS segment
  structure*, not document count — the nearest existing number, `TextIndexStats::document_count`
  (`stores/text.rs:1075-1086`), is a `COUNT(*)` and the two diverge freely with compaction.
- **`AGENTS.md` and `CLAUDE.md` are instruction-surface files bundled into a performance change.** Not
  a correctness problem, and the content is accurate — but see the Docs lint note below, because that
  edit is what turned the job red.
- **Tests.** The added tests are real and each names a mutation it catches — varint decoding, the
  row-10-not-`%_idx` plan, negative-starter/positive-continuation, round-robin order, busy accounting,
  persisted-`nMerge` resumption. What none covers: the invalid `nMerge > nSeg` record, a merge
  exceeding its budget, any wall-time bound, process death mid-statement, a setup-path failure, and an
  application writer racing while maintenance holds the lock. The last one is the H1 mechanism, so it
  is currently asserted only in prose.

## CI — three reds, and only one is yours

**Docs lint (yours, mechanical).** dprint reports "Found 2 not formatted files in 206 files". The
`db_diagnostics` row description got longer, which widens column 3 of that verb table, so every other
row needs re-padding — the diff hunks are identical text with more trailing spaces. The two files are
`AGENTS.md` and `CLAUDE.md`, both of which carry that table. Run the repo formatter and it is done.

**Component benches / packs — not yours.** This one is worth spelling out, because the surface reading
is misleading. The job did not fail a benchmark; it was **cancelled** at 40m07s against the
`timeout_minutes: 40` set for this component in `.github/workflows/bench-track.yml:212`. The dominant
cost is the bench-target compile, which the workflow's own comment records as ~16m and issue #940. This
branch is zero commits behind `main` and touches `khive-db`, a heavy shared dependency, so it
invalidates more of the bench compile cache than most — enough to push an already-marginal job over a
ceiling that other branches clear. Thirteen of the fourteen most recent pull-request runs of this
workflow passed; this is a budget boundary, not a regression in what you changed.

The more useful finding underneath it: the benches were already broken before the cancellation.

```
thread 'main' panicked at khive-pack-knowledge/benches/knowledge_bench.rs:67:14:
seed atoms: InvalidInput("bad params: unknown field `description`, expected one of `slug`,
`name`, `content`, `tags`, `properties`, `finalized`, `source_uri`, `source_type`")
```

That bench seeds atoms with a `description` field the params struct does not accept. It is not
specific to this branch — the `main` run at `a46914ae`, which is exactly this branch's merge base,
panicked identically at the same line and **still reported success**, because the Criterion step ends
in `|| true`. So the packs component has been publishing trend data with the knowledge benches dying
every run, and the exit code that would have said so is discarded. That deserves its own issue; the
workflow already carries a comment about a previous round of this exact problem leaving "the trend data
packs-blind", and this is the same class one layer down. Nothing for this PR to fix.

**Secret scan — unattributable.** That job scans every ref present in the checkout rather than the pull
request's own commits, so a red says nothing about which change is at fault until #2323 lands. Not
yours, and equally not ruled out for anyone else. The CI gate is the aggregate of the above and carries
no independent signal.

checked at the head named, with base comparisons read at `origin/main`; the CI conclusions come from
the run logs and job metadata cited.

ohdearquant:
Filed the bench finding as #2336 so it does not ride on this PR — nothing for you to do there. The Docs lint red is still yours (run the repo formatter over `AGENTS.md` and `CLAUDE.md`).

On the secret scan: #2323 has now merged, so the job is bounded to the pull request's own commit range and `-v` names the finding. That does not make the existing red above meaningful, and re-running it does not either — a re-run replays the workflow file pinned to the original run and scans wide again. The one-step discriminator: a scoped run has a step called `Resolve scan scope`; the wide one has `Scan git history for credentials`. Your next push produces a fresh scoped run, and a red at that point is worth reading.

ohdearquant:

`Docs lint` fails at `Check markdown formatting`: `AGENTS.md` and `CLAUDE.md` are not `deno fmt` clean.
wrapping, no content change.

ohdearquant:
Two things about the maintenance lane, plus a note on the telemetry. The required gate is also red
at this head — `CI gate` and `Docs lint` both fail — so that needs clearing regardless.

## Zero busy timeout does not give application writers priority

Before each merge the connection drops its busy timeout and then runs the merge synchronously:

```rust
// crates/khive-db/src/fts_maintenance.rs:332-341
conn.busy_timeout(Duration::ZERO)...;
let execution = conn.execute(table.merge_sql, params![requested_pages]);
let restore = conn.busy_timeout(Duration::from_millis(prior_busy_ms.max(0) as u64));
```

The docs describe the consequence as:

> Before each merge the dedicated connection temporarily sets `busy_timeout` to zero. **An
> application writer therefore wins immediately**

(`crates/khive-db/docs/api/checkpoint.md:100-103`, and the same claim in
`docs/adr/ADR-091-wal-snapshot-lifetime.md`.)

That holds for exactly one ordering. A zero timeout makes *maintenance* give up at once when another
writer already owns the SQLite write lock. It confers no queue priority and no cancellation once
maintenance has taken the lock, so an application writer arriving next simply waits for the merge to
finish. The conclusion is right for the case where the writer got there first and wrong for the case
where it did not, and the sentence does not distinguish them.

"The merge is short" would be a latency argument rather than a priority one, and there is no
execution-time bound on this path — the 500-page budget bounds SQLite work, not elapsed time.

**Suggested direction:** either add a real application-writer admission or cancellation mechanism
around the merge, or keep the current behaviour and describe it accurately as "maintenance yields to
an in-progress writer" rather than as writer preference. A test where maintenance takes the lock
first and an application write asserts its bounded wait would pin whichever is chosen.

## The merge runs inline on the checkpoint task, which can cost a checkpoint tick

`run_if_due` is called on the checkpoint task itself, on the same connection
(`crates/khive-db/src/checkpoint.rs:2074-2079`), and `run_merge_statement` executes synchronously
without yielding. The task's interval is configured with:

```rust
// crates/khive-db/src/checkpoint.rs:1969-1970
interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
```

So a merge that runs past the next scheduled tick does not delay that checkpoint — it *skips* it,
and the task waits for the following one. That extends how long WAL frames stay uncheckpointed,
which is the opposite of what the surrounding change is for, and it is at odds with
`checkpoint.md:88-92` describing the FTS cadence as independent of the 500 ms WAL tick.

A page budget bounds merge work; it does not bound scheduling. Moving the FTS work to an
independently supervised path, or preserving a hard checkpoint deadline, would keep the two cadences
actually independent. A regression covering a merge that spans a checkpoint interval is the test
that separates the two designs.

## Telemetry: a skipped stage and a fast stage are the same value

`RecallStageTimings` stores `Duration`s and renders them as bare integer milliseconds
(`crates/khive-pack-memory/src/handlers/common.rs:485-492`, `:511-529`). The exact sqlite-vec
fallback sets `fresh_tail_elapsed: Duration::ZERO` (`:1803-1811`), and the no-model path leaves the
embed and ANN timings at their defaults (`:1020-1033`), while the warning always emits all five
fields (`recall.rs:76-85`).

So `fresh_tail_ms=0` means either "did not run" or "ran in under a millisecond", and the reader
cannot tell which. Since `docs/api/recall-pipeline.md:97-106` promises the five stage fields, the
useful fix is to make not-run representable — omit the field, or use a nullable value — rather than
to collapse it onto a legitimate measurement.

authored it, so its output was treated as input evidence only. The busy-timeout sequence, the
documented claim it is paired with, the inline call site, and the `MissedTickBehavior::Skip`
configuration were each re-derived from the ref before posting.

ohdearquant (commented):

ohdearquant (commented):

ohdearquant (commented):

Verdict on head b9f8fedcb6a6c361a13df1da9e01c261b1b51990: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

ohdearquant (commented):

Verdict on head 3ad910cb809922e2b8851e79e65f73265ad7246a: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving at this exact commit. Verified before approving: the head matches the commit named in the request, all check runs at this commit have completed with no failures, no change requests are outstanding, and the approving account is not the authoring account.

PR #2333, opened 2026-08-30
fix(channel): persist outbound delivery backoff

### LINKED ISSUE #2272: Outbound delivery: implement ADR-122 transient-failure backoff and attempt tracking
ADR-122 (docs/adr/ADR-122-email-outbound-delivery.md) specifies per-message backoff for transient send failures: `delivery_attempts` incremented and `next_attempt_at` set (exponential in attempt count, bounded ceiling), with poll cycles skipping a pending note until `next_attempt_at` passes, and a successful send clearing both fields.

The current outbox loops implement the terminal-outcome subset only: `delivery = "delivered"` (with `delivered_at`, `transport_message_id`) and `delivery = "failed"` (with `failed_at`, `last_error`, e.g. allowlist rejections). Transient send failures still retry at raw poll cadence with no attempt tracking — a sustained soft failure (greylisting, mailbox-full 4xx) retries every cycle and rewrites nothing durable.

Scope:
- Add `delivery_attempts` / `next_attempt_at` handling to the delivery loops and the pending-scan predicate (`list_undelivered_outbound_messages`).
- Classify transport errors into transient vs permanent per the ADR (`delivery = "failed"` reserved for genuinely permanent classifications — configuration, authentication, allowlist — never attempt count).
- Tests: backoff skip behavior, clearing on success, and a fixture separating transient from permanent classification.

ohdearquant:
## Summary

- persist bounded exponential retry state for transient outbound email and Telegram failures
- classify SMTP, OAuth, and Telegram delivery failures so only permanent rejections terminate messages
- skip deferred pending messages and clear retry state after successful delivery

## Validation

- `cargo test -p khive-runtime outbound_ -- --nocapture`
- `cargo test -p khive-channel -p khive-channel-email -p khive-channel-telegram`
- `cargo test -p khive-mcp --features channel-email,channel-telegram`
- `cargo check --workspace --all-targets --all-features`
- `cargo test --workspace`

Closes #2272


ohdearquant:
Reviewed at head `5af1573500652701fddd6ce2ab6f0750bf4dbca1`. Durable backoff is the right change and most
of it is well built — the two-direction classifier, the 5s→30m ceiling, malformed deadlines failing
open, and clearing retry state only after the send returns success. One design decision needs to flip
before this leaves draft, and it shows up on three surfaces at once.

## Unknown errors default to permanent, and that is the unrecoverable direction

The new terminal path is genuinely new. On the base, every outbound send error took one branch —
`tracing::warn!("send failed; will retry next cycle")` (`origin/main:crates/khive-mcp/src/serve.rs:1631-1637`) —
and the note stayed pending. Now `record_outbound_send_failure` splits on
`ChannelError::delivery_failure_class()`, and `Auth`, `PermanentTransport`, `Config`,
`InvalidEnvelope` and `UnauthorizedSender` all map to `Permanent`
(`crates/khive-channel/src/lib.rs:274-282`) → `mark_outbound_message_failed` → `delivery="failed"`.

That state is terminal in the strong sense: the scan skips any note whose `delivery` is
`"delivered"` or `"failed"` (`crates/khive-runtime/src/curation.rs:1836-1841`), so the message is
never retried and never surfaces as pending. The only way back is a generic property patch —
`delivery`, `delivered_at`, `failed_at`, `delivery_attempts` and `next_attempt_at` are all absent from
`OWNER_ESTABLISHED_PROPERTIES` (`curation.rs:3677-3687`), so an arbitrary caller can rewrite them.
That is a separate pre-existing gap, and it would be the wrong thing to lean on as the recovery story.

So the two outcomes are asymmetric: a wrongly-pending message costs a retry every 30 minutes and stays
visible; a wrongly-failed message is gone with no operator-facing signal beyond a warning line. Where
the classifier is uncertain, it should choose the recoverable one. Today it does the opposite on each
of these:

- **SMTP.** `classify_smtp_send_error(error.is_permanent(), ...)` is the only discriminator
  (`crates/khive-channel-email/src/connector/smtp.rs:132-138`, used at `:251-253`). In the pinned
  lettre 0.11.22, `is_permanent()` is `matches!(self.inner.kind, Kind::Permanent(_))`, `Kind::Permanent`
  is documented as "Permanent SMTP error, 5xx reply code", and it is constructed solely from
  `Severity::PermanentNegativeCompletion`. It is a pure 5xx test carrying no stage information — and
  `transport.send(msg)` spans connection, AUTH and the DATA transaction in one await, because the
  builder above it does no I/O. An AUTH `535` is a 5xx permanent negative completion, so it is
  indistinguishable from a post-auth `550` and terminalizes the message.
- **OAuth.** The unmatched arm is `else { ChannelError::Auth(message) }`
  (`crates/khive-channel-email/src/oauth.rs:242-244`). `sanitize_oauth_error` deliberately reduces
  absent, malformed or unrecognized codes to the literal `"oauth_error"` (`:43-60`), so exactly the
  inputs the classifier understands least — an unlisted 4xx, an HTML gateway page, a provider code
  introduced next quarter — land in the terminal arm.
- **Telegram.** `let code = response.error_code.unwrap_or(status.as_u16())` then
  `(400..500).contains(&code)` → `PermanentTransport` (`crates/khive-channel-telegram/src/connector.rs:58-69`).
  The trailing `else { Transport }` only catches codes outside both ranges, so it is not the fallback
  for an unknown client error. A missing `error_code` on an HTTP 400, or a rate-limit response whose
  429 does not parse, is terminal.

The ask is one line of policy, three times: an explicit allowlist of codes known to be definitively
permanent, and everything else — unknown code, missing code, malformed body, unrecognized status —
classified as transport pressure and retried. Keep the existing timeout/network → `Transport` mapping;
that part is already right.

## The sharpest instance: a component failure is terminalized once per queued note

`SmtpConnector::send` fetches the OAuth token inside itself, per message
(`smtp.rs:240-248`), and the email outbox loop calls `record_outbound_send_failure` for every send
error (`serve.rs:1673-1697`). The loop over scanned notes contains no `break` — only `continue` — and
the scan pulls up to 200 notes a pass (`serve.rs:1507`).

So an expired client secret or a provider-side policy change is not one failure. It is N failures, one
per queued note, each independently classified `Auth` → `Permanent`, each writing `delivery="failed"`.
A single component-level credential problem can empty the queue into a terminal state in one pass,
with no component-level stop and nothing left pending to recover from. That is the opposite of what
ADR-122 asks for: a definitive auth failure should stop the supervised component visibly, while a
post-auth per-message 5xx may fail only its own note.

Worth a regression test with two queued notes and one token failure, asserting neither is
terminalized by it.

## Correction on Telegram bare-slug addressing — pre-existing, not this PR

`crates/khive-channel-telegram/docs/design.md:28-36` promises two accepted spellings,
`telegram:<slug>` and the bare `<slug>`, while the outbox selects with `Some("telegram:")` and guards
again with `starts_with("telegram:")` (`serve.rs:1959-1961`, `:1981-1986`). A bare-slug row is
therefore never selected: not sent, not failed, not retried. `TelegramChannel::send` does strip an
optional prefix (`channel.rs:121-139`), but the scan has already excluded the row before that code runs.

I want to be accurate about whose problem this is: **it is not this PR's.** Both halves are on the
base — `origin/main:crates/khive-channel-telegram/docs/design.md:30-31` carries the identical
bare-slug promise, and `origin/main:crates/khive-mcp/src/serve.rs:1892` already scans with
`Some("telegram:")` and re-guards with `starts_with` at `:1917`. The outbox loop is restructured here,
not introduced. So this is a real documented-but-unroutable address worth a follow-up, and it should
not hold up this change.

## What this PR does improve on that path

Unroutable non-maintainer slugs used to return `Ok(())` and be dropped silently
(`origin/main:crates/khive-channel-telegram/src/channel.rs:124-131` logs a warning and returns Ok);
they now return `InvalidEnvelope` and are recorded terminally. That is the right direction — a
misaddressed message becoming a visible terminal failure rather than a silent success.

## Also holds up

- Retry classification never depends on attempt count (`khive-channel/src/lib.rs:265-284`), so a
  transient error cannot be promoted to permanent by repetition.
- Malformed and unset deadlines fail open (`curation.rs:1843-1850`), so a bad legacy property does not
  strand a message. No migration is needed — notes without the fields are immediately eligible.
- Send confirmation precedes delivery marking and the state clear happens inside the subsequent marker
  (`serve.rs:1642-1654`, `curation.rs:2022-2039`), so a failed mark after a successful send leaves the
  note pending and re-sendable. At-least-once with duplicates, rather than silent loss, is the correct
  trade here and the code says so explicitly.
- Allowlist rejection is recorded terminally rather than skipped (`serve.rs:1544-1572`).

## Gaps worth closing while you are in here

- **No exhausted state.** Attempts saturate; after the 30m ceiling a transient message retries forever
  with no durable "exhausted" marker and no status surface. Visibility depends entirely on warning-log
  collection. That is consistent with ADR-122 and the amended docs, but it means a permanently
  unroutable-but-transient address is invisible.
- **A far-future `next_attempt_at` is accepted with no upper bound** (`curation.rs:1843-1849`), so
  corrupted state can park a note indefinitely. The added test covers future, overdue and malformed
  values (`curation.rs:4084-4130`) but not this.
- **Tests.** Each added test names a mutation it would catch, but the provider classifiers are only
  tested on their known branches — the SMTP test exercises the pure bool helper, not a real lettre
  error with an AUTH-versus-DATA stage. Missing: AUTH 535 vs post-auth 550, connection reset, timeout,
  `oauth_error`, unknown 4xx, malformed non-JSON body, missing Telegram `error_code`, and restart
  mid-backoff.
- **`docs/multi-backend.md:422-439`** says delivered state carries the transport message id for both
  channels. Telegram passes `None` (`serve.rs:1995-2001`) and the code notes it has no RFC 822
  Message-ID (`:1916-1919`). Narrow that to email, or "when supplied by the transport".
- **`docs/guide/communication.md:302-311`** describes authentication and SMTP 5xx as terminal, which is
  broader than what is safe — it should distinguish a component authentication failure from a
  per-message rejection after authentication.

## CI

Secret scan is red, and the red on **this run** is not attributable to this branch in either
direction — that job scanned every ref present in the checkout rather than the pull request's own
commits, so it says nothing about which change is at fault. Nothing to chase. The CI gate is the
aggregate of the same failure and carries no independent signal.

Update: #2323 has now merged, which bounds the scan to the pull request's own commit range and adds
`-v` so a finding names its file, line, commit and rule. That does **not** retroactively make the red
above meaningful, and re-running the job does not help either — I measured this: a re-run replays the
workflow file pinned to the original run, so it scans wide again. The discriminator is one step name.
A scoped run has a step called **`Resolve scan scope`**; the wide one has `Scan git history for
credentials`. Your next push to this branch produces a fresh run under the scoped logic, and at that
point a red is worth reading.

No builds, tests, benchmarks or database queries were run for this pass. Every claim above was checked
against source at the head named, with base comparisons read at `origin/main`.

ohdearquant:
Persisting delivery outcomes is the right move, and the transient/permanent split is the right
shape. Two of the classifiers land on the wrong side of it, and both fail in the direction that
loses mail.

## Unknown or malformed OAuth responses become terminal

The OAuth endpoint classifier allowlists what is retryable and treats everything else as an auth
failure:

```rust
// crates/khive-channel-email/src/oauth.rs:231-245
if status == StatusCode::REQUEST_TIMEOUT
    || status == StatusCode::TOO_MANY_REQUESTS
    || status.is_server_error()
    || matches!(error_code, "server_error" | "slow_down" | "temporarily_unavailable")
{ ChannelError::Transport(message) } else { ChannelError::Auth(message) }
```

and that class is terminal, in two more hops:

```rust
// crates/khive-channel/src/lib.rs:275-283
Self::Config(_) | Self::PermanentTransport(_) | Self::Auth(_)
| Self::UnauthorizedSender(_) | Self::InvalidEnvelope(_) => DeliveryFailureClass::Permanent,

// crates/khive-mcp/src/serve.rs:1412-1421
DeliveryFailureClass::Permanent => runtime.mark_outbound_message_failed(...)
```

So an HTTP 400 or 403 carrying an error code that is not one of the three known strings — or a 2xx
whose token body does not parse (`oauth.rs:283-286`) — removes the retry schedule and marks the
message `failed`. On `origin/main` the same condition logged *"send failed; will retry next cycle"*
and the message stayed queued.

The asymmetry is what makes this worth holding on rather than noting. Retrying a genuinely permanent
auth failure costs a few wasted attempts and stays visible. Terminating a transient one loses the
message and removes it from every subsequent scan, so nothing downstream ever reports it. An unknown
code is precisely the case where the classifier has no evidence of permanence, and it currently
defaults to the unrecoverable side.

**Suggested fix:** default unknown, missing, and malformed OAuth responses to retryable; reserve
terminal classification for protocol-defined permanent rejections. Worth covering with outbox tests
for an unknown code, a non-JSON error body, and malformed 2xx token JSON, each asserting `delivery`
stays pending with `next_attempt_at` written.

## A Telegram body code overrides the HTTP status, including 429

```rust
// crates/khive-channel-telegram/src/connector.rs:58-69
let code = response.error_code.unwrap_or(status.as_u16());
...
if code == 408 || code == 429 || code >= 500 { Transport }
else if (400..500).contains(&code) { PermanentTransport }
else { Transport }
```

Because the body code wins unconditionally, an HTTP 429 or a 5xx whose body carries
`error_code: 400` classifies as `PermanentTransport` and goes terminal — even though the transport
status is the strongest available signal of retryable pressure. A status/body disagreement is not
evidence of permanent rejection; it is evidence that one of the two signals is unreliable, and the
safe reading of an unreliable signal here is "retry".

The added tests at `:310-338` cover matching 429/429 and HTTP 200 with body 400, which are the two
agreeing cases. The disagreeing cases — 429 with body 400, 503 with body 400 — are the ones this
classifier decides, and they are untested.

**Suggested fix:** consult 408/429/5xx on the HTTP status *first* and keep the retryable
classification when the two signals conflict, retaining the permanent path for a validated
client-side rejection on a non-pressure status.

## Smaller

The communication guide describes per-message handling for startup configuration failures that the
implementation applies at startup rather than per message. Worth aligning the wording, since it is
the surface an operator reads when deciding whether a stuck message will retry.

authored it, so its output was treated as input evidence only. Every link in both chains above —
the OAuth fallback arm, the `Auth` to `Permanent` mapping, the `mark_outbound_message_failed` sink,
and the Telegram `unwrap_or` precedence — was re-derived from the ref before posting.

ohdearquant (commented):

Verdict on head 25f0eb3cde5a1726e7dd444bb33e33eaf6084aff: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

ohdearquant (commented):

Verdict on head e60ff9161aa7e838fcb9dffab4eac911d26bda9a: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 30ba8770ed9570a91cadd83062d1bca5a9c70e59: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approved at 30ba8770.

PR #2334, opened 2026-08-30
feat(audit): identify canonical operation arguments

### LINKED ISSUE #2047: Gate evaluates pre-canonicalization arguments; kind hooks and coordinator search canonicalize after the decision
## Observation (at main f0eb4d22c07c)

Gate evaluation happens at dispatch entry against the raw request envelope (`crates/khive-runtime/src/pack.rs:1346`, `:1413`, `:1656`). Two argument-mutation points then run after the decision:

1. `KindHook::prepare_create` receives `&mut params` inside the kg create handler (`crates/khive-pack-kg/src/handlers/create.rs:338`) — handler-side, post-gate by construction.
2. The coordinator search intercept passes the raw `args_value` to `dispatch_intercepted_with_metadata_with_identity` and runs `ValidatedSearchRequest::from_value` only inside the post-gate closure (`crates/khive-mcp/src/server.rs:2019`). The in-code comment states the ordering explicitly: the gate authorizes before handler-level validation runs.

## Why this matters

Today's ordering is consistent with normal registry dispatch: the gate is a namespace-level authorization seam, and canonicalization is handler-level. But any future policy conditioned on argument values (kind, relation, payload shape) would be decided against values that can differ from what the handler ultimately executes, because hooks and canonicalization rewrite params after the decision.

## Ask

Decide and document the contract: either (a) the gate is namespace-scoped only and semantic argument inspection is explicitly out of contract (document it at the gate seam), or (b) gates that inspect arguments must see canonicalized values, which requires moving canonicalization ahead of the gate or re-checking after it. This is a design question, not a current-behavior defect.

### LINKED ISSUE #2049: Audit and event rows cannot reconstruct which operation in a composed request produced them, nor what arguments executed
## Observation (at main f0eb4d22c07c)

Two attribution gaps in the audit/event plane:

1. **No per-operation index.** Every persisted event row gets a fresh id (`crates/khive-db/src/stores/event.rs:498`), so rows are individually addressable, but `request_id` is a request-group id shared by every op in a batch or chain. There is no generic operation-index field — the only `op_index` in the tree is the save-sink failures projection (`crates/khive-mcp/src/save_sink.rs:341`), which is a different surface. For one `request_id` with two same-verb operations in one composed request, nothing maps a row back to its parser position, or says whether a reference value was submitted literally or produced by `$prev` resolution.

2. **No argument capture.** `AuditEvent` (`crates/khive-gate/src/audit.rs`) carries timestamp, actor, namespace, verb, decision, deny reason, obligations, gate impl, and a correlation token — no argument field at all. Combined with the post-gate argument mutation described in #2047 (kind hooks and canonicalization rewrite params after the gate decision), the audit trail records neither the submitted nor the effective arguments of a mutated create.

## Why this matters

An audit plane that cannot answer "which op in this request wrote this row, with what arguments" cannot support forensic reconstruction of a composed request. Argument capture also has a disclosure trade-off (arguments can carry secrets), so the design decision is not "log everything" — it is choosing what minimal, masked argument identity to persist.

## Acceptance sketch

For one request id containing two same-verb operations where the second consumes the first's output by reference: the audit/event rows identify which row was operation 0 and operation 1, and whether the reference argument was literal or resolved.

ohdearquant:
## Summary
- document the gate contract as actor/namespace/verb authorization over the resolved transport envelope
- attach stable operation indices and literal/reference provenance to request-group audit rows
- persist secret-masked BLAKE3 identities for resolved and post-hook effective argument envelopes

## Verification
- git diff --check

Closes #2047
Closes #2049


ohdearquant:
Reviewed at head `a626181e6f91c4906cddacbeb24142bc08c0c3ca`. The gate ordering is right, the
operation indexing is sound, and the literal-versus-reference provenance genuinely shares the
resolver's representation rather than re-parsing — I checked that last one specifically because a
second parse is the usual way provenance goes quietly wrong, and there isn't one here.

Three things need to change before this lands, and the first is the one that matters.

## The "secret-masked" property does not hold, and one path stores plaintext

The order is correct and worth saying so: `canonicalize` → `serde_json::to_vec` → `mask_secrets` →
BLAKE3 over `masked.as_bytes()` (`crates/khive-runtime/src/audit_context.rs:77-85`). The digest is
over the masked serialization. That is the right construction.

The problem is what `mask_secrets` masks. It is a text scanner over a flat string, not a walker over
the JSON structure, and every entropy-based detection is gated behind a nearby credential trigger:

```rust
let near_trigger = contains_trigger(&window[..raw_start])
    || contains_trigger(&window[raw_end..])
    || has_inline_credential_trigger(raw_token);
...
if near_trigger {
    for run in token.split(|c: char| !c.is_ascii_alphanumeric()) {
        ...
        if shannon_entropy(run.as_bytes()) >= ENTROPY_THRESHOLD {
            return Some((token, "high-entropy-token"));
        }
    }
```

(`crates/khive-runtime/src/secret_gate.rs:967-969`, `:1023-1039`.) So a high-entropy value under a
benign key — `{"payload":"<32 random characters>"}` — has `{"payload":"` before it and `"}` after it,
neither of which is a trigger, and is never flagged. That is a reasonable design for the gate's
original job of scanning prose and diffs. It is not the property this feature advertises, because
here the input is machine-generated JSON where the *key* is the only context and benign keys are the
common case.

Two distinct consequences, and they are not equally bad:

- **The digest.** Hashing bytes that still contain the secret does not leak it, but it makes the
  stored digest a confirm-a-guess oracle for anyone who can reconstruct the rest of the envelope.
  Bounded, but real.
- **The key projection, which is plaintext.** `keys` is built from the *unmasked* `arguments` map,
  passing each key through `mask_secrets(key)` standing alone (`audit_context.rs:88-97`). A key
  evaluated in isolation has no surrounding text at all, so `near_trigger` can only be true if the
  key itself contains an inline credential label. A high-entropy key therefore reaches the durable
  row verbatim, truncated at `MAX_AUDIT_ARGUMENT_KEY_CHARS = 128` (`:16`) — which is longer than
  essentially every API key and bearer token in use. This is not an oracle; it is the credential,
  written to a store designed to be retained and read later.

The fix is to mask a `serde_json::Value` recursively at every leaf *and* every key, with a policy
that handles an unlabelled high-entropy leaf, then use that same masked structure for both the digest
input and the key projection. A JSON walker also gets you the nested and stringified cases for free,
which a flat scan cannot reach.

One note on the test that appears to cover this.
`argument_identity_is_order_independent_and_never_contains_values`
(`audit_context.rs:132-143`) asserts the digest does not contain the input value. A BLAKE3 hex digest
cannot contain it whether or not masking ran, so that assertion passes identically against a
completely disabled masker. The masking test has to assert the digest equals the digest of the
expected *masked* projection.

## Equal digests mean three different things

The other two findings share one root: `resolved == effective` is overloaded, and a reader cannot
tell which of three states produced it.

**The coordinator `link` path never records an effective envelope.** There are exactly three
production call sites of `record_effective_arguments` — `handlers/create.rs:372`,
`handlers/search.rs:280`, and `server.rs:2290`. The link route has none, even though the closure
immediately below it (search) does. Yet link parses `source_id`, `target_id` and `relation` into
typed values and supplies a defaulted `weight` at `server.rs:2232-2239`, then consumes exactly those
at `:2247-2249`. `dispatch_intercepted_with_identity` therefore completes the row from the
`gate_req.args` fallback (`crates/khive-runtime/src/pack.rs:1773-1780`), so a link that omitted
`weight`, or spelled a UUID differently, records equal digests while the coordinator consumed a
different envelope. That is the exact forensic question this feature exists to answer.

**A failed canonicalization is indistinguishable from one that was never needed.**
`capture_effective_arguments` seeds the task-local with `fallback`, awaits, and then returns whatever
is in it — unconditionally, whether or not the future succeeded and whether or not any handler ever
recorded (`audit_context.rs:53-64`). `complete_audit_event` then always wraps it in `Some`. There is
no "was this recorded" bit. So these three collapse into two outputs:

| What happened | What the row shows |
| --- | --- |
| Nothing to canonicalize | `effective == resolved` |
| Handler mutated, then failed | `effective == resolved` |
| Handler canonicalized and succeeded | `effective` differs |

`KindHook::prepare_create` takes `&mut params` and can mutate before returning an error
(`create.rs:365-366`), while the record call sits after the `?` at `:372` — so the middle row is
reachable today, not hypothetical. Make the task-local an `Option`, or carry an explicit status, and
represent failure as absent rather than as a successful-looking identity.

## What holds up

- **Operation indices are anchored and stable where it matters.** Zero-based parser position scoped
  to the request-group context, assigned before dispatch; a conflict does not reindex the remaining
  batch, and an aborted chain operation is skipped without shifting later positions. `MAX_OPS = 100`
  bounds the `u32` conversion for any parsed input. Worth stating in the docs that the index is
  stable for a replay of the same parsed request but is not a cross-retry identity — nothing binds it
  to an attempt.
- **Provenance agrees with the resolver rather than re-deriving it.** Classification and substitution
  share one `ArgValue` representation: an escaped `\$prev` becomes a literal `Value` and a valid path
  becomes `PrevRef`, both decided by the parser. There is no second parse to disagree with.
- **Resolution really does precede authorization** on every path I followed — chain substitution
  completes before dispatch, and the post-gate mutation points (create's kind hook, the search
  validator) are deliberate and documented. `DispatchHook::on_dispatch` receives an `EventView` and
  cannot mutate the envelope. ADR-018's new text matches this.
- Denied, gate-unavailable and unknown-verb rows correctly carry no handler effective identity.

## Smaller

The structured tracing copy is emitted right after the gate decision while the durable row gets
`effective_arguments` only after dispatch, so the traced event shows an absent effective arm where the
stored row has one. That is an absence rather than a wrong value, and it pre-exists this change — but
`crates/khive-gate/docs/api/audit-events.md:3-4` describes one event reaching both sinks, so either
the doc or the parity should move.

Test coverage gaps that follow directly from the above: a secret in a key position, a benign-key
high-entropy value, a nested or pre-stringified secret, a failed mutating hook, and the coordinator
link effective projection. Successful hook mutation is covered; hook failure is not.

## CI

The secret scan red on this run is **not attributable to this branch** — and unusually, that is
measured rather than assumed: the CI run for this branch's merge base `a46914ae` shows the same
`Secret scan (gitleaks)` job failing, while the comparable jobs pass. The finding is redacted, so the
leak itself stays unresolved in both directions.

Since then #2323 has merged, which bounds the scan to the pull request's own commit range and adds
`-v` so a finding names its file, line, commit and rule. Note that re-running the job does not pick
that up — a re-run replays the workflow file pinned to the original run. The discriminator is a step
name: a scoped run has `Resolve scan scope`, the wide one has `Scan git history for credentials`.
Your next push produces a scoped run, and a red at that point is worth reading.

ohdearquant:
The mask/hash ordering is right, and it is worth saying so first because it is the thing that would
be hardest to fix later. `argument_identity` canonicalizes, masks, and only then digests:

```rust
// crates/khive-runtime/src/audit_context.rs:78-86
let masked = crate::secret_gate::mask_secrets(...);
let digest = format!("blake3:{}", khive_types::Hash32::from_blake3(masked.as_bytes()));
```

The digest is over the masked envelope, not the raw one. That is the correct order.

## The masking population is narrower than the digest's durability requires

A high-entropy token is only flagged when it sits near a trigger word:

```rust
// crates/khive-runtime/src/secret_gate.rs:1104-1111
let entropy = shannon_entropy(token.as_bytes());
if entropy < ENTROPY_THRESHOLD { continue; }
// High-entropy token in trigger context — flag it.
if near_trigger { return Some((token, "high-entropy-token")); }
```

So a secret carried under a benign key — `data`, `payload`, `value` — is not near any trigger, is
not masked, and is therefore inside the bytes that get hashed. The digest becomes an offline oracle
for that secret: anyone holding the audit row can confirm a guess against it, forever.

Keys have the same gap from the other direction. They are masked by an independent call:

```rust
// crates/khive-runtime/src/audit_context.rs:88-96
map.keys().map(|key| crate::secret_gate::mask_secrets(key)...)
```

which sees only the key text, so a high-entropy secret used *as* a key has no adjacent context to
trigger on either.

Known prefixes, explicit credential labels, and credential-shaped hex are all handled, and because
the whole canonical JSON is scanned as text, nested and already-stringified values do get seen. The
gap is specifically the unrecognized-shape secret in benign surroundings.

Why this is worth holding rather than noting: under-masking here is not recoverable downstream. The
audit reader cannot un-hash a digest, and the row is durable by design. The existing assertion that
a digest does not contain plaintext is true of both the correct and the incorrect implementation, so
it cannot detect this.

**Suggested direction:** for the audit identity specifically, use a conservative projection —
mask any high-entropy secret-shaped leaf *and key* regardless of trigger context — and hash only
that projection. Audit identity can afford to over-mask in a way that a general-purpose text scanner
cannot, because nothing downstream needs to read the value back. Tests worth adding: a benign key
with a high-entropy value, a secret-shaped key, and a fixed masked-projection-to-digest vector.

This is the same underlying gap that shows up wherever the trigger-adjacency heuristic is asked to
carry a durability guarantee, so a fix here is worth coordinating with the other change touching
that predicate rather than solving twice.

authored it, so its output was treated as input evidence only. The mask-before-hash order, the
`near_trigger` condition on the entropy rule, the independent key-masking call, and
`audit_context.rs`'s absence from the base were each re-derived from the refs before posting.

ohdearquant:
Closing as stale: the branch conflicts with main after the audit and gate changes landed since it was opened, and the review found the argument-identity digests need a keyed construction and bounded caller keys before this can merge. The canonical-argument work returns as a fresh branch on current main; the two linked issues stay open.

PR #2335, opened 2026-08-30
perf(db): route file reads through pooled readers

### LINKED ISSUE #1987: git.digest fails with 'timeout during sql_bridge.reader_open' under concurrent MCP client load
## Summary

`git.digest` with `max_items=100` fails with `invalid input: timeout during sql_bridge.reader_open` when other MCP clients are concurrently active against the same daemon. Small digests (`max_items=5`) succeed under the same conditions. The failure is intermittent and load-dependent: measured ~30s wall time before the error returns.

## Environment

- kkernel 0.7.0, build `5b0b140af39da2ddc10930931f7e87555660d13e`
- WAL mode, shared warm daemon, several concurrent MCP stdio clients (3+ active sessions issuing reads/writes during the failure window)
- `KHIVE_CHECKOUT_TIMEOUT_SECS` unset (default 5s per `pool.rs`)

## Reproduction shape

1. Have several MCP clients active against the daemon (long-lived sessions issuing periodic ops).
2. Run a cold-cursor `git.digest` with `max_items=100` against a repository with pending backlog:
   `kkernel exec --strict '[{"tool":"git.digest","args":{"source":"https://github.com/<owner>/<repo>","max_items":100}}]'`
3. rc=1, per-op error `invalid input: timeout during sql_bridge.reader_open` after ~30s.
4. Same command with `max_items=5` succeeds immediately; the identical `max_items=100` call succeeds when the daemon is quiet.

## Diagnostics at failure time

`db_diagnostics()` immediately around a failing window reported `writer_acquisition_timeouts: 0`, `pooled_checkout timeouts` not elevated, checkpoint probe healthy (`busy: 0`), WAL logical frames low. The starvation is on the reader-admission path (`sql_bridge_reader_slots` semaphore / `checkout_timeout` in `open_cached_reader_handle`, `crates/khive-db/src/sql_bridge.rs`), not the writer lane, and it is not surfaced by any contention counter — the failure is invisible in diagnostics.

## Why this matters

- A long ingest op holds reader admission across per-item work, so its own footprint plus a few concurrent clients exhausts the reader budget; the op then fails as a whole after partial progress, and the caller sees an all-or-nothing error for a cursor-resumable operation.
- Observed error rate in a production-like setup: roughly a third of digest invocations failing during busy hours, all with this signature.

## Asks

1. Surface reader-checkout timeouts in `db_diagnostics` (`writer_contention` has counters; the reader path has none), so this failure mode is observable.
2. Consider bounding reader-slot hold time per digest item (acquire/release per item rather than across items), or letting `git.digest` degrade to a smaller batch instead of failing the op when reader admission times out — it is cursor-resumable by design.
3. Document `KHIVE_CHECKOUT_TIMEOUT_SECS` interaction: the observed ~30s wall before failure vs the 5s default suggests multiple sequential admission attempts inside one op; worth one line in the error text saying which phase timed out.

### LINKED ISSUE #2024: Route file-backed store reads through pooled readers (ADR-165 Slice 2)
The file-backed read arm of every store (text, vectors, note, entity, graph, event) opens a fresh standalone connection per read (`open_standalone_reader`); the pre-opened pooled readers (`ConnectionPool::reader`) are reached only by the non-file-backed arm. Costs: per-read connection setup, cold SQLite page cache per operation, and a churning standalone-reader population whose WAL snapshots pin checkpoint progress (observed: sustained WAL pressure past the high-water mark with PASSIVE checkpoints unable to reclaim).

Scope per ADR-165 Slice 2: route the file-backed read arm through `ConnectionPool::reader`; keep `open_standalone_reader` only for a closed exception list at the pool surface (boot-time schema probing, diagnostics needing an independent snapshot, ADR-091-documented constraints). Pool exhaustion must return the existing timeout error, never fall back to a standalone open. Lands with ADR-166 G2.

Refs: ADR-165 Slice 2 (proposed, PR #2022), ADR-091, ADR-166 G2.

ohdearquant:
## Summary

- Route all SQLite typed-store reads and ordinary raw-SQL reads through the bounded shared reader pool, preserving request cancellation and providing no standalone fallback.
- Keep an explicit deferred raw-SQL read transaction as the only request-path standalone reader and charge it against the same reader-admission limit; enumerate boot-schema probes and independent snapshot diagnostics as the closed infrastructure-only exceptions.
- Add pool-scoped reader route, saturation, timeout, active/peak/completed, and maximum completed hold-time evidence to `db_diagnostics`.
- Add focused TDD regressions plus the ADR-166 G2 production-shaped guard for entity, note, graph, event, text, raw SQL, and vector reads.

## Compatibility and operations

- Ordinary pool saturation returns retryable `AdmissionTimeout`; request cancellation remains `Timeout`.
- Explicit `BEGIN` reader acquisition retains the existing `sql_bridge.reader_open` timeout phase.
- `KHIVE_CHECKOUT_TIMEOUT_SECS` applies to each reader-admission attempt; a multi-read operation can accumulate more wall time across attempts.
- The current PASSIVE WAL diagnostic remains writer-owned and does not open a reader; infrastructure exceptions are explicit and never used as a fallback.

## Test plan

- `cargo check --workspace`
- `cargo test --workspace`
- `cargo test -p khive-db --lib hot_path_guard_g2_file_backed_read_suite_uses_only_pooled_readers`
- `cargo test -p khive-db --lib --features vectors hot_path_guard_g2_file_backed_read_suite_uses_only_pooled_readers`
- `cargo test -p khive-pack-kg --test integration db_diagnostics_runtime_audit_fields_are_additive`

Closes #2024
Closes #1987


ohdearquant:
Reviewed at head `9eed6e3c67adedf380ca9937d23c2243d0536fab`. Leading with the part that matters most,
because it is a negative result and negative results are easy to under-report: **the two hazards this
change most plausibly introduces are not present.**

- **No nested reader acquisition.** The shared helper (`crates/khive-db/src/stores/mod.rs:26-51`)
  takes one `ReaderGuard` and runs a synchronous `FnOnce(&rusqlite::Connection)`, so the closure
  structurally cannot await another store read. I walked the multi-read paths individually — entity
  count-then-data, the chunked graph and note loops, text term-statistics, vector batching, and the
  runtime `search_notes` fan-out — and each acquires and releases per operation rather than holding a
  permit across an inner read. The deferred `BEGIN` transaction reuses its retained permit rather than
  taking a second one (`sql_bridge.rs:969-991`, `:1062-1108`).
- **No ordinary standalone fallback survives in the request path.** A whole-tree direct-open scan with
  a must-match control resolves to one request-path standalone reader — the explicit raw-SQL
  transaction at `sql_bridge.rs:838-840` — with the remainder being pool constructors, migration and
  schema inspection, diagnostics, and the separate showcase crate.

Given the design removes the fallback, those two being clean is the substance of the change working.
The rest is smaller, but one item blocks and one is worth arguing about.

## Docs lint is failing, and it is yours — plus a collision you should know about

`deno fmt --check` reports two unformatted files: `docs/adr/ADR-023-declarative-pack-format.md` and
`AGENTS.md`. It is table repadding — the `db_diagnostics` row grew, which widens the column and leaves
every other row in that table short. `git diff --check` passes, which is why it is easy to miss; it
does not run the repository's formatter.

More usefully: **this PR and #2332 are on a collision course, and the Docs lint failure is a symptom of
it.** Both are adding fields to the same `db_diagnostics` surface, and they overlap on three files:

- `AGENTS.md` — **identical hunks**, `@@ -62,7 +62,7 @@` and `@@ -561,7 +561,7 @@` in both. Same two
  lines, same table row.
- `docs/adr/ADR-091-wal-snapshot-lifetime.md` — both append at the same anchor, `@@ -1496,3 @@`.
- `crates/khive-db/src/diagnostics.rs` — both add fields to `pub struct DbDiagnostics` and both extend
  `collect_with_runtime_audit_metrics_interruptibly` and `collect_inner`.

The doc overlaps will conflict outright. The `diagnostics.rs` additions are semantically additive and
should compose, but every construction site has to carry both sets of fields, so whichever merges
second needs a real rebase rather than a formatting pass — and will need the formatter run again,
because the column width changes a second time. Worth coordinating the order rather than discovering
it at merge.

## The admission bound is per attempt, and this change is what makes that bite

The docs are accurate that `KHIVE_CHECKOUT_TIMEOUT_SECS` applies per admission attempt
(`docs/guide/api-reference.md:802-804`). The consequence is worth stating explicitly, because removing
the fallback is what converts it from a latency note into a failure mode.

For a filtered note search the arithmetic is:

- `candidate_limit()` is `limit * 50` capped at `FILTERED_SCAN_CAP = 500`
  (`crates/khive-pack-kg/src/handlers/search.rs:7`, `:226-231`).
- The runtime overfetches: `let candidates = limit.saturating_mul(4).max(limit);`
  (`crates/khive-runtime/src/operations.rs:4077`), so up to 2,000.
- Hydration is one store call per candidate — `for id in &candidate_ids { note_store.get_note(*id).await? }`
  (`:4173-4175`) — over a union the code's own comment bounds at `≤ 2×candidates`, so up to ~4,000.

That is roughly 4,000 sequential admission attempts in a single request, plus the retrieval legs and
graph hydration chunks. Each is individually bounded; the operation is not. Under saturation the
worst-case admission wait scales with the candidate count rather than with the configured setting, and
an operator who sets that variable expecting it to bound a request will be surprised.

The loop pre-exists this change — `git show origin/main:crates/khive-runtime/src/operations.rs`
carries the same per-candidate hydration — so this is not a regression you introduced. But previously a
saturated pool degraded to a standalone reader here, and now it does not, which is precisely the
scenario where 4,000 attempts each wait out their ceiling. An operation-level budget, or bulk
hydration, would close it. Consciously accepting and documenting it is also a defensible answer; what
is not defensible is leaving the docs implying a single-window bound.

## The G2 guard is a sample, not the invariant ADR-166 describes

`crates/khive-db/src/backend.rs:920-988` exercises one operation each from entities, notes, graph,
events, text, optional vectors, and one raw read, snapshots `ReaderAcquisitionSnapshot`, asserts a
fixed pooled delta and a zero standalone delta. That is a good behavioral test and it would redden a
regression in those exact calls.

It is not a closed-world guard. A new unpooled read added tomorrow in a file or method it does not
name leaves it green, because its population is a hard-coded list of calls rather than an enumeration
of read seams. ADR-166 describes G2 as an enforcement invariant, which is a stronger claim than the
implementation supports — and this is the guard the whole no-fallback design leans on to stay true
over time.

Closing it means giving the guard a population it derives rather than one it hard-codes: a source-level
check over every `SqlReader`/typed-store read entry point, or a registration the seams must join. Keep
the counter assertions; they are the right behavioral half.

## Follow-ups, not blockers

- **Schema inspection bypasses the purpose enum.** `migrations.rs:996-1005` calls
  `open_read_only_snapshot_connection` (`pool.rs:2062-2065`) directly, reached from
  `kkernel/src/cli.rs:767-770`, and `migrations.rs:1455-1461` does the same for the model registry.
  These are infrastructure and CLI reads, not request-path, so the one-standalone-reader count stands —
  but the enum is presented as the closed classification mechanism and these are outside it and
  uncounted. Pre-existing; worth either routing through an explicit probe or documenting as a separate
  exception.
- **`AdmissionTimeout` is terminal in the audit batch classifier.** `audit_batch.rs:385-396` retries
  `Pool` and `Timeout` but not `AdmissionTimeout`, and its caller is the event append loop at `:440`.
  Pre-existing and on a background write lane, not the reader request path — but it is now the one
  place in the tree where the new retryable class is treated as fatal, so it deserves an explicit
  decision rather than being left as an omission. The ordinary paths get this right: `error.rs:573-617`
  marks it retryable, the retrieval adapter maps both to a transient query timeout, and knowledge
  search deliberately propagates `AdmissionTimeout` while fail-opening only on cancellation `Timeout`.

## Also verified sound

Store-by-store, the nine collapsed helpers preserve their prior SQL, mapping, filters and chunking;
none carried a store-specific timeout or transaction scope that the shared path drops. Error mapping is
centralized in `resolve_reader_checkout`, so exhaustion is `AdmissionTimeout` and pre-checkout
cancellation is `Timeout`. The `sql_bridge` use of `ReaderContract` for explicit-handle paths is a
deliberate compatibility classification, not an inversion. Counter coverage checked increment-site by
increment-site including the early-return and error paths, with cancellation-before-admission correctly
excluded from the timeout counter. ADR-005, ADR-091, ADR-165 and the API reference all match the code;
ADR-166's wording is the one seam.

## CI

`Secret scan (gitleaks)` is red with `leaks found: 1` and redacted output — no file, rule, line or
commit — so it is unresolved in both directions and licenses no conclusion. Since this run, #2323 has
merged, bounding the scan to the pull request's own commit range and adding `-v` so a finding names
itself. Re-running the job will not pick that up: a re-run replays the workflow file pinned to the
original run. The discriminator is a step name — a scoped run has `Resolve scan scope`, the wide one
has `Scan git history for credentials`. Your next push produces a scoped run.

`CI gate` is the aggregate and carries no independent signal. Everything else that completed is green.

ohdearquant:

`Docs lint` fails at `Check markdown formatting`: `AGENTS.md` and `docs/adr/ADR-023-declarative-pack-format.md` are not `deno fmt` clean.
wrapping, no content change.

ohdearquant:
The pooled-routing work looks right, and the guard is a good idea. The issue is what the guard's
assertion actually ranges over.

## The routing guard checks a hand-written list, not the population it stands for

The test constructs six stores plus optional vectors and a raw handle, then asserts:

```rust
// crates/khive-db/src/backend.rs:973
let expected_pooled_delta = 6 + u64::from(cfg!(feature = "vectors"));
```

over `entities`, `notes`, `graph`, `events`, `text`, `vectors`, and raw SQL.

Three production pooled-read seams are outside that set, and they are real ones — each routes
through the same helper the guard exists to protect:

```rust
// crates/khive-db/src/stores/agents.rs:272      StorageCapability::Sql
// crates/khive-db/src/stores/attachment.rs:147  StorageCapability::Attachments
// crates/khive-db/src/stores/sparse.rs:274      StorageCapability::Sparse
super::run_pooled_store_read(Arc::clone(&self.pool), ..., op, ...)
```

The consequence is the one that matters for a guard: reverting any of the six listed paths to the
old file-backed route reddens this test, but adding or changing an unpooled read in agents,
attachments, or sparse leaves it green — as does any new read seam added in a file the fixture does
not construct. So the test defends the six paths it names rather than the invariant it is written
for, and its failure mode is silence on exactly the case a guard is for: a *new* seam that someone
routes wrongly.

A fixed count over a hand-enumerated fixture is a population declaration, and this one is narrower
than the claim it is standing in for.

**Suggested fix:** either include all nine typed stores in the counter assertion, with the expected
delta covering nine typed reads plus the raw one, or derive the list mechanically so that adding a
new typed read seam without adding it to the guard is itself a failure. The second is more work now
and is the version that keeps working.

## The docs gate is red on a changed file

`Docs lint` fails at this head — `deno fmt --check` reports two unformatted files, including the
table changes in `docs/adr/ADR-023-declarative-pack-format.md` around `:158-169` and `:564`. The
formatting is the only thing at issue; applying the repository's Deno Markdown formatting to the two
reported files should clear it.

authored it, so its output was treated as input evidence only. The guard's constructed set and its
expected delta, and the three omitted `run_pooled_store_read` seams, were each re-derived from the
ref before posting.

ohdearquant (commented):

Verdict on head 8a8c973057d034bad010f5213ec48106e8a2f00a: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 791a4e1395457417b0ffb24b30978ba4130b38c8: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):

PR #2337, opened 2026-08-30
fix(mcp): preserve admitted cancellation outcomes

### LINKED ISSUE #2091: mcp: client cancellation after the daemon frame is written reports 'cancelled' while the operation commits; retries duplicate (frame carries no request identity)
## Observed

An MCP client that cancels an in-flight `request` tool call can receive a cancellation error while the daemon commits the operation anyway. A retry then duplicates the side effect. Measured with `comm.send`: three duplicate deliveries in one day, each pair = one cancelled call plus one retry, both committed.

## Mechanism (paths at main)

The bridge forward is a one-shot socket exchange: `try_forward_inner` in `crates/khive-mcp/src/daemon.rs` writes the request frame (`write_frame`, ~line 514) and then awaits `read_frame`. When the MCP client cancels the request, the in-flight tool future is abandoned between those two points. Nothing propagates the cancellation to the daemon, which completes dispatch and commit and writes its response to a socket nobody is reading.

The neighboring arm already names this ambiguity: when the daemon closes the connection before responding, the code returns a terminal-ambiguity error precisely because "the request may already have committed." Client cancellation is the same ambiguity class, but it produces no signal at all — the caller sees a plain cancellation, which reads as "did not happen."

`DaemonRequestFrame` (`crates/khive-runtime/src/daemon.rs:407`) carries ops, presentation, namespace, actor, provenance, and visibility — but no per-request identity. The daemon therefore cannot deduplicate a retried request even in principle.

## Proposed direction

1. **Request identity + dedup**: the bridge stamps a per-request UUID on every frame; the daemon keeps a short-lived seen-set for side-effecting verbs (at minimum `comm.send`/`comm.reply`) and answers a duplicate id with the original result — at-most-once commit per bridge request. The existing `request_id` audit-correlation parameter is adjacent infrastructure but is caller-supplied and optional; this needs to be unconditional and bridge-stamped.
2. **Shield the post-write window**: once the frame is written, complete the socket exchange on a detached task even if the tool future is cancelled, and surface the outcome as an audit event so a cancelled-but-committed call is observable rather than silent.
3. Related: #1843 covers interrupt propagation for in-flight storage queries on disconnect; this issue is the commit-ambiguity and retry-duplication seam above it.

## Workaround

After any cancellation error on a send-shaped op, read the target thread and confirm absence before resending.

### LINKED ISSUE #2222: MCP batch dispatch reports bare "request cancelled" while some ops committed
## Symptom

A multi-op MCP request can fail with a single opaque \`request cancelled\` error while some or all ops in the batch have already committed. The response carries no per-op results, so the caller cannot tell which ops landed.

## Observed reproductions (same day, independent callers)

1. A 6-op \`link\` batch returned bare \`request cancelled\`; inspection showed 3 of the 6 ops had committed. The other 3 had per-op validation errors (edge pair-restriction refusals) that the wrapper error completely masked — the caller saw "cancelled", not "invalid".
2. A 5-op \`comm\` batch (2 sends + 3 reads) returned bare \`request cancelled\` twice in a row; a later read of the destination thread showed the sends had committed on each attempt, so the natural retry delivered the same message three times.

## Why this is worse than a flaky transport error

- **Masked validation errors**: the real per-op refusal never reaches the caller, so a verbatim retry re-sends the same invalid ops and fails the same way.
- **Duplicate side effects**: for non-idempotent ops (sends, creates), "cancelled" reads as "nothing happened", making retry the obvious move — and the retry duplicates the committed ops.

## Expected

Either the batch is atomic under this failure (nothing commits when the wrapper reports cancellation), or the error surfaces per-op results: which ops committed, which failed, and each failure's actual error. Any partial-commit outcome must be distinguishable from a clean cancellation.

ohdearquant:
Summary:
- shield admitted daemon request frames from MCP handler cancellation
- return the daemon's terminal per-operation envelope instead of a bare cancellation error
- attach an opaque request identity to every bridge attempt and document reconciliation behavior

Verification:
- git diff --check

Closes #2222
Closes #2091


ohdearquant:
Reviewed at head `a02f3d93cc1e61d259e1f3e080ed3812e21b74d2`.

Leading with the part that took the longest to establish, because it runs opposite to the way this
change first reads: **the ownership change is correct, and it closes a real defect rather than
introducing one.**

Before this change the forward was a locally-pinned future (`let forwarded = forward_fn(...); tokio::pin!(forwarded);`)
and the cancellation arm did `return Err(McpError::internal_error("request cancelled", None))`, which
dropped that future. Dropping it mid-exchange abandons a socket write that may already have delivered
a complete frame, so the daemon could commit operations the bridge then reported as a bare
cancellation with no structure at all. Spawning an owned task and awaiting it under cancellation is
the right fix for that, and `tokio::spawn` detaching the work is what makes the guarantee real rather
than advisory.

I also checked the `retryable: true` classification at `crates/khive-mcp/src/server.rs:3031-3039`,
since a wrong answer there is a duplicate mutation. **It is sound**, but only through a chain that is
nowhere written down — see the last section, which asks you to write it down rather than change it.

Two things block, one of them in the ADR.

## The read deadline is no longer enforceable on this path

This is the one to fix. `scope_request_read_deadline` does not wrap anything in a timeout — it stores
a deadline in a task-local and awaits the future unchanged
(`crates/khive-storage/src/request_context.rs:172-193`). Enforcement is entirely cooperative: it
happens wherever code observes `request_read_is_cancelled()` or awaits
`wait_for_request_read_cancellation()`.

On the daemon-forward path there was exactly one such observation point, and this change converts it
from an exit into a wait. The cancellation arm now runs `forward_task.await` with no deadline of its
own (`server.rs:2991-2999`), and the forward it waits on is itself unbounded — `try_forward_inner`
calls `read_frame` with no timeout, and `read_frame` is a blocking `read_exact` pair
(`crates/khive-runtime/src/daemon.rs:643-661`). A daemon that accepts the connection and never
answers now holds the handler indefinitely, and the configured request read timeout has no remaining
mechanism to end it.

The unbounded daemon read pre-exists this change; what is new is that cancellation used to bound it
and no longer does. So this is a liveness regression traded for the correctness win above, and it
does not need to be a trade. The ownership change is what makes the cheap fix available: because the
task is spawned and detached, the handler can stop waiting without dropping the daemon exchange —
the frame still completes and the audit rows still land. Put a second, separate deadline on the
shielded await, and on expiry return `outcome: unknown, retryable: false`. That keeps the property
you are protecting — never retry, never fall back locally after a possible write — while restoring a
bound. Returning early is safe here in a way it specifically was not before.

## Amendment 8's heading was inserted above the previous amendment's closing sections

`docs/adr/ADR-049-khived-daemon.md` gained `## Amendment 8` at line 986, and the single hunk is
`@@ -983,6 +983,59 @@`. At the base ref, lines 986 and 994 were Amendment 7's `### Test obligations`
and `### What is unchanged`. The new heading landed above both, so they are now inside Amendment 8.

That produces three wrong readings, and the third is the one that matters:

- Amendment 7 no longer has test obligations. Its `### Test obligations` still exists in the file but
  now documents a different amendment.
- Amendment 8 has two `### Test obligations` headings, at 1030 and 1039, with different content.
- Amendment 8 now ends with a `### What is unchanged` reading "This amendment changes one exit code,
  one disposition string, and the reserved set." That is Amendment 7's scope statement. Amendment 8
  changes no exit code. A reader consulting the section that exists precisely to bound an amendment's
  blast radius gets another amendment's answer.

Moving the new `## Amendment 8` heading and its body below line 994 of the base text fixes all three.
Worth noting that nothing mechanical catches this: the insert is clean, the prose is well-formed, and
the diff is additive.

## The ADR states a property the code implements conditionally

Boundary 1 of the new amendment reads: "If the cancellation signal is already visible before daemon
admission, the bridge starts neither daemon forwarding nor local dispatch and reports a
not-dispatched cancellation." That is unconditional.

The guard implementing it sits inside `if p.save_to.is_none()` (`server.rs:2954`), which wraps the
entire daemon-forward block including the new pre-admission check at `:2959-2964`. A request carrying
`save_to` skips the block and falls through to `dispatch_request_wire`, so an already-cancelled
request does start local dispatch. The behavior itself is unchanged from the base — the `save_to`
bypass pre-exists and the check is new — so this is not a regression. But the ADR text is new, and it
claims more than the code delivers. Either scope the sentence to daemon-eligible requests, or hoist
the cancellation check above the `save_to` guard, which looks like the better answer: nothing about
the reasoning behind it is specific to daemon forwarding.

## The pre-admission cancellation is the one case with no envelope

`server.rs:2960-2963` returns `McpError::internal_error("request cancelled before daemon dispatch", None)`.
Its two siblings both carry structure — the join-failure path reports `outcome: unknown, retryable: false`
and the post-admission path reports `outcome: not_dispatched, retryable: true`.

The omission inverts the information gradient. Pre-admission is the case where the answer is
*certain*: nothing was written, nothing dispatched, the request is unambiguously safe to retry. It is
the one a caller can act on without judgment, and it is the one that arrives as prose. It is also not
a scoping constraint — `ensure_bridge_request_id` runs at `:2908`, so `request_id` is already in
scope at that return.

## Why `retryable: true` is sound, and why it should be written down

I went looking for a duplicate-execution hazard here and did not find one, but the argument is longer
than it should be for something load-bearing.

`forwarded == None` means the forward reported `NoSocket` or `KHIVE_NO_DAEMON`, and the contract at
`crates/khive-mcp/src/daemon.rs:2074-2078` asserts a `NoSocket` outcome never writes anything. The
implementation does not obviously honor that: `write_frame(...).is_err()` maps to `NoSocket`
unconditionally (`daemon.rs:537-539`), and `write_all` can return an error after delivering bytes.
It holds anyway, for two reasons that live in a different crate:

- A partially written frame is never dispatched. `read_frame` does `read_exact` on the four-byte
  length and then `read_exact` on exactly that many bytes, so a short frame stalls and then EOFs
  without reaching dispatch.
- A fully written frame cannot produce an error. `write_frame` is `write_all(len)` → `write_all(payload)`
  → `flush` (`crates/khive-runtime/src/daemon.rs:665-682`), and the call site's stream is a concrete
  `tokio::net::UnixStream`, whose `poll_flush` returns `Poll::Ready(Ok(()))`.

So the classification is right, and it is right because of an infallible flush on one concrete
transport. Introduce a buffering writer between the frame encoder and the socket and the second bullet
inverts silently: a fully buffered frame whose flush fails becomes `NoSocket`, becomes `None`, becomes
`retryable: true`, becomes a double-apply — with no test failing, because every test today uses a
transport that cannot flush-fail. Two cheap things close it: distinguish the write-failure case from
the connect-failure case at `daemon.rs:537-539` rather than folding both into `NoSocket`, and state
the invariant where the contract is already written at `:2074-2078` — that `NoSocket` is safe because
a partial frame is undispatchable and a complete one cannot error.

## Smaller and pre-existing

- The envelope carries no distinction between not-started, started-not-committed, and committed.
  `outcome: unknown` covers the middle case honestly, but a caller that could safely retry a
  never-started request cannot tell it apart from one that may have committed. Pre-existing, and a
  larger design question than this change.
- There is a window between the forward completing with `None` and the local fallback where a
  cancellation arriving in between is handled by the `cancelled_during_forward` flag rather than a
  re-check, so a cancellation landing after the select resolves is not observed. Pre-existing shape,
  and benign given the fallback path's own guards.

## Also verified sound

The spawned task takes ownership of the frame and pack list, so nothing it needs borrows the handler.
The join-error path is classified non-retryable with the request id attached. `ensure_bridge_request_id`
preserves a caller-supplied id unchanged and mints a nonzero id otherwise, folding both UUID halves
with `.max(1)`. The strict-fallback reshaping at the `Some(res)` arm is untouched and still routes
only marker-tagged errors into the per-operation envelope.

## CI

`Secret scan (gitleaks)` is red with a redacted finding — no file, rule, line or commit — so it
supports no conclusion in either direction, and the merge-base run shows the same job failing while
comparable jobs pass. Scanning was recently bounded to a pull request's own commit range with
verbose output, but re-running this job will not pick that up: a re-run replays the workflow file
pinned to the original run. The discriminator is a step name — a scoped run has `Resolve scan scope`,
the wide one has `Scan git history for credentials`. Your next push produces a scoped run, and a red
at that point is worth reading.

ohdearquant:
## The red `Secret scan (gitleaks)` is not from this change

`main` now resolves the scan scope before running gitleaks, so a pull request run judges only the
commits under review. Branches whose merge base predates that change still run the unscoped form,
which walks every ref in the checkout and reports a credential that has been in history
independently of this work — which is why the finding is identical on several unrelated branches
and why the log shows several thousand commits scanned rather than the handful in this diff.

This branch currently conflicts with `main`, so it cannot pick that up automatically. Resolving the
conflict and merging `main` in brings the scoped scan with it, and this check should go green with
no change to the code here.

Full-history coverage is unaffected either way: the nightly scheduled run takes no range and still
scans every ref, so a credential reachable from any branch is still found within a day.

ohdearquant (commented):

Verdict on head 05b2d860d7c84ea5f25656cf004f17550e18aa84: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving at c1c59ceceec2f5385dad241874f8107e698eca86. Checks are green at this exact commit.

I confirmed the following at this head rather than from the change description. One absolute deadline now bounds the connect, write and read phases of the socket exchange together, so none of the three can outlive the request. The post-cancellation shield waits only for the time remaining on that same deadline rather than starting a fresh ceiling, and returns immediately when no time is left. The pre-dispatch cancellation check now runs before the save-to branch is chosen rather than inside it, so a cancelled request carrying a save-to target can no longer reach local dispatch. The local fallback re-reads cancellation directly instead of relying on which select arm won, which closes the race where a no-outcome forward and an arriving cancellation land in the same instant.

PR #2338, opened 2026-08-30
docs(mcp): define read acknowledgement ordering

### LINKED ISSUE #1940: Batch execution lets mark-read succeed while the paired send in the same batch fails
Measured batch: `[comm.send(...), gtd.assign(...), comm.read(A), comm.read(B)]` in one request, under writer contention.

Both write ops failed with `database is locked`. Both `comm.read` ops succeeded and marked their messages read. Net effect: messages transitioned to read while the replies that justified marking them were never sent.

The batch surface currently gives no way to express "these reads should not advance state if the sends they are paired with fail": parallel batches have no ordering guarantee, and mark-read is a state-mutating write that rides on a nominally read-shaped verb, so it does not share the failure footing of the sends beside it.

Suggested direction (not a design): state-mutating reads (mark-read) should fail under the same conditions that fail writes — a locked store should refuse the mark rather than letting it race ahead of the batch's writes. Alternatively, a caller-visible contract that mark-read in a batch is independent of sibling ops, so callers know to sequence with an explicit chain.

Companion reporting defect filed separately: the failed mark currently reports ok:true.

ohdearquant:
Summary:
- declare comm.read and comm.mark_read as state-mutating acknowledgements at the MCP request boundary
- document that parallel siblings are independent and a failed send cannot roll back a committed read mark
- direct dependent callers to an abort-on-failure chain or the deliver-first comm.reply flow
- lock the caller-visible request-tool contract with an integration regression

Verification:
- git diff --check

Closes #1940

### DIFF
--- a/crates/khive-mcp/docs/design.md
+++ b/crates/khive-mcp/docs/design.md
@@ -22,6 +22,11 @@ decisions and rationale.
   callers must not infer full success solely from the absence of an RPC-level error.
 - Per-op failures do not abort siblings in Parallel mode; they do abort remaining
   ops in Chain mode (reported as `{"ok": false, "aborted": true}`).
+- `comm.read` and `comm.mark_read` are state-mutating acknowledgements, not
+  dependencies on sibling delivery operations. In Parallel mode they may commit
+  even when a sibling `comm.send` or `comm.reply` fails. Callers that need the
+  acknowledgement conditioned on delivery must use a Chain, or use `comm.reply`
+  for the common deliver-first/read-mark-second flow.
 - Invalid DSL (parse/lex failure) is preflighted before warm-daemon forwarding
   and returns an RPC-level `invalid_params` error; its error data carries
   `reason: "parse-error"` on both local and daemon-available paths. Per-verb
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -2815,6 +2815,13 @@ or `"writer_task_begin_busy"`)
 never rolls back a sibling that already committed. Inspect each result
 entry's own `ok` field rather than assuming batch-level atomicity.
 
+`comm.read` and `comm.mark_read` mutate delivery state. In a parallel batch,
+either acknowledgement does not wait for or depend on comm.send/comm.reply, so
+a read mark can commit even when the sibling send fails. When the mark must
+depend on a send, use a chain so a failed send aborts the mark. For the common
+reply-and-read flow, prefer `comm.reply`: comm.reply delivers first, then attempts
+the original message's best-effort read mark.
+
 `search` carries its own per-op `status` ("complete" | "partial") inside that
 op's `result` entry, separate from the top-level batch `status` above. A
 degraded-but-answered search stays ok:true with status="partial" plus a
--- a/crates/khive-mcp/tests/integration.rs
+++ b/crates/khive-mcp/tests/integration.rs
@@ -379,6 +379,32 @@ async fn request_tool_description_contains_dynamic_verb_catalog() -> anyhow::Res
     Ok(())
 }
 
+#[tokio::test]
+async fn request_tool_description_declares_comm_read_dependency_contract() -> anyhow::Result<()> {
+    let client = connect().await?;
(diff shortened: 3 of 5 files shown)


ohdearquant:
Reviewed at head `a705244f94437335301523a3460e2ca4896d2aac`. The documented contract is accurate — I
checked the load-bearing sentence against the code rather than against the other three copies of it.

`comm.reply` does commit delivery before attempting the mark. The reply is constructed and persisted
first, and the read-mark is a subsequent `set_note_property` whose failure degrades to
`marked_read: false` rather than failing the call — the code says so in as many words ("the reply is
already committed above, so a failed or no-op patch degrades to `marked_read: false` rather than
failing a delivered reply"). The mark is also skipped when the original is outbound or the caller is
not the addressee, which matches what the lifecycle doc already described. So "a delivery failure
cannot mark the original read" holds, and it holds for the stated reason.

Naming `comm.read` and `comm.mark_read` as mutations is worth doing. Both read as acknowledgements,
and the parallel-independence consequence is genuinely non-obvious from their names.

## The test guards the words, and the words are now a promise

`request_tool_description_declares_comm_read_dependency_contract` asserts three substrings are
present in the `request` tool description. As a documentation-presence guard that is sound, and the
choice of surface is right — the tool description is what a caller actually reads, so a doc-only
contract that lives elsewhere is a contract nobody sees.

But consider which mutation it reddens. Delete or reword the paragraph and it fails, correctly.
Reorder `handle_reply` to mark before it delivers and it stays green, because the description still
contains "comm.reply delivers first" — the sentence simply becomes false. The test cannot distinguish
a true promise from a false one, which is exactly the distinction this change introduces: before it,
the ordering was an implementation detail; after it, it is a documented guarantee callers are invited
to depend on.

The existing suite covers the mark's *value* — `marked_read=true` for an inbound original,
`marked_read=null` for an outbound one, and the addressee-scoping case. Searching the test tree for
`marked_read` I did not find one that fails the delivery and asserts the original stays unread, which
is the arm the new strongest sentence rests on. That negative is only as good as the search, so treat
it as a pointer rather than a finding: if that coverage exists under a name I did not match, this
section is moot. If it does not, one test — reply with delivery made to fail, assert the original is
still unread — would put behavior behind the promise.

## One wording suggestion

`crates/khive-pack-comm/README.md` recommends "a `send | mark_read` chain when that dependency is
required." The chain does give abort-on-failure sequencing, so the advice is correct. The trap is
that the chain idiom a reader will reach for next is `$prev`, and `comm.send(...) | comm.mark_read(ids=[$prev.id])`
marks the *newly sent* message rather than the original — the id that needs marking is not the one
the chain hands forward. Worth one clause saying the original's id must be supplied explicitly,
since the whole point of the paragraph is to steer callers toward a form that does what they meant.

The other three copies agree with each other and with the code. Nothing else to raise.

checked at the head named.

ohdearquant:
Approving.

Documentation-only (+55/-0) plus one description assertion. For a docs change the
only question that matters is whether the claims are true, so I checked the
load-bearing one against the implementation rather than against the prose.

The claim is that `comm.reply` commits delivery first and only then attempts the
original's read mark, so a failed delivery cannot leave the original marked read.
`handle_reply` does exactly that: `dual_write_message(...).await?` commits the pair,
and the `?` returns early on failure, so the mark below is unreachable when delivery
fails. The mark is then `set_note_property(...).unwrap_or(false)`, degrading to
`marked_read: false` rather than failing a delivered reply. The surrounding comment
in that function already stated the same ordering; this PR is putting it where callers
will actually see it, which is the right place for it.

The parallel-batch half is consistent with the documented batch semantics: a failed op
does not abort siblings, so a read acknowledgement placed beside a send can commit when
the send does not. Naming `comm.read` and `comm.mark_read` as mutations is worth doing
explicitly — both names are acknowledgement-shaped and read as queries.

On the test: asserting substrings of the tool description couples it to that prose, and
that coupling is deliberate here. The description is a shipped contract that callers
consume, and there is existing precedent in the same file for asserting its content.
I checked each asserted fragment against the text this PR adds to `server.rs`; all
four match.

No behaviour change, and the ordering guarantee being documented is one that already
holds in the code.

PR #2339, opened 2026-08-30
fix(runtime): enforce caller attribution boundaries

### LINKED ISSUE #1891: config: implement [gate] caller-enrollment enforcement (currently rejected fail-closed)
## Context

`KhiveConfig::load` rejects any config file containing a `[gate]` section (`ConfigError::UnsupportedGateSection`), because this build has no caller-enrollment enforcement and silently ignoring a requested policy would be fail-open.

Deployments that previously carried `[gate]` sections (e.g. `granted_actors` allowlists, `grant_unattributed = false`) written for builds that were expected to enforce them now refuse to start until those sections are removed. The refusal is correct; the enrollment intent behind those sections is real and currently has no supported expression.

## Ask

Implement config-declared caller enrollment so a `[gate]` section is accepted and enforced:

- `granted_actors`: allowlist of actor ids permitted to issue requests.
- `grant_unattributed`: whether requests with no resolved actor identity are admitted.
- Enforcement at the authorization seam (the `khive-gate` interfaces, ADR-018), not in storage or handlers.
- Startup must continue to refuse any `[gate]` key it does not fully enforce — partial support that ignores unknown keys reintroduces the fail-open this rejection exists to prevent.

## Acceptance sketch

- A config with `[gate]` + `granted_actors` boots, admits a listed actor, and rejects an unlisted one with a typed authorization error.
- `grant_unattributed = false` rejects unattributed callers; `true` admits them with the unattributed stamp.
- A config with an unknown `[gate]` key still refuses startup.

### LINKED ISSUE #2016: EventStore append accepts caller-chosen attribution: close the trait-boundary gap ADR-162 names
## Summary

ADR-162 states the attribution rule for the event plane at the trait boundary, not only at the
verb surface, and names the gap between the two as implementation work rather than an accepted
exception. This issue tracks closing it.

From `docs/adr/ADR-162-unified-event-plane-ownership.md`:

> **"Runtime-internal" is a claim about the trait boundary and not only about the verb surface,
> and the two are not in the same state today.** The verb surface satisfies the rule structurally,
> because no verb takes an event. The `EventStore` capability (ADR-005) is a different matter: it
> is an in-process trait whose append accepts a fully constructed event, including its namespace
> and actor fields, and any code linked into the runtime can call it.

and:

> **an in-runtime call site may compose an event, but it may not choose that event's attribution.**
> Namespace and actor are resolved from the same authenticated context that governs the causing
> work and are stamped from it, never carried in from values the composing code selected.

## Why this is not satisfied by the verb-surface argument

The verb surface is safe by construction: no verb accepts an event, so no external caller can
compose one. That argument says nothing about `EventStore::append`, which takes a fully
constructed event and is reachable from any code linked into the runtime. A rule stated only
where it is already structurally satisfied reads as enforced while the surface that actually
persists the row is unguarded.

## Scope of work

1. **Enumerate the call sites.** Every in-process caller of the `EventStore` append path, grouped
   by whether the event's `namespace` and `actor` are (a) stamped from the resolved request
   context, or (b) passed in as values the composing code selected. Group (b) is the conformance
   debt; its size is currently unmeasured, and the enumeration is the first deliverable rather
   than an estimate.

2. **Close group (b).** ADR-162 leaves the mechanism to the implementing decision and names two:
   - construction helpers that take the resolved identity and derive the fields, so a call site
     cannot express a wrong attribution; or
   - validation at the append boundary, rejecting an event whose attribution does not match the
     resolving context.

   The first is preferable where it applies, because it removes the ability to be wrong rather
   than detecting it. The second is needed wherever a construction helper cannot see the context.

3. **A test that fails before the fix.** An attempt to append an event with an attribution the
   calling context did not resolve must be rejected or unrepresentable, and the test asserting
   that must be observed failing against the pre-fix code. An assertion that has never been seen
   red is not yet a check.

## Out of scope

Caller-supplied *content*. ADR-162 forbids caller-composed events, not caller-supplied data inside
runtime-composed ones: a verb whose arguments carry a signal, a judgment, or a payload may have
that data recorded in the event the runtime composes, with the runtime stamping kind, attribution,
and context. This issue is about the attribution fields only.

## References

- ADR-162 — Unified Event-Plane Ownership (the rule and the named gap)
- ADR-005 — Storage Capability Traits (`EventStore`)
- ADR-018 — Authorization Gate (dispatch-time emission through the gate audit path)
- ADR-022 — Events Query Surface (namespace isolation on the read side)

## Maintainer

@ohdearquant

ohdearquant:
## Summary

- add an exact `[gate]` caller-enrollment schema with explicit attributed and anonymous grants, fail-closed validation, and an opaque daemon compatibility fingerprint
- make runtime event append capability token-scoped so caller-selected actor and namespace values are overwritten by resolved request attribution
- route registry audit, atomic, curation, brain, and KG projection append paths through their declared trusted attribution boundaries
- document the enrollment and EventStore attribution contracts and caller taxonomy

## Validation

- `cargo test --workspace -- --skip filtered_count_partitions_share_one_snapshot_during_concurrent_update` (pass, including doctests)
- `cargo test -p khive-db stores::note::tests::filtered_count_partitions_share_one_snapshot_during_concurrent_update -- --exact` (pass; isolated retry of unrelated flaky seam)
- `git diff --check`

Closes #1891
Closes #2016


ohdearquant:
Reviewed at head `60461011fc1d5a5efaa73e93ee58dc3abadf5f47`.

The attribution plumbing is genuinely well done, and I want to be specific about that before the
problem, because the problem is not in the part that was hard to build.

I enumerated the append surface independently rather than reading the diff: a whole-tree scan for
`append_event`/`append_events`/`append_events_idempotent` returns 129 lexical matches, which reduce to
27 production call sites once tests, trait definitions and forwarding layers are excluded. Twenty-one
go through `KhiveRuntime::events(token)` and therefore `AttributedEventStore`, which overwrites actor
and namespace on singleton, batch, preflight and idempotent paths with no fallback when the token is
absent. Three are registry audit calls composing from the resolved `GateRequest`. Three are
runtime-owned background writers, two of which use fixed daemon principals. The separate direct-SQL
family — atomic plan, brain fold-gate, brain persistence, and all three proposal projections — stamps
`EventAttribution::from_token` before insertion. No append family is missing from the inventory.

So the wrapper is mechanically correct. The issue is what it is wrapping.

## The gate reads a field the caller chooses, and this repository already says what that means

`CallerEnrollmentGate::actor_is_granted` decides on `self.granted_actors.contains(&req.actor.id)`
(`crates/khive-gate/src/enrollment.rs:40-45`). That `actor.id` arrives from the request frame:
`crates/khive-mcp/src/server.rs:3178-3184` serializes the client-supplied `namespace` and `actor_id`,
`crates/khive-runtime/src/daemon.rs:1170-1192` copies them into `RequestIdentity`, and
`crates/khive-runtime/src/pack.rs:1937-1945` copies them into the `GateRequest` the gate reads.

The doc comment on `peer_uid`, two crates away, already states the principle exactly:

> This is the only identity on this socket the caller cannot choose. Every identity field on the
> request frame — `namespace`, `actor_id`, `visible_namespaces`, `config_id` — is supplied by the
> connecting process, so none of them can answer "who is this". A check reading self-asserted fields
> is not a weak gate, it is not a gate: anyone who wants to pass it asserts the passing values.

That is `crates/khive-runtime/src/daemon.rs:308-315`, unchanged by this PR. The new gate is a check
reading self-asserted fields.

The consequence lands precisely on this change's stated purpose. A same-UID client sends
`actor_id` set to any id on the granted roster, plus any valid foreign namespace; the gate grants; and
`event_store_guard.rs:35-45` then faithfully stamps both self-selected values into the durable event.
The enforcement makes the record *more* confident without making it more true, which is worse than no
enforcement — a reader who knows appends are token-scoped will trust an attribution that was chosen
by whoever wrote it.

`grant_unattributed = true` has the same shape: it admits any same-UID caller that presents an
anonymous frame. That is a defensible posture, but it is a posture, not authentication, and the ADRs
should say which one they mean.

The fix has to bind enrollment to something off the frame — peer UID is already read on this socket
and is the obvious candidate — and derive namespace scope from that binding. If binding is out of
scope for this stage, then the honest alternative is to keep the policy but stop describing its values
as attribution, and say in ADR-018/129 that the roster is advisory until an authenticated principal
exists.

## Your macOS red is not a compile error — it is a stale lockfile in the excluded crate

Worth stating plainly because the check name points the wrong way. The workspace builds: the log ends
`Finished \`dev\` profile [unoptimized + debuginfo] target(s) in 1m 49s`. What fails immediately after
is:

```
error: cannot update the lock file .../crates/khive-merge/Cargo.lock because --locked was passed to prevent this
```

`crates/Cargo.toml:49` carries `exclude = ["khive-merge"]`, so that crate is outside the workspace and
keeps its own lockfile. It depends on `khive-runtime` by path
(`crates/khive-merge/Cargo.toml:21`), which reaches `khive-gate`. This PR changes
`crates/khive-gate/Cargo.toml` and regenerates `crates/Cargo.lock`, but
`crates/khive-merge/Cargo.lock` is not in the diff — so the excluded crate's lock no longer matches
its dependency graph and the `--locked` freshness step refuses it.

Regenerating that lockfile inside `crates/khive-merge/` and committing it should clear the macOS job
and the `Lockfile freshness` step on the Ubuntu shard, which was cancelled before it could report the
same thing. `CI gate` is the aggregate and carries no independent signal; both Ubuntu shards are
`cancelled` rather than `failure` — worth checking `.conclusion` from the check-runs API rather than
the summary column, which prints `fail` for both.

## ADR-162's inventory says closed where the code says conditional

`docs/adr/ADR-162-unified-event-plane-ownership.md:110-113` states that phase events use a fixed
daemon principal and that pending-schedule failures use verified immutable creator provenance. Both
are conditional in code.

`crates/khive-runtime/src/phase_events.rs:47-85` takes a `NamespaceToken` and stamps that token's
actor, and `crates/khive-pack-brain/src/handlers.rs:2270-2278` reaches it for `brain.mark_turn` — so
the principal is token-resolved, not fixed. `crates/khive-mcp/src/pending_events.rs:743-762` falls
back to `server.actor_id()` or `"local"` when creator provenance is absent, and appends through the
raw server event store at `:2665-2689`. That fallback is pre-existing; describing it as verified
provenance is not.

The claim at `:303-305` that every current request-caused append is token-stamped inherits the same
problem. This matters more than a normal doc nit because the base text was *more* accurate — it called
unvalidated append call sites "conformance debt" requiring implementation work, which is a correct
description of the current state. Replacing an accurate open item with an inaccurate closed one
removes the signal that would send the next reader to look.

Either narrow the status text to "verified when provenance exists, scheduler fallback for legacy
reminders" and describe phase events as token-resolved, or close the legacy path by refusing an
attributed failure append without verified provenance.

## Pre-existing, and not this change's to fix

An absent `[gate]` section still yields `AllowAllGate` (`crates/khive-runtime/src/config.rs:743-752`,
`.unwrap_or_else(|| base.gate.clone())`), which was already the base default. The present-section arms
are properly fail-closed, and I checked all four: a missing `granted_actors` defaults to an empty list
so named actors are denied, an empty list is the same explicit deny-all, an unknown key is rejected by
`#[serde(deny_unknown_fields)]`, and a malformed actor id fails `Namespace::parse` at startup. The
staged posture is disclosed in ADR-129, so this is a documented compatibility gap rather than
something introduced here.

Same for the legacy pending-reminder failure append: the raw call is identical at the base ref. It is
on the list only because this PR is what claims it is covered.

## Tests

The added tests redden real mutations — actor-set membership, the anonymous branch, fingerprint set
normalization, config-to-runtime installation, `deny_unknown_fields`, `Namespace::parse` validation,
and the decorator's three stamping paths. The projection forged-event assertions redden the direct SQL
stamp sites.

What is missing is every adversarial case, and for a change about attribution boundaries that is the
gap that matters: nothing sends a caller-supplied actor the caller does not own, nothing sends a
foreign namespace under an enrolled actor, nothing drives an unenrolled actor through the real
daemon/registry frame path, and nothing exercises a fingerprint mismatch end to end. Every current
test supplies a well-formed identity and checks that it is handled correctly, which cannot detect that
an ill-formed one is handled just as agreeably.

## Also verified

Config and protocol skew both fail closed at the warm daemon rather than degrading attribution:
`config_id` mismatch is refused before dispatch, `PROTOCOL_VERSION` is 4 and a mismatched request is
rejected before the handler, and the `metrics_only` path that precedes the config check is read-only
and cannot append. A pre-versioning old daemon can still execute a request before the new client
detects skew, which leaves that request's side effect ambiguous — but the client returns a hard
protocol error rather than falsely reporting a locally-attributed success.

`operations.rs`, `pack.rs` and `runtime.rs` remove no unrelated behavior. The `operations.rs` change
replaces an anonymous `for_namespace` token with the caller token while preserving record-namespace
routing, which is a correctness fix; `runtime.rs` moves store routing into `raw_events_for_namespace`
and wraps it. ADR-018, ADR-129 and ADR-143 all describe their shipped scope accurately, and no
amendment heading displaces a prior section's closing text.

ohdearquant:
## `macOS PR compile check` failure: a second, nested lockfile needs the same update

```
error: cannot update the lock file .../crates/khive-merge/Cargo.lock because --locked was passed to prevent this
```

(checkout prefix elided from the quoted error)

`crates/khive-merge` declares its own `[workspace]` table — deliberately, so the crate is
standalone-compilable without being a root workspace member — which means it carries an
independent `Cargo.lock`. The tree has exactly two tracked lockfiles and exactly two crates
declaring a `[workspace]` table:

```
crates/Cargo.lock              <- crates/Cargo.toml
crates/khive-merge/Cargo.lock  <- crates/khive-merge/Cargo.toml
```

This branch adds `sha2` to `crates/khive-gate/Cargo.toml` and updates `crates/Cargo.lock`, but
not the nested one. `khive-merge`'s lockfile still records the pre-change dependency set for
that crate:

```
name = "khive-gate"
version = "0.8.0"
dependencies = [
 "chrono",
 "khive-types",
 "serde",
 "serde_json",
 "thiserror",
]
```

Under `--locked` cargo refuses to reconcile that, so the step fails before anything is compiled.
Regenerating `crates/khive-merge/Cargo.lock` and committing it should clear the failure — the
change itself looks unrelated to the error text.

Worth noting the general shape, since it is not specific to this branch: any manifest change to
a crate the nested lockfile records will fail the same way, and nothing surfaces it while
authoring. The root lockfile updates as a side effect of an ordinary local build; the nested one
only updates if cargo is invoked inside `crates/khive-merge`, so the omission is invisible until
CI runs with `--locked`.

ohdearquant (commented):

oceanwaves630 (approved):
Approving on the reviewer report at this head. CI is green (25 passing, 0 failing, 0 pending). The API source-compatibility note is non-blocking and tracked as a follow-up.

PR #2340, opened 2026-08-30
fix(ann): release rotated mmap generations

### LINKED ISSUE #2081: kkernel: long-lived processes pin rotated ANN index generations (deleted-but-open files accumulate)
Long-lived `kkernel mcp` processes (multi-day stdio servers) keep open file descriptors on ANN index generation files (`khive.db.ann/vectors.bin` and sibling files) that were unlinked when the index was rebuilt. The process never re-opens the index after rotation, so each rebuild strands the prior generation on disk until the process exits.

Measured on one host: 36 deleted-but-open files totaling 5.3 GB, held across 7 long-lived processes, roughly 0.5-0.75 GB stranded per holder per rotation. Process lifetimes of several days make the accumulation effectively unbounded, and the space is invisible to directory-based disk accounting (only `lsof` against unlinked files sees it).

Ask, either of:
1. Release-on-rotation: when an ANN rebuild publishes a new generation, existing handles drop and re-open (or re-mmap) the new files instead of holding the unlinked generation indefinitely.
2. A documented lifecycle policy for long-lived processes across index rotations, if release-on-rotation is not tractable for mmap'd readers.

If a rotation/epoch signal already exists on the read path, the defect reduces to readers not acting on it by re-opening.

ohdearquant:
## Summary

- give each file-backed KHVVAMG2 checkpoint publication a unique commit identity while preserving legacy checkpoint reads and deterministic portable serialization
- run shutdown-aware memory and knowledge ANN rotation watchers that validate changed publications, reload under the existing lock order, atomically swap bridges, and release predecessor mmaps without waiting for another query
- evict invalid changed publications so normal warm/rebuild recovery can proceed instead of retaining deleted-but-open generations, and document the lifecycle contract

## Compatibility

Existing base records and 41-byte commit trailers remain readable. The new 16-byte publication nonce is additive for file-backed atomic saves; `to_bytes` keeps the deterministic legacy form. Watchers start only for writable warmed packs, poll metadata every five seconds, and stop with daemon shutdown or pack teardown.

## Tests

- RED then GREEN: `identical_checkpoint_rotations_have_distinct_commit_identities`
- RED then GREEN: memory and knowledge `rotation_tick_releases_predecessor_and_adopts_identical_peer_checkpoint`
- `cargo test -p khive-vamana -p khive-pack-memory -p khive-pack-knowledge`
- `cargo clippy -p khive-vamana -p khive-pack-memory -p khive-pack-knowledge --all-targets -- -D warnings`
- `cargo check --workspace`
- `cargo test --workspace`

Closes #2081


ohdearquant:
Reviewed at head `1b05bb4062d5603fe4290d2679a91b32f71a18f5`.

Leading with the safety question, because it is the one that would have blocked and it comes back
clean: **no reader can hold a reference into a generation that gets released.** The guard spans the
whole read, not just pointer acquisition — `search_loaded_with_seq` takes `ann.indexes.read().await`
and calls `bridge.search(query, k)` inside that guard's scope, replacement takes the write lock
before inserting, and `VamanaIndex::search` returns an owned `Vec` so no mmap-derived slice escapes.
Each map value is an owned `AnnBridge` owning its `memmap2::Mmap`, so the predecessor drops under the
write guard. Lock order agrees across the watcher, the rebuild path and the query path, and
`clear_namespace`'s `indexes`-then-`warm_states` order is not inverted anywhere the watcher runs.

That is the hazard a rotation-and-release change most plausibly introduces, and it is not present.

## Your red CI is not yours — please don't go hunting it

`CI (ubuntu-latest, shard 2/2)` fails on
`handlers::recall::tests::keyword_only_widening_ignores_vector_leg_candidates`, panicking at
`crates/khive-pack-memory/src/handlers/recall.rs:6987` with "keyword_only widening must not stop on
vector-leg-only candidates". That is a known-unstable fixture that has been reddening unrelated
shards.

This change cannot reach it. `git diff origin/main...HEAD --name-only` touches three files in
`khive-pack-memory` and **zero** in `handlers/recall.rs`. Worth stating the reasoning explicitly,
because the obvious check misleads here: the same job passes at the merge base, and base-green is
routinely read as proof that a red head is the branch's fault. For an unstable test that inference
does not hold — one base run is a single sample, and passing it is exactly what an intermittent
failure does most of the time. The file-level reachability check is the one that settles it.

The `CI gate` failure is the aggregate of that shard and carries no independent signal, and the two
cancelled shards are cancellations, not assertions.

## The nonce is not backward-incompatible; it is forward-incompatible, and the ADR should say so

The finding is real but the direction is worth getting right, because the two words license
different decisions.

- **Backward compatibility holds.** The new parser accepts all three lengths — `base_len`,
  `base_len + 41`, and the new `+ 16` publication trailer
  (`crates/khive-vamana/src/index.rs:2979-2988`). New readers read old records.
- **Forward compatibility breaks.** At the base ref, `parse_v2_commit` is an exact-length test
  accepting only `base_len` or `base_len + 41`. A 249-byte record emitted by the new writer is
  rejected by any older binary with `invalid_format`.

So `docs/adr/ADR-079-...:738`, "appends a random publication nonce to the backward-compatible
extended commit trailer", is not false — read strictly, "backward-compatible" modifies the 41-byte
trailer being extended, and that trailer genuinely is. The problem is what the sentence leaves out
next to a word that invites the opposite conclusion: **nothing in the ADR states that appending the
nonce makes the record unreadable to the previous release.** A reader deciding whether a rollback is
safe gets no answer, and the adjacent adjective suggests one.

The practical consequence is bounded but not free: an older binary reaching `load_or_build` recovers
by rebuilding, so this degrades to an expensive silent rebuild plus a rewrite back into the old
layout, rather than a hard failure. That is a real operational cost on a rollback and it deserves a
sentence rather than being left to be discovered. Either add the generation identity in a separately
versioned sidecar older readers ignore, or keep the nonce and document plainly that rollback forces
an ANN rebuild. Also worth a test that feeds a 249-byte publication to the previous parser and pins
whichever behaviour you choose.

## Invalidation clears a warm slot it does not own

In the watcher's error branch, removing the stale index is properly conditional — it only removes
when the installed bridge still carries `incumbent_digest`. But when it does remove, it then clears
the warm slot unconditionally:

```rust
if removed {
    warm_states_guard(&ann.warm_states).remove(key);
}
```

Nothing there checks whether a warm attempt currently owns that slot. `begin_warm` installs a
`Warming { attempt_id, generation }` owner and `run_warm_attempt` holds it across
`ensure_ann_for_model`, which on the ordinary full-scan path runs before the checkpoint lock — so the
attempt is not holding the per-key lock for its duration.

The reachable interleaving: attempt A claims the slot and starts its corpus scan; the watcher sees a
changed-but-invalid publication and clears the slot; a request sees no owner and starts attempt B.
When A finishes, `finish_warm_state` finds a different or missing attempt and deliberately no-ops
(`crates/khive-pack-knowledge/src/knowledge/vamana.rs:383-396`). Single-flight is defeated and one
key gets two full rebuilds. Publication locks and the generation check keep this to duplicate work
rather than an unsafe map, which is why this is a cost issue and not a correctness one — but "one
warm at a time" is a documented property and this path breaks it.

Invalidate through an owner-aware transition instead, or leave the permit in place and let its own
generation check make its completion stale.

## The tests exercise the liveness fix and cannot see the safety property

Each added test reddens a real mutation — `identical_checkpoint_rotations_have_distinct_commit_identities`
catches a writer that fails to make two identical saves distinct, and both
`rotation_tick_releases_predecessor_and_adopts_identical_peer_checkpoint` tests catch a refresh that
fails to detect the digest change, install the new map, or drop the predecessor.

All three rotate while quiescent. Nothing holds `indexes.read()` across a blocked `bridge.search`
while the watcher attempts its swap. So the suite covers the behaviour you added and is structurally
unable to detect the behaviour you must not break — and the lock discipline holding today is a
property of code someone will refactor. One test that parks a query inside the search and then forces
a rotation would put a guard behind the argument this change rests on.

## Smaller

An invalid publication is not distinguished from a merely *newer* one: a future-format record fails
validation and is evicted as corrupt. Given rollback is already the weak direction above, these are
the same question and one decision can answer both.

Recovery after eviction is request-triggered, so the gap between eviction and the next rebuild is
unbounded in wall-clock time if nothing asks. The five-second cadence bounds predecessor mapping
lifetime, not availability — the docs are precise about this and the ADR's phrasing could be too.

## Also verified

Watcher lifecycle matches its description: writable-warm start only, five-second `interval_at`,
shutdown observed through the cancellation token with the task tracked through return, panic or
cancellation, and a weak pack reference checked between ticks. A teardown landing mid-tick lets that
tick finish against the state it already upgraded and exits at the next one. Ordinary per-tick errors
retain the current bridge rather than evicting it. `to_bytes` still passes `None`, so the
deterministic legacy serialization is unchanged. The ADR amendment heading structure is intact —
the new rotation text sits before the prior closing section rather than displacing it.

ohdearquant:
The single real failure here is already fixed on `main`; this branch predates the fix.

`CI (ubuntu-latest, shard 2/2)` fails on
`handlers::recall::tests::keyword_only_widening_ignores_vector_leg_candidates`. That exact test
was stabilized by #2305, `test(memory): stabilize keyword widening fixture`, which landed on
`main` earlier today. This branch is 7 commits behind and does not contain it.

Merging `main` (or `gh pr update-branch`) should clear it. The other red rows are sibling matrix
shards cancelled by fail-fast, not separate problems.

Worth confirming rather than assuming, since I have not run it here: the reasoning is that the
failing test name matches the fix's subject and the fix is absent from this branch and present on
`main`. If it still fails after the merge, that is a genuinely different problem and worth saying
so rather than retrying.

ohdearquant:
## The red `CI gate` is one test, in a crate this branch changes

```
FAIL khive-pack-memory handlers::recall::tests::keyword_only_widening_ignores_vector_leg_candidates
Summary: 2534/4675 tests run: 2533 passed, 1 failed
```

The failing test lives in `khive-pack-memory` and this branch edits three files there, so this
is not the ambient pattern seen on other branches today. The assertion is that a keyword-only
widening ignores vector-leg candidates — a vector-leg boundary — which is adjacent to
releasing rotated mmap generations. Worth checking whether a released generation is leaving a
candidate visible to the keyword path.

ohdearquant (commented):

Verdict on head 1b05bb4062d5603fe4290d2679a91b32f71a18f5: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving on the owning seat's verification at this head under the standing lane. Static reviewer APPROVE at the exact head, CI green at that head (22 checks success, 6 skipped, none pending or failed, re-read at click time), and no daemon verdict at that head. The one daemon artifact on this PR sits at an older head and does not govern this one. Seat gates cited and accepted: fmt clean on a comment-only delta.

PR #2342, opened 2026-08-30
fix(mcp): default inbound email actor to "local"

ohdearquant:
## What

`default_inbound_actor_from_env()` resolves the actor that fresh, uncorrelated inbound email is
attributed to at server startup. When `KHIVE_EMAIL_DEFAULT_ACTOR` was unset or blank it fell back
to a hard-coded identity string rather than a neutral one. This changes that fallback to `"local"`.

Behaviour when the variable **is** set is unchanged.

## Why

The sibling resolver immediately above it in the same file, `ingest_namespace_from_env()`, already
falls back to `"local"`. Two adjacent startup resolvers, same shape, both reading an optional env
var, disagreeing about what "unset" means — and the one naming a specific identity is the odd one
out. `"local"` is the value the rest of the codebase already uses for "no attribution was
specified" (ADR-007: namespace is attribution, not isolation).

The practical effect of the old default: a deployment built with `--features channel-email` that
does not set `KHIVE_EMAIL_DEFAULT_ACTOR` attributes inbound mail to an identity that does not
exist in that deployment, and nothing surfaces the mismatch — the messages simply land under a
name their operator never chose.

This is a behaviour change for anyone depending on the old fallback, but depending on it would
mean depending on an identity string that was never theirs.

## Scope

- `crates/khive-mcp/src/serve.rs` — the fallback, its doc comment, and three regression tests
  (unset / set / blank-string).
- `.env.example`, `docs/khive-config-example.toml` — these *document* this default, so they move
  with it. The sample identities in the config example also become obvious placeholders, so they
  read as values to replace rather than values to copy.

## Verification

```
cargo test -p khive-mcp --features channel-email default_inbound_actor
test result: ok. 3 passed; 0 failed
```

The tests are load-bearing rather than decorative, checked by reverting the one-line fallback and
re-running: `default_inbound_actor_defaults_to_local` and `default_inbound_actor_ignores_blank_env_var`
fail on the assertion (`left: "<old default>", right: "local"`) while `default_inbound_actor_reads_env_var`
stays green, since an explicitly set variable never reaches the fallback.

- `crates/khive-channel-email/src/lib.rs`, `docs/guide/communication.md` — two further sites
  that stated the old value as the documented default. Found by grepping the variable across
  the tree at this ref rather than by code search, which under-reported.

## Trade-off: what `local` means

`local` is not an inert bucket. Actor resolution falls back to `local` for any caller that is
not running with an explicit identity, so `local` is also the pool that identity-less readers
see by default. This change therefore moves fresh, uncorrelated inbound mail — text the
operator did not author — out of one named identity's mailbox and into that shared pool.

That is a trade rather than a free win, and it is worth stating plainly:

- the old behaviour attributed external mail to a specific identity, which is wrong in any
  deployment that has no such identity and misleading in one that does;
- the new behaviour attributes it to nothing in particular, which is honest, but puts it where
  an unconfigured reader will see it.

`local` is chosen here because it matches `ingest_namespace_from_env`, the sibling resolver
immediately above it, and because inventing a new actor convention is out of scope for a change
whose whole point is that two adjacent resolvers disagreed about the same question. A dedicated
non-identity actor such as `channel:email` would give operator honesty *without* routing external
mail into the default-read pool. That is the better end state and it needs its own change, since
it introduces a convention rather than correcting a default.

Deployments that want inbound mail kept out of the default pool should set
`KHIVE_EMAIL_DEFAULT_ACTOR` explicitly. Behaviour with the variable set is unchanged.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,8 +42,8 @@ KHIVE_EMAIL_PASSWORD=your-app-password-here
 # KHIVE_EMAIL_INGEST_NAMESPACE=local
 
 # Actor that receives fresh inbound email (no In-Reply-To correlation match).
-# Default: lambda:leo
-# KHIVE_EMAIL_DEFAULT_ACTOR=lambda:leo
+# Default: local
+# KHIVE_EMAIL_DEFAULT_ACTOR=local
 
 # Comma-separated allowlist of recipient addresses the outbox loop may deliver
 # to. Defaults to KHIVE_EMAIL_MAINTAINER_ADDRESS when unset or empty.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `KHIVE_EMAIL_DEFAULT_ACTOR` now falls back to `local` when unset or blank,
+  matching `KHIVE_EMAIL_INGEST_NAMESPACE` and the adjacent startup resolver,
+  instead of a hard-coded identity with no meaning outside the deployment it
+  was named for. Behaviour with the variable explicitly set is unchanged.
+
 ## [0.8.0] - 2026-08-27
 
 ### Added
--- a/crates/khive-channel-email/src/lib.rs
+++ b/crates/khive-channel-email/src/lib.rs
@@ -30,7 +30,7 @@
 //!
 //! | Variable | Default | Description |
 //! |---|---|---|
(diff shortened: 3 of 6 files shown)


oceanwaves630 (approved):
Doc-default correction verified across all four sites; changelog entry present.

ohdearquant:
Follow-up filed as #2343 for the channel-actor question: uncorrelated inbound email lands
under a general-purpose identity rather than one scoped to the email channel, so
externally-originated content is not separable from locally-written content at read time.
That is pre-existing and independent of this change, which keeps this PR a one-line default
correction.

PR #2345, opened 2026-08-31
chore(ci): check that every ADR declares one status in its header

ohdearquant:
## What this adds

A lint that checks every `docs/adr/ADR-*.md` declares exactly one status, in its
document header, with a value from a known lifecycle vocabulary. It runs in the
existing `Format, SQL, and ADR lints` step alongside the reference lint, and it
ships with a self-test.

## Why

Reading an ADR's status looks mechanical and is not. The word "Status" appears in
two unrelated positions: the document header, stating the status of the record,
and inside a later section, where an amendment states its own. A parser that takes
the first status-shaped line in the file gets the second one whenever the header
uses a form it does not recognise.

That case is in the tree today. `ADR-051` declares its header status under a
`## Status` heading with the value on the following line:

```
# ADR-051: Section-level embeddings and hybrid compose scoring

## Status

Accepted (2026-06-07). **Fully implemented** (2026-06-08) — Phases 1 and 2 shipped.
```

and carries this two hundred lines below:

```
## Amendment 1: Blend KG entities into the compose candidate pool

Status: proposed.
```

A colon-anchored search skips the heading, finds the amendment, and reports an
accepted-and-implemented decision as a proposal. Nothing in the output suggests it
read the wrong line, which is what makes it worth a guard rather than care.

## What the rule is, and what it deliberately is not

The rule is about **scope**. Inside the document header — everything above the
first section heading that is not itself a Status heading — there must be exactly
one status declaration, and its value must begin with a known lifecycle word.
Below the header nothing is checked; an amendment stating its own status is
legitimate.

The rule is **not** about form. Three declaration forms are in live use and all
three are accepted:

```
**Status**: Accepted            bold, with or without a trailing hard-break \
- Status: Accepted              one bullet in a metadata block
- **Status:** Accepted          same, bold label
## Status                       heading, value on the next non-empty line
Accepted
```

Normalising these onto one spelling was the obvious alternative and it is the
wrong change. The bold form's trailing backslashes are markdown hard breaks
holding Status/Date/Authors together as a single rendered paragraph, so stripping
them silently joins the lines. The list form's status is one bullet among
`- Date:` and `- Supersedes:`, so rewriting only the status line strands a bold
non-bullet inside a list. Both are rendering changes across dozens of published
documents, and neither improves the parse, because form was never what made the
parse go wrong.

## Verification

- `python3 scripts/lint-adr-status.py` → `checked 174 records, 0 failing`
- `python3 scripts/lint-adr-status.py --self-test` → `27/27 cases correct`
- `sh scripts/ci.sh lint` passes end to end with both new steps wired in
- Mutation check on the real corpus: replacing `ADR-051`'s `## Status` heading
  with `## Overview` makes the lint report `no status declaration in the document
  header` and exit 1; restoring the file returns it to 0. The check is load-bearing
  rather than merely green.

The self-test covers the three failing shapes (a header with no status while a
later section has one, two header declarations, an unrecognised value), the three
accepted forms including `ADR-051`'s exact heading-plus-amendment layout, and an
empty corpus — which must be reported as a broken invocation rather than a clean
result, so pointing the checker at the wrong directory cannot read as a pass.


ohdearquant (commented):

Verdict on head 7e402ec0bfbd67e9283a6259ff47db82e2917dc9: REQUEST-CHANGES, 4 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 17ae345cee1820ba6654a8fcb2459ea61e4a80fa: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head f8027a92a390b8198c2333ad13512780e44a1997: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):

PR #2352, opened 2026-09-02
docs(kg-editor): describe the direct static load for sessions without an operator token

ohdearquant:
Documentation follow-up to #2298.

Since #2298 the showcase browser loads the approved curated asset directly when the session holds no operator token and the entry owns one, instead of first requesting the protected snapshot route that is guaranteed to answer 404 for unauthenticated callers. With a non-blank token the route-first order and the 404-only fallback of ADR-147 Amendment 3 are unchanged, and entries that exist only in the configured catalog still probe the route.

- `docs/adr/ADR-147-repo-showcase-bundle.md`: Amendment 5 records that order and restates the fallback boundary for authenticated requests.
- `apps/kg-editor/README.md`, `apps/kg-editor/DEMO.md`: the operator-token paragraphs no longer describe a snapshot request that is not made without a token.

No code changes.

### DIFF
--- a/apps/kg-editor/DEMO.md
+++ b/apps/kg-editor/DEMO.md
@@ -20,9 +20,10 @@ npm run dev
 ```
 
 Place the report at `$KHIVE_SHOWCASE_ANALYSIS_ROOT/khive/khive.repo.v1.json`.
-The protected route requires the bearer token on every request and fails closed
-to 404 without it, which silently selects the static fallback. Unlock the UI in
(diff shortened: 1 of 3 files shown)


oceanwaves630 (dismissed):
Approving on the reviewer report at this head, with CI green (20 passing, 7 skipped, 0 failing, 0 pending).

oceanwaves630 (approved):

PR #2353, opened 2026-09-02
docs(git): record the typed remote fetch failure contract for git.digest

ohdearquant:
Documentation follow-up to #2290.

Since #2290, `git.digest` returns `RemoteFetchError { remote, message }` when a remote source's clone or fetch setup fails, with credentials and query strings redacted from `remote`; a source that parses as neither a local path nor a remote URL stays `InvalidInput`, and storage failures keep their existing types. Neither ADR-088 Amendment 1 nor the API reference stated that contract.

- `docs/adr/ADR-088-amendment-1-git-digest.md`: point 5 under Remote-URL mode records the failure contract and the redaction guarantee.
- `docs/guide/api-reference.md`: the git pack overview names the typed failure.

No code changes.


oceanwaves630 (approved):

PR #2369, opened 2026-09-03
perf(db): route FTS subject lookups and deletes through a rowid map

ohdearquant:
## Problem

`fts_entities` and `fts_notes` declare `namespace` and `subject_id` as UNINDEXED FTS5 columns, so every delete and every subject lookup against them is a full virtual-table scan. On a 214,796-row table that costs about 0.7 s per write, and it grows with the table.

## Change

- A sidecar map per FTS table, `fts_{table}_rowids (namespace, subject_id, rowid)` WITHOUT ROWID, mirrors the FTS5 hidden rowid. Subject lookups and deletes go through the map's primary key; MATCH queries are unchanged.
- Every write path maintains the FTS row and its map row in one transaction: the writer task and plan paths already were, and the unmanaged file-backed fallback now runs its delete pair under `BEGIN IMMEDIATE`.
- Every map-routed read and delete re-checks `(namespace, subject_id)` on the FTS row, so a map entry left stale by a crash between the two writes can never return or delete another subject's document once FTS5 reuses the rowid.
- Migration 024 creates and backfills both maps. The backfill skips NULL-key rows (they stay reachable by MATCH and are never deleted), chooses the survivor among duplicate keys by `updated_at` with the higher rowid as tie-break, and sweeps orphans with a NULL-safe `NOT EXISTS` on `(id, namespace)`. Soft-deleted subjects are preserved.
- Backfill completion is recorded by a durable marker written in the same transaction as the backfill, so a partially populated map is reconciled on open and a completed one costs one O(1) probe.
- Before the marker is written, the map is reconciled key-aware inside one transaction: map rows whose FTS row is gone or carries a different `(namespace, subject_id)` are removed, every keyed FTS row is backfilled in survivor order, and keyed rows that lost the survivor race are swept. Migration 024's map cleanup checks key equality the same way. The invariant the marker certifies is that every live keyed FTS row has exactly one map row at its rowid and every map row points at a live FTS row with the same key; a parity test drives one malformed legacy state through both the migration and the runtime path and asserts identical results.
- Read-only databases use the map only when the marker is present; an unmarked or absent map opens in scan-fallback mode with one warning. Table keys ending in `_rowids` or `_rowids_state` are rejected so a sidecar name can never be requested as an FTS table.
- The test-only delete failure seam is armed per `(namespace, subject_id)` target, so seam tests run under the parallel test runner without interfering.

## Measurements

100 deletes by subject on a 200,000-row table: 27.5 s through the scan, 0.26 s through the map (ignored scale test, `scale_measurement_delete_by_subject`).

## Tests

EXPLAIN QUERY PLAN pin of the primary-key plan; `last_insert_rowid()` pairing; migration 024 backfill, dedup, orphan sweep, NULL-key survival, `updated_at` survivor and equal-timestamp tie-break, marker presence; stale-map protection for get, delete and batch upsert; rollback of the unmanaged delete pair when the map delete fails; partial-map reconciliation and marker short-circuit on open; writable legacy backfill NULL exclusion and survivor order; wrong-key map row removed before backfill; duplicate loser swept on open; migration wrong-key cleanup before the marker; migration/runtime parity on one malformed state; read-only scan fallback with an unmarked map; `_rowids` and `_rowids_state` key rejection; per-target seam isolation.


oceanwaves630 (approved):
Approved at 6a235f35277e5e48f228b0d8388bffb40d39cd93.

Checks at this commit: 29 check runs, all completed, none failed and none pending. No changes-requested reviews outstanding. Branch is mergeable against main.

Approval only; the merge is the author's.

PR #2370, opened 2026-09-03
perf(knowledge): drive candidate hydration off the primary key

ohdearquant:
## Summary

`hydrate_empty_hits` hydrates ANN hit shells from `knowledge_atoms` and `knowledge_domains` with one `id IN (...)` statement per chunk of up to 900 ids. On a store with no `sqlite_stat1`, SQLite's default estimate for an indexed equality is about ten rows, which beats a 250-probe primary-key lookup, so the planner drove the statement off `idx_knowledge_atoms_ns_created` and scanned the whole namespace once per chunk. On a large namespace every chunk then ran past the request read deadline and hydration reported `requested=N failed=N` on every request.

The two hydration statements now put the `IN` list first and write the namespace term as `+namespace = ?1`. The unary `+` keeps that column out of index selection, so the primary-key auto-index is chosen regardless of statistics. Projection, namespace scoping and the live-row predicate are unchanged.

`ANALYZE` was not used: it would make the plan depend on the store's statistics state rather than on the statement.

## Tests

- `hydrate_atoms_statement_plan_uses_primary_key_not_namespace_index` and its domains twin pin the plan through `EXPLAIN QUERY PLAN` on a statistics-free in-memory store with 250 ids: the primary-key auto-index must appear and no namespace index may.
- `hydrate_atoms_statement_still_scopes_by_namespace` seeds two namespaces, hydrates ids drawn from both against one namespace, and confirms the other namespace's row is never returned and is counted as unhydrated.

Two other statements in the same module keep the `namespace = ?1 AND ... IN (...)` shape (`load_domain_member_token_sizes`, `load_atom_body_line_counts`); they join through other tables and are left for a separate change.

### DIFF
--- a/crates/khive-pack-knowledge/src/knowledge/search.rs
+++ b/crates/khive-pack-knowledge/src/knowledge/search.rs
@@ -953,6 +953,61 @@ fn cosine_similarity(a: &[f32], b: &[f32]) -> f32 {
 // portable 999-variable ceiling.
 const HYDRATION_ID_CHUNK: usize = 900;
 
+/// Build the atom hydration statement for one id chunk.
+///
+/// The `IN (...)` list runs off `knowledge_atoms`' primary key, so the
+/// namespace predicate is deliberately kept out of index selection with
+/// SQLite's unary `+` (`+namespace`): a scratch or freshly vacuumed store has
+/// no `sqlite_stat1`, and without statistics the planner's default guess for
+/// an indexed namespace equality (~10 rows) beats a 250+ id primary-key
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Primary-key-first hydration reads equivalently: the unary + only removes namespace from index selection, so the predicate set is unchanged, and the added test that seeds two namespaces confirms cross-namespace rows still do not hydrate. Plan pins carry both a must-match and a must-not-match arm. 27 checks green at this head.

PR #2382, opened 2026-09-03
fix(knowledge): bound lexical term admissions per request

ohdearquant:
Long or decomposed knowledge searches can issue unbounded lexical work as query terms expand. A shared 32-term admission allowance now covers the full query and both subqueries, before rarity probes begin. Later retrieval stages reuse the admitted terms, and search responses expose `candidate_provenance.terms_truncated`.

The change preserves staged retrieval, eligibility filters, and timeout diagnostics. Added tests cover over-limit and under-limit searches, a shared decomposed-query allowance, and fallback probes restricted to admitted terms.

Validation: formatting and source consistency checks pass. Native tests have not run locally; hosted CI is required.


ohdearquant:
Measured against the current served build, before this lands as the fix for the reported
`degraded.lexical_timeout` behaviour.

**Term count does not drive the flag on the search path.** A nested-prefix ladder, holding the
topic constant so that every longer query contains every shorter one and only the term count
varies, produced no `degraded` key on any `knowledge.search` call: 18 runs from one to twelve
terms, across two interleaved passes. Carrier and result limit were excluded in the same way,
with the identical queries run over both the MCP and CLI paths at limits 2 and 3. Total request
wall time does not separate the two classes either; the slowest clean run was slower than most
flagged runs, which is what a stage-scoped allowance rather than a request-scoped one looks like
from outside.

**The cap would not bind the reported workload.** The queries in the incident expand to about 20
terms, below the 32-term cap this change introduces, and they are not decomposed, so their
lexical work would be unchanged by it.

**What does vary is how much the stage reads.** A single search returning three results was
measured making the serving process pull between 16 KB and 337 MB off disk, against a database of
roughly 9 GB. That four-order-of-magnitude spread, not the term count, is what tracks the flag.

None of this argues against the change on its own terms: bounding long-query fan-out is a real
and separate concern, and the per-candidate versus per-request distinction in the title is
correct. The narrow point is that merging it should not be recorded as resolving the reported
timeout, because two independent methods, a live measurement ladder and a source read of the
stage's own budget, agree that it would not.

ohdearquant:
Correction to my previous comment, on one sentence in it.

I wrote that a single search returning three results was measured making the serving process pull
between 16 KB and 337 MB off disk. That sentence attributes to one request a counter that belongs
to a whole process. The measurement differenced a per-process disk-read counter across an interval
enclosing each request, on a daemon that serves every client on the machine. The defensible
statement is narrower: the daemon's disk-read counter rose by up to that amount during an interval
enclosing a request. Attribution to the request, to any particular database file, or to the lexical
stage specifically does not follow from it, and I stated the shared-process caveat in my own notes
and then wrote the attributed version anyway.

The rest of that comment stands. The term ladder, the exclusion of carrier and result limit, and
the observation that the queries in question expand to about 20 terms and so would not be bound by
a 32-term cap are all direct readings of the response envelope rather than inferences from a shared
counter.

The narrow conclusion is unchanged and does not depend on the retracted sentence: this change
should not be recorded as resolving the reported timeout.

PR #2388, opened 2026-09-03
test(knowledge): wait for the rotation watcher by paused time, not a yield budget

ohdearquant:
The drop test in the knowledge pack advanced the paused clock and then gave the rotation watcher at most 100 yields to exit. The watcher builds its interval on its first poll, which can land after that advance, so the tick is never reached inside the budget and the test fails on a loaded runner; it failed on two unrelated changes in CI today.

The wait now runs under paused time with a 60-second virtual deadline. Under a paused runtime, sleep auto-advances the clock once every task is idle, so the tick is reached by virtual time rather than by a fixed yield budget. Twenty consecutive local runs pass before and after the change, so the failure is runner-load dependent and the change is justified by the mechanism rather than by a local reproduction.

### Second flaky test: `writer_stage_sample_attributes_a_slow_body`

The test slept 60 ms in the writer body and asserted the body dominated the stage sample; a commit fsync on a slow runner takes close to 200 ms and outweighed it. The body now sleeps 400 ms with a 350 ms floor, so the slow arm is the one measured on every runner. Twenty consecutive local runs pass; the whole `writer_task` suite passes.

### DIFF
--- a/crates/khive-db/src/writer_task.rs
+++ b/crates/khive-db/src/writer_task.rs
@@ -2851,9 +2851,12 @@ mod tests {
         }
         let handle = spawn(&pool, 8).unwrap();
 
+        // The delay must dwarf a COMMIT on a slow disk (an fsync of a few
+        // hundred milliseconds has been observed on shared runners), or the
+        // commit-stage comparison below reads the disk instead of the body.
         handle
             .send(|conn| {
-                std::thread::sleep(Duration::from_millis(60));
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Stamping to unblock the merge queue, not a substantive review.

This is a test-only change: 2 files, +16/-7, both hunks inside `mod tests`. No production code is touched, and every check is green at f693af9b.

PR #2391, opened 2026-09-04
fix(scripts): let the phase-order guard own its producer failure under set -e

ohdearquant:
Follow-up to #2283.

`check_ci_phase_order` in `scripts/lint-stub-markers.sh` read the producer status on the line after the command substitution. With `set -e` in force (the script enables it at the top), a direct call exited on the failed assignment before that status was read or reported; the existing callers only survived because they invoke the guard inside an `if`, where the shell suspends `set -e`.

- Capture the status on the assignment itself (`|| phase_rc=$?`) so the helper reports the failure and returns nonzero in every calling context.
- Add a self-test arm that calls the guard directly in a subshell with `set -e` enabled and asserts both the report line and the nonzero return. The subshell is kept out of any `if`, `&&` or `||` list on purpose: the shell suspends `set -e` inside those and the suspension reaches into a subshell that re-enables it, so the arm would pass vacuously against the old helper.

Verification (macOS `sh`, bash 3.2 in POSIX mode): the new arm fails against the previous helper with `exited before reporting the producer's status (got status 3)` and `sh scripts/lint-stub-markers.sh --self-test` passes with the fix.

### DIFF
--- a/scripts/lint-stub-markers.sh
+++ b/scripts/lint-stub-markers.sh
@@ -61,11 +61,13 @@ ROOT="$SCRIPT_DIR/.."
 # stale first phase the way a text scan of the source could. The producer's
 # exit status is read before its output: a list printed by a script that then
 # failed is not an answer, and a pipeline ending in head would discard that
-# status.
+# status. The helper owns that failure: the producer's status is captured on
+# the assignment itself, so a direct call with `set -e` in force still reaches
+# the report below instead of exiting the caller on the failed substitution.
 check_ci_phase_order() {
     phase_ci_sh="$1"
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Approved at cf644c78a4d5809d32f55a06ef9403e9c06fd5d5.

Verified at this exact head: 27 checks reported, 20 pass and 7 skipped, none failing or pending; base `main`; mergeable and up to date; no outstanding change requests from any account.

Stamping rather than reviewing — the technical judgment on the change is the author's, and this approval records that the merge preconditions hold at this head.

PR #2392, opened 2026-09-04
perf(db): index the hot note property paths and page the inbox by key

### LINKED ISSUE #2390: note page queries cost seconds of CPU per inbox or task listing on a large store
On a store with about 95,000 notes, a single `comm.inbox` or `gtd.tasks` call costs roughly 2.4 s of daemon CPU. Measured on the running daemon (0.8.0, revision 21e2a916) by reading its CPU time around a burst of three `comm.inbox` and two `gtd.tasks` calls with the audit plane confirming no other dispatches in the window: 12.2 s CPU over 14 s. A 10 s stack sample of the daemon attributes the active time to `query_note_page_snapshot` stepping rows in `khive-db/src/stores/note.rs` (the `query_notes_filtered` / `query_notes_filtered_count_free` path).

Mechanism, from the code: property filters are compiled to `json_extract(properties, '$.…')` predicates (`build_note_filter_where`), so every candidate row under the kind filter is parsed; the only property-aware index is the unread-probe recipient index, which the general page query does not use. The inbox post-filter path pages with `LIMIT 200 OFFSET n` and rescans from the start on every page, so a filtered read over a large namespace is quadratic in pages. Over a day of fleet traffic (about 1,500 dispatches per hour, of which inbox and task listings are about 170) this keeps the daemon at 50-60 percent of one core continuously.

Proposal: (1) generated columns or expression indexes for the hot property paths the comm and gtd packs filter on (recipient actor, read flag, assignee, status) so the page query is index-driven; (2) keyset pagination (created_at, id) for the inbox post-filter loop instead of offset rescans; (3) a per-verb CPU cost sample in the audit payload so the next regression of this kind is visible from `brain.event_counts` instead of from a stack sample.

ohdearquant:
Fixes #2390.

Two cost sources sat behind the seconds of CPU per `gtd.tasks` / `comm.inbox` call measured in the issue: the GTD task listing pushes its status/assignee predicates into SQL but had no index on those expression paths, so every listing narrowed to `idx_notes_kind` and then JSON-parsed every `task` row in the namespace; and the inbox post-filter loop re-issued its page query with a growing `OFFSET`, re-walking every earlier row on every page.

## Changes

- **V27 migration** (`crates/khive-db/sql/027-notes-hot-property-indexes.sql`, registered in `migrations.rs`; the same `CREATE INDEX IF NOT EXISTS` statements are added to the V1 baseline `schema.sql` and to `notes-ddl.sql` so fresh and already-migrated databases converge): two partial expression indexes, `idx_notes_task_status` and `idx_notes_task_assignee`, on `notes(namespace, kind, json_extract(properties, '$.<path>'), created_at DESC, id ASC) WHERE deleted_at IS NULL`. The expressions are copied from the SQL `build_note_filter_where` compiles, since an expression index only applies when it matches the compiled predicate byte for byte. The existing partial unread-probe index is untouched.
- **Keyset paging for the inbox post-filter loop**: `NoteFilter` gains `after: Option<NoteSeekAfter>` (`khive-storage`), a boundary over the store's default `created_at DESC, id ASC` order; `query_notes_filtered_count_free` dispatches it to a new `fetch_notes_after` that runs two single-direction seeks (the exact-timestamp tie group ordered by `id`, then strictly older timestamps), because neither the OR-form boundary nor a row-value comparison gets index-seek treatment for a mixed `DESC, ASC` order on this SQLite build. `after` is rejected together with a custom `order_by` or a non-zero offset. `query_inbox_response` now carries the last row's `(created_at, id)` between pages instead of an offset; rows, order and `has_more` are unchanged.
- The GTD scan cap (`TASK_SCAN_MAX_ROWS`) stays as a safety bound: reading `handle_tasks` and `fetch_all_matching_tasks` showed every status/assignee/priority predicate already pushed into `NoteFilter`, so the cost was the missing index, not un-pushed filtering.

## Query plans (`EXPLAIN QUERY PLAN` on the compiled SQL)

Task listing, status Eq + assignee Eq, before: `SEARCH notes USING INDEX idx_notes_kind (namespace=? AND kind=?)` + `USE TEMP B-TREE FOR ORDER BY`. After: `SEARCH notes USING INDEX idx_notes_task_assignee (namespace=? AND kind=? AND <expr>=?)`.

Unread-default inbox listing: `SEARCH notes USING INDEX idx_notes_unread_probe_recipient_direction (...)` before and after; a regression test pins it.

Page 50 of 50 on a 5,000-row fixture at page size 100: offset form 20,317 VM steps, seek form 638 (tie branch 21, older branch 617).

## Tests

- `migrations_tests.rs`: V27 applies to a pre-V27 database (indexes dropped to simulate it) and to a fresh one; `latest_schema_version` advances to 27.
- `note_tests.rs`: plan assertions for the task-listing shapes (with a dropped-index control so the assertion cannot pass vacuously), the unread-probe regression guard, seek paging equals offset paging on a 900-row fixture, `after` + `order_by` rejected, and the VM-step comparison above.
- `khive-pack-comm/tests/integration.rs`: 300-row fixture with a `from_prefix` post-filter rejecting 50 rows; the keyset loop returns the same 250-row partition, order and `has_more` boundary as offset paging, across external offset chunks.

## Not in this change

- A general `(to_actor, direction)` index for the read/all-status mailbox listing was drafted and dropped: without `ANALYZE` statistics SQLite preferred it over the tuned partial unread-probe index for the unread-default listing, which is the path the issue measured; four existing unread-probe plan tests caught the regression. That listing shape remains unindexed; a test guards against reintroducing the index without first resolving the planner ambiguity.
- Per-verb `cpu_ms` in the audit payload: `cost_unit` is a deterministic cost model, not a timer, and the payload shape is pinned by a test; adding real timing needs dispatch-boundary instrumentation and a payload schema change, which belong in their own change.

### DIFF
--- a/crates/khive-db/sql/027-notes-hot-property-indexes.sql
+++ b/crates/khive-db/sql/027-notes-hot-property-indexes.sql
@@ -13,17 +13,34 @@
 -- `crates/khive-db/src/stores/note.rs`) -- an index only helps when its key
 -- expression matches the compiled predicate byte-for-byte.
 --
--- A general `(to_actor, direction)` index for the comm mailbox listing was
--- also drafted for this change but dropped: with no ANALYZE statistics,
--- SQLite sometimes chose it over the existing partial
--- `idx_notes_unread_probe_recipient_direction` for the default unread-status
--- listing, even though the partial index (scoped by read status) is
--- strictly cheaper for that case -- confirmed by `EXPLAIN QUERY PLAN`
--- flipping to the new index and by VM-step regressions in
--- `crates/khive-db/src/stores/note_tests.rs`'s existing unread-probe tests.
--- Shipping it would have risked regressing the one path #2390 actually
--- measured (`comm.inbox` at unread-default status). The read/all-status
--- listing case it would have served remains unindexed.
+-- Both indexes are partial on `deleted_at IS NULL` only, not additionally on
+-- `kind = 'task'`: every caller binds `kind` as a parameter rather than a
+-- literal, and SQLite can use a partial index only when it can prove the
+-- index's WHERE clause at plan time, which a bound parameter never
+-- satisfies (`crates/khive-db/src/stores/note_tests.rs`'s
+-- `narrowing_hot_property_index_to_kind_task_is_not_chosen` measures this).
+--
+-- These indexes cover only the GTD task-listing path. Two related paths
+-- remain unindexed. The comm inbox listing's `status="all"`/`status="read"`
+-- case (no `$.read` predicate, or `$.read = true` rather than the tuned
+-- `$.read` absent-or-false probe) and the default `gtd.tasks()` listing (no
+-- `status=` filter, which compiles to an open `$.status NOT IN
+-- ('done','cancelled') OR $.status IS NULL` exclusion with no bounded set to
+-- seek) both still cost a full partition scan. A general
+-- `(namespace, kind, created_at DESC, id ASC)` index would serve the first
+-- of those, but it is also a legal plan for `comm.inbox`'s default
+-- `status="unread"` listing, and with no ANALYZE statistics SQLite prefers
+-- it there over the purpose-built partial
+-- `idx_notes_unread_probe_recipient_direction`, even though the partial
+-- index is cheaper for that case -- confirmed by `EXPLAIN QUERY PLAN`
+-- flipping to the general index and by VM-step regressions in
+-- `crates/khive-db/src/stores/note_tests.rs`'s
+-- `candidate_created_at_id_seek_index_flips_unread_probe_plan`. A narrower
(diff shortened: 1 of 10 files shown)


ohdearquant (commented):

Verdict on head 9da2407570c71b2ba957d672a5f0075b5bfaec15: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Approving at this exact commit. Verified before approving: the head matches the commit named in the request, all check runs at this commit have completed with no failures, no change requests are outstanding, and the approving account is not the authoring account.

PR #2397, opened 2026-09-05
fix(runtime): keep the events daemon's SQLite locks across sidecar hardening

ohdearquant:
## Summary

The events daemon released SQLite's advisory locks on its own database at boot. An external connection that closed later (a backup tool, an inspection shell) then took itself for the last connection, checkpointed, and unlinked the `-wal`/`-shm` sidecars while the daemon kept writing to the unlinked inodes. Every other opener saw a stale sidecar pair over a newer main file and read the database as malformed, while writes through the daemon socket kept landing.

Mechanism: `run_events_daemon` opened the pool, ensured the schema, and only then ran the owner-only hardening of the database and its sidecars. The hardening opens and closes a descriptor on each file (`O_NOFOLLOW` open, `fchmod`). POSIX advisory locks are per process and per inode and are released by the close of any descriptor for the inode, so that close dropped the SHARED lock SQLite keeps on the database in WAL mode and the shared lock on the `-shm` DMS byte, while SQLite still believed it held both.

## Change

- `run_events_daemon`: hardening runs before the pool opens. A pre-existing database and any sidecars an earlier process left behind are tightened first; sidecars SQLite creates afterwards inherit the database file's mode. After the open, the new `verify_events_db_owner_only_unopened` checks the file set with `lstat` only and refuses to serve on a loosened or replaced entry, so nothing opens a live store file outside SQLite.
- `harden_events_db_sidecars` documents its precondition (no open connection to the database in this process). The embedded arm already ran it before the open; the comment now says the order is load-bearing.
- Cross-process regression test `crates/kkernel/tests/events_daemon_keeps_sqlite_locks.rs`: runs the real `events-daemon` subcommand as a child process and probes the database SHARED range and the `-shm` DMS byte from the test process with `F_GETLK`, asserting a shared lock held by the daemon's pid at bind and after an idle period. A process cannot see its own locks through `F_GETLK`, so the check has to be cross-process. The test kills the daemon on unwind (an RAII guard), so a failed assertion leaves no child holding descriptors against a removed directory.
- Unit test for the unopened check: a loosened `-wal` is refused naming the file and its mode; a symlink planted at `-shm` is refused without being followed.
- Embedded (direct) backend registry: the registry was keyed by path spelling and mode, so a second key could name a file this process already holds (another spelling, the other `read_only` flag, or the path after a rename) and run the hardening, or a read-only pool's header read, on it. Each registry entry now pins the held file's inode with a read-only descriptor opened after the pool and never closed (a close is the only descriptor operation that releases the process's locks), and `direct_backend` resolves the requested path by `stat` against those pinned identities before anything else. The same mode reuses the existing backend under any spelling and after a rename. A read-only open of a database this process holds writable is refused: its header read would close a descriptor on the held file, and a read-only pool admits no live sidecar set anyway. A writable open beside a read-only holder proceeds through SQLite alone, with no hardening of the held files: a read-only main beside a writable secondary is a supported topology, and in direct mode the read-only handle may reach the shared events database first. The spelling itself is never a key: a file created at a held path after a rename is another database with its own pool. Only a file nobody in the process holds is hardened. Five regression tests: a second spelling through a directory link resolves to the same backend and leaves a loosened sidecar untouched; a renamed held database resolves to the same backend and is not touched under its new name; a writable open beside a read-only holder succeeds and leaves the file's mode alone; a read-only open of a writable-held database is refused; a new file at a held spelling after a rename gets its own pool and a link planted there is refused.
- CHANGELOG entry under Unreleased.

## Verification

- `cargo test -p khive-runtime --lib events_split::tests::`: 45 passed.
- `cargo test -p kkernel --test events_daemon_keeps_sqlite_locks`: passes on this branch. With the old order restored (hardening after the open) the same test fails with `F_GETLK type 2` (`F_UNLCK`) on the database range, and passes again after the restore.
- Scratch reproduction: an events daemon built from this branch holding a scratch database, then `sqlite3_rsync origin replica` (the local-replica backup form). Shared locks held by the daemon on both ranges before and after; sidecars still on the path with the same inodes. The released 0.8.0 binary under the same procedure: no lock before, sidecars unlinked after.
- Two runs of the local-replica backup script against a scratch registry and backup root, with the daemon from this branch holding the origin: both `success`, `wal_after_bytes` equal to `wal_before_bytes`, sidecars on the path with the same inodes, shared locks still held, replica and origin `quick_check` ok.
- `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean; `cargo test -p khive-runtime` green (1457 passed).

## Not changed

`khive-db`'s read-only open path reads the database header through a plain file open before any connection of that pool exists; it serves read-only backends and snapshot inspection and holds no live connection of its own, so it is outside this fix. The embedded registry keeps that path off any database this process holds writable by refusing the read-only open first; the same rule applies to any other caller that reaches it from a process holding a connection to the same file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   instead of a hard-coded identity with no meaning outside the deployment it
   was named for. Behaviour with the variable explicitly set is unchanged.
 
+### Fixed
+
+- The events daemon no longer releases SQLite's advisory locks on its own
+  database. Permission hardening of the `-wal`/`-shm` sidecars ran after the
+  database was opened and closed a descriptor on each file, which under POSIX
+  drops every lock the process holds on that inode; a backup or inspection
+  connection closing afterwards then took itself for the last connection,
+  checkpointed, and unlinked the sidecars while the daemon kept writing to the
+  unlinked files. Hardening now runs before the open, the post-open check uses
+  `lstat` only, and a cross-process test asserts the locks are held.
+
 ## [0.8.0] - 2026-08-27
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   moved away, a writable open there is refused while its sidecars remain,
   whatever file has since appeared at the spelling, a database this process
   holds read-only included, and however the path is written: the directory is
-  resolved before the spellings are compared.
+  resolved before the spellings are compared, and the sidecars are known by
+  the identity the holder pinned at its open, so the same files reached under
+  another name, a hard link or a spelling the filesystem folds onto theirs,
+  are refused the same way.
 
 ## [0.8.0] - 2026-08-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   open, so a held database is found under any spelling and after a rename: a
   second spelling reuses the open backend, a read-only open of a database the
   process holds writable is refused, and a writable open beside a read-only
-  holder goes through SQLite alone, with no hardening of the held files.
+  holder goes through SQLite alone, with no hardening of the held files. A
+  writable open at a spelling where `-wal`/`-shm` sidecars remain beside no
+  database is refused rather than creating a database over them: they belong
+  to a database moved or removed from that spelling, possibly one still open.
 
 ## [0.8.0] - 2026-08-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   checkpointed, and unlinked the sidecars while the daemon kept writing to the
   unlinked files. Hardening now runs before the open, the post-open check uses
   `lstat` only, and a cross-process test asserts the locks are held. The
-  embedded backend registry resolves a database by file identity before it
-  hardens anything, so a second spelling of a held database reuses the open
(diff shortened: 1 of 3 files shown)


ohdearquant (commented):

Verdict on head fb87ace0aab9c92aa93ecca3d4dffb3b8d72983b: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head cd6c8316a021599640c4f08895e4c70dc658f774: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head f713d0b65b01494abb9604f663fbea8662ac4b9e: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head e2d395b5355dc2160c75924dd17c6e4ff1a7c6a9: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head cbfe176fc1abab46fa4171a917d9e27fc1b25b94: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Superseded by #2398, which is merged; the broader sidecar-identity hardening from this branch is tracked as a follow-up.

PR #2398, opened 2026-09-06
fix(runtime): keep the events daemon's SQLite locks across sidecar hardening

ohdearquant:
## Summary

The events daemon ran the owner-only hardening of its database and `-wal`/`-shm` sidecars after the pool had opened them. Hardening opens and closes a descriptor on each file, and POSIX advisory locks are per process and per inode, released by the close of any descriptor for the inode, so the SHARED lock SQLite keeps on the database in WAL mode and the shared lock on the `-shm` DMS byte were dropped while SQLite believed it still held them. An external connection closing later (a backup tool, an inspection shell) took itself for the last connection, checkpointed, and unlinked the sidecars underneath the daemon, which kept writing to the unlinked inodes; every other opener then read the database as malformed.

## Changes

- Hardening runs before the pool opens; a post-open check verifies the file set with `lstat` only and refuses to serve on a loosened or replaced entry; the hardening function documents its precondition.
- A cross-process test runs the real `events-daemon` subcommand and asserts with `F_GETLK` that the daemon holds both shared locks at bind and while idle; the lock constants are typed once so the test compiles on macOS and Linux.
- CHANGELOG entry.

Supersedes #2397, narrowed to the fix and its regression test. The broader sidecar-identity hardening explored on that branch (held databases resolved by inode, moved-away sidecars, read-only holders) is deferred to a follow-up.

Assessed and not changed: the read-only backend header probe in the database pool opens the header outside SQLite before that pool has connections; latent for a same-process live holder, outside this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   instead of a hard-coded identity with no meaning outside the deployment it
   was named for. Behaviour with the variable explicitly set is unchanged.
 
+### Fixed
+
+- The events daemon no longer releases SQLite's advisory locks on its own
+  database. Permission hardening of the `-wal`/`-shm` sidecars ran after the
+  database was opened and closed a descriptor on each file, which under POSIX
+  drops every lock the process holds on that inode; a backup or inspection
+  connection closing afterwards then took itself for the last connection,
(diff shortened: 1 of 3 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Events split verified at 2efa8d52: CI green at head, merge-base equals main.

PR #2406, opened 2026-09-07
feat(knowledge): expose bounded lexical timeout diagnostics

ohdearquant:
## Summary

Add `degraded.lexical_timeout_details` to `knowledge.search` and
`knowledge.suggest` when a lexical read times out in a publicly disclosable phase.
The list contains at most one record per pass and at most three records total.

Every response with `degraded.lexical_timeout=true` also carries
`degraded.lexical_timeout_instrumented=true`, regardless of whether any phase is
withheld. This marker identifies build instrumentation capability, not hidden
execution details. Healthy lexical responses omit it entirely; it is never false.

| Field | Meaning |
| --- | --- |
| `pass` | `full`, `subquery_1`, or `subquery_2` |
| `phase` | `reader_open` or `term_frequency` in public responses |
| `stage_elapsed_ms` | Time since the lexical read stage began, captured at the failed read |
| `operation_elapsed_ms` | Time spent in that failed read attempt, including waiting and cleanup |
| `configured_budget_ms` | The configured lexical-stage budget |
| `effective_budget_ms` | The remaining deadline allowance at stage entry, respecting an earlier parent deadline |

Durations use a monotonic clock and are truncated to whole milliseconds. They
are captured before subsequent reranking or metadata work, not at serialization.
Healthy responses omit the details list. A timed-out response also omits the list
when all of its timeout records are operator-only.

## Compatibility

`degraded.lexical_timeout` remains the same boolean with the same meaning.
Consumers reading that boolean see no change, including when no public details
are available. Existing candidate selection, partial-result retention, stage
budgets, timeout predicates, and other degradation fields are unchanged. Other
storage errors still propagate as before.

## Disclosure Design

Public phase labels are limited to `reader_open` and `term_frequency`: whether
these phases are entered does not depend on matches returned by the corpus.

The remaining labels are operator-only: `phase_a_rowids`, `phase_b_hydration`,
`eligibility_fallback`, `namespace_membership`, and `recent_fallback`. Their
reachability depends on global-index matches before namespace filtering.
Changing only foreign-namespace matches can select a different failure phase
while leaving local results unchanged. Exposing those labels would therefore
reveal match presence even without returning a foreign row or a match count.
These omissions are intentional, not missing instrumentation.
The public list can omit records and is not a complete execution ledger, even
when it is non-empty. The capability marker does not disclose whether a hidden
pass timed out or completed normally.

All seven phases emit the same structured timing fields as warning-level tracing
events. The diagnostic records contain no SQL, raw errors, query text, row
identifiers, namespace names, match counts, touched-row counts, or byte counts.
This change does not eliminate pre-existing wall-clock timing side channels.

## Interpretation Limits

A phase identifies where the timeout was observed, not why the allowance was
exhausted or where the entire budget was spent. Earlier work may have consumed
the budget before an almost-immediate failed read. Pooled-reader waiting occurs
inside query calls, so operation elapsed cannot separate waiting from execution.
Elapsed time also cannot distinguish deadline expiry from explicit cancellation.
This change adds diagnostics; it does not fix or establish the underlying timeout
cause.

## Validation

Required package validation is `cargo test -p khive-pack-knowledge`, and it passes at
this revision, including the capability marker and its mixed-pass test.

- Deterministic failures at every read site check the phase label and matching structured tracing fields.
- Clock controls distinguish individual-probe time from whole-stage time, preserve earlier parent deadlines, cover cancellation and expired deadlines, and verify that downstream work cannot alter captured values.
- Public dispatch checks healthy omission, the configured budget, all three decomposed pass identities, and preservation of the existing boolean and other response fields.
- Partial-result fixtures check the exact retained candidates under both query orders; existing candidate-selection regressions remain covered.
- A foreign-match fixture changes the withheld failure phase at equal elapsed time while requiring identical public output and unchanged local fallback results.
- A decomposed-query fixture toggles a real foreign match to add an operator-only timeout while retaining the same public timeout record. Both responses must serialize to identical bytes with the capability marker present and true in both.
- Two mutation controls were run against this revision. Disabling timeout-record propagation fails at the missing-record assertion, and making the capability marker depend on withholding fails at the mixed-pass marker assertion. Both sources were restored byte for byte afterwards and the package suite, the workspace check, the formatter and the linter all pass again.

### DIFF
--- a/crates/khive-pack-knowledge/docs/design.md
+++ b/crates/khive-pack-knowledge/docs/design.md
@@ -132,7 +132,11 @@
   lexical-only degradation with request time left to spare still gets a full rerank pass; the
   `lexical_timeout` degradation flag is still attached to the response whenever the stage itself
   timed out, independent of whether the rest of the request completed normally.
-- `degraded.lexical_timeout` remains a boolean. When a timed-out read's phase can be disclosed
+- `degraded.lexical_timeout` remains a boolean. On instrumented builds, every response with
+  `degraded.lexical_timeout=true` also carries `degraded.lexical_timeout_instrumented=true`,
+  regardless of whether any phase is withheld. This marker identifies build capability, not
+  hidden execution details; healthy lexical responses omit it entirely, never emitting false.
+  When a timed-out read's phase can be disclosed
   without exposing global-index match presence, `degraded.lexical_timeout_details` adds at most
   one record per lexical pass (`full`, `subquery_1`, `subquery_2`), at most three records total.
   Each record contains `pass`, `phase`, `stage_elapsed_ms`, `operation_elapsed_ms`,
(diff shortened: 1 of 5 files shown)


oceanwaves630 (approved):
Approved at 4e02bd2c.

PR #2407, opened 2026-09-07
docs(adr): ADR-172 versioned notes, compare-and-set, keyed create, durability option

ohdearquant:
## ADR-172: versioned notes — compare-and-set updates, keyed create-if-absent, and a durability option

Proposed. Design only; no code in this PR.

A state layer that keeps one versioned document per key on top of khive needs three things the note surface does not export today: a conditional update the caller can key on the revision it read, a unique key with create-if-absent, and listing by key prefix, last-write time and all-of tags. The runtime already performs a snapshot compare-and-swap internally; this ADR exposes a database-maintained `version` (trigger, covering every `UPDATE notes` site), `expected_version` on `update`, an optional `fence` precondition on a second keyed note in the same writer transaction, `key` with a partial unique index, three list predicates with a keyed keyset cursor, and a per-deployment `synchronous` / `fullfsync` option with its measured commit cost.

Acceptance is stated in the ADR before implementation, including the cross-process races and the mutation arms.

### DIFF
--- a/docs/adr/ADR-172-versioned-notes-compare-and-set.md
+++ b/docs/adr/ADR-172-versioned-notes-compare-and-set.md
@@ -126,9 +126,15 @@ calls a note, a `key` is what a program looks it up by.
 
 `list(kind="note", ...)` gains:
 
-- `key_prefix`: rows whose `key` starts with the string. Implemented as a range predicate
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Approved.

PR #2408, opened 2026-09-07
docs(adr): ADR-173 chunked blob upload for objects larger than one wire frame

ohdearquant:
## ADR-173: chunked blob upload for objects larger than one wire frame

Proposed. Design only; no code in this PR.

`blob.put` accepts objects up to 64 MiB but carries them as base64 inside one request, and the request parser caps the whole `ops` string at 1 MiB under the 8 MiB frame, so the largest object a client can put over the wire today is 768 KiB; reads already cover the gap through `blob.get` ranges. This ADR adds `blob.begin` / `blob.put_part` / `blob.commit` / `blob.abort`: a server-side staging object addressed by a capability id, sequential parts bounded by a limit derived from both caps and hashed incrementally, a tail resend acknowledged only on matching length and digest, commit through the one publish routine `blob.put` uses, expiry enforced at the verbs and by a daemon-owned sweeper with the staging area invisible to the orphan GC on both backends, and the put ledger written at commit in hosted mode. Frame size, parser cap and wire protocol are unchanged.

Acceptance is stated before implementation, including the tail-retry, expiry, shared-publish and part-limit mutation arms.


oceanwaves630 (approved):
Approved.

PR #2409, opened 2026-09-07
docs(adr): ADR-174 ordered streams, dense per-stream append with expected_seq and a fence

ohdearquant:
Adds ADR-174: an ordered per-stream append primitive for notes.

- `stream.append(stream, record, expected_seq=None, fence=None)` inserts a note and a row in a new per-stream ledger in one writer transaction; `seq` is dense from 1 per stream; `expected_seq` conditions the append and fails with `seq_conflict` writing nothing; the fence is ADR-172 §2b unchanged.
- `stream.read(stream, after, limit)` returns entries in `seq` order off the ledger's primary key with `head_seq` and `next_after`; `stream.stat` returns the head.
- Entries are immutable and streams dense, enforced by two triggers on `notes` (`stream_member`).
- Reconciliation after a lost acknowledgement uses `expected_seq` plus a read at that `seq`; no server-side idempotency token.
- Retention (drop/truncate) is explicitly out of scope.

Docs only. Depends on ADR-172 (#2407) for `version`, `key` and `fence`.


oceanwaves630 (approved):
Approved.

PR #2410, opened 2026-09-07
ci(workflows): reject empty filtered test selections

ohdearquant:
`cargo test <filter>` exits 0 when the filter selects zero tests, so a workflow step that runs a filtered selection stays green forever once the filter stops matching (a renamed module, a removed `#[ignore]`). Four steps in this repository run such selections: the tz-database audit, the bench-embedder invariant in the bench pipeline, and the two Windows lanes in ci.yml.

This change adds `scripts/assert-tests-ran.sh`, which reads cargo's own `test result:` summary lines from a tee'd log and fails the step when the number of tests that reported a result is below the floor the step declares. All four steps declare a floor of 1: the guard detects an empty selection without pinning the suite size. `scripts/tests/test_ci_workflows.py` discovers every filtered `cargo test` step in the workflows and asserts each one is guarded, and exercises the guard's own arms (zero selection, malformed summary, a test path beginning with `result::`, CRLF and colored output).

No product code changes. The full workspace suite was run at this head on macOS with a newer stable toolchain than the CI pin, 0 failures.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+scripts/assert-tests-ran.sh text eol=lf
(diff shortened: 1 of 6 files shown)


oceanwaves630 (approved):
Approved.

PR #2411, opened 2026-09-07
fix(python): preserve native pages and structured operation errors

ohdearquant:
Two contract defects in the Python client facade, both found by reading the daemon's wire shapes against the client's parsing.

- `list` results in cursor mode arrive under a collection key (`entities`, `notes`, `edges`) with `next_after` and `scan_incomplete`; the client read only `items`/`results`, so a cursor page parsed as empty. `Page` now carries the collection under its key plus `next_after`, `scan_incomplete` and the limit-clamping metadata, and does not invent `next_offset` or counts the server did not send.
- Per-op error objects (`kind`, `code`, `stage`, `retryable`, `request_state`, `task_terminated`, the unavailable-arm fields, `details`) were flattened to a string, which lost write finality. `OpResult.error` is now a validated `OpError` model preserving every field verbatim; string errors stay strings. `Session` now routes through the shared envelope validation instead of bypassing it.

Tests: offline fixtures derived from the server's error and page shapes, plus live controls against a scratch daemon (including a three-note cursor walk to a terminal page). Reverting either fix makes the corresponding tests fail.

### DIFF
--- a/python/docs/RESPONSE_WIRE_CONTRACT.md
+++ b/python/docs/RESPONSE_WIRE_CONTRACT.md
@@ -1,0 +1,201 @@
+# Response wire contract
+
+This document describes the native JSON shapes currently emitted for substrate
+`list` operations and per-operation failures. It records source behavior; it does
+not add server guarantees or describe every transport-level error. Examples omit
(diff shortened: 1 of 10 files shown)


oceanwaves630 (approved):
Approved.

PR #2412, opened 2026-09-07
ci(python): require the client suite before merge

ohdearquant:
Runs the Python client suite (`python/tests`) as a required job in the CI gate. The job builds a debug `kkernel` so the suite's live controls run against a scratch daemon of the same revision; pytest exits nonzero when nothing is collected, so no separate selection floor is needed.


oceanwaves630 (approved):
Approved.

PR #2413, opened 2026-09-07
fix(marketplace): point the plugin manifest at the one plugin that exists

ohdearquant:
The plugin manifest listed seven plugin sources under `./marketplace/<pack>` that do not exist in the tree, so a marketplace client following it finds nothing. The only plugin is `./marketplace/khive`, whose `plugin.json` already covers every pack. The manifest now lists that one entry, taking its name and description from the plugin's own manifest.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,39 +7,9 @@
   },
   "plugins": [
     {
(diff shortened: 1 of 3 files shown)


oceanwaves630 (approved):
Approved.

PR #2414, opened 2026-09-07
python: bump khive-py to 0.1.1

oceanwaves630:
khive-py 0.1.0 covered two response contracts: before and after the page and error-object changes that landed on main. A client cannot distinguish the two by version, so this bumps the package to 0.1.1 so a dependant can declare a floor that names the current contract.

No code change; the two version strings only.


ohdearquant (approved):
Approved.

PR #2415, opened 2026-09-07
docs(adr): ADR-172 Amendment 1, keyed cursor tiebreaker and after_key ambiguity

oceanwaves630:
ADR-172 Amendment 1. Four corrections to the keyed-note sections, no verb signature changes:

- the keyed listing cursor gains `id` as a tiebreaker, because a key may be held by one live note per kind and a listing that omits `note_kind` spans kinds, so `(updated_at, key)` alone is not unique;
- `after_key` on a key held by more than one kind fails with `key_ambiguous` instead of picking one;
- the keyed refusals disclose occupancy within the caller namespace only and are admitted under the read decision of the gate;
- migration 028 sits in the versioned chain every database walks from V1, so fresh and upgraded databases carry the same objects.

Two acceptance arms added: the tiebreaker walk with its mutation, and the fresh-build schema read-back.


ohdearquant (approved):
Approved.

PR #2417, opened 2026-09-07
fix(db): synchronize filesystem blob publication directories

oceanwaves630:
## Summary

- Synchronize the shard directory, its parent and the blob root before acknowledging a Unix publication.
- Repeat synchronization on deduplicated writes, including retries after a post-rename failure.
- Synchronize a newly opened writable root and its parent. Keep read-only opening nonmutating.

## Compatibility

`FsBlobStore::new` now creates only the root directory. Its parent must already exist; missing parents produce an I/O NotFound error naming the path. Custom nested roots must have their parent provisioned before startup. Existing root directories and configured symlink spellings remain supported.

File synchronization is unchanged. Non-Unix publication retains its previous file-sync and atomic-rename behavior; it does not gain the Unix directory barriers.

## Known Boot Resolution Boundary

`crates/khive-mcp/src/serve.rs:3843` suppresses blob resolver errors when `[storage.blob]` is absent. A `KHIVE_BLOB_ROOT` beneath a missing parent can therefore be ignored with a debug log, and boot may continue without a blob store. Explicit `[storage.blob]` configuration errors abort. This existing routing remains unchanged and needs a separate fix.

## First-Boot Durability Boundary

The default blob root is `blobs` beside the database. Its parent is the database directory, which daemon startup may create recursively (`crates/khive-mcp/src/serve.rs:3891` and `:4007`). Startup does not establish that directory's own durability.

This change covers publication beneath an existing, durable root parent. It does not establish whole-daemon machine-loss durability or provide a power-loss guarantee for arbitrary filesystems and devices.

## Tests

Coverage includes fresh publication, deduplicated retries, file/rename/directory failures, initialization retries, missing-parent refusal, descriptor identity, root-path replacement, kernel error propagation and fresh-process verified reads.

- Workspace tests: 9,769 passed, 20 ignored on macOS.
- Workspace check, formatting, and all-target/all-feature Clippy with warnings denied passed.
- Nine barrier-removal, retry-bypass, descriptor and recursive-creation mutations failed their expected controls; restored code passed.

Windows/Linux execution and filesystem power-cut behavior were not tested.

### DIFF
--- a/crates/khive-db/src/stores/blob.rs
+++ b/crates/khive-db/src/stores/blob.rs
@@ -8,7 +8,11 @@
 //! path (guaranteeing same-filesystem rename), the written length is checked
 //! against the input length, then an atomic rename publishes the entry —
 //! crash-safe (a crash mid-write leaves an orphaned temp file, never a
-//! partially-committed blob).
+//! partially-committed blob). On Unix, publication also synchronizes the
+//! shard directory chain before acknowledging the reference. Initialization
+//! synchronizes the root and its parent, which must already exist and be
+//! durable. This is not a power-loss guarantee for arbitrary filesystems or
+//! devices. Non-Unix publication does not provide these directory barriers.
 
 use std::collections::HashMap;
 use std::fs;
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
Host gate green (fmt, clippy, focused blob tests, full workspace); review verdict archived locally.

PR #2419, opened 2026-09-07
docs(adr): ADR-178 plan-only request, grammar check without dispatch

oceanwaves630:
Adds ADR-178 (Proposed): a plan-only form of the `request` tool that runs the existing DSL parser and returns the parsed shape without dispatching anything.

- `request(ops, plan=true)`, `kkernel exec --plan`, and `Session.plan(ops)` return one result: `parsed`, `mode`, `stage_count`, per-stage `verb`, `pack`, `known` (catalog read), normalized `args`, `prev_refs`, and the parser's `limits`. A grammar error is the same `DslError` text dispatch would return for the same string.
- Nothing is dispatched: no identity, no gate, no audit event, no store access. `plan=true` composes with `ops` only; presentation, output format, sink and request id beside it are refused.
- Explicitly not admission: the record forbids keying any authorization decision on a plan beyond "parses, and how many stages". Dry-run of effects, resolved `$prev` targets and gate pre-checks are out of scope.
- Motivation: runtimes that let a model author request strings were porting the grammar to refuse malformed strings early; one parser stays one parser.
- Acceptance stated before implementation: seven arms including corpus-equal error text, zero-effect control, stage accounting, unknown verb not an error, envelope isolation, three surfaces one result, and a dependency assertion that the plan handler reaches no gate, store or identity constructor.

### DIFF
--- a/docs/adr/ADR-178-plan-only-request.md
+++ b/docs/adr/ADR-178-plan-only-request.md
@@ -1,0 +1,135 @@
+# ADR-178: Plan-Only Request — Grammar Check Without Dispatch
+
+- **Status**: Proposed
+- **Date**: 2026-09-07
+- **Extends**: [ADR-016](ADR-016-request-dsl.md) (Request DSL: the one `request` tool, three
+  syntactic forms, the parser this record reuses)
+- **Depends on**: [ADR-096](ADR-096-warm-daemon-per-request-identity.md) (per-request identity; a
+  plan-only request carries none)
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

Verdict on head fdb11ef7db4738d60fc380861a1aec8d514db008: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (approved):
Approved.

PR #2421, opened 2026-09-08
fix(memory): avoid blocking recall on store acquisition

oceanwaves630:
`memory.recall` could exceed its request deadline on a loaded machine: the memory pack called synchronous store constructors (`notes`, `events`, `vectors_for_model`) from async code, and each may wait for SQLite's writer connection. When the same task also owned an open read snapshot, the constructor's wait blocked the executor that had to run the snapshot's release, so the wait ran to the writer checkout timeout, once per model and per site. Under test-suite parallelism this surfaced as one CI shard failing `ns733b_recall_verbose_multi_model_breakdown_excludes_off_namespace_candidates` with `memory.recall exceeded its 30000ms deadline`; the same call sites are reachable from the daemon's dispatch path, where it is a latency defect.

Fix: `store_access::acquire_store` runs a store constructor on a blocking worker after the request read phase is confirmed active, carrying the request context so a stopped request is refused instead of admitted; the four call sites (recall note store, ANN event store, ANN fingerprint store, ANN vector store) use it. Request deadlines and the detached ANN build lifecycle are unchanged; no deadline, readiness timeout or poll count was altered; `recall.rs` is untouched.

Regression control: a current-thread test opens a deferred snapshot, then polls store acquisition ahead of the snapshot's release in a biased join, for each of the three stores. On the previous code it fails with a writer checkout timeout; with the fix all three acquisitions complete while the sibling releases the snapshot, and the writer timeout counter does not move.

Measured under matched load (64 fresh test processes, 32 concurrent, 32 CPU hogs, 12 logical CPUs): before, 1 of 64 hit the deadline with first-dispatch median/max 301 ms / 35.2 s; after, 0 of 64 with 467 ms / 769 ms.

Gates on the CI-pinned toolchain: fmt, clippy (workspace, all targets, deny warnings), memory pack tests (331 + 89), full workspace tests.

### DIFF
--- a/crates/khive-pack-memory/src/ann.rs
+++ b/crates/khive-pack-memory/src/ann.rs
@@ -1171,7 +1171,13 @@ async fn emit_ann_warm_phase_event<P: serde::Serialize>(
     payload: P,
 ) {
     // A missing event store means auditing is unconfigured, not that warming failed.
-    let Ok(store) = rt.events(token) else {
+    let result = crate::store_access::acquire_store("memory.ann.event_store", {
+        let runtime = rt.clone();
+        let token = token.clone();
+        move || runtime.events(&token)
+    })
+    .await;
+    let Ok(Ok(store)) = result else {
         return;
     };
     let payload_value = match serde_json::to_value(&payload) {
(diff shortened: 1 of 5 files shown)


ohdearquant (commented):

ohdearquant (approved):
Approved.

PR #2423, opened 2026-09-08
fix(scripts): data guard reads staged blobs from the index

oceanwaves630:
## Defect

`scripts/check-json-data.sh` in pre-commit mode listed the staged paths from the index but judged each one by its working-tree file: `check_file` returned early when the path had no file on disk, and the size check read `wc -c` on the working-tree copy. Two staged states therefore passed the guard and reached the commit:

- a staged `.jsonl` whose working-tree file was removed after `git add` (the guard exits 0, the commit contains the JSONL);
- a staged JSON above the size ceiling whose working-tree copy was truncated after `git add` (the guard measures the 1-byte file, the commit contains the 300 KB one).

Reproduced in a scratch repository; the fixture test below carries both arms.

## Change

Pre-commit mode now reads the staged blob: existence is the index listing itself (`--diff-filter=ACMR`), and the size is `git cat-file -s :<path>`. A staged JSON whose object cannot be read is refused rather than skipped. `--all` mode is unchanged: it scans the working tree, which is what a CI checkout has.

## Test

`scripts/tests/check-json-data-test.sh`, seven cases in scratch repositories with nothing committed: staged JSONL present (refused), staged JSONL with the working-tree file removed (refused), staged oversize JSON with the working-tree copy truncated (refused, message names the staged size), small JSON (passes), `KHIVE_ALLOW_DATA=1` bypass (passes), `--all` on an untracked JSONL (refused), benchmark-path JSONL with the file removed (passes). Against the previous script the same test fails at case 2. The test runs in the existing data-leak-guard CI job after the tree scan.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -399,6 +399,8 @@ jobs:
       - uses: actions/checkout@v7
       - name: Scan whole tree for corpus files
         run: bash scripts/check-json-data.sh --all
+      - name: Guard fixture test (staged blobs judged from the index)
+        run: bash scripts/tests/check-json-data-test.sh
 
   secret-scan:
     name: Secret scan (gitleaks)
--- a/scripts/check-json-data.sh
+++ b/scripts/check-json-data.sh
@@ -5,7 +5,10 @@
 # an entire memory corpus as JSONL onto a public PR.
 #
 # Modes:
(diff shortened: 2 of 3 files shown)


ohdearquant (commented):

ohdearquant (approved):
Approved.

Issue #2424, opened 2026-09-08
runtime: an audit-obligation error discards the committed domain result; carry a three-valued domain disposition on the error


ohdearquant:
## Observed

When a dispatch succeeds and its audit-obligation row then fails to commit, the caller receives the audit error and nothing else. `fold_audit_obligation` in `crates/khive-runtime/src/pack.rs` has the domain result in hand in that arm and discards it:

```rust
(Ok(value), Ok(())) => Ok(value),
(Ok(_), Err(audit_err)) => Err(audit_err),
(Err(err), _) => Err(err),
```

The domain verb ran before the audit append (dispatch at ~2464, audit append and fold at ~2632-2648), so in this arm the domain effect has usually already committed. The error the caller sees carries no disposition of that effect. Every error raised after the dispatch is therefore the same class for a caller, whatever reason string it names: not a retry permission. #2399 (`AdmissionDeadlineExpired` after the effect committed) and #2256 (`StoreFailure` after the write committed) are two observations of this one class.

The consequence lands on the caller: a client that retries "carefully" after such an error produces a duplicate it cannot see, and a client that does not retry has to re-read an outcome it has no key for. ADR-133 D4's fourth forbidden mode (duplicated accounting) is pushed one layer out, onto the domain records, where nothing checks it.

## Expected

The never-false-success policy of ADR-133 D2 to D4 stays: the op still reports `ok: false`. The error object becomes machine-readable about the domain effect:

- `domain_disposition`: one of `committed`, `not_committed`, `unknown`. Three values, because a caller reads an absent or two-valued field as `not_committed` exactly when the runtime could not tell, which reinstates the defect with more confidence attached.
- `committed` on the obligation arm above, together with the domain result the fold was holding, so the caller learns the same identifiers a success would have returned.
- `not_committed` on refusals raised before any dispatch (unknown verb, invalid parameters, write-time gate refusal).
- `unknown` everywhere else, including any error the transport loses.

Same shape on the MCP `request` envelope and on the daemon frame; the Python client preserves the field on the per-op error object.

## Scope

Contract change on the error type; specified by an amendment to ADR-133 before code. Operation identity for a caller-side outcome re-read is a separate ask and is not replaced by this: response loss before the fold stays `unknown`.

Issue #2399, opened 2026-09-06
comm verbs report AdmissionDeadlineExpired from the audit obligation after the effect committed


ohdearquant:
## Symptom

Under load, a verb call can return

```
internal: audit obligation commit failed for verb "comm.read": AdmissionDeadlineExpired
```

with `ok: false`, although the verb's own effect has already committed. Observed three times today on one busy daemon (load average above 80):

- `comm.send`: the error was returned, yet the recipient's poller saw the message about 30 s later. A caller that retries on the error sends a duplicate.
- `comm.read` (twice, distinct messages): the error was returned; a subsequent `comm.reply` on the same message reported `marked_read: true`, and `comm.thread` read the body normally.

Other calls in the same minute succeeded, so this is the audit obligation's own deadline expiring after the primary write, not a general refusal.

## Expected

One of two contracts, stated explicitly in the response:

1. The audit obligation is part of the verb's commit, so a failed obligation rolls the verb back and `ok: false` means nothing happened; or
2. The verb commits first and the obligation is best-effort, in which case the response says the verb committed (for example `ok: true` with an `audit: deferred` marker) so callers do not retry.

Today the response reads as (1) while the behaviour is (2).

## Reproduction

Busy daemon (two cargo builds plus several agent processes running), then `kkernel exec 'comm.read(id="<fresh inbound message id>")'`. Three of about ten calls in a busy minute returned the error; all three effects had landed.

PR #2425, opened 2026-09-08
docs(adr): ADR-133 amendment 3, domain disposition on the obligation error

oceanwaves630:
## What

Amendment 3 to ADR-133. The fold that applies D2 (`fold_audit_obligation`) fails a would-be-success dispatch when its obligation row does not commit, and discards the domain result it is holding. The caller's error therefore says nothing about a domain effect that usually already committed (#2399, #2256). The never-false-success policy stays; the amendment states what the error must say and what a caller may do with it.

- **A3.1** Every per-op error object carries a machine-readable `domain_disposition` with exactly one of `committed` (with `domain_result`), `not_committed` (pre-dispatch refusals), `unknown` (everything else). Three values, because an absent or two-valued field is read as `not_committed` exactly when the runtime could not tell. Same shape on the MCP envelope, the daemon frame and the Python client.
- **A3.2** Consumer rule: a post-dispatch obligation error is never retry permission; `committed` resolves from the error, `unknown` only by an outcome re-read keyed on a caller-chosen identity, which is a separate record.
- Seven acceptance arms, including the honest one: a caller that replays after a committed-arm error produces a second record today, asserted until the identity record lands.

Tracked as #2424. Docs only; implementation follows the signed text.

### DIFF
--- a/docs/adr/ADR-133-incidental-writes-off-the-request-hot-path.md
+++ b/docs/adr/ADR-133-incidental-writes-off-the-request-hot-path.md
@@ -762,8 +762,11 @@ carries `domain_disposition` with exactly one of three values:
   well as by parameter validation, and the secret gate is the standing example: it is invoked inside
   handlers (`crates/khive-runtime/src/operations.rs`, the code and git ingest
   handlers) at points where a handler may already have written, so a `SecretDetected` raised there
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
Approved.

PR #2426, opened 2026-09-08
docs(adr): ADR-179 operation identity on memory.remember (keyed create)

oceanwaves630:
## What

ADR-179 (Proposed): a caller-chosen `key` on `memory.remember`, unique among live `memory` notes in the write namespace. A replay of the identical request after a lost acknowledgement fails with `key_conflict` carrying `existing_id`, so the caller holds the id of the one memory it intended without an idempotency store of its own. Two writers racing on one key get exactly one success; the unique index is what refuses.

Storage is ADR-172 §3's `notes.key` column and partial unique index, landed here verbatim as a first slice (migration 028); ADR-172's compare-and-set, fence, `create(key=)`, `get(key=)` and keyed listings are not landed and stay in ADR-172, which reconciles its migration text to this one when it lands.

The key is immutable, released by soft or hard delete (so a pruned key may be reused; stated as a consequence), composes with `source_id` (edge only on the successful create), and constrains nothing about the source. A `key_conflict` carries `domain_disposition: "not_committed"` under ADR-133 Amendment 3, and an `unknown` disposition on the replay is resolved by replaying again.

Nine acceptance arms: lost acknowledgement (with the no-key control producing two notes), concurrent recovery across two processes, decoy annotation on the same source, unchanged behaviour without a key, release on delete, key validation, namespace scope, disposition interplay, three surfaces. README index row added. Docs only.

### DIFF
--- a/docs/adr/ADR-179-operation-identity-memory-remember.md
+++ b/docs/adr/ADR-179-operation-identity-memory-remember.md
@@ -30,8 +30,8 @@ supplies that identity for the write that a memory-backed agent loop issues most
 The store already refuses duplicates by identity where it has one: `try_insert_note` is an
 `INSERT OR IGNORE` followed by an `external_id` verification, and the audit lane appends by generation
 id. `memory.remember` exposes no identity of its own. ADR-172 proposes a general `key` on notes with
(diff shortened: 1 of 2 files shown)


ohdearquant (approved):
Approved.

PR #2428, opened 2026-09-08
fix(memory): cancellation-aware store acquisition, worker released on cancel or drop

### LINKED ISSUE #2422: memory: make store acquisition cancellation-aware (blocking worker held until writer checkout timeout)
Follow-up to #2421, which moved memory-pack store acquisition (`notes`, `events`, `vectors_for_model` constructors) onto a blocking worker so a recall no longer blocks the executor while waiting for SQLite's writer.

Two bounded gaps remain in `crates/khive-pack-memory/src/store_access.rs`:

1. **Cancellation does not stop an in-flight acquisition.** When a recall is cancelled or times out while `acquire_store` is waiting on the writer, dropping the `JoinHandle` does not stop the blocking task; it keeps a blocking-pool worker until the writer checkout timeout expires (up to one checkout timeout per abandoned task). Under concurrent cancellation and writer contention this can amplify tail latency and, in the limit, crowd the blocking pool. Acquisition should be cancellation-aware (check the request context inside the wait, or bound the number of concurrent acquisitions per request phase).
2. **A cancelled join is classified as an internal failure.** A cancelled `spawn_blocking` task surfaces as `RuntimeError::Internal`, but the ANN shutdown classifier only recognises cancelled joins wrapped in `StorageError::Driver`, so a cancelled warm phase reads as a completed failure rather than `PhaseCancelled`.

Minor: `handlers/common.rs` repeats the `memory.recall.note_store` label in two literals for `await_request_read_phase` and `acquire_store`; bind it once.

Acceptance: a test that cancels a recall while acquisition is blocked on the writer and asserts the blocking worker is released before the checkout timeout, and a test that a cancelled acquisition reports `PhaseCancelled`. The existing regression test for #2421 (which fails when the fix is reverted) stays as is.

oceanwaves630:
## Defect

`memory.recall` acquires its note, event and vector stores on a blocking worker. When the store constructor has to wait for the SQLite writer, that wait was a plain `try_lock_for(checkout_timeout)`: a cancelled or dropped recall left the worker parked for up to the full checkout timeout, and a cancelled `JoinError` from the acquisition task surfaced as `Internal` rather than the stored `PhaseCancelled` terminal event. Closes #2422.

## Change

- `ConnectionPool::writer_until(should_stop)`: the constructor path waits for the writer in slices of at most 2 ms, checks the stop predicate between slices, keeps the original absolute checkout deadline, and re-checks once more after acquiring so a cancellation that landed during the final slice still wins. Once admitted, constructor DDL is not interrupted.
- The memory store acquisition opts into that path through an explicit scope on the captured request context (`scope_store_acquisition`); ordinary accessors and ordinary writer admission are unchanged. Dropping the awaiting future closes a watch sender so the pending constructor stops even when no inherited cancellation fired.
- A stopped acquisition maps to the existing request-phase `StorageError::Timeout`; a cancelled join keeps its concrete `JoinError` inside `StorageError::Driver`, which the ANN shutdown classifier already reads as `PhaseCancelled`. The terminal-event emitter was extracted unchanged so a test can observe the stored event.

No storage trait, wire shape, schema or ADR changes.

## Tests

- Pool: an in-flight wait stops on the second predicate check without a pool timeout; a request deadline shorter than the checkout timeout stops the wait; an uncancelled wait still times out with the configured checkout timeout and the timeout counter moves.
- Backend: ordinary accessors ignore request cancellation; an admitted constructor finishes its DDL after a cancellation fired inside real SQLite work.
- Memory: cancelled and dropped acquisitions release the blocking worker while the writer is held (notes, events, vectors); an uncancelled acquisition waits and succeeds; a cancelled acquisition join emits exactly one `PhaseCancelled` event. The existing writer-contention regression test is unchanged and still passes.
- Mutation controls: with the stop predicate ignored inside the wait, the worker-release tests fail at their timing assertion; with the join error flattened as before, the terminal-event test fails on `PhaseCompleted`.

Gates run on Rust 1.95.0: fmt, clippy `--all-targets -D warnings` for the touched crates, and the `khive-pack-memory` and `khive-db` test suites pass. The full workspace test run stopped at a comm hold-time timing assertion while the host was under heavy unrelated load; that gate is re-run on a quiet host before merge and is recorded here rather than weakened.

### DIFF
--- a/crates/khive-db/src/backend.rs
+++ b/crates/khive-db/src/backend.rs
@@ -562,6 +562,20 @@ impl StorageBackend {
         )))
     }
 
+    fn constructor_writer(&self) -> Result<crate::pool::WriterGuard<'_>, SqliteError> {
+        let context = khive_storage::capture_request_read_context();
+        let Some(operation) = context.store_acquisition_operation() else {
+            return self.pool.try_writer();
+        };
+        self.pool
+            .writer_until(|| context.blocking_stop_reason().is_some())?
+            .ok_or_else(|| {
+                SqliteError::RequestReadStopped(khive_storage::StorageError::Timeout {
+                    operation: operation.into(),
+                })
+            })
+    }
+
     /// Get a NoteStore. Applies the notes DDL if not already present.
     ///
     /// Idempotent — safe to call multiple times.
@@ -582,7 +596,7 @@ impl StorageBackend {
             ));
         }
         if !self.is_read_only() {
-            let writer = self.pool.try_writer()?;
+            let writer = self.constructor_writer()?;
             note::ensure_notes_schema(writer.conn())?;
 
             // The anti-join repair is a full `notes` scan -- gate it to run at
@@ -630,7 +644,7 @@ impl StorageBackend {
             ));
         }
         if !self.is_read_only() {
-            let writer = self.pool.try_writer()?;
+            let writer = self.constructor_writer()?;
             event::ensure_events_schema(writer.conn())?;
         }
 
@@ -732,7 +746,7 @@ impl StorageBackend {
             )?));
         }
 
-        let writer = self.pool.try_writer()?;
+        let writer = self.constructor_writer()?;
 
         // Detect old-schema vec0 tables that predate the `field` column.
         // vec0 virtual tables do not support ALTER TABLE, so we must drop and recreate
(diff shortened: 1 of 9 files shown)


ohdearquant (approved):
Approved.

PR #2429, opened 2026-09-08
docs(adr): ADR-174 Amendment 1, stream.batch as one fenced transaction with per-op errors as values

oceanwaves630:
## Summary

ADR-174 Amendment 1 (Proposed): `stream.batch`, one request that runs a list of appends and keyed document writes with authority checked once and per-stream numbering in list order, in one of two named modes. Atomic mode (the default under a fence) runs the list in one writer transaction, checks a caller-supplied `observed` version set before the first write, and refuses the whole batch on any member's conflict, writing nothing. Per-member mode (no fence) runs each member in its own transaction in list order and returns a member's refusal as that member's value while its siblings stand.

## Why

§5 leaves a caller two shapes: the request array (siblings commit, numbers in admission order, nothing about list position) and the chain (ordered, but the remainder aborts after a failure and keeps what committed). A state layer built on streams needs both an all-or-nothing form under a fence and a best-effort form without one, and neither can be composed from the array or the chain on the caller's side because the fence is per transaction and checking authority once is the point. The consumer's suite proves both forms; its unfenced conformance case is quoted in the amendment as arm 1, and its fenced tests (nothing lands after a stale fence; every observation check precedes every insert) shape arms 2 and 3.

## What does not change

The request array and the chain keep §5's semantics; `stream.append` is unchanged; the density invariant holds inside an atomic batch by construction and across a per-member batch by serial execution. Reads are not batch members in either mode.

## Acceptance

Nine arms in the amendment: per-member order and values, atomic refusal writes nothing, observed checked before the first write, per-member refusal keeps siblings, authority once, two processes atomic (adjacent numbers), two processes per-member (dense union, list order, no adjacency, interleaving control), mode-argument refusals, and five mutation controls.

### DIFF
--- a/docs/adr/ADR-174-ordered-streams-append.md
+++ b/docs/adr/ADR-174-ordered-streams-append.md
@@ -336,8 +336,11 @@ keyed document write of ADR-172 §2 and §3). Common to both modes:
   as `existing_id` (ADR-179 D5), so the consumer rule of ADR-133 Amendment 3 reads it without a
   special case.
 - Authority is checked once for the batch, on the caller's namespace, before the first write.
-- Appends to one stream take consecutive numbers in list order; appends to different streams are
-  independent.
+- Appends to one stream are numbered in list order, so a batch's numbers on one stream increase
+  with list position; appends to different streams are independent. Adjacency is a property of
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
Approved.

PR #2430, opened 2026-09-08
docs(adr): ADR-180 tool pack, registry, discovery and use policy

oceanwaves630:
## ADR-180: Tool Pack

Proposed record for a `tool` pack on the kg pack: a namespace-scoped registry of tools, skills, plugins and verbs (project entities), capability concepts joined by `implements` edges, `tool.suggest` that merges registry search with capability expansion and carries the caller's decision, a policy table with a fixed resolution order (grant, then most specific policy, then a side-effect default), and asynchronous approval with a fast path.

Eight acceptance arms and the known rough edges are in the record. The code lands in a separate PR.


ohdearquant (approved):
Approved.

PR #2432, opened 2026-09-08
docs(adr): ADR-181 and ADR-182 Amendment 1, run contract and credential rule

oceanwaves630:
## ADR-181 and ADR-182, Amendment 1

Adopted on the first whole-file reads of the two records. ADR-181 gains the run contract the loop driver binds to: `cwd` relative to the tree, caller-supplied environment filtered by allow-listed keys, `declared_write_paths`, tail-preserving capture with byte counts, a receipt for every refusal, session identity, a sandbox identity object, and a hard refusal of version-control binaries. ADR-182 states the credential rule as an explicit amendment of ADR-108 hard rule 4, fixes the enforcement order (Gate, then tool policy, both in the receipt), refuses approval and merge on fork pull requests without a named policy row, and scopes `git.checkout` to a read-only object read. Acceptance arms added to both.


ohdearquant (approved):
Approved.

PR #2433, opened 2026-09-08
fix(daemon): include forwarded actor in default read scope

### LINKED ISSUE #2427: memory: fold forwarded actor namespace into default daemon read visibility
## Problem

A protocol-v4 daemon client can supply `actor_id` with an empty `visible_namespaces` list. Episodic `memory.remember` then writes to the actor namespace, but ordinary default reads receive a token whose visible set contains only `local`. The same client can therefore write a memory that its default recall and list requests cannot see.

Source inspection at `bf79b03dec371977b7479a209d2f809825c70e43`:

- `crates/khive-pack-memory/src/handlers/remember.rs:34-53` redirects an implicit episodic write to the actor namespace.
- `crates/khive-runtime/src/daemon.rs:1168-1174` copies frame visibility into `RequestIdentity` unchanged.
- `crates/khive-runtime/src/pack.rs:2407-2433` uses that visibility plus `local` for default reads.
- Config-based identity already folds the actor into read visibility at `crates/khive-runtime/src/config.rs:755-770`, as asserted by `engine_config.rs:1728-1761`.

This report identifies the source inconsistency; a current-revision daemon-wire regression has not yet been executed.

## Reproduction and Expected Behavior

Use an isolated daemon with the kg and memory packs, protocol v4, `namespace="local"`, a named actor, and `visible_namespaces=[]`.

1. Create an observation, then remember an episodic memory with `source_id` referencing that observation and a unique tag.
2. Recall using that tag and an exclusive `created_before` bound later than the memory's creation time.
3. List the observation, memory, and its annotation edge through the same identity.
4. Repeat the reads with the actor namespace explicitly present in visibility as a positive control.

Default reads should include both local records and the caller's actor namespace, consistent with configured identities. Ordinary default writes must remain in `local`; episodic routing and explicit namespace overrides must retain their existing behavior.

## Repair Boundary

Normalize forwarded actor visibility once where a daemon frame becomes `RequestIdentity`, before both ordinary registry dispatch and coordinator-intercepted search. Do not widen deliberately constructed internal identities: scheduled replay explicitly keeps its event scope, and exact-token reconstruction must remain exact.

A registry-only change is insufficient because coordinator search resolves visibility separately in `crates/khive-mcp/src/server.rs:2564`. An unconditional change to every `RequestIdentity` would also alter scheduled replay at `server.rs:3712-3721`.

## Acceptance

- A real daemon-wire regression fails before the fix and passes afterward, covering recall, memory listing, and annotation listing.
- Explicit actor visibility already works and does not duplicate records.
- Anonymous frames retain local-only visibility; another actor does not acquire the first actor's implicit read scope.
- Explicit namespace reads remain exact, and default observation writes remain local.
- Coordinator wire reads share the normalized visibility; scheduled replay and exact-token identities are unchanged.
- Existing namespace-scoped recall tests remain green. Reverting only the ingress normalization makes the new regression fail again.

oceanwaves630:
## Summary

A client that identifies with a non-`local` actor and forwards an empty visibility list to the warm daemon could write episodic memory into its actor namespace and then not read it back with a default-scope recall, because the daemon frame copied the forwarded visibility verbatim while direct configuration loading folds the actor id into default read scope. Closes #2427.

## Change

- `khive-runtime` daemon ingress applies the same fold once: a forwarded `actor_id` that parses as a non-`local` namespace is added to `visible_namespaces` when absent. Explicit `namespace=` operations and scheduled-replay token identities are unchanged.
- ADR-007 records the ingress fold in one sentence.

## Tests

- New wire regression against a file-backed scratch daemon (kg + memory packs): the three default-read omissions are red at the parent commit and green here; anonymous and other-actor default reads still do not see the actor's memory; explicit `local` does not widen; repeated actor visibility does not duplicate hits.
- Server controls: scheduled replay does not gain actor or daemon visibility; coordinator search consumes the normalized identity without widening internal identities.

### DIFF
--- a/crates/khive-mcp/src/daemon.rs
+++ b/crates/khive-mcp/src/daemon.rs
@@ -26,6 +26,9 @@ use tokio::net::UnixStream;
 use crate::tools::request::RequestParams;
 
 #[cfg(test)]
+mod memory_namespace_tests;
+
+#[cfg(test)]
 mod test_harness;
 
 #[cfg(test)]
--- a/crates/khive-mcp/src/daemon/memory_namespace_tests.rs
+++ b/crates/khive-mcp/src/daemon/memory_namespace_tests.rs
@@ -1,0 +1,335 @@
+use std::path::Path;
+use std::time::Duration;
+
+use futures::FutureExt;
+use khive_runtime::daemon::{run_daemon, DaemonRequestFrame, PROTOCOL_VERSION};
+use khive_runtime::{KhiveRuntime, RuntimeConfig};
+use serde_json::{json, Value};
+use serial_test::serial;
+use tokio::time::timeout;
+
+use super::test_harness::{clear_daemon_env, connect_when_ready, exchange, RecoveryTestGuard};
+use crate::server::KhiveMcpServer;
+
+const ACTOR: &str = "test:memory-writer";
+
+async fn request(
+    socket: &Path,
+    config_id: &str,
+    actor: Option<&str>,
+    visible: &[&str],
+    ops: String,
+) -> Vec<Value> {
+    let frame = DaemonRequestFrame {
+        ops,
+        presentation: Some("verbose".into()),
+        presentation_per_op: None,
+        namespace: "local".into(),
+        actor_id: actor.map(str::to_owned),
+        process_ref: None,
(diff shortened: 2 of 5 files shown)


ohdearquant (approved):
Approved.

PR #2435, opened 2026-09-08
fix(runtime): preserve domain outcomes in request errors

### LINKED ISSUE #2256: Write verbs hard-fail on AdmissionDeadlineExpired after the domain write has already committed

### LINKED ISSUE #2399: comm verbs report AdmissionDeadlineExpired from the audit obligation after the effect committed

### LINKED ISSUE #2424: runtime: an audit-obligation error discards the committed domain result; carry a three-valued domain disposition on the error

oceanwaves630:
# fix(runtime): preserve domain outcomes in request errors

A domain write can succeed before its required audit append fails. Request errors now report `domain_disposition` and retain the exact canonical `domain_result` when the operation’s own handler succeeded. Predispatch refusals report `not_committed`; handler failures and uncertain transport outcomes remain `unknown`. Errors remain `ok: false` and gain no automatic retry permission.

Per-operation errors use structured objects with `error.message`. Daemon protocol v4 keeps the existing text `error` and adds optional `error_detail`. MCP and Python preserve these details; aborted operations and response depth/byte limits retain their disposition. Existing direct runtime APIs strip only the new dispatch-stage wrapper; the post-success audit-obligation error is intentionally new.

Validation completed:

- Workspace tests, workspace check/clippy, full runtime/MCP/CLI test reruns, and formatting checks pass.
- Final focused disposition run: 30 passed. Audit fault-injection/read-admission controls: 28 passed.
- One actual committed-send failure passes through the shared MCP request body and native frame handler; Python preserves the captured response’s exact error and canonical UUID. The control checks one native send, two physical notes, and no automatic replay. Python uses captured native response bytes with a synthetic handshake; this is not a deployed-server test.
- Removing disposition/result fields makes the committed-send regression fail after domain persistence is confirmed; exact restoration passes.
- Python full suite: 345 passed with the newly built local CLI explicitly selected, including 10 live cases against an isolated temporary scratch daemon. External reader fixtures pass five compatibility cases and nine synthetic retention cases.

Final checks and commit hooks pass. The committed source builds successfully, and the 345-test Python suite also passes against that exact binary in an isolated scratch daemon.

Closes #2424
Closes #2399
Closes #2256


ohdearquant (approved):
Approved.

PR #2436, opened 2026-09-08
feat(request): inspect request plans without dispatch

oceanwaves630:
`request(ops, plan=true)`, `kkernel exec --plan '<ops>'`, and `Session.plan(ops)` return the parsed plan of a request, its ops, verbs, parameters and dependencies as resolved against the grammar and the loaded verb catalog, without executing any operation. A malformed request reports its parse error in the same envelope shape; unknown verbs and unknown parameters are reported per op.

- `khive-request` gains a pure planner, `plan_request`, shared by every surface.
- The MCP `request` tool accepts `plan: bool` (default false) and returns the plan envelope.
- Daemon protocol moves to v5. Plan frames are a new request kind; a mismatched client or daemon refuses with `version_mismatch`, and the CLI plan path requires an already-running matching daemon.
- Python: `Session.plan(ops)` and transport support.
- Docs: request guide, daemon API reference, kkernel usage.

Validation: workspace fmt, clippy (all targets, `-D warnings`), check and tests; the Python suite; seven behavioral acceptance arms in which the MCP, CLI and Python surfaces produce identical plans for the same corpus.

Implements ADR-178.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2018,6 +2018,7 @@ dependencies = [
  "serde_json",
  "serial_test",
  "sha2",
+ "syn",
  "tempfile",
  "tokio",
  "tokio-util",
@@ -2447,6 +2448,7 @@ dependencies = [
  "khive-types",
  "serde",
  "serde_json",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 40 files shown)


ohdearquant (approved):
Approved.

ohdearquant:
Superseded by #2438, the same change rebased onto the current main.

PR #2437, opened 2026-09-08
docs(adr): ADR-181 and ADR-182 Amendment 2

oceanwaves630:
Two amendments adopted from the loop driver's contract tests.

ADR-181 Amendment 2: limits the platform cannot enforce per run are refused at config load with `unsupported_on_platform`; the file-size limit's truncate-then-signal semantics; `exec.identity` as the sandbox identity a client checks before running.

ADR-182 Amendment 2: exact compare on every ref move (`expected`, `expected_head`, `expected_local`, `expected_remote` with explicit `null` for an absent remote branch), commits by plumbing, actor-only credentials through a validated resolver, no hooks or filters or configuration, repository and reviewer identity at the platform login, `disposition` with a read-only `git.reconcile`, caller-scoped receipts with `gate` and `policy`, tree and diff reads, and the test-only `contract-faults` feature.


ohdearquant (approved):
Approved.

PR #2439, opened 2026-09-08
docs(adr): ADR-181 and ADR-182 Amendment 3

oceanwaves630:
Amendment 3 to two accepted records, docs only.

ADR-181 (exec):
- `declared_write_paths` entries cover themselves and every path below them at a `/` boundary; a change is any byte, mode, addition or deletion difference against `tree_in`; undeclared changes keep the input entry in `tree_out` and make `success` false.
- The per-session `seq` is allocated inside the receipt insert and backed by a unique index over `(namespace, session_id, seq)`; the column is what every reader reports.
- The sandbox profile denies `process-exec` for `git`, `gh`, `git-*` and every path in the `[exec] never` set, beside the existing registered-binary refusal.

ADR-182 (git):
- The `[git_write.actors]` row is selected by the actor the runtime resolved for the request, never by a request field.
- The fork-merge decision is the `git.pr_merge.fork` policy row; recorded as an argument, no rule change.

Acceptance arms 20-22 (ADR-181) and 27 (ADR-182) added. The matching exec code lands in the exec pack PR.


ohdearquant (approved):
Approved.

PR #2440, opened 2026-09-08
docs(adr): ADR-174 Amendment 2

oceanwaves630:
Amendment 2 to ADR-174, docs only, found on the source audit before the streams schema lands.

- The stream ledger gains a sixth trigger refusing any `UPDATE` on `note_streams`; a ledger row is append-only once written. Acceptance 7 adds the direct-statement arm, acceptance 8 counts six triggers.
- `CHECK (seq > 0)` is overdetermined by the gap trigger for `seq = 0`, so its mutation arm now strips the gap trigger as well, with the trigger-only removal as the control.
- `stream.stat` reads count and head from one statement over the same rows.


ohdearquant (approved):
Approved.

ohdearquant (approved):
Approved.

PR #2441, opened 2026-09-08
docs(python): cite the daemon protocol constant at version 5

oceanwaves630:
The response wire contract table cites the daemon protocol constant by its source line. The constant moved to 5 with the plan-only request change and the citation check on main went red. One citation updated; the check passes again.


ohdearquant (approved):
Approved.

PR #2442, opened 2026-09-08
docs(adr): ADR-182 Amendment 4

oceanwaves630:
Amendment 4 to ADR-182, docs only, recording rulings given while the local slice of the git verbs is being implemented.

- Every `update-ref` compare-and-swap takes `--no-deref`; a symbolic branch ref refuses with `ref_symbolic`.
- Ref moves write a reflog entry carrying the receipt id; `git.reconcile` settles an `unknown` row only on that evidence, never on head equality.
- The new verbs require the tool pack at call time and fail closed as `policy_unavailable` when its tables are absent; the legacy verbs keep their kg-only dependency.
- Read verbs gate on a repo-only allow-list match; `git.receipts` is an actor-scoped audit read outside the allow-list; `git.branch` carries no credential.
- The daemon fingerprint covers the whole `[git_write]` section.

Acceptance arms 28 to 30 added.


ohdearquant (approved):
Approved.

PR #2443, opened 2026-09-08
docs(adr): ADR-182 Amendment 5

oceanwaves630:
Amendment 5 to ADR-182, docs only, two tightenings found by reading Git and the runtime source.

- Reconcile: Git's files backend appends the reflog entry before installing the ref and a failed install keeps the entry, so the receipt marker is necessary, not sufficient. An `unknown` row settles to `committed` only with the marker present and the sha at or below the current head. No crash-atomicity claim anywhere.
- Policy: pack backends may differ, so the decision comes through the registry dispatch of `tool.check` with the caller's identity, never by reading the tool tables through the git pack's own backend handle; taken before any write, no lock held across it.

Acceptance arms 31 and 32 added.


ohdearquant (approved):
Approved.

PR #2444, opened 2026-09-08
fix(runtime): assert the served protocol version in the legacy decode test

oceanwaves630:
The legacy-decoder test in the daemon module pinned the response's version
field to the literal `4`. The served constant is now `5`, so the assertion
failed against a frame that was otherwise correct, and because the suite runs
fail-fast the failure also stopped roughly 1400 later tests from running.

What the test exists to show is that a response struct shaped for the older
protocol still decodes a current frame, including the version field it carries.
Asserting the constant preserves that subject and does not go stale on the next
bump. The neighbouring fixtures that carry literal mismatched versions (`0`,
`3`, `99`) are untouched: there the literal is the input under test, not an
expectation about what the daemon serves.


ohdearquant (approved):
Approved.

PR #2445, opened 2026-09-08
docs(adr): ADR-091 Amendments 17 and 18, stream write scopes

oceanwaves630:
Two rows in ADR-091's write-transaction external-work inventory, one per stream caller.

**Amendment 17 (2026-09-08)** enters `KhiveRuntime::stream_append`. Amendment 11's audit table is normative and exhaustive, and its review guard requires any new `SqlAccess::atomic_unit` caller to be entered in it; the single-append path is one, so this records its row. The row is added by appending a new amendment rather than editing Amendment 11's table, which keeps the earlier text readable as what was true when it was written.

**Amendment 18 (2026-09-10)** enters `run_prepared_stream_batch`, the batch surface that landed after Amendment 17 was written. Same owner, same guard, its own row. It is SQL-only for the same reason the append is: every member arrives as a prepared action, the closure binds and executes statements and folds bounded results, the stream head is read once per distinct stream and memoized, and every refusal returns from the closure. The deadline comparison reads the writer's clock through SQLite rather than the process clock, so one transaction has one time. Per-member mode runs the same closure with a one-element member list, so each writer hold covers exactly one member.

Amendment 18 also names something the audit found rather than implying it is handled: in atomic mode the statement count inside the writer hold grows with the member and observation counts the caller sends. The verb refuses an empty list and does not cap a large one, and the only ceiling today is the daemon's 8 MiB frame, which is a transport limit rather than a decision about writer hold time. Every comparable path in this document names its own number instead. That gap is stated in the amendment, not resolved by it: a verb-level cap changes the stream contract and belongs to that contract's own amendment.

Docs only. `deno fmt --check` clean, `scripts/lint-adr-refs.sh` and `scripts/lint-adr-status.py` both clean at the branch head. Main is merged in, so the checks run against a current merge ref.


ohdearquant (approved):
Approved.

ohdearquant (commented):

Verdict on head a1f4335fbc405c77506796682f5b56a0442ea32a: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2450, opened 2026-09-08
docs(vamana): plain reference to the private build_thread_count

oceanwaves630:
The public docs of `VamanaIndex::build` linked the private `build_thread_count`, which the doc build rejects under `-D warnings` (`rustdoc::private_intra_doc_links`). A code span says the same thing without a link. `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` passes locally on this head.


ohdearquant (approved):
Approved.

PR #2451, opened 2026-09-08
docs(kg,mcp): state the envelope contract of get and neighbors in help

oceanwaves630:
## What this states

Callers read the verb help through `verbs()` and the MCP instructions, and three shapes were not written down anywhere a caller reads:

- `get` returns the bare record with no envelope. `kind` is the granular kind (`concept`, `task`, `observation`, ...), `entity_type` is the governed subtype when one is set, and an entity's vocabulary type lives at `properties.type`.
- `neighbors` returns a bare array of hits, while `list` returns the `{"items": [...]}` envelope. Each hit carries `origin_id` for the queried node, `edge_id`, `relation`, `weight`, and the neighbor's `id`, `kind` and `name`; `include_entity_type=true` adds `entity_type` when the neighbor has one.
- The kg pack's verbs are unprefixed; every other pack's verbs are written `pack.verb`.

Measured against the serving binary before writing: `list(kind="entity", limit=1) | get(id=...) | neighbors(node_id=..., include_entity_type=true, limit=2)` returned an envelope, a bare record with `kind: concept`, `entity_type: null` and `properties.type` set, and a bare array of two hits with exactly the keys listed.

## Where

`crates/khive-pack-kg/src/handler_defs.rs` for the two verb descriptions, `crates/khive-mcp/src/server.rs` for the instructions sentence, `docs/guide/api-reference.md` for the same two statements. No behaviour changes.

### DIFF
--- a/crates/khive-mcp/src/server.rs
+++ b/crates/khive-mcp/src/server.rs
@@ -4710,7 +4710,10 @@ fn build_instructions(catalog: &str, builtins: &str) -> String {
     format!(
         "khive — request-only MCP surface. One tool, `request`, \
          dispatches verbs through the loaded pack registry. Configure packs via \
-         KHIVE_PACKS or --pack (built-ins: {builtins}). Verbs registered on this \
+         KHIVE_PACKS or --pack (built-ins: {builtins}). The kg pack's verbs are \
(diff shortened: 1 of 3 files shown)


ohdearquant (approved):
Approved.

PR #2505, opened 2026-09-09
feat(kg): ordered fence lists on note writes, stream.append and batch append members

oceanwaves630:
## Summary

A write may depend on several keyed notes at once. Singleton note `create`, note `update` and `stream.append` now accept either the existing fence object or a non-empty ordered list of fence objects. Every entry is checked in supplied order inside the same writer transaction as the guarded write, before its statements; the first missing or stale entry refuses the whole write.

Batch append members of `stream.batch` take the same guards. In atomic mode all member fences are checked before any write and a refusal names the outer member plus the inner list index; in per-member mode the refusal is returned at the member position and successful siblings are preserved.

## Wire contract

- The object form's message and details are unchanged.
- The list form always adds `details.index`, a zero-based decimal string, including for a one-element list. That is the only difference between an object refusal and the equivalent one-element-list refusal.
- An explicit null, an empty list, a malformed entry, an unknown entry field, an invalid key or note kind, or a non-positive expected version refuses with `invalid_input` before the write transaction opens. A list cannot name the same `(kind, key)` twice; the error names both indices.
- A confirmed fence refusal carries `domain_disposition: not_committed`. Storage failures with unknown outcomes keep their existing disposition.
- The Python client accepts a dictionary or a list of dictionaries and preserves entry order.

ADR-172 Amendment 3 records the contract (Proposed).

## Verification

Workspace fmt, all-target clippy and check are clean; runtime, kg and MCP suites pass (3,053 passed, 9 ignored); Python units pass (484). Two-daemon integration of note and stream fences passes (13). Object-form controls are extended with one-element and two-element lists, valid, stale and missing entries, first-failure ordering, unchanged target content and version, unchanged stream head, and the duplicate-key refusal.

## Client capabilities that change at merge

- `multi_head_fence` (a list-valued fence on note `create` and `update`) becomes available.
- `stream_fence` becomes available for `stream.append` and for the append members of `stream.batch` only. A batch-wide fence, the `observed` member and the keyed `write` member keep their current availability and land with the batch lane.


ohdearquant (commented):

Verdict on head 9a7a831164291378b50e6fc1791b5d2d3fc25fd4: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2607, opened 2026-09-11
docs(adr): ADR-183 batched write disposition, commit what passes

oceanwaves630:
## Context

A verb that takes a list of records validates each one in a loop and returns on
the first refusal. Every other record in the call is refused with it, nothing is
written, and the single error describes only the offending value, so the caller
holds N records and one error about text that lives in whichever record failed.

Measured on `knowledge.upsert_atoms`: a run parked 97 atoms on one refusal.
Re-sent one atom per call, 90 of the 97 committed and 7 refuse on their own. Of
the 97, 57 do not contain the reported trigger word anywhere in their own fields
— the word was in a sibling record of the same call. The per-record scan is
correct; the coupling is the control flow.

## Decision

A batched write commits the records that pass and returns the refused ones with
the record and the field that refused. `status` is `partial` exactly when
`refused` is non-empty. A caller who wants all-or-nothing passes `atomic=true`
and gets today's behaviour by name, with the refusing record still named.

`stream.batch` already has this shape: an `atomic` flag, defaulted by whether the
call carries a fence, and an explicit refusal of the contradiction ("a fence
admits no partial commit"). This ADR generalizes that one pack's contract and
names the verbs it binds today, as a census at one revision rather than as part
of the rule.

Half of it is true either way and lands first, separately: a per-record
validation failure inside a batch loop carries the record's identity. An
all-or-nothing batch whose single error cannot be located is unusable whether or
not partial commit exists.

## Scope

Status is Proposed. Docs only — no verb changes ride with it. The acceptance arms
are stated in the ADR before implementation, including the mutation control
(restoring the early return must fail the partial arms and leave the `atomic=true`
arm green, since that arm is the behaviour being restored).

This is not the request DSL's op-level batching. That rule gives each op its own
result and stops at the verb boundary; inside a verb a record list had no
disposition, and this gives it one.

## Lints

`deno fmt --check`, `scripts/lint-adr-refs.sh` (428 files, 242 titled references
OK) and `scripts/lint-adr-status.py` (186 records, 0 failing) all pass.

### DIFF
--- a/docs/adr/ADR-183-batched-write-disposition.md
+++ b/docs/adr/ADR-183-batched-write-disposition.md
@@ -61,7 +61,9 @@ A partial disposition is a successful op. Its result body is:
 }
 ```
 
-- `status` is `partial` exactly when `refused` is non-empty, and `ok` otherwise.
+- `status` is `ok` when every record in the request committed; `partial` when every record was
+  attempted and at least one refused; `incomplete` when the call stopped early and some records
+  were never attempted (§6). `incomplete` takes precedence over `partial`.
 - A record is identified by its **zero-based position in the request list**, never by a
   caller-supplied name. The slug is itself a scanned field, so echoing it would return the very
   text a secret-gate refusal exists to withhold; and two records in one payload may carry the same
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

Verdict on head 8508918bffbad33a32274142e2b8d67602f344ee: REQUEST-CHANGES, 8 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head 7899d995955cee7d1c43d15c4a344a192c203c4a: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2611, opened 2026-09-11
feat(knowledge): report search candidate provenance

oceanwaves630:
## Summary

`knowledge.search` reports where its candidates came from. The response gains `candidate_provenance` (`lexical` state, `fallback`), and every result gains `score_provenance` (`sources`, `embedding_rerank`, `normalization`, `calibrated`), so a caller can tell a lexical miss from a fused result and read a score as a request-relative rank rather than a calibrated probability.

## Behaviour

- A genuine FTS miss returns no lexical candidates instead of ranking unrelated recent corpus rows; `candidate_provenance.lexical` reports `matched`, `no_match`, `filtered`, `partial_timeout`, or `timed_out`.
- `candidate_provenance.fallback` is `ann` only when the returned set has ANN evidence and no returned result has lexical evidence; otherwise `none`.
- A hit found by both the lexical and ANN legs keeps both labels after RRF fusion; `embedding_rerank` records whether a successful rerank transformed the score.
- When no embedder is configured the rerank step is skipped and the original scores and provenance are kept.

## Docs

ADR-047 and the API reference describe the new response fields and the score interpretation.

## Tests

Integration arms for the true lexical miss, the filtered and timed-out candidate states, and both-source retention through fusion; unit arms for provenance merging under decomposition and for the skipped-embedding rerank path. The no-embedding fixture builds an in-memory runtime and never resolves a store from the environment.

### DIFF
--- a/crates/khive-pack-knowledge/README.md
+++ b/crates/khive-pack-knowledge/README.md
@@ -7,9 +7,10 @@ rerank, and composed into markdown briefings under a token budget.
 ## Features
 
 - **TF-IDF corpus search** (`knowledge.search`) over atom name/tags/content, with
-  score bands (`>=0.46` reliable, `0.42-0.46` mixed, `<0.42` mostly off-target),
-  query decomposition, and RRF fusion against an ANN pass when an embedder is
-  configured
+  request-relative scores, per-result lexical/ANN score provenance, query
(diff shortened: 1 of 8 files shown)


ohdearquant (commented):

PR #2612, opened 2026-09-11
fix(comm): bound probe count and preserve unseen pages

oceanwaves630:
## Summary

`comm.probe` now pages a burst and bounds its stale count.

- `new_messages` is the earliest 100 unseen matching rows in commit-sequence order (all remaining rows if fewer), displayed ascending by `created_at`. `cursor_us` is the maximum sequence among the rows actually returned, floored by the honored caller cursor, so a caller drains a burst larger than 100 messages by round-tripping the cursor without skipping rows. An empty page does not advance the cursor.
- `stale_unread_count` is `min(exact stale unread count, 1000)`: values below 1000 are exact, 1000 means at least 1000. It is independent of the arrival cursor and the returned page, uses the strict `created_at < cutoff` predicate, and counts through the existing partial unread-probe index with a row limit before aggregation.
- The page and the count still come from one SQL statement in one snapshot; `cursor_reset` semantics and the caller-cursor floor are unchanged. No schema, index, or `comm.inbox` change.

## Tests

- A 150-message burst drains as a page of 100 then 50 with no cursor skips.
- The stale count saturates at 1000 and stays independent of the page cursor and unread eligibility.
- An empty page preserves a negative caller cursor.
- The production SQL is exercised directly: the cutoff is strict, and `EXPLAIN QUERY PLAN` shows the count using the partial index with the cutoff and no sequence join.

## Docs

The comm pack README and the probe-cursor API note describe the paging contract and the count cap.

### DIFF
--- a/crates/khive-pack-comm/README.md
+++ b/crates/khive-pack-comm/README.md
@@ -41,8 +41,9 @@ make a filtered call return early with an empty page.
 
 `comm.probe` is a strictly read-only verb built for frequent polling (e.g.
 every 30s by many monitors): it never mutates the `read` flag or writes any
-row. It runs a single indexed query (`INDEXED BY idx_comm_message_to_actor`)
-over inbound messages addressed to the given `actor`.
+row. The page and stale unread count are computed in one SQL statement and
+share a snapshot. Both include only live inbound messages whose recipient
+exactly matches the given `actor` in the caller's namespace.
 
 Args:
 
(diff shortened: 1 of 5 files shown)


ohdearquant (commented):

Verdict on head cfe3c6e0a2634d2f6c4cae4d36eed4dcc91f12bb: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2613, opened 2026-09-11
feat(runtime): support explicit absence in note fences

oceanwaves630:
## Summary

Note fences can assert absence. Every fence entry now requires `expected_version`: an integer of at least 1 asserts the live holder's exact version in the write namespace, and an explicit `null` asserts that no live note holds that `(kind, key)`, so a missing or soft-deleted note satisfies it. `version` is accepted as an input alias; supplying both names, omitting both, or adding unknown fields is invalid. Serialization always emits `expected_version`, including the explicit null.

## Behaviour

- Applies to `stream.append`, singleton note `create`/`update`, and `stream.batch` append members (object or an ordered list of at most 100 distinct `(kind, key)` entries, checked in order in the same writer transaction).
- An absence conflict returns `fence_conflict` with `expected_version="absent"` and the live `current_version`; list refusals carry a zero-based string `index`, and an atomic member refusal also names its `member` position.
- Oversized lists are refused with `invalid_input` naming the cap and the count sent, before entry interpretation or writer admission. Empty lists, duplicate entries and outer `fence:null` remain invalid on these surfaces.
- The batch-wide `stream.batch(fence=...)` keeps its object-only contract; top-level `fence:null` still means no fence. `observed` entries are unchanged.

## Tests

Runtime and KG regressions for missing, live and soft-deleted holders, the input alias and its refusals, the no-write parameter refusal, rollback on conflict, batch inheritance, and the distinction between an unchanged `observed` entry and a null top-level batch fence.

## Docs

The KG pack README and the handler parameter definitions describe the contract.


ohdearquant (commented):

PR #2615, opened 2026-09-11
docs(adr): ADR-180 Amendment 4, a policy row can name what it is about

oceanwaves630:
## Summary

A policy row answers for every call an actor makes to a tool. There is nowhere to put "in this repository", so an operator who wants a bounded rule either writes an unscoped allow and grants more than intended, or writes nothing and gets `ask` for work that was authorized. The only `scope` in the pack today is on a grant request, where it is free text recorded with the request and decides nothing.

Amendment 4 adds a nullable `scope` to a policy row, in the same pattern language the row already uses for actor and tool.

## Contract

- A row **without** a scope is consulted for every call. Every stored row keeps its current meaning, so a table with no scopes decides every question exactly as it does today, ties included.
- A row **with** a scope is consulted only when the call carries a scope its pattern matches. A call naming no scope is decided by the unscoped rows alone, so scope narrows and can never widen.
- `tool.check` takes the scope the caller is acting in and echoes it back; `tool.request` uses the scope it already accepts for its pre-check.
- Specificity gains a fourth state below actor and tool: absent, `*`, prefix, exact. The deny-over-ask-over-allow ranking and the `created_at ASC, id ASC` tiebreak are untouched.
- Row identity for Amendment 3's upsert becomes `(namespace, actor, tool, scope)`.

## What it deliberately does not do

A grant's scope stays free text. Binding it would change what rows already marked `granted` allow, which is a migration of live authorization state rather than an addition, and it is named as the next question instead of being folded in here. A run names no scope, so `exec.run` is decided by unscoped rows; inventing a scope at that call site would be the widening the amendment forbids.

## Acceptance

Arms 26 through 33: the no-change arm first, the three-call narrowing arm, an unscoped-deny beside a scoped-allow fixture that fails in opposite directions if either rule is wrong, pattern arms, the echo arm, a migration arm covering both an existing table and an absent one, an `exec.run` arm, and a mutation control that reverses the match predicate to the widening direction and predicts which arms go red.

Docs only. The implementing change is a separate PR.

### DIFF
--- a/docs/adr/ADR-180-tool-pack.md
+++ b/docs/adr/ADR-180-tool-pack.md
@@ -283,3 +283,85 @@ than a refusal**, because a refusal is information and this is not.
 25. Mutation control, stated before running: making the upsert insert a second row instead of
     replacing turns arms 17 through 20 red and leaves 21 through 24 green. That separation is what
     proves the single-live-row property is what those arms test, rather than the ordering.
+
+## Amendment 4 (2026-09-11): a policy row can name what it is about
+
+A policy row answers for every call an actor makes to a tool. There is no way to write "this actor
+may push in this repository" without also writing "this actor may push anywhere", because the row
+has nowhere to put the repository. The only place the word scope appears today is on
+`tool.request(tool, actor, scope, reason, notify)`, where §Approval defines it as free text stored
+with the request, so it records what was asked for and decides nothing.
+
+An operator installing rules for a bounded demonstration has two choices today, and both are wrong:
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

Verdict on head 2bdfff373c7656b8d7a4436b05e19915735cd8ba: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

Issue #2575, opened 2026-09-11
telemetry pack: the channel table and the rollup, since the durable half already works


oceanwaves630:
## Summary

khive has the event plane already: `stream.append` / `stream.read` / `stream.batch` are the durable
log, and `brain.event_counts` proves the read side is buildable — it returns windowed counts grouped
by kind, actor and verb. What is missing is a **telemetry pack**: the same shape over an arbitrary
stream rather than over brain's own plane, plus the one piece that is genuinely absent, a declared
assignment of event kinds to carriers.

This is deliberately a small ask. Most of it is already served; the argument is for naming the
service so consumers stop each building the aggregation half in process.

## What a consumer has today, as one worked example

One runtime's telemetry package is about 1,100 lines across six modules. Its parts, and where each
would land:

| part | what it does | khive today |
|---|---|---|
| typed envelope | validates an event's payload against a per-kind schema before it is emitted | nothing |
| channel table | assigns each kind a carrier, a failure posture and a cursor kind | nothing |
| durable append | event-first append to the log, ahead of any subscriber | `stream.append` |
| ephemeral ring | bounded broadcast, never stored; a reconnect from before the window sees a gap | nothing |
| counters | per-run and per-job rollups so a reader answers "what happened" without replaying | `brain.event_counts`, but only over brain's plane |
| cursors | resume a reader from where it stopped | `stream.read` |

Two of those six are already khive. One (counters) exists for exactly one plane. Three do not exist.

## The part that is a design decision, not a port

**The channel table is configuration, not code.** Each event kind is assigned three things:

- **carrier** — `durable` (goes to the log, readers get everything since their cursor) or `ephemeral`
  (goes to a bounded ring, a reader that fell behind sees a gap).
- **failure posture** — `stop` (a refused append raises out of the emitting task) or `gap` (a dropped
  event surfaces as a gap at reconnect, never as an error).
- **cursor kind** — `log` or `ring`, which follows from the carrier but is worth stating because it
  is what a reader's resume logic keys on.

The reason this belongs in a table rather than at the emit site is that the choice is an operator's,
not an emitter's: whether a streaming delta is worth durable storage depends on the deployment, and
an emitter that decides it has hard-coded one deployment's answer. A worked table looks like:

```toml
[telemetry]
stream = "telemetry"
default_carrier = "ephemeral"         # fail-cheap: an unclassified kind is not silently made durable

[[telemetry.channels]]
[...]

PR #2617, opened 2026-09-11
docs(adr): ADR-184 telemetry pack, channel table and emit outcome

oceanwaves630:
Proposes a `telemetry` pack over the durable log khive already has, and settles the three contract questions an implementation cannot settle for itself.

**khive owns a drop, not a broadcast window.** `carrier = "ephemeral"` means the daemon accepts the event and drops it. A cross-process ring is a publish-subscribe bus with its own liveness, backpressure and delivery semantics, and adopting one would make the daemon a message broker in addition to a store. The weaker contract is honest, and adding a held window later breaks no caller, because nothing may depend today on an ephemeral event being readable at all.

**An unclassified kind takes a default the operator declared.** `telemetry.default_carrier` has no compiled default and the configuration fails to load when the table omits it. Defaulting to durable pays storage quietly for a kind nobody meant to store; defaulting to ephemeral discards data silently. Both hide the same classification bug. Every emit result carries `classified`, so an emitter that has quietly started producing an unclassified kind sees it in its own result rather than only in the operator's file.

**The emit outcome is one required enum, and ambiguity has its own value.** A durable append can fail while the row is still enqueued, so the domain effect is undetermined and the event cannot truthfully be reported as dropped. A nullable `dropped` field was rejected: a code path that forgets to set it writes exactly the same null as a path that tried and cannot say, so the value cannot carry the distinction it was chosen for. `outcome` is `recorded`, `dropped` or `unknown`; `unknown` carries the original structured error unmodified; `seq` appears only with `recorded`. Neither failure posture retries, because a retry of an append whose effect is open is how one ambiguous event becomes two recorded ones.

Two smaller rulings follow. Read `coverage` replaces the ring's `gap` flag rather than keeping a familiar name whose meaning changed. `telemetry.counts` is computed at read time against a head pinned at the start of the scan, with an incremental rollup deferred until a measurement asks for one.

Twelve acceptance arms are stated before implementation, each naming its control in the same test. Arm 5 is the rule-separating pair: an append refused before any domain effect must return a different value than an append whose effect is undetermined, or the fixtures are not separating the rules.

Docs only. No code, no schema, no verb surface changes in this PR.

Refs #2575

### DIFF
--- a/docs/adr/ADR-184-telemetry-pack.md
+++ b/docs/adr/ADR-184-telemetry-pack.md
@@ -309,6 +309,55 @@ against the pinned toolchain from `crates/`, with `--no-fail-fast`.
 
 Nothing merges before this record is accepted.
 
+## Amendment 1 (2026-09-11): configuration scope, coverage semantics, read reproducibility, disposition mapping
+
+Four questions the record above left answerable two ways. Each is settled here rather than in the
+implementation, because an implementation that answers them is making the decision.
+
+**A1.1 — the configuration requirement is pack-scoped.** D3 says a configuration omitting
+`telemetry.default_carrier` fails to load. Read globally that refuses a deployment which does not
+load the telemetry pack at all, for a key belonging to a pack it does not run. That is wrong and it
+is not what D3 decides. The requirement attaches to the pack's own configuration table: a
+deployment whose loaded pack set excludes telemetry needs no telemetry table and no key, and loads
+clean. A deployment that loads the pack and omits the key fails, naming the key. The absent-pack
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

Verdict on head 64dd6fe6310a653fabb8ac65d691978944b5d746: REQUEST-CHANGES, 7 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2667, opened 2026-09-13
docs(adr): a client must not take the socket a supervisor is for

ohdearquant:
ADR-049 §2 gives the thin client one job beyond dispatch: spawn the daemon if no responsive socket
exists. It says nothing about a deployment that also puts the daemon under a process supervisor, and
there the two starters race for the socket and PID rendezvous.

The client wins that race structurally. It spawns on demand, at the instant of a call, from whatever
process calls first; the supervisor starts at load and then on a throttle interval. The daemon's own
first-writer-wins refusal, which is correct in isolation, then turns every supervisor attempt into a
non-zero exit against the incumbent. The supervisor never owns the daemon it exists to supervise, and
the daemon runs with the configuration of whichever client happened to call first.

### Measured, not inferred

On a supervised restart of a deployment carrying both starters:

- Every client process known at the time was frozen, the incumbent stopped, the new binary installed,
  and the supervisor kickstarted **first**, with the clients still frozen. That ordering is the
  obvious remedy and it was not sufficient.
- 69 seconds later the supervisor's job still had no pid, and the socket was already held by a daemon
  carrying a *different* client's configuration — one that started after the enumeration and so was
  never frozen. The supervised job's last exit code was 1, its state "spawn scheduled".
- Stopping that daemon by hand handed the socket to a third client within 45 seconds.

A freeze covers the clients that exist when it enumerates them. Where clients are started by anything
asynchronous, the set is open, so ordering cannot close it.

### What this record proposes

That a deployment be able to declare daemon startup is owned elsewhere, and that the client then
**refuse** rather than spawn, naming the owner, with the refusal distinguishable in kind from "the
daemon is starting". Auto-spawn stays the default; the single-starter case ADR-049 describes is
untouched.

Socket activation is recorded as the stronger form where the platform offers it, and two weaker
alternatives are recorded with the reason each is insufficient: adopting a canonical configuration
fixes the configuration half and leaves the daemon unsupervised, and disabling auto-spawn per client
holds only until one client is started without it, which is the state that produced the measurement
above.

Docs only. No code changes, no behaviour change; this is a proposed record for review.


oceanwaves630 (approved):
Approved.

PR #2668, opened 2026-09-13
Name a refusal's receipt on the error, not only in its text

ohdearquant:
A refused `exec.run` writes a durable receipt and has always named it inside the refusal sentence as
`receipt_id=<id>`. That made the wording of a message part of the contract: a consumer had to recover
the id with a regular expression over prose, and the first reword breaks it silently, because a failed
match looks exactly like a refusal that wrote no receipt.

The refusal now returns a typed error carrying the id as its own field. The projected envelope reads:

```json
{
  "kind": "runtime_error",
  "code": "exec_refused",
  "message": "exec.run refused: <reason> (receipt_id=<id>)",
  "receipt_id": "<id>",
  "domain_disposition": "not_committed"
}
```

The message is byte-identical to what it was, so a reader parsing the sentence today keeps working.
`not_committed` is correct by construction for this error: the receipt exists precisely to record that
nothing ran, so the refusal carries its own proof of no-write and overrides whatever disposition the
enclosing boundary guessed.

### Arms

- The projection arm asserts the field, the code, the unchanged message and the disposition override.
- An ordinary invalid input is its control and must not grow a `receipt_id`; without it, stamping the
  field unconditionally would pass.
- The pack arm resolves the envelope's id through `exec.receipt` and asserts it names the row the
  sentence names, so the field and the text cannot drift into pointing at different receipts.
- A refusal that wrote no receipt must name none. That arm's zero is only a reading if the listing it
  reads can see a receipt at all, so it makes one through the same listing in the same test.

Mutation control: restoring the old untyped refusal reddens exactly the envelope arm and leaves the
other nine receipt tests green, the existing sentence-parsing one included.

Gate on this revision: `cargo fmt --all -- --check`, `cargo check --workspace --all-targets --locked`,
`cargo clippy --workspace --all-targets --locked -- -D warnings` and `cargo test --workspace --locked
--no-fail-fast` all zero, 10937 tests passed across 250 binaries, no failures.

The `exec.run` and `exec.receipt` descriptions now say where the id lives and that the field, not the
wording, is the contract.


oceanwaves630 (approved):
Approved.

Issue #2669, opened 2026-09-13
Listing notes sorts the whole namespace into a transient b-tree before paging


ohdearquant:
## Symptom

A serving daemon on a workstation sits at roughly two thirds of one core indefinitely, with resident
memory swinging between about 1.2 GiB and 5.7 GiB. Measured over a 15 minute window at a 2 second
sampling interval: 613 s of CPU over 899 s of wall clock, 68.2% of one core, across 55 memory
excursions with a mean duration of 13.1 s and a mean start-to-start cadence of 16.6 s. 90% of the
daemon's CPU falls inside those excursions; outside them the daemon costs 6.9% of a core.

The store held 103,425 notes.

## Where the time goes

Five stack samples, one of 4 seconds and four of 3 seconds spread over ninety seconds, are all
dominated by the same path:

```
NoteStore::query_notes_count_free
  -> sqlite3_step
    -> sqlite3BtreeInsert
      -> accessPayload -> copyPayload -> memmove
```

A b-tree insert underneath a read is SQLite materializing a transient b-tree. In every repeat sample
the `sqlite3_step` count equals the `query_notes_count_free` count, so essentially all SQL stepping in
the process is that one query. The vector index path was included in the same search as a control and
is absent.

## Mechanism

`query_notes_count_free` issues, with `deleted_at IS NULL` and an optional `kind`:

```sql
SELECT id, namespace, kind, status, name, content, salience, decay_factor,
       expires_at, properties, created_at, updated_at, deleted_at, key, version
FROM notes
WHERE namespace = ?1 AND deleted_at IS NULL [AND kind = ?2]
ORDER BY created_at DESC, id ASC
LIMIT ? OFFSET ?
```

The indexes available on `notes` are `(namespace)`, `(namespace, kind)` and `(created_at DESC)`. None
of them produces rows in `created_at DESC, id ASC` order under that filter, and none is partial on
`deleted_at IS NULL`. So the ordering has to be established by a sort, and the sorter carries the
selected columns, which include `content` and `properties`. That is both the CPU and the multi-GiB
resident excursion, and it happens before `LIMIT` and `OFFSET` are applied.

The consequence is that listing notes costs time and memory proportional to the size of the namespace
rather than to the page requested. `list` is a common call, so any caller that polls it turns this
into a periodic load.

## What would settle it

`EXPLAIN QUERY PLAN` for that statement against a store with a representative number of notes, before
[...]

PR #2671, opened 2026-09-13
ADR-186: one partial index for the note listing order

ohdearquant:
Record for the schema change behind #2669. Proposed, not implemented: no migration in this PR.

Listing notes establishes `created_at DESC, id ASC` with a transient b-tree because no index on
`notes` produces that order under `namespace = ? AND deleted_at IS NULL [AND kind = ?]`. The sorter
carries `content` and `properties` for the whole matching set, before `LIMIT` and `OFFSET` apply.

The decision is one index, `(namespace, created_at DESC, id ASC)` partial on `deleted_at IS NULL`,
and the ADR carries the measurement that chose it over two. The kind-bearing variant is faster for a
kind-filtered page, 245 VM steps against 1,929, but it cannot serve the unfiltered listing at all, and
`notes` is the table a serving process writes most, so the remaining 1,684 steps are written down as
the number that would reopen the question rather than spent now.

Two things the record is careful about. The rare-kind case is not an improvement and not a regression:
the planner declines the new index there and keeps its small sort, so the acceptance asserts the
unchanged plan rather than a speedup. And the comm recipient and unread indexes lead with the same
namespace column, with pre-ANALYZE selection depending on catalog order, so their existing plan tests
are part of the acceptance rather than an afterthought.

The measurement itself is #2670, which lands the arms as a test with no schema change, so this ADR's
acceptance is a flip of arms that already run rather than new instrumentation written alongside the
fix.


ohdearquant:
Hold before merging this. The implementation was written against the record and it goes red, in a way
the record does not currently account for.

Migration V34 creates exactly what the Decision says, `idx_notes_namespace_created ON notes(namespace,
created_at DESC, id ASC) WHERE deleted_at IS NULL`, declared with the generic note indexes ahead of the
comm block. Three of the six `comm_filter_plan_tests` fail on the same assertion: the unread inbox query
plans as

```
SEARCH notes USING INDEX idx_notes_namespace_created (namespace=?)
```

instead of `idx_notes_unread_probe_recipient_direction`. The three are the fresh-bootstrap arm, the
fresh half of the recreated-unread arm, and the arm whose purpose is that unread work stays bounded as
other mailboxes grow. The pre-analyzed-upgrade arm passes.

So the boundary is statistics rather than catalog order: with `sqlite_stat1` present the planner keeps
the selective comm index, and with no statistics it takes the broader one. The kind-bearing variant was
tried on the chance that sharing the comm indexes' `(namespace, kind)` prefix would keep the cost
comparison honest; it fails identically, with the plan reading `(namespace=? AND kind=?)`.

What the regression costs is the defect this ADR is about, moved to another query: an unread listing
for a recipient with nothing unread would walk the namespace's whole message history newest-first
rather than touching only that recipient's unread inbound rows, which is what the partial index exists
to prevent.

The acceptance in this ADR is what caught it, so the record's own arm did its job. The Decision is what
needs amending: it reads as though the index is safe to add on its own, and the measurement says it is
not without either pinning the comm seeks' plans or guaranteeing statistics. Amendment to follow rather
than a merge of the text as written.

PR #2690, opened 2026-09-13
fix(python): allow tool arguments in operation construction

### LINKED ISSUE #2608: khive-py: op(tool, **args) cannot express the seven verbs that take a `tool` argument
`khive.ops.op` is declared `op(tool: str, **args: Any)`, so the first parameter shares a name with
an argument that seven served verbs take. Any call that names that argument collides in Python
before the request is built:

```python
>>> from khive.ops import op
>>> op("exec.run", tree=t, tool="my.tool", actor=a)
TypeError: op() got multiple values for argument 'tool'
```

The affected verbs, enumerated from the pack vocabularies at `2a17870be` (every `HandlerDef` whose
params carry a `ParamDef` named `tool`, directly or through `P_TOOL`):

| verb | the `tool` argument |
|---|---|
| `exec.run` | required — which registered tool to run |
| `exec.runs` | optional — tool name filter |
| `tool.check` | required |
| `tool.describe` | required |
| `tool.policy` | required — tool name or pattern |
| `tool.request` | required |
| `tool.requests` | optional — restrict to one tool name |

So the exec pack's write verb and most of the tool pack's grant path are exactly what `op` cannot
address, and there is no keyword spelling that works: `op("exec.run", **{"tool": "my.tool"})` is
the same call.

The operation shape itself is fine — the dictionary is two lines to build by hand:

```python
keep_null = {"record", "fence"} if verb == "stream.append" else set()
{"tool": verb, "args": {k: v for k, v in kwargs.items() if v is not None or k in keep_null}}
```

which is what a caller has to fall back to today. What is lost in the fallback is the
`stream.append` null-preservation rule, which lives only inside `op` and is easy to drop silently
because it only shows up when a record or fence is explicitly null.

Suggested fix, either shape works and both keep every non-colliding call source-compatible:

- make the verb positional-only: `def op(tool: str, /, **args: Any)`, which frees the keyword and
  is a one-character change; or
- take the arguments as a mapping: `op(verb, args=None, /, **kwargs)`, merging `args` over
  `kwargs`.

The first is smaller and fixes all seven at once.

ohdearquant:
Closes #2608.

`op(tool, **args)` could not express the seven verbs that take a `tool` argument (`exec.run`, `exec.runs`, `tool.check`, `tool.describe`, `tool.policy`, `tool.request`, `tool.requests`): Python rejected the call with a duplicate-argument error before any operation was built. The verb is now taken positionally, so `tool=` in the keyword arguments reaches the encoded `args` untouched; with no positional verb, `op(tool="stats")` still names the verb as before. Behaviour for every other verb is unchanged, and a `tool=None` filter is still omitted.

Tests: one parametrized case per affected verb asserting the encoded operation, the keyword-verb form, the positional-verb-with-tool-argument form, the missing-verb rejection, and the `None` omission control.


ohdearquant (commented):

Verdict on head ebc721fdbb24b5c6487437c7893572fee6b1e6bf: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

PR #2691, opened 2026-09-13
fix(ci): preserve cache fixture exemptions across commits

### LINKED ISSUE #2344: gitleaks: fingerprint-exact ignores decay when a finding is re-attributed
## Problem

`.gitleaksignore` records finding-specific ignores as exact fingerprints of the form
`commit:path:rule:line`. Every component is positional, so an ignore stops matching whenever
the finding is re-attributed — a rebase, a squash, or an edit that shifts the line. The
adjudicated fixture then reappears as a brand-new finding.

This is not hypothetical, and the file already documents one occurrence: two hex cache-lock
test fixtures in `crates/khive-pack-git/src/cache.rs` were first ignored at lines 1405-1406,
then needed a second pair of entries at lines 1407-1408 after a later change shifted them and
re-attributed the finding to a different commit. Both pairs are still listed, for the same two
values.

A scan over a wider set of refs reproduces it a third time: the same two fixtures match under
another commit that carries those lines, which no existing entry covers. The allowlist is
correct for the exact commits enumerated in it and silently incomplete for every other commit
carrying identical content.

## Why it matters

The failure mode is a false positive on a value that has already been adjudicated as a
non-credential. That is the most expensive kind of noise for a secret scanner, because the
correct response — "we already looked at this, it is a fixture" — is invisible to whoever hits
it next. They either re-derive the adjudication or, worse, learn to wave the gate through.

The maintenance cost also grows monotonically: the file accumulates a new pair of entries per
re-attribution and never sheds the stale ones, so it becomes a log of past line numbers rather
than a statement about which values are acceptable.

## Options

- Move the adjudication from the fingerprint to the value or its site: an allowlist rule
  (`[[rules.allowlist]]` / a path- or regex-scoped stopword) states "this fixture is not a
  credential" once, and keeps holding when the line moves.
- Or keep fingerprints but treat drift as a maintenance signal: fail the scan when a listed
  fingerprint matches nothing, so a stale entry is visible rather than silently carried.

The two are complementary — the first stops the recurrence, the second stops the file rotting.

## Verification

A test that shifts an adjudicated fixture to a different line and asserts the scan still
passes. The current arrangement fails that test by construction, which is the point: the
assertion should be about the value being acceptable, not about where it currently sits.

ohdearquant:
Closes #2344.

The two non-credential hex values in the cache-lock serialization fixture were exempted by exact `commit:path:rule:line` fingerprints in `.gitleaksignore`, so every commit that moved those lines re-attributed the finding and reddened the credential scan again. The exemption now lives in `.gitleaks.toml` as a path- and rule-scoped allowlist (`generic-api-key`, `crates/khive-pack-git/src/cache.rs`, exactly the two values), independent of commit and line number; the stale fingerprints are removed and the two documentation fingerprints stay.

Tests: `scripts/tests/test_gitleaks_config.py` runs the installed scanner against disposable fixtures (never the checkout): the exact values at the exact path are exempt, moved lines and commits stay exempt, and other paths, other values, other rules and provider-shaped credentials are still reported. CI runs it right after the pinned scanner install.


ohdearquant (commented):

PR #2703, opened 2026-09-14
fix(kg): restore refuses a merge tombstone and names the kept id

### LINKED ISSUE #2698: kg restore: a merge-tombstoned entity can be restored as a live row while still carrying merged_into
At 4c8615c1, entity merge (`crates/khive-runtime/src/curation.rs:3256-3265`) tombstones the source row by setting `deleted_at`, `merged_into` and `merge_event_id`. `restore_entity` (`crates/khive-runtime/src/operations.rs:4924-4946`) accepts any caller-owned deleted entity and clears only `deleted_at` and `updated_at`, leaving both merge provenance columns set. After that, `get_entity` (`operations.rs:1760-1772`) takes its live-row fast path and returns the source entity instead of the "was merged into" disclosure, so a merged source can be brought back as a live duplicate of the kept entity while still carrying `merged_into`.

Expected: `restore` refuses a row whose `merged_into` is set, with a specific merge-tombstone error naming the kept id, unless and until a transactional merge rollback exists that restores every preimage and index. Test: merge A into B, restore A, assert the refusal and that `get(A)` still reports the merge.

ohdearquant:
## Summary

`restore` on an entity that `merge` had folded into another entity used to clear `deleted_at` and bring back a second live copy. The row still carried `merged_into`, so `get` disclosed it as merged while `list` and `search` served it as live.

- `restore_entity` now refuses a row whose `merged_into` is set, with a `merge_tombstone` conflict whose details carry the kept id (`merged_into`).
- The tombstone keeps its "was merged into" disclosure on `get`; nothing about `merge` changes.
- New runtime test: merge, then restore the losing side, assert the refusal names the kept id and the tombstone is unchanged.
- API reference: restore section states the refusal.

## Test plan

- [x] `cargo fmt --check`, `cargo clippy` on runtime, kg and kkernel, `cargo test -p khive-runtime -p khive-pack-kg` locally
- [x] `restore_refuses_a_merge_tombstone_and_keeps_the_disclosure` passes; the existing restore and merge tests are unchanged

Closes #2698


ohdearquant (commented):

PR #2722, opened 2026-09-14
memory: create(kind=memory) stores the same form memory.remember stores

### LINKED ISSUE #2684: kg `create` of a `memory` or `task` note skips the defaults the specialised verbs apply, so the same record reads differently by path
**Observed** (2026-09-13). `create(kind=memory)` produced a note whose `get` view has `salience: null`, `decay_factor: null` and no `memory_type`, while `memory.recall` rendered it as episodic with the default salience; `create(kind=task, content=...)` stored the title as the content.

**At source.** The task hook (`khive-pack-gtd/src/hook.rs`, `stored_content_is_title_fallback`) runs on generic create; `memory.remember`'s defaults are applied in the memory pack's own handler and not by a hook, so generic create leaves them unset and the recall path fills them at read time.

**Ask.** Either apply the memory defaults through the same hook path the task kind uses (one canonical stored form), or refuse generic `create` for kinds that own a specialised writer, naming the writer.

ohdearquant:
Closes #2684.

A memory note written through the generic `create(kind=memory)` path stored `salience` and `decay_factor` as null and carried no `memory_type` property; `memory.recall` filled the gaps at read time, so `get` and `recall` disagreed about the same record depending on which verb wrote it.

The memory pack now declares a kind hook for the `memory` note kind. On `prepare_create` it resolves `memory_type` (from `properties.memory_type`, default episodic), `salience` and `decay_factor` through the same resolver `memory.remember` uses, with the same defaults and range checks, and writes the resolved values back into the create arguments. `create` accepts `decay_factor` for notes (refused for entities, validated finite and non-negative) so the resolved value reaches storage; stream batch writes carry it through as well. Namespace selection stays with `memory.remember`, since `create` resolves the namespace before any kind hook runs, and the reference says so.

Tests: the memory integration suite creates one note through `memory.remember` and one through `create` and asserts `salience`, `decay_factor` and `memory_type` match field by field, plus semantic defaults, explicit values kept, and the three refusals. The kg suite stores an explicit `decay_factor` on a plain note and refuses it on an entity, and a help test pins the two new parameter descriptions.


ohdearquant:
Closing: the memory pack's contract for generic memory creation is the explicit refusal that directs callers to memory.remember (ADR-021 amendment in preparation), not normalization on create. Superseded by that lane.

PR #2775, opened 2026-09-15
feat(kg): order search by recency and render updated_at on hits

### LINKED ISSUE #2680: kg search: hits carry `created_at` and score but not `updated_at`, and there is no `order_by`, so 'latest state of X' cannot be asked deterministically
**Observed** (2026-09-13). A note updated to version 10 still ranked by its original `created_at` and its semantic score against unrelated records; "the current handoff" stayed a heuristic.

**At source.** `khive-pack-kg/src/handlers/search.rs` renders hits without `updated_at` and accepts no ordering parameter.

**Ask.** Add `updated_at` and `version` to search hits, and an `order_by` of `score` (default) | `updated_at` | `created_at` with `desc` default for the time orders. Keyed singleton notes (already in the runtime) are the other half of a deterministic "current X" read; this issue is only the search half.

ohdearquant:

Closes #2680.

## The defect

`search` returned relevance order only, and its hits carried neither `updated_at` nor `version`. A caller asking "what changed most recently" had no way to express it, and a caller holding a hit could not tell how old it was without a second `get` per row.

## The change

- `handlers/params.rs`: `SearchParams` accepts `order_by`, so the parameter survives the wire boundary.
- `handlers/search.rs`: `SearchOrder` with `score` (default), `updated_at` and `created_at`; parsed and refused in `ValidatedSearchRequest::from_value`, so an unknown value is an error rather than a silent fallback to relevance. A stable descending sort is applied to both entity and note hits, and both hit shapes render `updated_at`.
- `handler_defs.rs` documents the parameter in the `search` verb help.

Design call worth a reviewer's attention: a time order widens the backend candidate window to the same bounded window the property and tag filters already use. Re-ranking only the top-`limit` scored hits would answer "the most recent of the most relevant few", which is not what the parameter says. The default `score` order keeps the narrow window, so default behaviour and its cost are unchanged.

Entity hits render `"version": null`. There is no version column on `entities` (`crates/khive-db/sql/entities-ddl.sql`), no field on `khive_storage::entity::Entity`, and `update` already refuses `expected_version` for non-notes; the null is row-shape parity, stated in a code comment, rather than an invented number. `updated_at` is real on both substrates.

## Tests (9 added)

Unit: `search_params_accepts_order_by`, `search_params_order_by_absent_is_none`.

Integration: `search_default_order_is_the_unchanged_relevance_order`, `search_order_by_updated_at_returns_the_most_recently_updated_first`, `search_order_by_created_at_returns_the_most_recently_created_first`, `search_entity_order_by_updated_at_returns_the_most_recently_updated_first`, `search_refuses_an_unknown_order_by_value`, `search_note_hits_carry_updated_at_and_version_across_an_update`, `search_entity_hits_carry_updated_at_and_a_null_version`.

The two `assert_ne!` arms confirming that a time order genuinely disagrees with the score order fire and pass, so the fixture is not ordering-degenerate.

## Verification state, stated plainly

`cargo test -p khive-pack-kg` and `cargo fmt --all -- --check` passed, but on a tree that differs from this one by a single `.to_string()` removal in one integration assertion. `cargo clippy -p khive-pack-kg --all-targets -- -D warnings` was **red** on that same assertion (`clippy::cmp_owned` at the comparison) and has not been re-run since the one-line fix. CI runs all three here for the first time.

## Known gap, left out of scope

`crates/khive-mcp/src/server.rs` renders its own entity and note rows for the multi-backend coordinator fan-out path, parsing through the same `ValidatedSearchRequest`. It will accept `order_by` and silently ignore it. Fixing it belongs with that path's own row-shape work rather than riding here.


ohdearquant:
Pushed `c685558` after the shard 2/2 failure at `68b62f8`.

The failure was a single test, `kkernel coordinator::tests::multi_backend_and_direct_search_rows_have_exact_key_set_parity`, and it was a real defect in this change rather than a flake. 5500 of 5501 tests passed; that one asserted that both search routes return the same row keys and got:

```
  left: {"created_at","entity_kind","id","kind","name","score","snippet","source","title"}
 right: {"created_at","entity_kind","id","kind","name","score","snippet","source","title","updated_at"}
```

`left` is the multi-backend route, `right` is the single-backend one. This change added `updated_at` to entity and note rows in the single-backend serializer, but the multi-backend path builds its row shape independently, so the two routes began disagreeing on the key set. A caller would have seen `updated_at` or not depending on how many backends were configured, which is exactly what that test exists to prevent.

The fix carries the field through the coordinator the same way `created_at` already travels: collect `updated_at` beside `created_at` when hydrating each merged hit, add the two maps to the fan-out result, and render the field on both coordinated rows. 32 added lines across three files, no behaviour change on the single-backend path.

The workspace build and clippy were not run locally for this push; the shared build lane on the development machine was occupied. Formatting was checked per file. CI is the gate here.

PR #2777, opened 2026-09-15
fix(mcp): preserve committed stream receipt timestamps

### LINKED ISSUE #2537: Agent presentation compacts result-carried timestamps, so a contract requiring exact values is not met on the default presentation
## What happens

The default Agent presentation rewrites any field named `updated_at` (or `created_at`) that is not nested under an object-valued `properties`, either into a relative string or into a minute-truncated absolute form:

```rust
// crates/khive-runtime/src/presentation.rs
/// - Within the last 24 hours: relative form (e.g. `"3m ago"`, `"2h ago"`).
/// - Older: minute-granularity absolute form `"YYYY-MM-DDTHH:MM"`.
fn compact_timestamp(s: &str, now: i64) -> String {
```

The suppression list beside it is two fields wide:

```rust
const PAYLOAD_TIMESTAMP_FIELDS: &[&str] = &["trigger_at", "due"];
```

So a verb result that carries a timestamp as a *value the caller consumes* rather than as record metadata loses its seconds and its offset before the caller sees it. `trigger_at` and `due` are on the list precisely because that already happened to them once; the list has no other members, and nothing marks a result object as payload the way a stream entry's `record` is marked.

This matters for any contract that specifies an exact timestamp on a result. A write result whose documented shape is ISO 8601 with microseconds does not carry microseconds under the default presentation: the caller reads `3m ago`. The value is intact under `verbose` and `human`, so the contract holds on one presentation and not on the one most callers get.

## Why the obvious fix is wrong

Adding `updated_at` to `PAYLOAD_TIMESTAMP_FIELDS` would suppress compaction on every record in every verb, which is the opposite of what the compaction is for, and existing arms assert the compacted form for record metadata.

## Suggested shape

Mark the *object* as payload rather than the field name globally, the way `is_stream_entry` already exempts a stream entry's `record` from the reduction that surrounds it. A result member is a value the caller round-trips; a record's metadata is not. Either a shape guard on the result envelope, or an explicit marker the handler sets, keeps the two apart.

Whichever is chosen, the arms worth having: the exact value survives on a result member under Agent mode, record metadata still compacts in the same response (control), and `verbose` is unchanged.

ohdearquant:
`stream.append` and `stream.batch` receipts now retain their exact committed timestamps in Agent and Human presentation. The registered verb policy protects only the receipt's specified string fields; unrelated metadata keeps the existing presentation behavior. Canonical chain values and Verbose output retain their existing contract.

Includes ADR-045 Amendment 6 and regressions for append/batch receipts, nested caller data, aliases, local dispatch, output formats, chaining, frame/depth limits, and expiry followed by another writer.

Validation: 16 focused tests passed across types, runtime and MCP, including all seven wire-level receipt tests and the expiry/later-writer regression. Normal commit hooks passed, including workspace/all-targets Clippy, Rust formatting, ADR links and document formatting.

Closes #2537.


ohdearquant:
Reviewed at `dc2f805`. No blocking finding in the code.

What I checked and what convinced me:

**The policy is closed, and the test proves it is closed rather than asserting the two happy cases.** `stream_receipt_policies_are_closed_to_the_registered_writers` runs `stream.batch.other` and `stream.read` through the same match and asserts they fall through to `Standard`. For a name-matched dispatch that decoy is the control that matters, because a prefix or substring mistake would light up on exactly that input and on nothing else.

**Receipts do not leak into descendants.** `ReceiptContext` is a five-state machine and the only transitions that preserve a raw string are `AppendRoot` plus key `created_at`, and `BatchMember` plus key `created_at` or `updated_at`. `BatchRoot` promotes to `BatchResults` only for the `results` key when the value is an array, and everything else resets to `None`. The baseline test then asserts `results[0].details.updated_at` is still relativised to `"3m ago"` while the member's own `updated_at` is preserved, which is the case a looser implementation would get wrong.

**The tests compare against the store, not against the implementation.** The integration oracle dispatches an independent `get(id=...)` in verbose mode and compares receipt strings to the stored record. That is the right shape: an assertion built from the presenter's own output cannot fail when the presenter is wrong.

**The `AlwaysVerbose` short-circuit survives.** Both call sites still force `PresentationMode::Verbose` before the policy is threaded through, so the existing exemptions are untouched.

One process item, not a code objection: this change flips ADR-045 Amendment 6 from `Proposed` to `Accepted` in the same commit as the implementation that depends on it. Whether an amendment to an already-accepted ADR needs its own sign-off before dependent code merges is not mine to decide, so the merge is held pending that answer rather than on anything in the diff. CI at this head is 28 checks, 7 pending, zero failures.

Issue #2346, opened 2026-08-31
db_diagnostics: counters are per-process but the payload carries no process identity (pid / start time)


ohdearquant:
## Problem

The `db_diagnostics` writer/reader counters (`crates/khive-pack-kg/src/handlers/db_diagnostics.rs`, backed by `crates/khive-db/src/diagnostics.rs`) are in-process atomics: they reset to zero whenever the daemon process restarts, and the payload says nothing about which process produced them or how long it has been alive.

That makes any quoted counter unfalsifiable on its own. `writer_acquisition_timeouts = 0` over an assumed multi-day window and `= 0` forty seconds after a crash-restart are the same artifact. The payload already carries build identity, so the reader can tell which *code* produced the numbers — but not which *process lifetime* they cover, and a restart (including a crash-restart, exactly when these counters matter most) silently truncates the window.

## Proposal

Add to the `db_diagnostics` payload:

- `process`: `{ pid, started_at }` (or `uptime_seconds`) — the process that owns the atomics.

With that, the counter's scope travels in the artifact instead of depending on the quoter remembering to check daemon uptime separately. A reader can then state every counter as "N over this process lifetime since `started_at`" mechanically.

## Acceptance

- `db_diagnostics` response includes pid + process start time alongside the existing build identity.
- After a daemon restart, the payload visibly shows a new `started_at`/pid next to the reset counters, so a zeroed counter is self-explaining.

PR #2781, opened 2026-09-15
docs(adr): define diagnostics process and pool identity

ohdearquant:
Define the accepted `db_diagnostics.process` contract for #2346: the serving PID, OS creation time, and a monotonic main-pool generation distinguish counter windows across process restarts and pool reconstruction. Unavailable start times require a nonempty reason naming the platform limitation. Reader and writer acquisition/task counters use the identity triple; checkpoint counters remain process-global.

This ADR amendment precedes the implementation. Formatting, ADR references, and normal documentation commit hooks passed.

Part of #2346.


ohdearquant:
The shape is right — one file, status Proposed, contract only, no implementation riding along. Two points on the text, one of which I think has to be settled before this is accepted.

**The last sentence reopens the hole the amendment exists to close.**

#2346's argument is that `writer_acquisition_timeouts = 0` over an assumed multi-day window and the same `0` forty seconds after a restart are the same artifact, so the counter's scope must travel in the payload. `pid` and `started_at` close that at the process boundary.

Then the amendment says:

> Checkpoint counters are process-global, while reader and writer acquisition/task counters can reset when the main pool is reconstructed within the same process.

Within the same process, `pid` and `started_at` are unchanged by definition. So after a pool reconstruction a reader sees identical process identity beside zeroed counters, and cannot distinguish that from a quiet window. That is exactly the unfalsifiable artifact #2346 describes, relocated from the process boundary to the pool boundary, and the amendment names it without giving the reader a field that detects it.

What would close it: a monotonic pool generation alongside the process identity, incremented whenever the main pool is reconstructed, so any counter that can reset is read against the generation that produced it. A consumer comparing two payloads then distinguishes "no activity" from "counters reset under me" mechanically, which is the property #2346 asks for. Naming which counters are generation-scoped and which are process-scoped is worth stating explicitly too, since the amendment already establishes that they differ.

**Second, smaller.** `started_at` is `null` with `started_at_unavailable_reason` when unavailable. Worth stating that the reason is required whenever the value is null, so that a null with no reason reads as a defect in the producer rather than as an unsupported platform. As written, a producer that fails to populate either field is indistinguishable from one correctly reporting that the platform cannot supply a creation time, and the unavailable path is the one least likely to be exercised.

Neither point is a reason to hold the split; the amendment is correctly separated from its implementation. The generation question is the one I would want answered in the text rather than discovered by whoever first quotes a counter across a pool rebuild.

PR #2782, opened 2026-09-15
fix(tool): validate visible registrations and grant schema readiness

### LINKED ISSUE #2767: tool registration resolves in the owning namespace only, so tool.list and tool.describe disagree and a pre-pin grant reads allow
**At source** (verified at the merge head of #2614).

`registration_snapshot` in `crates/khive-pack-tool/src/pin.rs` resolves a tool
registration with `WHERE namespace = ?1`, bound to the caller's own namespace.
The path it replaced went through the entity read that applies the token's
visible namespaces as `namespace IN (...)`. So a tool registered in a namespace
the caller can see, but does not own, resolved before and is `NotFound` now.

Two consequences.

1. `tool.list` still enumerates over the visible set, so it lists a tool that
   `tool.describe` then reports as unregistered. One surface says the tool
   exists and the next says it does not.

2. The decision path treats that `NotFound` as "no registration". A grant
   written before pinning existed has no registry id and no digest, and the
   active check reads `(None, None, None)` as active, so `tool.check` and
   `tool.request` answer `allow` with source `grant` for a tool that is in fact
   registered and whose definition may have changed. Execution still refuses at
   preflight, which resolves the registration properly, so enforcement stays
   closed; it is the advertised decision that is wrong.

**Ask.** Resolve the registration over the caller's visible namespaces, the
same set the listing and the execution preflight use, rather than the owning
namespace alone.

**Acceptance.**

- A tool registered in namespace `b`, read by a caller in namespace `a` whose
  visible set contains `b`, resolves: `tool.describe` returns it, and
  `tool.check` reports the pinned decision rather than a legacy allow.
- The same caller with `b` absent from its visible set still gets `NotFound`.
- `tool.list` and `tool.describe` agree on membership for both cases, asserted
  in one test so they cannot drift apart again.
- A pre-pin unpinned grant for a registered tool reports the invalidated
  decision on every surface, with the execution preflight refusal as the
  control that enforcement was closed all along.

### LINKED ISSUE #2768: a swallowed pack schema plan failure leaves every grant query selecting columns that do not exist
**At source** (verified at the merge head of #2614).

Pack-auxiliary schema plans are applied at startup. The single-backend server
path calls the non-propagating applier in `crates/khive-mcp/src/server.rs`,
whose body logs a failure and continues:

```rust
if let Err(e) = backend.apply_pack_ddl_statements_with_columns(plan.statements, additions) {
    tracing::warn!(pack = plan.pack, error = %e, "failed to apply pack schema plan at startup (non-fatal)");
}
```

The multi-backend path propagates the same failure and refuses to boot.

That asymmetry was harmless when a plan was `CREATE TABLE IF NOT EXISTS` and a
swallowed failure meant a no-op on an already-correct database. It is not
harmless now. The grant column list selects `registry_id`,
`definition_digest`, `invalidated_by_registry_id` and `invalidated_at`
unconditionally, and the plan is one transaction that rolls back in full, as
the schema test asserts. So on the single-backend path a failed plan leaves the
columns absent and every grant query fails with `no such column: registry_id`
for the lifetime of the process. The authorization surface is down, and the
only evidence is one warning line at boot.

A read-only open has the same shape from a different direction: the applier
returns early for a read-only backend, so a read-only open of a database that
predates the columns keeps the old table and every grant query errors.

**Ask.** Either make the single-backend path propagate as the multi-backend
path does, or make the grant queries degrade against a table without the
columns instead of erroring. Boot must not reach a serving state in which the
authorization surface answers every question with a SQL error.

**Acceptance.**

- A plan forced to fail on the single-backend path produces a refusal that
  names the pack, and the process does not serve grant verbs in that state.
- A read-only open of a database without the columns is decided deliberately:
  either refused at boot with a message naming the missing columns, or served
  with the grant queries degrading, and a test pins whichever is chosen.
- A successful plan on both paths is unchanged, with a before-and-after table
  shape comparison as the control.
- The existing rollback test stays green, so the failure arm is still total.

ohdearquant:

Tool registration and ingestion retain separate registrations and capability links in each namespace, even when another namespace exposes a tool with the same name. Name-based reads and grant pinning prefer the caller's own registration, then use the existing exact-case and recency ordering among visible registrations. A newer foreign registration cannot replace an own-namespace grant pin; a foreign fallback never inherits the retired registration's grant.

MCP startup propagates pack schema failures in single-backend mode and refuses read-only startup when required columns are missing or incompatible. Read-only validation reports all missing and incompatible columns in one actionable error, uses no writer, and preserves underlying database errors. Writable upgrades retain transactional rollback.

Validation: 52 native tests passed across schema diagnostics, the full tool suite, execution preflight, and MCP startup/read-only controls. Regressions include separate same-name registrations, own-namespace pin precedence, aggregated column incompatibilities, and unchanged database error propagation. Normal commit hooks passed, including workspace/all-target Clippy, Rust and Markdown formatting, ADR references, and secret checks.

Closes #2767. Closes #2768.


ohdearquant:
The code is coherent and the scope is clean — 13 files, one commit, every path on the grant/schema subject, and the body describes both issues it closes. One thing to settle before this lands, and it is about the ADR file rather than the code.

The `ADR-017` hunk is a normative change, not a wording repair. Before it, the accepted text said the registry "carries declarations to the pack's assigned backend and preserves read-only startup's writer-free behavior" and was silent on failure. After it, the text commits the system to two behaviours it did not previously state:

- both single- and multi-backend hosts **refuse boot**, naming the owning pack, when its plan fails;
- read-only startup **refuses** missing or incompatible columns, naming the table/column pairs, before exposing handlers.

Those are the behaviours #2768 exists because the system lacked. So the decision is changing here, and it is changing by an in-place rewrite of accepted body text rather than as a dated amendment with a status. The consequence is that a later reader cannot tell the contract moved, or when, or what it said before — which is the one thing an ADR's history is for. Nothing was silently dropped: writer-free read-only startup survives the rewrite, and I checked that specifically.

The repair is the shape already used for ADR-023 in #2781: the two sentences become an `## Amendment (2026-09-15)` section carrying its own status, in its own pull request, and this one drops `docs/adr/ADR-017-pack-standard.md` and keeps the twelve code and design files. That also means a question about the specification cannot park a 31-test fix for two issues.

Worth stating plainly since it cuts the other way: the code itself reads as the right implementation of what those sentences describe, and read-only validation acquiring a reader and naming the missing pairs is a better error than the one it replaces. This is a sequencing point, not a rejection.

PR #2783, opened 2026-09-15
docs(adr): define pack schema readiness at startup

ohdearquant:
Defines a dated ADR-017 amendment requiring single- and multi-backend startup to refuse boot with the owning pack's name when its schema plan fails. Read-only startup remains writer-free, validates declared columns through a reader, and refuses boot on missing or incompatible columns with the owning pack and affected `table.column` pairs.

The amendment defines missing tables/columns and the full compatibility predicate: a nullable column with the declaration's SQL type, no default, no primary key, and no hidden/generated value. Existing ADR text remains intact; the amendment is **Accepted**.

Related: #2768. Implementation: #2782.

Validation: the definitions match the schema validator. Normal documentation commit hooks passed, including ADR references, Markdown formatting, and secret checks. Rust hooks correctly skipped because this change touches only ADR-017.


ohdearquant:
The form is right: a dated section appended, status Proposed, the accepted body left alone, and both the issue and the implementing pull request cited so a reader can follow it in either direction. That is what was missing when these four sentences were an in-place edit.

One substantive question, and it is the kind that is cheap to settle now and expensive after two readers have each assumed a different answer.

The amendment states two refusals that do not obviously have the same blast radius:

> Both single- and multi-backend hosts refuse boot with the owning pack's name when its plan fails.

> Read-only startup ... refuses missing or incompatible columns before exposing handlers.

The first is unambiguous: the host does not come up. The second says "before exposing handlers" without saying whose. #2782's own description says it "validates required columns on read-only backends before exposing **affected verbs**", which is a partial refusal — the offending pack's surface is withheld and the rest of the host serves. Those are different contracts. A reader deciding whether a read-only deployment with one pack's columns missing is a degraded host or a dead one cannot get that from this text, and it is the question an operator will actually ask.

Worth pinning in the same sentence, since it is the other term a later reader will have to guess at: what makes a column *incompatible* as opposed to missing. Declared type, nullability, both, something narrower. The missing case is self-evident and the incompatible case is the one that will be implemented from whatever the first implementer assumed.

Neither point is an objection to the decision. If the intent is partial refusal, saying "the affected pack's verbs" rather than "handlers" is a one-word repair and makes the amendment agree with its own implementation.

Issue #2718, opened 2026-09-14
entity update: carry the ADR-172 Amendment 5 fenced-write and no-op rules once entities have a version (ADR-188)


ohdearquant:
ADR-172 Amendment 5 (#2717) states the rule for notes: an accepted fenced write advances the version by exactly one, and an unfenced update whose normalized patch equals the stored value is a disclosed no-op (`unchanged: true`). ADR-188 says that whatever is decided for the no-op case applies to entities unchanged, so the two substrates stay symmetric.

Entities cannot carry that rule yet: `entities` has no `version` column and `update` refuses `expected_version` for anything but notes (`crates/khive-pack-kg/src/handlers/update.rs`, "expected_version, fence and embed apply only to notes"), because ADR-188 is still proposed. So today an identical entity update has no version to leave unchanged and no fence to mint against, and no test can assert either half of the rule.

Ask, to land with ADR-188's implementation: the entity `update` path takes the same two arms notes now carry, an identical fenced patch advances the entity version by one and never reports `unchanged`, and an identical unfenced patch returns `unchanged: true` with the version and `updated_at` untouched, asserted against one shared definition rather than two literals.

Issue #2673, opened 2026-09-13
kg: `update` has no version fence for entities (`expected_version` is notes-only), so two stale writers to one entity both succeed


ohdearquant:
**Observed** (MCP consumer surface, 2026-09-13). Two writers holding the same observed version of a record both updated it; the second silently replaced the first. For notes `expected_version` exists and refuses the stale write with `version_conflict`; entities have no equivalent.

**At source.** `khive-pack-kg/src/handler_defs.rs:464` declares `expected_version` on `update` as "Notes only". `khive-pack-kg/src/handlers/update.rs:49-52` refuses it for entities: "expected_version, fence and embed apply only to notes".

**Ask.** Extend the fence to entities: a positive `expected_version` on an entity update is checked inside the writer transaction and a stale value fails without mutation with the same `version_conflict` / `expected_version` / `current_version` details notes return. Entities already carry `version` on read, so the contract is symmetric.

**Acceptance.** Stale entity update refused, record unchanged; current version succeeds and bumps; a note arm as the control for the shared refusal shape.

ohdearquant:
Correcting the premise before the work starts, because it changes the shape of the fix.

Entities carry no version today. At the current main:

- `crates/khive-db/sql/entities-ddl.sql` declares `entities` with no `version` column.
- The only `ALTER TABLE entities` in the migration set adds `content_ref`.
- `khive_storage::entity::Entity` has no `version` field.

So this is not a guard that exists and is switched off for one substrate. It is a schema change, then a read-surface change, then a guard change, in that order: a fence has nothing to compare against until the column and the trigger exist and reads return the value.

The design is recorded in ADR-188, which also settles that `merge` and `delete` carry the parameter on the same terms as `update` — a fence on `update` alone leaves the race open to a competing writer that happened to call `merge`.

PR #2785, opened 2026-09-15
docs(adr): clarify entity version fences and no-op semantics

ohdearquant:

Refines Proposed ADR-188 for #2673 and #2718: merge's `expected_version` fences the surviving `into_id`, force cannot bypass the fence, and dry runs check it without mutation. The proposal also defines shared entity/note no-op behavior, exact response-shape compatibility, local upsert revisions, and atomic rollback of stale no-op assertions.

The migration measurement procedure explicitly requires a fixture delivered by the implementation revision. It rejects zero or ambiguous test selection, requires one actually passed test, and requires both empty and requested-population measurement records. This ADR-only PR does not publish the fixture or report measurements. Status remains Proposed; implementation validation, measurements, and acceptance remain pending.

Validation: document formatting, ADR references, and normal documentation commit hooks pass. No runtime implementation changes or issue closures are included.

Refs #2673 and #2718.


ohdearquant:
One file, one commit, status held at Proposed with the measurements explicitly not taken. The migration procedure is the strongest part and I want to say why before the criticism, because it is the thing worth copying: it names the executable fixture and its env var, states that only `run_migrations` is timed, declares 9,285 as a caller-visible lower bound excluding tombstones and other namespaces, says in its own text that neither population is the complete fleet count, and ends with "No measurement has been performed." That is a procedure that cannot be quoted as a result by accident.

Two gaps, both in **Consequences**, and both are caller-visible costs of the no-op rule that the rule itself discloses elsewhere.

**1. `unchanged=true` is a second response-shape change and only the first is listed.** Consequences says "Entity read responses gain a field. Additive for a client that ignores unknown fields, and a change for anything asserting an exact shape." The no-op rule adds `unchanged=true` to canonical and atomic *write* results — a different surface, with the same breakage for anything asserting an exact write-response shape. The one sentence that exists reads as covering the whole shape exposure, which is exactly why the second one is easy to miss.

**2. A no-op can abort an atomic unit, which makes "mutation-free" an incomplete description of its cost.** The record says the writer transaction rechecks the prepared snapshot, so an earlier operation in the unit can invalidate a no-op, and the acceptance list has the arm for it. But an operation defined as mutation-free that can roll back other operations' mutations is a consequence, not just a test case. A caller reading "an identical patch with no `expected_version` is a mutation-free assertion" will reasonably conclude it is safe to include one in an atomic unit for free, and it is not.

**Smaller, and it is a trap rather than a gap.** "Entity tags live in the tags column; note tags live in `properties.tags`. An entity's custom `properties.tags` remains an ordinary ordered property." So on one entity, `tags` is order-insignificant and `properties.tags` is order-significant, and the two are spelled the same. A caller who moves a list between them silently changes whether reordering counts as an identical patch. It is disclosed, but as a subordinate clause in a paragraph about comparison helpers; it would earn its place as its own sentence naming the hazard.

Nothing here touches the decisions. The survivor-only fence in particular is argued correctly — one parameter cannot name two revisions is the whole of it, and `force` bypassing the safety floor but never the fence is the right split.

PR #2789, opened 2026-09-15
fix(brain): name the parameter, shape and an example when auto_feedback results are malformed

### LINKED ISSUE #2772: brain.auto_feedback refuses a bare id array with a raw deserializer error naming an internal struct
**Observed** (2026-09-15, serving build). An agent trying to discharge
retrieval feedback after a `memory.recall` gave up after three attempts and
logged an observation instead, because no shape it tried was accepted and the
refusal did not say what was wrong.

Reproduced here. Passing the ids as strings, which is what a caller reaches for
after reading `results[].id` out of a recall response:

```
brain.auto_feedback(query="...", results=["1e8807ef-...-2287a746f"],
                    target_id="1e8807ef-...-2287a746f",
                    served_by_profile_id="khive-recall-v1", signal="useful")
```

returns:

```
invalid input: invalid type: string "1e8807ef-...-2287a746f", expected struct AutoFeedbackResult
```

The same call with `results=[{"id": "..."}, ...]` succeeds. So the accepted
shape exists and is one character away from the rejected one, but the message
names an internal struct, does not name the field it was reading, does not say
that an object with an `id` key is what it wants, and does not show the
accepted form. A caller that cannot see the source has nowhere to go from it.

This sits on the retrieval feedback path, which is the one that stops being
exercised the moment it is inconvenient, so a refusal that cannot be acted on
costs every later ranking improvement.

**Ask.** Refuse with a message that names the parameter, the accepted element
shape, and one example of it. Keep the refusal itself; only the message
changes.

**Acceptance.**

- An array of bare id strings is refused with a message that contains the
  parameter name, the required element shape and a literal example.
- The message names no Rust type.
- The accepted object form still succeeds, asserted in the same test so the two
  cannot drift apart.
- Other malformed shapes for the same parameter, an array of numbers and an
  object where an array is expected, get the same class of message rather than
  a raw deserializer error.

ohdearquant:

`brain.auto_feedback` refused a `results` array of bare id strings with the deserializer's own
message, which named an internal type and neither the parameter nor the accepted shape. An agent
that had just read `results[].id` out of a recall response had no way to get from that refusal to
the object form one character away, and the retrieval feedback path is the one that stops being
exercised the moment it is inconvenient.

The refusal is unchanged. The message now names the parameter, the required element shape and a
literal example:

```
auto_feedback: `results` must be an array of result objects, each with a string `id`; pass the
whole result set, for example results=[{"id": "1e8807ef"}, {"id": "c3f21b90"}]. Element 0 is a
string.
```

The same message covers an array of numbers, an object where an array belongs, and an object whose
`id` is not a string. The check only improves the message for input the deserializer rejects anyway:
it adds no rule, and anything it admits still passes through the same deserializer.

One test carries all of it, including the accepted object form as its control, so the message and
the contract cannot drift apart. Without that arm the assertions are satisfied by a handler that
refuses everything. A refused call is asserted to write nothing.

Closes #2772.

Scope: 2 files, 1 commit.


ohdearquant:
The `Python client tests` failure on this head is #2794, not this change. The diff here is confined to `crates/khive-pack-brain`, and the failing assertion is `test_observed_id_arm9_socket_second_process_recreation` comparing a database-wide `SELECT COUNT(*) FROM events WHERE kind != 'audit'` across a window a background task can enter. The refusal the arm exists to check was itself correct in that run: `conflict`, `not_committed`, the expected `details`. Every other element of the snapshot tuple matched; the events column differed by one.

PR #2797, opened 2026-09-15
fix(db): preserve contention budgets and diagnostic evidence

### LINKED ISSUE #2346: db_diagnostics: counters are per-process but the payload carries no process identity (pid / start time)

### LINKED ISSUE #2364: WAL-pin diagnostics: align the legacy forecast cadence with housekeeping and update the ADR-023 emission wording
Two loose ends in the WAL-pin operator diagnostics:

1. **Legacy forecast cadence.** In `crates/khive-db/src/walpin.rs` the read-only inspection derives its legacy fallback window from the interval it is handed (`fallback_window_secs = stale_window_from(sweep_interval)`), and the diagnostics runtime hands it the checkpoint interval (default 500 ms) while ordinary housekeeping uses the session sweep interval (5 s). A legacy producer temp can therefore be forecast as reapable after roughly 3 s while housekeeping would not reap it until roughly 15 s. Pass the housekeeping fallback cadence into diagnostics (or resolve it in one place both paths call) and add a regression test with distinct checkpoint and session intervals.

2. **ADR-023 emission wording.** `docs/adr/ADR-023-declarative-pack-format.md` still says WAL-pin cleanup counters are emitted only when the request ran the cleanup enumeration, and that a skipped measurement is omitted rather than serialized as `0`/`false`. The read-only inspection now reports `false`/`0` for a completed pass that ran no mutating cleanup, as ADR-091 Amendment 15 describes. Amend ADR-023 to describe the read-only sidecar measurement and forecast semantics, reserving omission for a skipped or failed pass.

### LINKED ISSUE #2386: db: a failed busy_timeout update must not extend the BEGIN retry budget
In the bounded BEGIN IMMEDIATE retry (merged in #2316), lowering the connection `busy_timeout` to the remaining budget can fail; the code then continues with the previous full timeout for the following attempts, so under persistent contention the serialized writer can stall for up to three full timeout windows instead of the documented one-window bound (`crates/khive-db/src/writer_task.rs`, the retry loop around the `busy_timeout` update).

Treat a failed timeout update as budget exhausted (surface the busy refusal instead of retrying), or retry only when the update succeeded.

Related test gaps in the same file: replacing the retry backoff iterator with an empty iterator leaves all writer-task tests passing, so the second BEGIN attempt, the absorbed counter and the timeout restoration are not pinned; the transient-contention test can release its lock before the writer reaches BEGIN, so an uncontended success satisfies it. Add a test that holds the lock until the first BEGIN refusal is observed and asserts the absorbed counter and the restored timeout.

ohdearquant:

Writer retries and diagnostic snapshots now preserve the limits and lifetimes they report:

- If reducing SQLite's busy timeout fails, stop before another `BEGIN IMMEDIATE` and retain the original busy error. Count only retries actually admitted by the remaining budget. A busy refusal counts as absorbed only after the reduced timeout is installed successfully.
- Forecast stale WAL sidecars with housekeeping's compiled five-second session fallback, so a different checkpoint interval or local sweep override does not change the diagnostic answer.
- Identify diagnostic counter windows with the serving PID, OS process creation time, and main-pool generation. A rebuilt main pool advances the generation even before diagnostics first observes it; shared handles and secondary pools retain the correct lifetime. An unavailable OS start time stays null with an explicit platform reason.

The checkpoint counters remain process-wide; pool reader and writer counters can reset with a new main-pool generation. Documentation explains that distinction and the read-only sidecar forecast.

Validation: all 12 selected native regressions passed, followed by normal commit hooks including workspace/all-target Clippy, Rust and Markdown formatting, ADR links, and secret checks. The selected regressions cover retry refusal and recovery, distinct maintenance cadences, process identity across collectors and a fresh process, and pool reconstruction with reset counters.

The khive-db documentation build also passes with `RUSTDOCFLAGS="-D warnings"` after qualifying the identity type links.

Closes #2386. Closes #2364. Closes #2346.


ohdearquant:
The `Doc build (-D warnings)` failure here is real and reproduces locally: `cargo doc -p khive-db --no-deps` with `RUSTDOCFLAGS=-D warnings` exits 101 with `unresolved link to BuildIdentity` and `unresolved link to ProcessIdentity`.

Both types exist in the same file as the link, so the cause is the link form rather than a missing item. The module doc previously wrote the link with an explicit path and this change replaced it with the shorthand:

```
-//! [`BuildIdentity`](crate::diagnostics::BuildIdentity), so a reading is
+//! Every payload carries [`BuildIdentity`] and [`ProcessIdentity`] for the
```

Restoring the explicit paths on both names is sufficient: with `[`BuildIdentity`](crate::diagnostics::BuildIdentity)` and `[`ProcessIdentity`](crate::diagnostics::ProcessIdentity)` the same command exits 0.

Why the shorthand does not resolve, since it is surprising in a module that defines both types: `pub mod diagnostics;` in `lib.rs` carries an outer `///` doc comment, and with that present the module's links resolve outside module scope. Removing that outer comment makes the bare names resolve and instead makes the file's other explicit-path links report as redundant link targets. That is a diagnosis, not a suggestion — the explicit-path form is the right fix here, because the rest of the module doc already uses it and would otherwise all have to change together.

The rest of the change reads correctly to me. Three notes, none of them blocking:

1. The shared `busy_timeout` budget across `BEGIN IMMEDIATE` retries is the right shape, and the behaviour change where a failed timeout reduction now surfaces the original busy error instead of retrying with the full timeout is the part that actually bounds the window. Worth calling out in the body, since it is a behaviour change rather than a refactor.

2. `record_writer_task_begin_busy_absorbed()` moved from before the `busy_timeout` reduction to after it, so a refusal whose reduction fails is no longer counted as absorbed. That is the correct meaning for the counter's name and it matches the new comment, but it is a second semantic change in the same hunk and a reader of the diff can miss it behind the extraction.

3. `SessionSweepConfig::default().interval` and `TruncateState::default().legacy_walpin_fallback_interval` both resolve to `DEFAULT_SESSION_SWEEP_INTERVAL` today, through two independent paths, and nothing asserts they agree. If one moves off that constant the diagnostics report and the housekeeping pass would disagree about the cadence without any test noticing.

I checked the ADR-091 Amendment 6 citation rather than taking it: the amendment does say the compiled default and not the checkpoint cadence or an environment override, `DEFAULT_SESSION_SWEEP_INTERVAL` is 5 seconds as the amendment states, and Amendment 15 supersedes Amendment 6 only on the sidecar-enumeration clause, leaving the cadence rule current.

PR #2806, opened 2026-09-15
ci: pull requests skip the macOS shard that runs no tests

### LINKED ISSUE #2805: On pull requests, CI (macos-latest, shard 2/2) runs one 48-second scan and holds a runner slot whose median queue is 34.5 minutes
## Measured

Over the last 12 `pull_request` CI runs, per-job queue time versus run time:

```
macos:  n=13  queue median 34.5 min, p90 60.5   run median 0.9 min, p90 19.3
ubuntu: n=16  queue median  0.1 min, p90  0.8   run median 12.4 min, p90 48.6
```

macOS jobs spend two orders of magnitude longer waiting for a runner than ubuntu jobs do. At the
median, 97% of a macOS job's wall time is queue.

Every macOS job in that sample with a run time under 3 minutes is `shard 2/2`, at ~48 seconds,
concluding `success`.

## Why

On a pull request, `CI_SUITE` resolves to `macos-pr` for `macos-latest`
(`.github/workflows/ci.yml`, job `ci`, `env.CI_SUITE`). Every `macos-pr` step is gated
`env.CI_SUITE == 'macos-pr' && matrix.shard == 1`, and every other step is gated
`env.CI_SUITE == 'full'`. So on a pull request, `CI (macos-latest, shard 2/2)` executes exactly one
step: the ungated no-stub placeholder scan.

That is correct in the workflow's own terms — the comment on `shard` says shard 2 "exists only to
carry the second test partition", and the `macos-pr` lane has no second partition to carry. The
cost is that it still claims a macOS runner slot in the scarcest queue in the pipeline, to repeat a
48-second scan that both ubuntu shards and macOS shard 1 already run.

With nine pull requests open, that is nine macOS slots held for no test coverage.

## Proposal

Drop `{os: macos-latest, shard: 2}` from the matrix on `pull_request` events only, by building the
matrix with `fromJSON` on a conditional expression. Push, nightly and manually dispatched runs keep
the full four-job matrix, because those run `CI_SUITE == 'full'` where shard 2 does carry a real
test partition.

Nothing else changes:

- `CI (macos-latest, shard 2/2)` is not a required status check; the required contexts are
  `Secret scan (gitleaks)` and `CI gate`.
- `CI gate` aggregates `needs.ci.result`, which is computed over whichever jobs the matrix
  produced, so a smaller matrix does not weaken it.
- The placeholder scan stays guarded on every lane that compiles pull-request-controlled code:
  both ubuntu shards and macOS shard 1 still run it.

## Acceptance

- A pull request produces three `CI (...)` checks, not four, and `CI gate` still gates on all of
  them.
- A push to the default branch still produces four.
- Re-measure queue time on the following day's pull request runs and report before and after.

ohdearquant:

Closes #2805.

## Measured

Per-job queue time versus run time over the last 12 `pull_request` CI runs:

```
macos:  n=13  queue median 34.5 min, p90 60.5   run median 0.9 min, p90 19.3
ubuntu: n=16  queue median  0.1 min, p90  0.8   run median 12.4 min, p90 48.6
```

At the median, 97% of a macOS job's wall clock is spent waiting for a runner. Every macOS job in
that sample with a run time under three minutes is `shard 2/2`, at about 48 seconds, concluding
success.

## Why that job is empty

On a pull request, `CI_SUITE` resolves to `macos-pr` for `macos-latest`. Every `macos-pr` step is
gated `env.CI_SUITE == 'macos-pr' && matrix.shard == 1`, and every other step is gated
`env.CI_SUITE == 'full'`. So `CI (macos-latest, shard 2/2)` executes exactly one step on a pull
request: the ungated no-stub placeholder scan.

That is correct in the workflow's own terms. The `shard` comment already says shard 2 "exists only
to carry the second test partition", and the `macos-pr` lane has no second partition. The cost is
that the job still claims a macOS runner slot to repeat a 48-second scan that both ubuntu shards
and macOS shard 1 already run.

## The change

The matrix becomes an explicit `include` list, three entries on `pull_request` and four otherwise.
Written as two literal lists rather than an `exclude` expression so each branch names every job it
produces.

Push, nightly and manually dispatched runs are `CI_SUITE == 'full'`, where macOS shard 2 does carry
a real nextest partition, so they are unchanged.

## Why this does not weaken the gate

- `CI (macos-latest, shard 2/2)` is not a required status check. The required contexts on the
  default branch are `Secret scan (gitleaks)` and `CI gate`.
- `CI gate` aggregates `needs.ci.result`, which is computed over whichever jobs the matrix
  produced, so a three-job matrix is gated exactly as a four-job one is.
- The placeholder scan is deliberately ungated because both suites compile pull-request-controlled
  code. It still runs on both ubuntu shards and on macOS shard 1, so no lane that compiles that
  code loses it.
- A repo-wide search finds nothing outside this workflow that names the macOS shard check. The
  other `macos-latest` hits are `release.yml` and `serving-artifact.yml`, which carry their own
  matrices, and a packaging table in ADR-026.

## Verification

- `scripts/tests/test_ci_workflows.py`: 37 passed, 69 subtests passed, with the edit in place.
- That suite was proved to read the edited file rather than a copy: mutating a string it asserts on
  turned exactly one test red, and the file was restored by inverse mutation to its pre-mutation
  SHA-256 before committing.
- Both JSON branches of the expression parse, and the workflow parses as YAML with the `ci` job's
  matrix resolving to the `include` expression.

Scope: 1 file, 1 commit.


ohdearquant:
This pull request tests itself: a `pull_request` run uses the workflow from the head, so its own
run is the acceptance arm.

At `9c3c2708`, the run enumerates three matrix jobs and not four:

```
CI (macos-latest, shard 1/2)
CI (ubuntu-latest, shard 1/2)
CI (ubuntu-latest, shard 2/2)
```

`CI (macos-latest, shard 2/2)` is absent, which is the change. The other half of the acceptance,
that a push to the default branch still enumerates four, can only be read after this merges; it
will be checked on the merge commit's own run.

PR #2813, opened 2026-09-15
fix(email): reuse authenticated SMTP connections

### LINKED ISSUE #2372: email outbox: reuse the SMTP transport across deliveries in one pass
`LettreSmtp::deliver` builds a new `AsyncSmtpTransport` for every message (`crates/khive-channel-email/src/connector/smtp.rs`). With the connection-preamble split in #2333 each delivery now pays a connect + TLS + AUTH handshake plus a NOOP round trip, and the pooled connection is dropped with the transport at the end of the call.

Proposal: keep one transport per credential set for the lifetime of an outbox pass (or across passes until the OAuth token changes), so consecutive deliveries reuse the pooled connection. The preamble check then runs once per pass instead of once per message. Acceptance: a test with a scripted transport counts one connection for N deliveries in one pass; the AUTH-stage classification from #2333 is unchanged.

ohdearquant:

Sequential email deliveries reuse an authenticated SMTP connection while credentials remain unchanged, avoiding repeated connection, TLS, AUTH and NOOP exchanges. Changing OAuth credentials discards the previous connection. Failed or cancelled exchanges discard the in-flight connection so a later retry reconnects through the explicit authentication stage.

Connection setup retains implicit TLS on port 465, mandatory STARTTLS on other ports, and lettre's 60-second network/command timeout. A connection closed while idle produces a retryable transport error; the next attempt reconnects. Authentication and recipient errors retain their separate classifications.

Validation: all 197 email library tests passed, including the seven new connection lifecycle regressions; normal commit hooks passed, including workspace/all-target Clippy, Rust formatting, and secret checks; path-filtered hooks with no applicable files were skipped. Seven scripted in-memory SMTP regressions cover connection reuse, credential changes, AUTH and NOOP rejection, recipient rejection, idle closure, and cancellation.

Closes #2372.


ohdearquant:

Holding this one on a behavioral question rather than on the diff, which is sound in itself.

`EmailChannel::from_env()` is constructed once at daemon start (`crates/khive-mcp/src/serve.rs`),
so the `Mutex<Option<AuthenticatedConnection>>` this adds lives for the whole process. SMTP servers
close idle sessions in minutes. After a quiet stretch the cached connection is dead, the next
delivery reuses it, `send` fails, and the caller gets a `ChannelError`; the session is correctly
dropped and the delivery after that reconnects. On a mailer that sends a handful of notifications
an hour, that is a large share of sends failing.

The previous code could not reach that state. It constructed the transport inside every call, so
each send started from an empty pool and `test_connection` opened a live connection immediately
before the delivery. Paying a handshake per send is the cost this removes, and it was also what
made a stale connection impossible.

`smtp_closed_connection_retries_without_hiding_reconnect_auth_rejection` covers the case and
asserts the failure: it scripts a server close, requires the next delivery to error, and requires
the one after it to reconnect. The name reads as recovery; the assertions say the caller must
retry, and nothing on the send path does.

The stated reason not to reconnect is right and scoped narrower than it reads. "Calling the
connection directly cannot reconnect inside MAIL/RCPT/DATA" and "a new handshake always passes
through the connection/AUTH classifier" are both true, and neither rules out reconnecting *after* a
failed exchange on a connection that was already suspect, through that same classifier.

Suggested shape: record whether the session in hand came from the cache or was authenticated in
this call. On a send failure from a cached session, discard it, run connect and auth and classify
once, and retry the send. On a send failure from a freshly authenticated session, return the error
unchanged, so the retry is bounded at one. An AUTH rejection on the reconnect stays an auth error,
which is the property the current arm protects.

Acceptance in the harness already here: script a session that delivers once then closes, and
require the next delivery to succeed with two connections, two auths and two messages. Keep a
separate case scripting the reconnect's AUTH as 535 and requiring the auth error, so the retry
cannot swallow the classification. Add a must-not arm where a fresh session's send is refused and
no second connection is opened, or the bound is not really one.

Separately, and not a hold: holding the mutex across the exchange serialises concurrent senders
behind one connection. That is the right default at this volume and it is a change from the pooled
transport, so the struct's doc comment should say it rather than leave "sequential" to imply it.

ohdearquant:

Concrete resolution for the hold, because the behavioural question has an answer that is smaller
than it looked and one of the two halves is a naming fix.

### Probe the reused connection, do not retry the failed send

The obvious repair is to retry a failed `send` on a fresh connection. That one is not available
here: `send` drives MAIL, RCPT and DATA, and a transport error after the server has already accepted
DATA is indistinguishable at this layer from one before it. Retrying then delivers the message
twice. A duplicate notification is worse than a late one.

What is available is a probe, because it is idempotent. In the cache-hit arm of
`send_message_with_connect`, before reusing the session:

```rust
Some(session) if session.credentials == credentials => {
    if session.connection.test_connected().await {
        session
    } else {
        // The far side closed an idle session. Drop it and reconnect; the
        // message has not been offered yet, so this costs a handshake and
        // nothing else.
        let mut connection = connect().await?;
        connection.auth(mechanisms, &credentials).await.map_err(smtp_connection_error)?;
        classify_smtp_preamble_status(connection.test_connected().await)?;
        AuthenticatedConnection { credentials, connection }
    }
}
```

The asymmetry between the two `test_connected` calls is deliberate and worth a comment in the code:
on a **fresh** connection a refused NOOP is a Transport error, which is what
`classify_smtp_preamble_status` is for and what `smtp_noop_refusal_never_sends_and_next_attempt_reconnects`
pins. On a **reused** connection the same `false` means the session aged out, which is an expected
state and not an error. Same value, two meanings, separated by which arm read it.

Cost is one NOOP round-trip per delivery on the reuse path, against the connect, TLS and AUTH this
PR exists to avoid. On a mailer sending a handful of notifications an hour that trade is the whole
point of the change.

### The arm named for recovery asserts the loss

`smtp_closed_connection_retries_without_hiding_reconnect_auth_rejection` does not describe what it
runs. Its three deliveries are: success, `ChannelError::Transport`, `ChannelError::Auth`, and its
own counts say `(connections, auth, noop, messages) = (2, 2, 1, 1)`. One message delivered out of
three attempts, and the second was consumed by a dead socket. The test is correct about today's
code; the name is the part that travels, and a reader skimming names sees a covered case.

Two changes:

- rename it to state the loss, e.g. `smtp_closed_connection_costs_the_delivery_that_finds_it`, with
  the `(2, 2, 1, 1)` tuple staying exactly as it is. Keeping this arm matters: it is the control
  that shows the probe is load-bearing.
- add the recovery arm the current name implies: a CLOSE between two deliveries, and
  `counts.messages == 2`. Against the unpatched code it returns 1, which is the mutation that makes
  it worth having.

ohdearquant (commented):

ohdearquant:

Pushed `a154197b`, one commit on top of `fff6e81f`, a fast-forward verified against the remote tip
immediately before the push.

## What it does

The connector cached one authenticated connection and reused it for every later delivery on the same
credentials. An SMTP server closes an idle session on its own schedule, so the first delivery after
that close failed against a socket the connector still believed in. The cached session is now probed
with `test_connected()` before reuse, and a session that does not answer is dropped and replaced by
a fresh connect and AUTH.

The same `false` means two different things in the two arms, and the asymmetry is deliberate: on a
fresh connection a refused NOOP is a real failure and `classify_smtp_preamble_status` turns it into
an error, while on a cached one it means the session aged out, which is an ordinary state. The
comment at the site says so.

Retrying a failed SEND is deliberately not added. `send` drives MAIL, RCPT and DATA, and a transport
error after DATA was accepted cannot be told from one before it, so a retry there can deliver the
message twice. The probe is safe precisely because it runs before any of that.

## Tests

The arm named `smtp_closed_connection_retries_without_hiding_reconnect_auth_rejection` promised a
recovery it never ran — nothing in it closed a connection. It is replaced by two arms that each pin
one thing:

- `smtp_closed_connection_reconnects_and_still_delivers` — the probe's own check.
- `smtp_closed_connection_surfaces_a_refused_reauth_as_auth` — the visibility control: a reconnect
  whose AUTH is refused must still surface as an auth error rather than be swallowed by the retry.

The scripted test server now emits a NOOP reply before every delivery after the first, which is what
the probe consumes, and four existing count tuples move with it.

## Gate, on the host

    cargo fmt --all -- --check                                          rc=0
    cargo clippy -p khive-channel-email --all-targets -- -D warnings    rc=0
    cargo test -p khive-channel-email                                   rc=0   198 passed, 0 failed

## Mutation check

Written down before it ran, then run: no-op the probe so a cached session is reused unconditionally,
and `smtp_closed_connection_reconnects_and_still_delivers` must fail.

    mutated    rc=101   the arm FAILED, at the delivery count
    restored   rc=0     both arms green, file sha256 back to the pre-mutation value exactly

Without the probe the connector reuses the dead session and the delivery count reads 1 instead of 2,
which is the behaviour the arm exists to forbid.

PR #2817, opened 2026-09-15
fix(memory): distinguish skipped recall stages from fast work

### LINKED ISSUE #2387: Slow-recall telemetry collapses skipped stages into zero
The slow-recall warning introduced with the FTS segment maintenance work reports five stage timings as bare integer milliseconds (`RecallStageTimings` in `crates/khive-pack-memory/src/handlers/common.rs`). A stage that was skipped and a stage that completed in under one millisecond both read `0`:

- the exact sqlite-vec fallback sets `fresh_tail_elapsed` to zero;
- the no-model path leaves the embed and ANN timings at their defaults;
- millisecond quantization turns sub-millisecond work into zero.

`docs/api/recall-pipeline.md` promises all five fields carry stage semantics, so a reader cannot tell "not run" from "ran quickly".

Suggested shape: represent stage state explicitly (`Option`, a status enum, or an adjacent `*_attempted` field) and cover the exact sqlite-vec fallback, the no-model path, and a sub-millisecond completed stage as separate tests.

ohdearquant:

Slow recall warnings now pair each stage duration with an `*_attempted` flag. A skipped stage remains `0 ms` for compatibility but reports false; an entered stage that finishes below one millisecond reports true. Sequential widening preserves attempted state while accumulating durations, and parallel model stages retain their maximum duration.

No-model recall reports skipped embedding and ANN; exact sqlite-vec retrieval reports ANN work and skips fresh-tail. Existing failure reasons and degraded-result behavior are preserved. This improves attribution and does not claim a speedup.

Validation: all six focused stage timing, no-model, exact sqlite-vec, and warning regressions passed; normal commit hooks passed, including workspace/all-target Clippy, Rust formatting, and secret checks.

Closes #2387.


ohdearquant:

The red on both shards was one assert, and it was the census rather than the arm:

```
config-ledger-reaching pack tests must use #[serial(config_ledger)]; offenders:
  ["crates/khive-pack-memory/src/handlers/common.rs:
    request_cancellation_tests::no_model_recall_records_fts_without_inventing_vector_stage_work
    (reaches config-ledger seam via `collect_recall_candidates`)"]
```

Linux shard 2/2 exited 100 and the macOS shard exited 101 on the identical panic, so one offender
accounts for both. Pushed the attribute the census asks for, spelled the way the neighbouring arms
in this crate spell it, which needs no import: the census accepts the bare and the fully qualified
path equally, at `serial_attribute_keys`.

Worth naming rather than leaving implicit: the census is right and the arm did reach the seam. The
new call it added, `collect_recall_candidates`, is the reaching one, so this is the census catching
the change rather than a pre-existing gap.

The remaining red is the packs bench shard, which is not a required check and is the separate
compile-budget problem tracked in #2834.

Gate not run locally on this commit — the build lane is occupied — so this rests on the run this
push triggers rather than on anything I verified by executing. It is a one-line attribute, and the
parse was checked.

PR #2818, opened 2026-09-15
fix(knowledge): name the bounded topic candidate count

### LINKED ISSUE #2732: knowledge: topic query total is the candidate-window length, not the corpus count
## Problem

`knowledge.topic` with a `query` reports `total` as the length of the candidate window it hydrated (the effective limit times four, filtered by domain, before the final take), not the number of concepts in the corpus that match. A caller reading `total` to decide whether more work exists, or to size a page walk, gets a number bounded by the request rather than by the corpus. Topic listing without a query already reports the full matching count, so the two totals under the same key mean different things.

The 2026-09-14 ADR-047 amendment (knowledge list and topic limit reports) documents this as the existing behavior and keeps it, so that the limit-report change stays additive.

## Ask

Either a corpus-true `total` for the query branch (a count over the matching concepts independent of the candidate bound), or a distinct field name for the window length so the two meanings stop sharing `total`. A test with a corpus larger than the candidate bound must distinguish the corpus count, the window length and the returned row count.

ohdearquant:

Queried `knowledge.topic` responses now report `candidate_window_count` instead of `total`, making the bounded, hydrated candidate count distinct from both the corpus size and the returned result count. Listing requests without a query keep `total` unchanged, including explicit null queries.

Domain filtering, retrieval limits, and result order stay unchanged. Help and API documentation describe the response-key change. The regressions distinguish a 17-concept corpus, a 12-candidate window, and three returned results, and cover filtered, empty, and zero-limit cases.

Validation: all eleven topic count, listing preservation, query, routing, and missing-hydration regressions passed; normal commit hooks passed, including workspace/all-target Clippy, Rust formatting, and secret checks.

Closes #2732.


ohdearquant:
This is a breaking response-shape change and the description should say so: a queried `knowledge.topic` no longer returns `total` at all, so a client reading that key gets nothing rather than a smaller number.

That is the right call — the field was reporting a bounded candidate window as if it were a corpus count, and keeping it around during a deprecation window would mean keeping the wrong number answering the old question. Worth stating plainly in the release notes rather than leaving a reader to infer it from a renamed key.

Issue #2815, opened 2026-09-15
bench(vamana): the quick profile panics inside Criterion's statistics layer at build/10000


ohdearquant:
The `ann-query` component benchmark exits 101 on every run. Criterion panics inside its own statistics layer after `build/VamanaIndex::build/10000`, at `criterion-0.5.1/src/stats/univariate/sample.rs:31`, whose precondition is a sample slice with more than one element and no `NaN`.

Observed on run [35021154889](https://github.com/ohdearquant/khive/actions/runs/35021154889), job 104557604845: compile exits 0, the Criterion pass exits 101, and the partial estimates produced before the panic are retained and uploaded.

### Why it is invisible today

The component workflow ran `cargo bench ... || rc=$?` and then emitted a warning, so the job stayed green whatever Criterion did. #2336 removes that for the knowledge bench and makes the exit code fatal, which is what surfaced this one. The failure itself is older than that change and is unrelated to it: `ann-query` covers `khive-hnsw`, `khive-vamana`, `khive-quant`, `khive-query` and `khive-request`, none of which #2336's change touches.

### Mechanism

`bench_build` (`crates/khive-vamana/benches/vamana_bench.rs:59`) sets `group.sample_size(10)` and runs `VamanaIndex::build` at n = 1000, 5000 and 10000. Under the `--quick` profile the pass is bounded by time rather than by sample count, so a single iteration slower than the budget yields a one-element sample, and the statistics layer asserts on it.

That reading accounts for the size dependence — the panic follows `build/...​/10000` and not the smaller sizes — but it has not been confirmed by running the profile locally, so treat the mechanism as untested and the observation as measured.

### Ask

Make the quick profile produce a usable sample for this group, and make a sample too small to summarise a stated outcome rather than a panic inside a dependency. Either bounding the sizes the quick profile runs, or an explicit sampling mode and measurement time for the group, would do it; the acceptance is that `cargo bench -p khive-vamana --bench vamana_bench -- --quick` completes with every expected benchmark id and exits 0.

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

The mechanism in the issue body is marked untested, and having now run the profile locally three ways, it does not hold. All three exited 0.

**1. The quick profile as CI runs it.** `cargo bench -p khive-vamana --bench vamana_bench -- --quick`, on a clean target with no prior baseline:

```
[...]

PR #2832, opened 2026-09-16
fix(ci): run the quick bench profile without plots

ohdearquant:

Closes the standing failure in #2815, and the mechanism is now executed rather than reasoned.

### What was happening

The `ann-query` component has read red on every main commit since `9e5f0fdaa`, and the same panic sits
in the logs of the runs before it that reported **success** — the shard swallowed the exit code then:

- red, run [35040601613](https://github.com/ohdearquant/khive/actions/runs/35040601613): `phase=criterion elapsed_seconds=158 exit_code=101`
- green, run [35025968809](https://github.com/ohdearquant/khive/actions/runs/35025968809): `phase=criterion elapsed_seconds=159 exit_code=101`

### The mechanism, from a local reproduction with the backtrace

Running the component's own package list in one `cargo bench` invocation reproduces it. The frame
names the culprit, and it is not the statistics of the benchmark:

```
panicked at criterion-0.5.1/src/stats/univariate/sample.rs:31:
assertion failed: slice.len() > 1 && slice.iter().all(|x| !x.is_nan())
   3: <criterion::plot::plotters_backend::PlottersBackend as criterion::plot::Plotter>::pdf
   4: <criterion::html::Html as criterion::report::Report>::measurement_complete
   5: criterion::analysis::common::<criterion::measurement::WallTime, usize>
```

It is the HTML report's density plot. The quick profile has an early exit for a benchmark whose single
iteration already exceeds the budget: it returns one measured point, printed as the identical
lower/point/upper triple you see beside `build/VamanaIndex::build/10000`. The measurement itself
completes and prints; the report drawn from it is what asserts.

That also explains why a single benchmark run never reproduced it: the panic needs the report path to
run over such a sample, which happens after the measurement has already succeeded.

### The change

`--noplot` on the quick pass. CI measures and records a trend here; it does not publish a report. The
trend recorder reads `{new,base}/estimates.json`, which the run still writes.

### Verified by execution, both directions

Same command, same five packages, same target directory:

- without the flag: exit 101, one panic, the backtrace above;
- with the flag: exit 0, no panic, **237** `estimates.json` files written and **0** plot files.

The second half is the one that matters for the trend ledger, so it is a count and not an assurance.


ohdearquant:

Audit after, discharged against the predicate this body stated: *"the next main run of the component bench workflow is read for a criterion phase reporting exit 101 anywhere."*

**The predicate fired, and it is the wrong instrument.** The next main run of the component bench workflow at this merge commit reports:

```
phase=compile   elapsed_seconds=388 exit_code=0
phase=criterion elapsed_seconds=480 exit_code=101
```

but the 101 is not the statistics-layer assertion this change closed. It is a bench asserting on its own measurement quality:

```
thread 'main' panicked at khive-pack-memory/benches/p95_gate.rs:406:5:
assertion `left == right` failed: [four-model fan-out (beyond gate, informational)]
a memory.ann_warm phase event fired during the timed window
  left: 498  right: 494
```

"Exit 101 anywhere" cannot separate a criterion-internal panic from a bench's own assert, so it answers a broader question than the one it was written for. The narrower predicate — an exit 101 raised **inside criterion's statistics layer**, with no user assert in the trace — is the one that tests this change, and it is not present in that log. The three shards that carried the original defect all report `exit_code=0`.

**What the reachable 101 actually is.** The packs shard's exit code only became load-bearing in #2812; before that a panicking bench left the step green. Two benches were dead behind it. The first, a seed that passed a field the atom params struct denies, is already fixed and its run at `552e30a0f` is green end to end. The second is `p95_gate`, and its own source comment states the condition it is asserting on:

> That call can enqueue a detached rebuild but cannot await it; the event-count assertion below rejects the run if that maintenance persists `memory.ann_warm` events inside the timed window (emission is best-effort).

So the bench is refusing to record a baseline it believes is contaminated. That is the right call about a measurement and the wrong exit status for a shard: a refused baseline is an unrecorded data point, not a broken build, and today it reds the component bench workflow on every push to main. Filed separately.

Two runs carry it so far, at different configuration cases (`four-model fan-out` on main, `three-model fan-out` on a branch head), and the run immediately before this one was green. The correlation with this change is **not** claimed: n is one green against two reds, `--noplot` does not run inside `p95_gate` (which has its own `main`, not the criterion harness), and the mechanism named in the bench's own comment needs no help from this diff to fire on a shared runner.

PR #2835, opened 2026-09-16
fix(ci): pin the build stamp to the commit under test

### LINKED ISSUE #2834: bench: the packs component runs to 90% of its 600s budget and a slow runner times out

The `packs` benchmark component now runs to about 90% of its 600-second criterion budget, and a
slower runner crosses it.

### Measurements

| run | phase=criterion | outcome |
| --- | --- | --- |
| main `a53e0c253` | `elapsed_seconds=541 exit_code=0` | passed with ~10% headroom |
| PR run [35043903533](https://github.com/ohdearquant/khive/actions/runs/35043903533) | `elapsed_seconds=600 exit_code=124` | killed by `timeout` |

### Why it changed

Two of the component's benchmarks used to abort in their own seed, and `cargo bench` stops at the
first binary that panics, so the binaries behind it never ran at all. #2828 fixed both seeds, which
means the component now compiles and runs seven benchmark binaries where it previously stopped at
the second. The budget was set against the shorter pass and was never re-derived.

So this is not a regression in any benchmark; it is the component doing the work it was always
supposed to do, against a bound that predates it.

### Options

- raise the criterion budget for this component specifically, which is honest but leaves the next
  benchmark that lands here to rediscover the limit;
- split the packs component, so each half has its own budget and a timeout names a smaller set;
- give the component list a per-component budget instead of one number shared by every component,
  which is the same work as the first option done once.

The third is the one that survives the next benchmark. The measurement to attach to any of them is
the per-binary time, which the log already prints.

### Acceptance

The `packs` component completes inside its budget on two consecutive runs with its exit code
forwarded, and the run log shows each benchmark binary reached — a pass that ends early because
something aborted is the state this issue exists to keep visible.

ohdearquant:

Closes #2834.

**2 files and 1 commit against main.** Both are workflow files; no crate changes.

`crates/khive-runtime/build.rs` is the workspace's only build script, and it emits

```
println!("cargo:rustc-env=KHIVE_BUILD_TIME={build_time}");
```

where `build_time()` falls back to `Utc::now()` whenever `SOURCE_DATE_EPOCH` is unset. It is unset
everywhere under `.github/`. So every re-run of that script emits a value that has never been seen
before, which dirties `khive-runtime` and every crate above it. A restored cache is then worth far
less than it looks, because the one crate that can never be reused sits near the bottom of the graph.

This sets `SOURCE_DATE_EPOCH` from the timestamp of the commit under test in the two jobs where the
cost was measured. The script already honours the variable; nothing in it changes.

## What it costs today, in the main CI job

Counted in the ubuntu shard-1 log of the CI run on the head this was written against, with the
`Compiling <crate>` lines grouped by the `scripts/ci.sh` phase that emitted them:

| phase | crates compiled | of those, workspace | third-party | `khive-runtime` |
| lockfile | 1 | 1 | 0 | 1 |
| forward-deployed | 329 | 12 | 317 | 3 |
| no-stubs | 2 | 2 | 0 | 2 |
| clippy | 1 | 1 | 0 | 1 |
| tests | 49 | 49 | 0 | 1 |
| tests-doc | 26 | 26 | 0 | 1 |
| channel-email | 38 | 38 | 0 | 1 |
| no-default-features | 1 | 1 | 0 | 1 |
| release | 49 | 49 | 0 | 1 |

Twelve `khive-runtime` compiles, and the third-party column is what separates the causes. A phase
that recompiles because its mode changed — `check` to `clippy`, dev to test to release, a different
feature set — has to rebuild third-party crates under the new mode as well. `forward-deployed` is
exactly that and shows what it looks like: it builds `khive-merge`, which declares its own
`[workspace]` and therefore resolves into its own uncached target directory, and it pays 317
third-party crates for the privilege. Its three are structural and this change does not touch them.

Every other row rebuilt workspace crates and **not one third-party crate**. The cache was otherwise
perfect. The rebuild starts at `khive-runtime` and propagates upward, which is the signature of the
crate's own build script, not of the phase's flags. `tests` is the clearest: 49 workspace crates, 0
third-party, one `khive-runtime`.

## What it costs in the component bench job

That job compiles the bench targets and then runs them, which is two cargo invocations over one
checkout, and the second one is inside a 600-second budget it also has to benchmark in. On the same
head, the packs component reported:

```
phase=compile   elapsed_seconds=505  exit_code=0
phase=criterion elapsed_seconds=600  exit_code=124
```

against a sibling component that finished its criterion phase in 184 seconds. The budget is being
spent on a rebuild the compile phase already paid for.

## Acceptance

Two counts, both read after this lands rather than asserted here:

1. The packs component's criterion phase at the same head, which should stop recompiling and
   complete inside its budget.
2. `Compiling khive-runtime` in one completed ubuntu shard log, against today's **twelve**. The
   three `forward-deployed` compiles are expected to survive, as are the genuine feature and
   profile variations — `channel-email`, `no-default-features` and `release` — since those are
   different builds and are supposed to be.

## What this does not do

It does not change what `KHIVE_BUILD_TIME` means for a local build, which still stamps the wall
clock. The variable is the supported lever for this and the build script was already written to
read it; only CI is pinned.


ohdearquant:

**Negative result. This change does not do what the body claims, and the count is identical.**

The acceptance stated in the body was the `Compiling khive-runtime` count on this pull request's own ubuntu shard, against twelve before. The shard completed. Escapes stripped in the same read, with a control on the same line so a zero cannot come from the ANSI sequences:

```
lines                       11001
Compiling  (any crate)        496      <- control, non-zero
SOURCE_DATE_EPOCH lines        41
Compiling khive-runtime v      12      <- acceptance, was 12
```

The stamp is pinned and reaching the job: `SOURCE_DATE_EPOCH: 1789526757` appears in the environment block of every step. It changed nothing.

Per phase, workspace against third-party, which is the split that separates a mode change from a stale build-script output:

| `ci.sh` phase | workspace | third-party | khive-runtime |
| --- | ---: | ---: | ---: |
| lockfile | 1 | 0 | 1 |
| forward-deployed | 12 | 317 | 3 |
| no-stubs | 2 | 0 | 2 |
| clippy | 1 | 0 | 1 |
| tests | 49 | 0 | 1 |
| tests-doc | 26 | 0 | 1 |
| channel-email | 38 | 0 | 1 |
| no-default-features | 1 | 0 | 1 |
| release | 49 | 0 | 1 |
| **total** | **179** | **317** | **12** |

Within runner noise this is the same table as before the pin, so the compile work is the same work.

## Where the reasoning went wrong

The premise was that `build.rs` emits a never-before-seen `KHIVE_BUILD_TIME` on every run, and that a changed value is what dirties `khive-runtime` and everything above it. Pinning the value does make the emitted output stable. It does not stop the script from **running**, and a build script that re-runs invalidates its crate whether or not its output changed. Nothing in this diff addresses the re-run.

That is also why the nine phases that rebuild workspace crates and not one third-party crate still look exactly the way they did. They were never evidence of a changed stamp; they are evidence of a rebuild that starts at the only crate in the workspace with a build script and propagates upward. Pinning the value and stopping the re-run are different changes, and only the second one is the one the measurement asks for.

## What the next experiment is

`build.rs` registers, as `rerun-if-changed`, `.git/index`, `HEAD`, `packed-refs`, the current symbolic ref, and — via `register_tracked_files` — **every tracked file in the repository**. Any of those moving between phases re-runs the script. `git ls-files`, which the script itself calls, refreshes the index and writes it back when stat data is stale, so the script has a path to invalidating itself.

The decisive read is a no-op `cargo check -p khive-runtime` against a warm target, then `touch .git/index`, then the same command again. If the second one recompiles, the registration is the mechanism and the fix is to the inputs the script declares, not to the value it emits. That runs locally and costs one build, so it does not need CI to answer it.

This PR stays open and stops claiming #2834. A deterministic build stamp is worth having on its own terms — the same commit should produce the same `KHIVE_BUILD_TIME` — but that is a different justification and it is not worth landing under a mechanism the measurement just refused.

ohdearquant:

Closing unmerged. The measurement already said this changes nothing; the follow-up on #2834 says there was no anomaly for it to change.

The per-phase table I published in the comment above, and in this pull request's body, came from a grep for `Compiling ` only. `cargo` also prints `Checking` for metadata-only units, and counting both turns "nine phases rebuild workspace crates and not one third-party crate" into an ordinary cached-dependency pattern: `cargo clippy` runs `clippy-driver` on workspace members only via `RUSTC_WORKSPACE_WRAPPER`, and the dependency artifacts are restored from cache before the first phase. Each `ci.sh` phase runs a different cargo mode, so one `khive-runtime` per phase is one per mode. Corrected table and reasoning are on #2834.

That leaves this diff with no justification I am willing to land it under. Pinning the stamp is defensible on reproducibility grounds — the same commit should produce the same `KHIVE_BUILD_TIME` — but that is a different proposal with a different acceptance, and it should be argued on its own rather than inherited from a mechanism that turned out not to exist.

The branch stays if the reproducibility argument is worth making later.

PR #2839, opened 2026-09-16
feat(gtd): refuse completion while a task's dependencies are unresolved

### LINKED ISSUE #2749: gtd: completing a task with unresolved depends_on succeeds, and done is terminal
`gtd.complete` (and the terminal transitions) succeed on a task whose `depends_on` dependencies are still open. The completion path loads the task, checks the terminal and lifecycle guards, and patches the status; it never reads the task's dependency edges or its derived dependency state (`crates/khive-pack-gtd/src/handlers.rs`, `handle_complete`; no reference to `depends_on` or `dependency_state` on the complete or transition path). Because `done` is terminal, a mistaken completion of a blocked task cannot be undone, and its dependents are released as if the work had happened.

Acceptance:
- `gtd.complete` on a task with unresolved dependencies is refused with a structured `dependency_blocked` error that lists the ids of the open dependencies.
- An explicit override exists only through an argument the verb publishes in its help; no implicit bypass.
- Dependents of the blocked task stay blocked; the refusal leaves the task unchanged (version and status).
- The same rule covers every path that reaches a terminal status (complete and the terminal transitions).

ohdearquant:

Closes #2749.

### The defect

`gtd.complete` and the terminal transitions loaded the task, checked the terminal and lifecycle guards, and patched the status. Neither path read the task's dependency edges or its derived dependency state. Because `done` is terminal, a mistaken completion of a blocked task could not be undone, and its dependents were released as if the work had happened.

### The guard

Completion now runs the dependency diagnostic the pack already computes and refuses with a conflict that names each blocker and its state. The refusal carries `reason=dependency_blocked`, the task id, the target status, the derived dependency state, the blocker ids and count, and the full `blocked_by` payload, so a caller can act on it without a second round trip.

A **cancelled or missing blocker is refused on the same grounds**. That is the decision worth stating, because it reads as strictness and is not: neither a cancelled dependency nor a dangling id establishes that the depended-on work happened, and admitting them would make the guard say "ready" for exactly the case where the dependency record has rotted.

### Where it deliberately does not apply

`ensure_completion_dependencies_ready` returns immediately unless the target is `done`. **Cancelling a blocked task stays legal**, and so does every non-terminal move. The issue asks for "every path that reaches a terminal status", and `cancelled` is terminal — this diff answers that with an argument rather than by extending the guard: cancelling a task whose dependencies are unresolved is the normal way to abandon blocked work, and a guard that refused it would turn a correctness check into a lock with no way out. Two arms hold that line so it cannot be closed by accident.

### The override is published, not implicit

`ignore_dependencies` is a declared parameter on **both** `gtd.complete` and `gtd.transition`, with its own `ParamDef` and description, and both verb descriptions now say that done requires ready dependencies unless it is passed. There is no environment variable, no header, and no status spelling that bypasses the check. Using it is audited.

### Both paths, including the atomic one

`kkernel`'s ADR-099 `--atomic` apply calls the same `prepare_transition` / `prepare_complete` functions, so the flag is threaded through there rather than duplicated, and the guard runs during preparation — before any write. One arm asserts exactly that for both verbs: the atomic prepare refuses and the audit row count is unchanged.

### Arms

Fifteen new arms across the pack tests and `kkernel`. The ones that carry the weight:

- a blocked completion is refused **and the task is unchanged** — status, version, revision and properties all asserted, so a half-applied refusal fails;
- dependents of the blocked task stay blocked, which is the downstream half of the acceptance;
- cancellation of a blocked task stays legal, and non-terminal moves stay legal — the controls that keep the guard from becoming a lock;
- a `done` blocker plus the ready path still completes, so the guard is not simply refusing everything;
- cancelled, missing, and otherwise broken blockers are each refused **without mutation**;
- the override completes and is audited;
- the override is published on both verbs, asserted against the handler definitions rather than against the prose;
- the atomic path refuses before writes, its override is audited, and its readiness is evaluated at preparation time.

### Verification

Local workspace gate on this branch, and the numbers are in the comment below rather than in this body so they carry their command.


ohdearquant:

Workspace gate on this branch, run on the host against the merge of current main into it.

```
fmt rc=0
test rc=0
== END rc=0
```

263 test binaries, 11271 passed, 0 failed, 31 ignored.

The nine arms this change adds, by name, all passing:

```
dependency_completion_blocked_cancellation_stays_legal ... ok
dependency_completion_cancelled_blocker_is_refused_without_mutation ... ok
dependency_completion_done_blocker_and_nonterminal_moves_stay_legal ... ok
dependency_completion_missing_blocker_is_refused_without_mutation ... ok
dependency_completion_other_broken_blockers_are_refused_without_mutation ... ok
dependency_completion_override_is_explicit_and_audited ... ok
dependency_completion_override_is_published_on_both_verbs ... ok
dependency_completion_refusal_preserves_task_and_keeps_dependents_blocked ... ok
dependency_completion_transition_done_refuses_but_cancelled_succeeds ... ok
```

Scope, from the API rather than from the commit list: 9 files changed, +800 -35.

The three that carry the behaviour are the refusal arms paired with a must-stay-legal arm: a refusal
leaves the task untouched and its dependents blocked, cancelling a blocked task still works, and a
non-terminal move is unaffected. Without the legal arms a change that simply refused everything
would satisfy the refusal arms on its own.

PR #2842, opened 2026-09-16
fix(runtime): drop build-script rerun inputs whose resolved path is absent

### LINKED ISSUE #2840: khive-runtime is never fresh: the build script registers a loose-ref path that does not exist when the ref is packed

`khive-runtime` is never fresh in a working checkout. Its build script re-runs on every single
cargo invocation, with no edit of any kind, because it registers a `rerun-if-changed` path that does
not exist. Cargo treats a missing registered file as permanently stale.

## Measured: the controls are identical to the treatments

Six `cargo check -p khive-runtime -v` runs against one warm target, each treatment followed by its
own control. Counting both `Compiling` and `Checking` lines, plus the number of
`build-script-build` invocations:

| run | what changed first | units | khive-runtime units | build script ran |
| --- | --- | ---: | ---: | ---: |
| A | warm-up after a test-mode build | 211 | 1 | 1 |
| B | **nothing** | 1 | 1 | 1 |
| C | touched an unrelated tracked non-Rust file | 1 | 1 | 1 |
| D | **nothing** | 1 | 1 | 1 |
| E | touched `.git/index` | 1 | 1 | 1 |
| F | **nothing** | 1 | 1 | 1 |

The no-edit rows are the finding. B, D and F rebuild exactly as much as C and E do, so no edit is
required and no edit makes it worse. Anything that looked like "this edit costs a rebuild" was this
baseline all along.

## Cargo names the cause

```
prepare_target{package_id=khive-runtime target="build-script-build"}:
  dirty: FsStatusOutdated(StaleItem(MissingFile {
    path: ".../.git/refs/heads/<current-branch>" }))

prepare_target{package_id=khive-runtime target="khive_runtime"}:
  stale: missing ".../.git/refs/heads/<current-branch>"
  dirty: FsStatusOutdated(StaleDepFingerprint { ... })
```

## Why the path is missing

`register_git_inputs` in `crates/khive-runtime/build.rs` collects four inputs — `HEAD`, `index`,
`packed-refs`, and whatever `git symbolic-ref -q HEAD` returns — resolves each with
`git rev-parse --git-path`, and prints `cargo:rerun-if-changed=` for the result **without checking
that it exists**.

`git rev-parse --git-path refs/heads/<branch>` answers with the loose-ref path whether or not the
ref is loose. When the ref is packed there is no such file. In this checkout 621 branch refs are
packed and 3 are loose, which is the ordinary state of a repository that has been gc'd or freshly
cloned, so the missing case is the common case rather than the exotic one.

Note that `packed-refs` is already in the same list. The packed case is therefore already covered by
an entry that does exist, and the loose registration is redundant in exactly the situation where it
is absent.

## Cost

27 of the 49 workspace crates depend on `khive-runtime` directly, and the rest of the graph above
them follows. Every `cargo check`, `cargo clippy`, `cargo test` and `cargo build` in a checkout whose
branch ref is packed pays that rebuild before it starts on whatever changed.

## Not a claim about hosted CI

This is measured on a working checkout only. Whether it fires on a hosted runner depends on whether
that checkout leaves `HEAD` symbolic and its branch ref loose, which differs between the push and
pull-request cases and which I have not measured. The discriminator is one line in a job log: run
cargo twice in a row with no edit between and see whether the second one reports `khive-runtime` as
fresh.

## Fix

Skip an input whose resolved path does not exist. The other three always exist, and `packed-refs`
already carries the packed case, so nothing stops being observed:

```rust
let path = /* resolved as today */;
if !path.exists() {
    continue;
}
println!("cargo:rerun-if-changed={}", path.display());
```

An arm that fails today: with the current branch's ref packed, assert that the build script's
emitted `rerun-if-changed` set contains no path that is absent from disk. It is a pure function of
the repository state and needs no build to check.

ohdearquant:

Closes #2840.

Branch diff against main: 3 files, 1 commit, 3 files changed, 111 insertions(+), 21 deletions(-).

## The defect

`khive-runtime`'s build script registers the current branch's ref file as a
`cargo:rerun-if-changed` input:

    git symbolic-ref -q HEAD          -> refs/heads/<branch>
    git rev-parse --git-path <that>   -> .git/refs/heads/<branch>

`git rev-parse --git-path` answers with the LOOSE ref path whether or not the ref is loose. Once
the refs are packed, which is the ordinary state after a `git gc` and in a fresh clone, that file
does not exist.

**A `rerun-if-changed` path that does not exist is not "unchanged" to cargo. It is permanently
stale.** The unit re-runs its build script on every single invocation, and since a build script that
re-runs invalidates its crate, so does everything above it. 27 of the 49 workspace crates depend on
`khive-runtime` directly and the rest of the graph follows, so every `cargo check`, `clippy`,
`test` and `build` in such a checkout pays that rebuild before it starts on whatever changed.

Cargo names the cause itself, with `CARGO_LOG=cargo::core::compiler::fingerprint=info`, which is
worth more than the reasoning above because it is the tool's own account rather than a derived one.

## The fix

Skip an input whose resolved path is absent. The registration logic moves out of `build.rs` into
`git_rerun_inputs` in `build_info_support.rs`, which both `build.rs` and the test module include
by `#[path]`, so the function the build script runs is the function the tests exercise.

Dropping an absent path costs no signal, and the doc comment says why rather than leaving it to be
rediscovered: `packed-refs` is already one of the registered inputs and it does exist, so a packed
ref that moves still invalidates the stamp through that file. The loose entry is redundant in
exactly the case where it is missing.

## It is a CI fix as well, which this description previously denied

The first version of this section said hosted runs were already safe: pull-request runs check out a
detached HEAD, so the symbolic ref is never registered, and push runs create the loose ref, so the
registered path exists. Both sentences are true about the symbolic ref and the conclusion drawn from
them is wrong, because `packed-refs` is registered unconditionally and is absent on every hosted
runner.

The runner builds its checkout with `git init`, `git remote add origin`, `git config --local gc.auto
0`, a fetch, and `git checkout --progress --force refs/remotes/pull/<n>/merge` — read from a job log,
not assumed. Reproducing exactly that sequence against a local remote leaves `HEAD` and `index`
present and `.git/packed-refs` absent; an ordinary clone has all three, which is the control. Nothing
in that sequence ever writes `packed-refs`, and `gc.auto 0` means nothing packs the refs later.

Cargo's own fingerprint log on such a checkout, with nothing edited between two `cargo check -p
khive-runtime` invocations:

    stale: missing "<repo>/.git/packed-refs"
    dirty: FsStatusOutdated(StaleItem(MissingFile { path: "<repo>/.git/packed-refs" }))

and `khive-runtime` recompiles. Creating an empty `packed-refs` and changing nothing else: the next
invocation recompiles nothing. Removing it again: the recompile returns. Four arms, both directions.

What it costs, from two runs of the `packs` component bench job, each of which runs `cargo bench
--benches --no-run` to completion and then `cargo bench --benches -- --quick --noplot` in the next
second:

| run | compile phase | rebuild inside the criterion phase | criterion phase |
| --- | --- | --- | --- |
| pull request 35114166920 | 491s, exit 0 | 7m35s | killed at its 600s cap, exit 124 |
| push to main 35110403410 | 383s, exit 0 | 5m50s | 502s, exit 0 |

The rebuild set is `khive-runtime` and the twelve crates above it, which is the shape this defect
makes. So the component bench shards have been paying for that build twice in one job, and the
benchmarks have been running in whatever is left of 600 seconds.

## Tests

Two arms, and the second is the reason the first is trustworthy.

- `git_rerun_inputs_drop_the_loose_ref_path_once_the_ref_is_packed` is the arm that fails against
  the unpatched code. It builds a repository with one commit, runs `git pack-refs --all`, then
  **asserts the precondition** that the loose path is now absent rather than assuming the packing
  worked, and asserts that no returned path is absent from disk. It also asserts `packed-refs` is
  still returned, so a "fix" that returns an empty vector cannot pass it.
- `git_rerun_inputs_keep_the_loose_ref_path_while_it_exists` is the must-KEEP control: a fix that
  simply stopped registering the symbolic ref would pass the first arm and fail this one.

Both are pure functions of repository state and need no build.

## Gate

    cargo fmt --all -- --check                              rc=0
    cargo clippy --workspace --all-targets -- -D warnings   rc=0
    cargo test -p khive-runtime                             rc=0

1927 passed, 0 failed, 10 ignored, across 18 test binaries. Both new arms by name:

    test build_info::tests::git_rerun_inputs_drop_the_loose_ref_path_once_the_ref_is_packed ... ok
    test build_info::tests::git_rerun_inputs_keep_the_loose_ref_path_while_it_exists ... ok


ohdearquant:

Mutation check, run after the gate rather than promised in the description.

The guard this change adds was no-opped, the two arms were re-run, and the file was restored by the
inverse edit:

    pre-sha       86a9c30f2872a8a0408bfd1e0602754260b8e5a90d736c6f0e2edbe88ae5502b
    mutation      `if !path.exists() {`  ->  `if false {`   (line 55)
    mutated-sha   c8392d6720e2d79c64b7f0280de22c0ec2373fef45c0cfcfc0a364d564eabd10

Under the mutation, filtered to the two arms:

    test git_rerun_inputs_drop_the_loose_ref_path_once_the_ref_is_packed ... FAILED
    test git_rerun_inputs_keep_the_loose_ref_path_while_it_exists ... ok
    test result: FAILED. 1 passed; 1 failed; 0 ignored; 1743 filtered out

After the inverse edit:

    post-sha      86a9c30f2872a8a0408bfd1e0602754260b8e5a90d736c6f0e2edbe88ae5502b   (equal to pre)
    test git_rerun_inputs_drop_the_loose_ref_path_once_the_ref_is_packed ... ok
    test git_rerun_inputs_keep_the_loose_ref_path_while_it_exists ... ok
    rc=0

Two things are being checked and only one of them is the exit code.

The run asserts the arms **by name**, not that the suite went red. A compile error also exits
non-zero, and a check that reads only the exit code would call a broken build evidence that the fix
is load-bearing. So the requirement was the specific arm FAILED and the control arm stayed green,
and both were counted from the output.

The control staying green under the mutation is what says the mutation hit the guard and not
something shared. And the restore is checked by sha256 equality rather than by rerunning the tests,
because a passing suite does not prove the file came back byte for byte.

ohdearquant:

Correcting this pull request's own description rather than editing it quietly, because the
description becomes the commit message.

It said, under "Not a CI fix": *"This does not change what a hosted run does ... Both hosted
shapes were already safe, for two different reasons."* That is wrong. It reasoned about one
of the four registered inputs, the branch's ref file, and generalised from it. `packed-refs`
is registered unconditionally and is absent on every hosted runner, so the permanently-stale
condition this change fixes is present on CI too — on pull-request runs and on pushes alike.

The section is replaced with the measurement: the runner's own checkout sequence from a job
log, the same sequence reproduced locally showing `HEAD` and `index` present and
`packed-refs` absent against an ordinary clone as the control, cargo's fingerprint log
naming `missing ".git/packed-refs"` as the reason `khive-runtime` recompiles, and the
both-directions arms — create the file and the recompile stops, remove it and the recompile
returns.

It is worth real time on CI. In two `packs` component bench jobs, the criterion phase began
rebuilding `khive-runtime` and the twelve crates above it in the second after a successful
`cargo bench --benches --no-run` finished: 7m35s in one run, which then hit the 600s cap and
exited 124, and 5m50s of a 502s phase in the other.

I do not have the after-arm on a hosted runner, because the bench job only builds the
component whose crates a pull request touches and this branch touches none of them. The
first `packs` run after this merges is that arm, and it is worth reading rather than
assuming.

PR #2843, opened 2026-09-16
fix(knowledge): bound the FTS rarity probe so it cannot spend the stage budget

### LINKED ISSUE #2766: knowledge.suggest: the lexical arm times out on every query against a production-sized corpus (term_frequency phase)
**Observed** (2026-09-15, release build, a corpus of production size).

Every `knowledge.suggest` call returns hits, and every one of them carries
`degraded.lexical_timeout = true`. The lexical arm contributes nothing; the
vector arm carries the whole result. A caller reading only the hits sees a
healthy response, which is why this can run for a long time unnoticed.

The instrumentation names one phase:

```
"degraded": { "lexical_timeout": true },
"lexical_timeout_details": [
  { "phase": "term_frequency", "pass": "full",
    "configured_budget_ms": 2000, "effective_budget_ms": 1999,
    "stage_elapsed_ms": 2002, "operation_elapsed_ms": 73 }
]
```

A single frequency probe costs 73 ms against this corpus, and the stage spends
its entire budget, so roughly 27 probes fit inside it. The term budget is
larger than that, so the stage cannot finish on this corpus at any query.

**Ask.** Make the full pass fit the budget on a production-sized corpus, and
say by how much with a measurement. Not a larger budget: a larger budget makes
every call slower and still dies once the corpus grows past it.

**Acceptance.**

- On a fixture corpus whose per-probe cost is in the same range as the one
  above, a full-pass `knowledge.suggest` completes with `lexical_timeout`
  absent and the lexical arm contributing candidates.
- The per-probe and whole-stage elapsed numbers are reported before and after,
  measured the same way, on the same corpus.
- A corpus deliberately built to exceed the budget still degrades cleanly:
  `lexical_timeout` set, details naming the phase, hits still returned.
- `arm_participation` reflects the lexical arm's real contribution in both
  cases, so the degraded state stays visible to a caller who looks.

ohdearquant:

Closes #2766.

Branch diff against main: 1 file, 1 commit, 304 insertions and 8 deletions.

## What was spending the budget

`rarest_fts_terms_first` fetches no candidates. It issues one bounded `count(*)` per term,
sequentially, and its entire product is an ORDERING of the terms the real fetch then queries.

It ran against the whole lexical stage budget. On a production corpus a five-to-seven term query
could spend all 2000 ms deciding what order to ask in, and the caller then returned an empty
candidate list for a stage that had not yet asked for a single candidate. The search returned
nothing, having done no searching.

Two things are wrong with that and only one of them is the timeout. An optimization must not be able
to consume the budget of the work it optimizes. And when it runs out of time it has to degrade to
the unoptimized path, not to no path.

## The bound

The probe now runs under its own deadline nested inside the stage's.
`scope_request_read_deadline` keeps whichever deadline is earlier and restores the wider one on the
way out, so the probe is bounded and the remainder of the budget stays with the fetch.

The probe budget is derived from whatever stage budget is in force rather than declared as a second
constant. Two constants drift apart, and a test override then moves only one of them.

A quarter, rather than a half or a tenth, and the constant's doc comment carries the argument so a
later change has something to argue against. The probe's value is concentrated in the first few
terms, because querying the rarest term first is what bounds the rowid window the later terms
widen, so it does not need most of the budget to deliver most of its benefit. And the fetch it
precedes is the part that returns nothing useful when it is short of time, so the remainder belongs
to the fetch.

## What expiry now does

The terms are used in the order the candidate list produced them. Every term is queried either way;
the ordering is a hint about which to query first, so the fallback is the unoptimized path rather
than a failure path. That order is not the caller's word order: the candidate list is sorted, so
two spellings of the same query fall back identically.

## Reporting the degradation is the second half, not a free property

The classifier that decides the fetch outcome read a single flag, set by the per-term fetch loop. A
cut-off probe is a second and independent way for the stage to have degraded, so it now sets its own
flag and the classifier reads both.

Without that, a probe that expired before any candidate was fetched returned a clean no-match, which
collapses the two answers a caller most needs to tell apart: a miss, and a stage that ran out of
time. With candidates in hand the outcome is `PartialTimeout` with the phase named as the
term-frequency probe; with none it is `TimedOut`, exactly as a per-term expiry already was.

## Tests

- `rarity_probe_expiry_keeps_the_lexical_arm` fails against the unpatched code, where the atoms are
  empty and the state is `TimedOut`. It asserts all four things that changed: candidates present,
  timeout recorded, state `PartialTimeout`, phase named as the probe.
- `rarity_probe_decides_the_order_and_its_expiry_falls_back_to_arrival` is what makes the first arm
  mean something. Over the same corpus and the same per-term expiry it runs both with the probe
  intact and with it cut off, and asserts that the intact run queries the rare term first whatever
  the caller wrote while the cut-off run follows the candidate-list order. One arm cannot show both
  that the probe decides the order and that its expiry falls back, so both run here, differing only
  in whether the probe was allowed to finish. It runs both spellings of the query, and both falling
  back to the same term is the evidence that the fallback is the candidate order and not the
  caller's.
- The existing lexical-timeout arms are the control on the classifier half. They went red under the
  bound alone, because an expired probe read as a clean miss, and they are green again unmodified.
  What they encode is the contract this change restores, not one it replaces.

The term pair in the ordering arm is chosen rather than incidental. The seeded corpus writes
`discusses topic termN`, so every token matches 50 rows except `term1`, whose FTS prefix family also
covers `term10` through `term19` and therefore matches about 550. A common token has to be paired
with one outside its own prefix family or the two match sets are nested and no assertion can
separate them.

The probe loop's test seam is a separate task-local from the fetch loop's. They are different
boundaries, and a test that wants the probe to expire must not also move the fetch's clock.

## Gate

    cargo fmt --all -- --check                              rc=0
    cargo clippy --workspace --all-targets -- -D warnings   rc=0
    cargo test -p khive-pack-knowledge                      rc=0

Test totals across the package's 11 binaries: 517 passed, 0 failed, 3 ignored.


ohdearquant:

The red check on this head is `Component benches / packs / Criterion quick profile`, exit
124: the criterion phase hit its 600s cap. 7m35s of that phase was cargo rebuilding
`khive-runtime` and the twelve crates above it, starting in the second after the job's own
`cargo bench --benches --no-run` compile phase finished successfully — the build-script
staleness described in #2834 and fixed by #2842. The same rebuild takes 5m50s on a push run,
which is why that one finishes inside the cap and this one does not.

It is not a required check and it is not about this diff, which touches one file under
`khive-pack-knowledge`. Recorded here rather than left for a reader to re-derive.

Issue #2837, opened 2026-09-16
bench: a refused p95 baseline exits 101, so the packs shard reds main on a timing race


ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

The `packs` component bench shard is red on main. The failure is not a broken build and not a regression in the code under measurement: it is `khive-pack-memory/benches/p95_gate.rs` declining to record a baseline it believes was contaminated, and exiting 101 to say so.

```
thread 'main' panicked at khive-pack-memory/benches/p95_gate.rs:406:5:
assertion `left == right` failed: [four-model fan-out (beyond gate, informational)]
a memory.ann_warm phase event fired during the timed window — an ANN graph (re)build
started mid-measurement ... Refusing to record this baseline.
  left: 498  right: 494
```

## The bench is asserting on a condition it documents as outside its control

Immediately above the assert:

> Let any durable-epoch debounce check already due from seeding fire before opening the timed window, then confirm one more clean call. **That call can enqueue a detached rebuild but cannot await it**; the event-count assertion below rejects the run if that maintenance persists `memory.ann_warm` events inside the timed window (emission is best-effort).

So the bench knows a detached rebuild can land inside the window, cannot wait for it, and treats it as grounds to refuse. As a statement about the measurement that is right — a p95 sampled across a graph rebuild is not the warm-path number the gate exists to track. As an exit status it is wrong: a refused baseline is a data point that was not collected, and today it fails the job, fails the workflow, and leaves main red until a run happens to get a quiet window.

## Why it only started showing

The shard's exit code became load-bearing in #2812. Before that a panicking bench left the step green, so this has been able to fire for as long as the guard has existed without ever being visible. Two benches were behind that swallowed code. The other one — a seed passing a field the atom params struct denies — is fixed, and the run at `552e30a0f` is green end to end, which is what makes this one legible now.

## Observations

| run head | criterion phase | outcome |
| --- | --- | --- |
| `552e30a0f` | `elapsed_seconds=479 exit_code=0` | green |
| `49ecf08da` | `elapsed_seconds=480 exit_code=101` | `four-model fan-out (beyond gate, informational)` |
| branch head off `49ecf08da` | `elapsed_seconds=484 exit_code=101` | `three-model fan-out (baseline case M=3 queried)` |

[...]

PR #2845, opened 2026-09-16
bench(memory): let the p95 gate refuse a row without discarding the run

ohdearquant:

Answers questions 1 and 2 of #2837. Question 3 is not addressed.

1 file, 1 commit: `crates/khive-pack-memory/benches/p95_gate.rs`, +117 -40.

## What changes

`GateConfig` gains a `role`, either `Gating` or `Informational`. The two baseline
configurations are gating; the four-model fan-out is informational, which is what the
footer this binary prints has always said about it and what nothing enforced.

`bench_configuration` and `wait_until_ann_warm` return `Result<_, String>` instead of
asserting. Every condition they checked is a statement about the sample just taken, so the
reason travels back to the caller and the row owns it.

The table is printed as it is produced rather than collected and printed after the loop. A
contaminated row prints `n/a` in the three percentile columns, `0` for `n`, and its role,
so the row is present and visibly empty rather than absent.

The exit status is decided once, after every row is on screen, from the roles of whatever
came back contaminated. A contaminated informational row is reported and the run exits 0.
One or more contaminated gating rows print a `GATE FAILED: k of n gating configuration(s)`
line and the process exits **2**.

Two is the point of it. The previous exit code was 101, which is what every Rust panic
exits with, so a caller could not tell "this baseline was refused" from "this binary
crashed" — and those want opposite responses, re-run on a quiet box versus investigate.
Exit 2 is reachable only from this one line.

## What it is measured on

A warm host run of all three configurations, file-backed WAL, warm ANN route, 200 timed
samples each:

| configuration | p50 ms | p95 ms | p99 ms | n |
| one-model (M=1) | 4.930 | 5.634 | 6.155 | 200 |
| three-model fan-out (M=3) | 12.234 | 17.593 | 18.921 | 200 |
| four-model fan-out (M=4, informational) | 22.482 | 25.545 | 26.885 | 200 |

Exit 0, every row printed, no failure line — which exercises none of the new handling, and
is why the two arms below exist. Both were predicted in writing before being run, and both
came back as predicted.

**Only the informational row reports contamination** (the condition replaced with
`config.role == Role::Informational`): exit **0**. Both gating rows measure and print
their numbers (5.062 / 6.142 / 8.630 and 16.068 / 21.630 / 27.799); the M=4 row prints
`[CONTAMINATED, informational]`; the footer says `informational, does not fail this run`;
no `GATE FAILED` line. This is the arm that cannot pass on the code before this change,
where any contaminated row ends the process regardless of which row it was.

**Every row reports contamination** (the condition replaced with `true`): exit **2**. No
row measures, three `contaminated:` lines, two of them marked `gating, fails this run`, and
`GATE FAILED: 2 of 2 gating configuration(s) could not be measured against a clean warm
route`.

The file was restored by inverse edit after each arm and its sha256 checked back to the
pre-mutation value exactly. The clean run above was measured on those same bytes; the only
difference between them and what is committed here is a doc comment moved back above the
struct it describes, confirmed by reconstructing the measured bytes from the committed file
and matching the hash.

## What it does not do

Question 3 asks whether the window can be made quiet — awaiting the enqueued rebuild rather
than checking that it did not happen — and what bounded retries would cost against the
600s criterion budget. Nothing here does that. A contaminated gating row still fails, it
just fails legibly and without taking the rows beside it down. For the cost side of that
question: this binary accounts for 122s of the phase in the job log, and writes no criterion
estimate of its own.

There is also no automated regression check to separate this refusal from. The gate's rule
— at most 1.0ms absolute and at most 5% of the matching baseline — is prose in the footer
that a reader applies by eye against a previous run. "Refused" and "regressed" never shared
an exit status because only one of them has ever had one. The refusal now has a code of its
own; the regression side is still a human reading two tables.


ohdearquant:

A run this morning produced the case this change is about, in a shape #2837 did not
anticipate: the process did not panic, it was killed, and the result is the same.

`packs` component bench shard, criterion phase of run 35114166920. Phase start 15:26:39Z
with `timeout_seconds=600`. The bench profile then compiled for 7m35s inside that budget —
after a separate compile phase that had already spent 491s and exited 0 — leaving 127s for
the benches. Seven of them ran in 18s. `p95_gate` started at 15:34:32Z and measured:

| configuration | p95 ms | finished |
| --- | --- | --- |
| one-model (M=1, gating) | 6.449 | 15:34:53Z |
| three-model fan-out (M=3, gating) | 13.844 | 15:35:40Z |
| four-model fan-out (M=4, informational) | — | killed while seeding |

Seeding the informational configuration took 56.9s on that runner, against 39.4s for M=3.
At 15:36:39Z the phase hit 600s and exited 124.

Both gating rows had been measured and both were lost, this time to a SIGKILL rather than
a panic, for the same reason: the table was printed after the loop finished. With this
change each row is printed as it is produced, so that run would have ended with the two
rows the gate reads on screen and the informational row simply absent. Streaming the rows
is not a cosmetic half of this change — it is the half that survives a process death the
binary cannot catch.

It also puts a number on what the informational row costs where the budget is tight: 56.9s
of seeding, in a phase that had 127s left when the first bench started.

ohdearquant:

This branch's own `packs` component bench job ran green on CI, which is the hosted witness the
host run could not give. Run 35116603726, criterion phase `elapsed_seconds=527 exit_code=0`:

| configuration | p50 ms | p95 ms | p99 ms | n | printed at |
| --- | --- | --- | --- | --- | --- |
| one-model (M=1, gating) | 6.539 | 6.997 | 7.280 | 200 | 15:53:01Z |
| three-model fan-out (M=3, gating) | 14.027 | 14.944 | 15.780 | 200 | 15:54:00Z |
| four-model fan-out (M=4, informational) | 13.279 | 14.511 | 15.150 | 200 | 15:54:59Z |

Two things in that log are worth stating beside the numbers.

The margin was **73 seconds** out of 600, and 359s of the phase was the build-script rebuild
described in #2834 and fixed by #2842. So the difference between this green run and the red one
on another branch an hour earlier is about a 12% swing in runner speed against a budget that is
two thirds spent before the first benchmark starts. That is the shape of a flake, and it is not
a property of what this binary measures.

Each row is printed at the moment it is produced — the timestamps above are nearly a minute
apart. Under the previous shape all three appeared together after the loop, so a phase killed
between 15:53:01Z and 15:54:59Z would have shown none of them. That is the behaviour this
change is for, and this run is what it looks like when the process survives; the red run on the
other branch is what it looks like when it does not.

PR #2851, opened 2026-09-16
fix(kg): let merge dry_run predict the safety floor instead of failing on it

### LINKED ISSUE #2850: merge: a dry run cannot preview a pair the entity safety floor would refuse, and the refusal names a developer-only override

`merge(kind="entity")` runs a cheap safety floor before it merges: the two records must share an entity kind, carry similar names, and not be owned by disjoint projects. Two problems with how that floor meets a caller.

## 1. A dry run of a pair the floor would refuse fails instead of predicting the refusal

`dry_run`'s documented contract is "return the planned summary without mutating records or emitting an event", unconditionally. The floor does not read `dry_run`:

```rust
// crates/khive-pack-kg/src/handlers/merge.rs
if !force {
    validate_entity_merge_floor(&into_entity, &from_entity)
        .map_err(entity_merge_guard_error)?;
}
self.runtime
    .merge_entity_with_reason_and_force(token, into_id, from_id, policy, content_strategy, dry_run, reason, force)
```

So `merge(into_id=A, from_id=B, dry_run=true)` on two dissimilar records returns the conflict error, not a plan. The merges a caller has the most reason to preview are exactly the ones it cannot preview, and the text and the behaviour disagree, which a connector directory reads as a broken parameter.

A dry run is a request for the plan, and when the floor would refuse, the refusal is the plan.

## 2. The refusal's remedy names a parameter the caller reading it should not reach for

```
entity merge refused by name_similarity guard; use force=true only when the caller accepts responsibility
```

`force` exists for a developer integrating this runtime. Handing its name to a model at the moment of failure buys a blind retry with `force=true` instead of a look at the two records, and the sentence never says what the check compared, which is the one fact that would let the caller fix the records instead.

## What would close this

- The floor branches on `dry_run` and reports that it would refuse, with what it compared.
- Whatever field carries that verdict is present on every dry-run response, so a caller reads a value rather than a missing key.
- The refusal sentence names the check and what it compared, gives an action a caller can take, and leaves the override to the operator.

ohdearquant:

`merge(kind="entity")` runs a safety floor before merging: the records must share an entity kind, carry similar names, and not be owned by disjoint projects. Two things about how that floor met a caller.

## A dry run of a pair the floor would refuse now predicts the refusal

`dry_run` promises the plan without mutating records or emitting an event. The floor did not read it:

```rust
if !force {
    validate_entity_merge_floor(&into_entity, &from_entity)
        .map_err(entity_merge_guard_error)?;
}
```

So `dry_run=true` returned the conflict error for exactly the pairs a caller has the most reason to preview, and the parameter's text and its behaviour disagreed. A dry run asks for the plan, and when the floor would refuse, the refusal is the plan.

The floor now branches on `dry_run` and returns

```json
{"dry_run": true, "would_merge": false, "refused_by": "name_similarity",
 "compared": "the records' names", "into_id": "...", "from_id": "...",
 "into_value": "Quantum Annealing", "from_value": "Renaissance Painting",
 "detail": "entity merge refused by the name_similarity check, ..."}
```

Two properties this shape is chosen for:

- **`would_merge` is on every dry-run response**, true when the plan is a merge. The alternative asks a caller to read a missing key as a verdict, which is the reading that fails silently.
- **The preview acquires nothing.** It returns before the runtime call, so no id, lock or row is claimed. A validate-only mode that acquires state is a partial execution wearing a preview's label.

`force` still skips the floor, so a forced dry run plans the merge the floor would have refused. The preview fork is reached by the refusal, not by `dry_run`.

## The refusal says what was compared instead of naming the override

Before:

```
entity merge refused by name_similarity guard; use force=true only when the caller accepts responsibility
```

The remedy handed a caller the name of a parameter meant for a developer integrating this runtime, which buys a blind retry rather than a look at the two records, and it never said what the check compared. Now:

```
entity merge refused by the name_similarity check, which compared the records' names; make the records agree on that check before merging, or ask an operator to authorize an override
```

`details.guard` is unchanged. `details.override`, which carried the parameter name, is replaced by `details.compared`.

## Scope

- The transactional reread inside the merge stays authoritative for a real merge. A dry run whose rows change between the handler's read and that reread still surfaces the conflict error; closing that would mean teaching the merge transaction to return a refusal as a plan.
- The proposal-apply path merges under the legacy same-kind policy and never reaches this floor, so it is unaffected. It is the only other caller of the merge family.
- Note merges have no safety floor; they gain `would_merge` and nothing else.

## Verification

Host run on this branch, one cargo lane, out-of-tree target:

```
fmt rc=0
clippy -p khive-runtime -p khive-pack-kg --all-targets -D warnings   rc=0
cargo test -p khive-pack-kg      289 + 330 + 11 + 8 + 5 + 3 + 1 passed, 0 failed
cargo test -p khive-runtime --lib merge_entity      40 passed, 0 failed
```

Mutation control, predicted before running: no-op the dry-run branch and exactly one arm should redden, because `force` skips the floor in the second arm, the third never reaches it, and the fourth tests the refusal sentence rather than the branch.

```
fixed      16 passed, 0 failed
mutant     15 passed, 1 failed  <- merge_dry_run_previews_the_floor_refusal_for_every_guard
restored   16 passed, 0 failed  (source restored byte-identical, sha256 compared)
```

Four new arms: one covering every guard (the compared-value pair is a string for kind and name and an array for projects, so a single fixture would not have exercised the shape), one for the passing floor, one for the forced dry run, and one asserting the refusal names the check and what it compared and does not name the parameter. That last arm carries a control proving its predicate can see the parameter name in a string that contains it.

6 files, 2 commits. The second commit is `deno fmt` on the table this change adds to the API
reference: the repository's docs gate is a formatter, and a hand-padded markdown table fails it.

Closes #2850


oceanwaves630 (approved):
Approved. All 29 checks at this head concluded, none failed, and both required contexts are green.

PR #2854, opened 2026-09-16
fix(kg): resolve an omitted merge kind from into_id, as the parameter documents

### LINKED ISSUE #2852: merge: an omitted kind is assumed to be entity, though the parameter documents resolving the substrate from into_id

`merge`'s `kind` parameter documents:

> Optional substrate or granular kind hint. Omit to resolve the substrate from `into_id`.

The handler does not do that:

```rust
// crates/khive-pack-kg/src/handlers/merge.rs
let raw_kind = p.kind.as_deref().unwrap_or("entity");
let spec = resolve_kind_spec(raw_kind, registry)?;
```

So an omitted `kind` is not resolved from `into_id`, it is assumed to be `entity`.

## What a caller sees

`merge(into_id=<note>, from_id=<note>, dry_run=true)` refuses with `not found: entity <id>`. Measured against a running server on two freshly created notes, with the control in the same pass:

```
merge(into_id=A, from_id=B, dry_run=true)
  -> not found: entity a93e5721-ecfe-4513-b277-b8176deb1332

merge(into_id=A, from_id=B, kind="note", dry_run=true)
  -> {"dry_run": true, "kept_id": "a93e5721", "removed_id": "54a57096",
      "content_appended": true, "edges_rewired": 0, "properties_merged": 0}
```

Both records exist, the refusal is the default rather than a bad id, and the text names a substrate the caller never chose.

That is the expensive shape of a wrong message: it describes the records as missing, so a caller's next move is to go looking for ids that were correct all along.

## Second, smaller, and created by fixing the first

Once the substrate is resolved from `into_id`, the two ids can disagree in a way they could not before, because previously both were read as entities. A merge whose sides are an entity and a note should say so and name both sides, rather than reporting the second record as missing under the first record's substrate.

## What would close this

- An omitted `kind` resolves the substrate from `into_id`, as documented. An explicit `kind`, including a granular one, keeps its current meaning.
- A cross-substrate pair is refused with a message naming what each id resolved as, and that refusal does not read as a lookup failure.
- The existing diagnostic for a record owned by another pack, which directs the caller to that pack's own verbs, still fires on the inference path.

ohdearquant:

Closes #2852.

## What changes

`merge`'s `kind` parameter documents "omit to resolve the substrate from `into_id`". The handler did not:

```rust
let raw_kind = p.kind.as_deref().unwrap_or("entity");
let spec = resolve_kind_spec(raw_kind, registry)?;
```

An omitted `kind` now resolves the substrate from `into_id` through `infer_kind_from_uuid`, the same helper `delete` uses for the same question, wrapped in the same `diagnose_private_merge` that directs a caller holding a record owned by another pack to that pack's own verbs. An explicit `kind` keeps its meaning exactly, granular hints included.

Measured before the change, against a running server, with the control in the same pass:

```
merge(into_id=A, from_id=B, dry_run=true)                  -> not found: entity a93e5721-…
merge(into_id=A, from_id=B, kind="note", dry_run=true)     -> {"dry_run": true, "kept_id": …,
                                                               "content_appended": true,
                                                               "edges_rewired": 0}
```

Both records existed. The refusal named a substrate the caller never chose, so the caller's next move was to go looking for ids that were correct all along.

## The case this change creates, handled in the same commit

Resolving the substrate from `into_id` makes a disagreement between the two ids reachable without the caller having typed anything, because both were previously read as entities. That pair is now refused with a message naming what each id resolved as:

```
cannot merge across substrates: into_id <id> resolves as entity, from_id <id> resolves as note;
merge joins two records of one substrate, so pass the pair you meant or name the kind explicitly
```

The test asserts the refusal does not read as a lookup failure, and runs the reversed pair as a control so the message cannot be reporting whichever side happens to be second.

## Scope

- No documentation change: the parameter text was already the contract, and the code was the deviation.
- ADR-061's Amendment 1 requires a live private-record resolver claim to direct callers to the owning pack. The inference path is wrapped in that same diagnostic, so that behaviour is preserved. No test covers that path at HEAD, and this change does not add one.
- The proposal-apply path merges under the legacy same-kind policy and never reaches this code.
- All 19 `merge` call sites in the pack were enumerated: none passes an omitted `kind` expecting an entity default, and the whole-pack run confirms it rather than the grep alone.

## What the first attempt got wrong, fixed in the second commit

Two refusals that the call could answer without reading anything changed shape, and both are
restored:

- An id that resolves to no record. The entity default made that a `entity <uuid>` refusal, because
  the lookup failed in the entity substrate. Inference replaced it with `not found: <uuid>`, which
  is a different sentence and one the knowledge pack's own tests pin. Inference now answers only
  when an id resolves to a record; when either side resolves to nothing the historical default
  stands.
- An invalid `strategy` or `content_strategy`. The inference reads records, so running it first
  answered a malformed argument with a record's problem. Both arguments, and the explicit-kind
  lookup, are now parsed before any read.

The added test takes both arms and names the id each arm expects in the message, because the entity
default checks `into_id` first: with a present `into_id` and a missing `from_id`, the refusal names
the present one.

## Verification

Branch diff against `main`: 2 files, 2 commits.

Gate on this branch's tree: `cargo fmt --check` clean; `clippy --all-targets -D warnings` clean;
the whole `khive-pack-kg` pack green at 289 + 329 + 11 + 8 + 5 + 3 + 1 with 0 failed, so no test
anywhere in the pack was relying on the entity default.

The refusal wording is pinned outside this pack, which the first gate did not build:
`cargo test -p khive-pack-knowledge --test integration issue_558` passes 6 of 6, including
`issue_558_merge_missing_and_validation_errors_unchanged`, the arm that names the expected
`entity <uuid>` string.

Mutation, with the expected arms named before the run: restoring the two-line default reddens
exactly `merge_with_kind_omitted_resolves_the_substrate_from_into_id` and
`merge_across_substrates_with_kind_omitted_names_both_sides` (12 passed, 2 failed) and leaves every
pre-existing merge arm green. The inverse restore is byte-identical by sha256 and the suite returns
to 14 passed, 0 failed.


oceanwaves630 (approved):
Verified independently at b3ba69ab.

`infer_kind_from_uuid` returns `specific: None` on every arm, and `resolve_kind_spec("entity")` is `Entity { specific: None }`, so for two entities with `kind` omitted the resolved spec is byte-identical to the previous default — the granular-hint path is unreachable from inference and `ensure_entity_kind` receives the same `None` it always did.

The argument order holds: the explicit-kind lookup and both strategy parses stay ahead of every read, so a malformed argument is still answered without touching a record.

The `NotFound` arm falls back to the entity default before either id is looked up in anger, which is what keeps the historical `into_id`-first refusal order. The non-`NotFound` error arm returns the error rather than swallowing it, and the pair where one side is a non-`NotFound` error and the other is `NotFound` falls back to the entity default — identical to the pre-change path, which never inferred at all.

Two edges with `kind` omitted now reach `merge(kind="edge") is unsupported` instead of a missing-entity refusal. Still a refusal, and a truer one.

Required checks green at this head from a `pull_request` run.

PR #2855, opened 2026-09-16
ci: run the markdown formatter from the committed gate

### LINKED ISSUE #2853: ci: the committed gate has no markdown formatter arm, so a docs-only defect only appears on the hosted side

The hosted docs job runs `deno fmt --check` from the repo root over the set named in
`deno.json` (228 markdown files today). `scripts/ci.sh` has no formatter arm at all: its
lint phase covers `cargo fmt`, SQL, and the ADR linters, and nothing in the script ever
looks at markdown formatting.

So a docs-only defect passes every phase a contributor can run and first appears on the
hosted side. Measured this week: a hand-padded markdown table in `docs/guide/api-reference.md`
passed `scripts/ci.sh lint` locally and failed the hosted docs job, which needed a second
commit to fix.

The two sides also carry two copies of the same command, which is how they drifted in the
first place.

**Ask**

1. A `docs-fmt` phase in `scripts/ci.sh` that runs `deno fmt --check` from the repo root,
   dispatched like every other phase and included in the `all` run.
2. The hosted docs job runs that phase instead of its own copy of the command, so the local
   gate and the hosted gate execute one carrier.
3. A missing `deno` refuses loudly and non-zero. A skip would exit 0 and check nothing, and
   the `deno-tests` phase needs the same binary anyway.

ohdearquant:

Closes #2853.

## What changes

The hosted docs job ran `deno fmt --check` over the set named in `deno.json` (228 markdown files
today). `scripts/ci.sh` had no formatter arm at all: its lint phase covers `cargo fmt`, SQL and the
ADR linters, and nothing in the script ever looked at markdown. So a docs-only defect passed every
phase a contributor can run and first appeared on the hosted side — measured this week, a
hand-padded table in `docs/guide/api-reference.md` that needed a second commit to fix.

- `scripts/ci.sh` gains a `docs-fmt` phase running `deno fmt --check` from the repo root,
  dispatched like every other phase, listed in `--print-phases`, and placed second in the `all`
  order so a three-second failure does not wait behind the cargo phases.
- The hosted docs job runs `scripts/ci.sh docs-fmt` instead of keeping its own copy of the command.
  Two copies of one check is how the local and hosted sides came to disagree.
- A missing `deno` refuses non-zero and names itself. A skip would exit 0 and check nothing, and
  the `deno-tests` phase needs the same binary, so a checkout that cannot run one cannot run the
  other either.

## Verification

Branch diff against `main`: 3 files, 1 commit.

Three measured arms of the phase itself:

```
sh scripts/ci.sh docs-fmt                 -> Checked 229 files, rc=0
# with one deliberately misformatted doc
sh scripts/ci.sh docs-fmt                 -> error: Found 1 not formatted file in 229 files, rc=1
PATH=/usr/bin:/bin sh scripts/ci.sh docs-fmt
                                          -> deno not found on PATH: install deno v2.x …, rc=1
```

Two contract tests in `scripts/tests/test_ci_workflows.py`: one asserts the hosted job runs the
phase and carries no second copy of the command, one exercises the absence arm against a stub
`deno` that records its argv, so the pair shows the refusal comes from the missing binary rather
than from the phase never running, and that the phase checks rather than rewrites.

Both mutations were reconciled arm by arm against a prediction stated first: giving the workflow
back its own `deno fmt --check` reddens only the wiring test; deleting the `command -v` guard
reddens only the absence test (127 instead of 1, with the shell's own "command not found" in place
of the named refusal). Each restore is byte-identical by sha256 and both tests return green.

Every committed test that reads `scripts/ci.sh` was run: `test_ci_workflows.py`,
`test_binary_resolution.py`, `test_writer_census.py`, `test_verify_local_artifact.py`, and the
phase-order self-test in `lint-stub-markers.sh` — which asserts `no-stubs-scan` stays first, and
still does.


oceanwaves630 (approved):
Approved. All 27 checks at this head concluded, none failed, and both required contexts are green.

PR #2859, opened 2026-09-16
request: a dry-run merge declares no write-conflict keys

### LINKED ISSUE #2856: request: a dry-run merge declares write-conflict keys, so two previews in one batch refuse each other

A batch of two `merge(..., dry_run=true)` ops is refused:

```
conflict: writes overlap with another op in this batch
```

Both had to be issued as sequential singleton calls. Neither op writes anything.

**Mechanism**

`write_keys_for_op_pub` (`crates/khive-request/src/conflict.rs`) pushes `entity:<into_id>` and
`entity:<from_id>` for every `merge` op. The builder never reads `dry_run`; a grep for it in that
file returns nothing, so a preview and a real merge declare identical write keys.

`crates/khive-request/docs/api/write-conflicts.md` opens by saying the keys are derived "for
operations that can target the same stored record". A dry-run merge does not target a stored
record: it reads the pair, evaluates the safety floor and the strategy, and returns a prediction.

**Census, so the scope is stated rather than assumed**

The wire surface carrying `dry_run` is exactly two verbs: `merge`
(`crates/khive-pack-kg/src/handlers/params.rs`) and `memory.prune`
(`crates/khive-pack-memory/src/handlers/prune.rs`). `conflict.rs` has no `prune` arm, so `merge`
is the only keyed verb the defect can reach.

**Ask**

A `merge` op whose `dry_run` argument is the literal boolean `true` contributes no write key.
Absent, `false`, non-boolean and dynamic (`$prev`) values keep today's keys, which matches the
rule the doc already states for statically unknowable arguments: unknown means conservative.

The consequence belongs in the doc in the same change: a dry run may then sit in a batch beside a
write to the same record, and its prediction describes the state it read. That is the documented
behaviour of a read beside a parallel write, not a new hazard, but a reader should not have to
derive it.

ohdearquant:

Closes #2856.

Branch diff against main: 2 files, 1 commit.

## What changes

`write_keys_for_op` (`crates/khive-request/src/conflict.rs`) built `entity:<into_id>` and
`entity:<from_id>` for every `merge`, preview or not. A dry-run merge reads the pair and returns a
prediction without writing, so those keys could only ever refuse a sibling operation. Two previews
of the same pair in one batch refused each other, and a preview refused a delete of a record it was
never going to touch.

The builder's own contract, in `docs/api/write-conflicts.md`, says keys exist for operations "that
can target the same stored record". A preview is not one of them, and this makes the code agree with
the sentence that was already there.

## The predicate is a literal `true`, deliberately

Absent, `false`, and a non-boolean value all keep the keys. The rest of this builder decides on what
it can know statically from the argument, and an argument it cannot read as a preview stays
conservative rather than guessing. A fix written as "the `dry_run` key is present" would be wrong in
the direction that drops a real conflict.

## Tests

Four arms, after the existing `merge_from_id_conflicts_with_delete`:

- two dry-run merges of the same pair in one batch do not conflict;
- a dry-run merge does not conflict with a delete of its `from_id`;
- `dry_run: false` keeps both keys;
- a non-boolean `dry_run` keeps both keys. This is the arm that fails a fix keyed on presence
  rather than on the value, which every other arm here would accept.

## Scope

`memory.prune` is the only other verb carrying `dry_run` on the wire, and this builder has no prune
arm at all, so its behaviour is unchanged. The document gains one table row and one paragraph
stating the literal-boolean rule and what it means inside a batch.


oceanwaves630 (approved):
Approved at 55d4773fca871b38139f3e0eba3413d11aef96bb. Required checks complete and green at this head; identity legs verified.

PR #2860, opened 2026-09-16
ci: reject unexpected skipped dependencies in the aggregate gate

### LINKED ISSUE #2807: CI gate accepts a skipped dependency, so a needed job that stops running keeps the required check green
`CI gate` (`.github/workflows/ci.yml`, job `ci-gate`) is one of the two required status checks, and
its aggregation step counts `skipped` as passing:
`select(.value.result != "success" and .value.result != "skipped")`. That is deliberate — several
needed jobs are conditional and a globally-required context that is never reported would wedge most
pull requests — but it means the required check cannot distinguish "this gate ran and was happy"
from "this gate did not run", and the failure direction is toward looking safe. Two cases from one
day make the shape concrete. A pull request head whose only concluded check was
`Enable auto-merge: skipped` reads, to any predicate counting concluded-and-not-failed, exactly like
a head whose whole battery passed. And one level up, if a job in the `needs` list starts skipping
itself — a path filter that stops matching, an `if:` whose expression silently resolves false after
an unrelated edit — `CI gate` goes green with that gate no longer running, and nothing in the
required-check surface reports the change. Worth a deliberate answer from whoever touches this
workflow next: either the aggregation distinguishes jobs that are *expected* to skip from jobs that
merely *did*, by naming the conditionally-run set explicitly rather than accepting any skip, or the
acceptance of skips is documented at the step with the list of jobs it is meant to cover, so a new
`needs` entry does not inherit the exemption by default.

ohdearquant:

CI gate now accepts skipped results only for `automerge-push-guard`, `dependency-review`, and `coverage-ratchet`, the three dependencies with job-level conditions. Every other dependency must succeed; failures and cancellations still fail the gate for every job. New dependencies receive no automatic skip exemption.

The workflow tests execute the aggregation script across all current dependencies, a new dependency, and mixed results. They also check that the named exceptions match the conditional jobs. All three new tests fail against the previous predicate; restoring that predicate reproduces the same 15 failing assertions.

Validation: `uv run pytest scripts/tests/test_ci_workflows.py -q` — 41 passed, 147 subtests passed. Normal commit hooks passed; Rust hooks skipped because no Rust files changed.

Branch diff against main: 2 files, 1 commit.

Closes #2807.


oceanwaves630 (approved):
Approved.

PR #2861, opened 2026-09-16
ci: bound main-push concurrency without replacing queued commits

### LINKED ISSUE #2822: CI: main-push runs accumulate without bound and starve pull-request gates

## What was measured

At 2026-09-15 18:10 ET this repository had 29 workflow runs in a non-completed state. Sixteen of
them were `CI` runs triggered by `push` on `main`, the oldest created at 18:39 the previous hour,
two of them `in_progress` for more than ninety minutes. They compete for the same macOS runners
that every pull request's `CI gate` waits on, and the median macOS start delay measured on this
repository earlier the same day was 34.5 minutes.

Eleven pull requests were open at the time. Every one of them was waiting.

## The main-push half is deliberate

`.github/workflows/ci.yml` (concurrency block):

```yaml
group: ci-${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

The comment above it states the intent: a concurrency group holds at most one pending run, so a
single group for `main` would cancel intermediate `main` runs, and keying non-pull-request runs on
`github.run_id` gives each its own group so nothing on `main` queues behind or displaces anything.

That is a real trade and the reasoning is sound. What is missing is a bound. One group per push
means the number of live `main` runs equals the number of merges in the window, and a merge-heavy
evening produces sixteen of them against a runner pool that cannot absorb one.

## The pull-request half looks unintended

Two branches each had a `CI` run for a superseded head sitting in `queued` state next to the run for
their current head, in the same concurrency group, under `cancel-in-progress: true`:

| branch | superseded head | current head |
|---|---|---|
| `khive/namespace-move-mover` | `ab5e9d59` | `4fd99953` |
| `khive/adr122-auth-classification` | `873798cb` | `ea0a33a8` |

A run that has been admitted by the concurrency gate but has not yet been given a runner appears to
sit in a state neither rule reaches: `cancel-in-progress` does not apply because it is not in
progress, and the one-pending-slot rule does not apply because it is no longer pending. The REST API
distinguishes the two states directly — cancelling a run that has not reached `queued` returns

```
HTTP 409: Cannot cancel a workflow run that has not been queued yet.
```

while the same call on a `queued` run succeeds. So under runner starvation, pull-request runs
accumulate per branch as well, which is the case the group key is written assuming cannot happen.

## Why it matters

A required check is read from the newest run at the head commit. Every older run in the queue is
spending a scarce runner producing a verdict no merge decision will read. The queue looks like
capacity pressure and is partly garbage, which makes the obvious remedy (wait, or buy runners) the
wrong one.

## Options

1. Bound the `main` push group by time or count rather than by `run_id` — for example key it on
   `github.run_id` only for the first N minutes, or key on the calendar hour so at most one
   intermediate run per hour survives. Keeps most of the coverage the current comment protects.
2. Restrict the expensive part of the matrix on `push` to the macOS leg only when the diff touches
   platform-sensitive paths, so a merge-heavy window costs Linux minutes rather than macOS slots.
3. Add a scheduled sweep that cancels `push` runs which are not at the branch tip and
   `pull_request` runs whose head is not the current head of an open pull request. This is a
   remedy rather than a fix, but it is small and it is the one that works today.

Cancelling 18 such runs by hand at 18:12 was reversible and immediately freed the pool; the cost is
that the intermediate `main` coverage the concurrency comment was written to preserve is gone for
that window, which is exactly the trade option 1 would let the workflow make deliberately.

ohdearquant:

Main pushes currently get independent concurrency groups and can fill the runner queue. They now share one admitted run and up to 100 waiting runs using GitHub's `queue: max`, which preserves older waiting commits. Pull requests retain their ref-based cancellation; integration pushes, nightly runs and manual runs retain independent groups. [GitHub concurrency semantics](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency).

If the queue fills, GitHub cancels additional arrivals: those landed commits need a rerun to obtain a completed CI gate. Waiting also delays post-merge results; this bounds concurrent demand without reducing the work needed to test every commit. The workflow comments state both limits and distinguish a REST run's `queued` status from concurrency-pending state.

Validation: `uv run pytest scripts/tests/test_ci_workflows.py -q` passed with 39 tests and 79 subtests. The new regression failed on the original workflow and again when the fix was removed; restoring the exact edit restored its SHA-256 and passing results. GitHub's workflow parser accepted the syntax, its expression evaluator passed nine event cases, and normal commit hooks passed.

Scope against main: 2 files, 1 commit.

Closes #2822.


oceanwaves630 (approved):
Approved at 1f6fc2ac9f7a51d83750cb5c16f43bc729b7bf71. Required checks complete and green at this head; identity legs verified.

Issue #2793, opened 2026-09-15
A pooled reader held one of 8 admission slots for 64.7s while admission was saturated, and the diagnostics maximum names no read


ohdearquant:
## Measurement

A live `db_diagnostics` reading on a busy store, taken under real multi-seat load (26 concurrent
processes holding the database open, host 1-minute load average 14.7), over a 4h35m daemon lifetime:

```
reader_admission_capacity            8
peak_active_pooled_reader_checkouts  8      <- admission saturated
pooled_reader_checkouts          36514
reader_checkout_timeouts             0
max_completed_reader_hold_micros  64748317  <- 64.75 seconds
standalone_reader_opens            176
```

Writer-side contention over the same window was clean: `writer_acquisition_timeouts` 0 across 8749
writer acquisitions, with every BEGIN-busy, request-failure and side-effects-unknown counter at 0.

## The problem

One pooled reader held its slot for **64.75 seconds**, against a pool of 8. For the duration of a
hold like that, the store serves concurrent readers from 7 slots, and two of them would leave 6.
Nothing was refused in this window — `reader_checkout_timeouts` is 0 — so today this is a latency
and headroom fact rather than an outage. But saturation and a minute-long hold both happened, and
the only reason they did not compound is that they did not overlap enough times.

The diagnostics surface cannot say which read it was, and that is the second half of the issue: the
maximum is reported as a bare number with no attribution. An operator who sees 64.75s has no next
step from the response alone.

## Ask

1. Attribute the maximum: report which label or verb held the longest completed checkout beside the
   number, the same way the writer counters are already split by class. A single unlabelled maximum
   cannot be acted on, and it is the field most likely to be the first sign of a real stall.
2. Report the hold-time distribution, not only the maximum — a p50/p95 beside it distinguishes one
   pathological read from a pool that is generally held too long.
3. Then decide whether a long-running read belongs in the pooled path at all. A read that can run
   for a minute is a different resource class from one that runs in milliseconds, and sharing an
   8-slot pool between them is what turns one slow read into everyone's latency.

## Acceptance

A fixture with a deliberately slow read and several fast ones, asserting the reported maximum is
attributed to the slow one by name, and that the distribution shows the fast reads separately. A
test asserting only that the maximum is non-zero passes today and distinguishes nothing.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant._

[...]

PR #2862, opened 2026-09-16
db: name the operation that held the longest completed reader checkout

ohdearquant:

Part of #2793. It answers ask 1 and leaves asks 2 and 3 open, so the issue stays open.

Branch diff against main: 3 files, 1 commit.

## What changes

`max_completed_reader_hold_micros` was a bare number. The live reading that opened #2793 was 64.75
seconds against an 8-slot pool, and an operator reading it has no next step from the response alone,
because the field cannot say which read it was.

Every typed-store read resolves its checkout through `ConnectionPool::resolve_reader_checkout`,
which already holds the operation name in order to build a timeout error from it. The guard now
carries that name, and it is reported as `max_completed_hold_operation` beside the micros figure, in
`ReaderAcquisitionSnapshot`, in `ReaderContentionDiagnostics`, and in the `db_diagnostics` payload.

## One lock, not two atomics

The maximum and the name of what held it are one reading. `max_completed_hold_micros` was an
`AtomicU64` updated with `fetch_max`; it is now a small struct behind a mutex, written and read as a
unit. Kept apart, two concurrent completions can leave a duration from one checkout beside a label
from another, which is worse than no label: it is a wrong answer shaped like a measurement.

## `None` is an answer, not a gap

A checkout that never passed the resolve route carries no operation name and reports `None`. The
pool-internal and raw-SQL callers are that case. The one such route with a stable identity, the
explicit SQL read transaction, labels itself rather than going unattributed.

## Tests

- `the_longest_completed_hold_names_the_operation_that_held_it`: three fast checkouts and one slow
  one, all through the resolve route. The fast ones are the control, because a test asserting only
  that some label is present passes on any of the four. It asserts `completed_pooled_checkouts == 4`
  first, so the attribution is not read off a population of one, and it asserts no duration at all:
  the hold figure is diagnostic evidence and never a timing gate.
- `a_checkout_taken_outside_the_resolve_route_reports_no_operation`: the unlabelled route must
  report no operation rather than borrow the name of whatever ran nearby.
- The `db_diagnostics` wire-contract test asserts the new key is present and is a string or null,
  separately from the numeric-field loop beside it, because it is the one value in that block that
  is not a number.

## What is deliberately not here

Ask 2, the hold-time distribution, and ask 3, whether a long-running read belongs in the pooled path
at all. The distribution is a follow-up on the same fields; ask 3 is a design decision that wants the
distribution in hand first.

## Verification on the host, one process, out-of-tree build directory, under the machine lock

- `cargo fmt --all --check` rc=0
- `cargo clippy -p khive-db --all-targets -- -D warnings` rc=0
- `cargo test -p khive-db` rc=0: 1075 passed, 0 failed, 2 ignored in the crate's own suite, and 0 failed
  across its other targets
- `cargo test -p khive-pack-kg --test integration db_diagnostics` rc=0: 1 passed, 0 failed

The label is load-bearing, shown by a mutation rather than by assertion. With
`label_operation` changed to record nothing, the two arms separate exactly as predicted before the
run:

```
the_longest_completed_hold_names_the_operation_that_held_it   FAILED   (pool.rs:5488)
a_checkout_taken_outside_the_resolve_route_reports_no_operation  ok
test result: FAILED. 1 passed; 1 failed; 1075 filtered out
```

The second arm staying green is the point of stating the prediction first: it asserts `None`, which
a build that never labels anything also produces, so it cannot distinguish the fix and is not
evidence for it. The restore was an inverse copy verified byte-identical, and both arms return to
`2 passed; 0 failed`.


oceanwaves630 (approved):
Approved at b74f2ac2d1abbaf8e5bebd711f1a3fcf2b9b5a2a. Required checks complete and green at this head; identity legs verified.

PR #2880, opened 2026-09-16
tool: resolve the grant decision in SQL and match patterns by bytes

### LINKED ISSUE #2596: tool.check consults only the newest 1000 policy rows, so an old deny silently stops being enforced
## Summary

`tool.check` resolves a decision from at most the newest 1,000 policy rows in the namespace. Older rows are not consulted, are not reported, and produce no marker. On an authorization surface that direction is fail-open: an old `deny` silently stops being enforced.

`decide()` (`crates/khive-pack-tool/src/policy.rs`) fetches its policy set with

```rust
let policies = list_policies(rt, ns, None, 1000).await?;
```

and `list_policies` is

```sql
SELECT ... FROM tool_policy WHERE namespace = ?1 ORDER BY created_at DESC LIMIT ?2
```

so the 1,001st-newest row and everything behind it is invisible to every decision.

## Why the cap is reachable rather than theoretical

`tool_policy` is append-only. `insert_policy` mints a fresh UUID on every call, there is no upsert, and the pack exposes no delete: `tool.revoke` takes a grant id, and the base `delete` verb does not resolve a policy id. So the only way to change a policy is to append another row, which means the row count grows monotonically with every correction, and the cap is reached by using the surface as designed.

The grant side has the same shape with the opposite sign: `decide()` reads `list_grants(.., Some("granted"), .., 500)`, so an old active grant past the newest 500 stops being honoured. That fails closed, which is the safer direction and still wrong.

## Reproduce

1. Write a `deny` policy for an actor/tool pair, and confirm `tool.check` answers `deny` with `source: "policy"`.
2. Append more than 1,000 policy rows in the same namespace.
3. `tool.check` for the original pair now answers from the default, with no indication that a matching `deny` exists in the table.

## Expected

A decision surface either consults every row that could match, or it says it could not. Two shapes that both work:

- resolve the decision in SQL with the match predicate pushed down, so the cap bounds the returned rows and not the population considered;
- keep a cap and make saturation explicit in the response, the way the comm unread count reports `saturated` rather than a silently truncated number.

The first is preferable here: a policy decision is not a count, and a caller has nothing useful to do with "the answer may be wrong".

## Related

The correction mechanics that drive the row count are #2597.

ohdearquant:

Closes #2596.

## Change

The policy half of this issue was fixed by resolving the deciding policy in SQL. The grant half was left capped: `decide()` read the newest 500 `granted` rows and tested each one in Rust, so an active grant with 500 newer grants in front of it stopped being honoured and the check fell through to policy or the default. That fails closed, and it is still a wrong answer, since the caller holds an approval the decision cannot see.

`select_active_grant` moves the whole activity test and both pattern matches into one statement that returns at most one row:

- status `granted`, unexpired at the decision time, not invalidated by a registration;
- a registration pin that is absent on both the row and the tool, or names the tool's current registry id and definition digest (a half pin matches neither arm, as before);
- the actor and tool patterns.

The order is unchanged: newest request first. `id ASC` is added as a final key here and in `list_grants`, so a timestamp tie resolves the same way every time and a check cites the grant `tool.requests` lists first. Before this change `decide()` read that listing, so the two agreed by construction; they are now separate statements, and the shared key keeps them agreeing. `GrantRow::is_active` is removed, since the statement is now the only expression of that test.

While writing the shared predicate, a second defect in the policy statement turned up. SQLite's `LIKE` and `length()` read a TEXT value only up to its first NUL, so the statement read a stored `a\0b*` as `a`: not a prefix pattern, and no match for `a\0bcd`. A `deny` written on such a pattern stopped applying, while the Rust matcher it replaced did apply it. `pattern_match_sql` and the specificity rank now compare `CAST(.. AS BLOB)`, which compares the same bytes `pattern_matches` compares. A lone `*` gets its own arm, because `substr` of a zero-length blob is NULL rather than an empty blob, so the prefix arm alone would not let `*` match an empty value.

```
tool.policy(actor="agent:a", tool="a\0b*", decision="deny")
tool.check(tool="a\0bcd", actor="agent:a")
  before: decision ask, source default
  after:  decision deny, source policy
```

## Files

- `crates/khive-pack-tool/src/policy.rs`: `pattern_match_sql`, byte-wise rank in `select_deciding_policy`, `select_active_grant`, `decide()` calling it, and `id ASC` in `list_grants`.
- `crates/khive-pack-tool/tests/verbs.rs`:
  - `an_active_grant_still_allows_after_five_hundred_later_grants`: one grant, then 520 later grants for other actors; the check still cites the first grant.
  - `of_two_active_grants_the_newer_request_is_cited`: a wildcard-actor grant and an exact-actor grant for one pair; five checks all cite the newer request.
  - `a_nul_bearing_prefix_pattern_matches_by_bytes_for_policies_and_grants`: a `deny` on `a\0b*` decides `a\0bcd`; an exact `allow` on `a\0bcd` then outranks it; a grant on `a\0b*` allows `a\0bcd`; `a\0xcd` matches neither.
- `crates/khive-pack-tool/tests/grant_pins.rs`:
  - `an_active_grant_survives_five_hundred_newer_rows`: with and without a registration, an old grant still decides behind 550 newer unrelated grants and 550 newer expired copies of itself, and stops deciding once revoked.
  - `grant_matching_keeps_literal_case_sensitive_byte_prefixes`: exact, prefix and `*` patterns on both columns, with case, `%` and `_`, a non-trailing `*`, a non-ASCII prefix and NUL-bearing patterns.
  - `grants_keep_newest_request_precedence_and_namespace`: the newest active grant decides; expired and invalidated grants drop out; a revoked grant never decides; another namespace's grant decides only there.
  - `equal_request_times_keep_the_existing_grant_selection`: with two grants requested at the same time, the check cites the grant `tool.requests` lists first.

## Verification

All arms ran at `8997d12b8` in one process, in the order below.

```
commit rc=0        (the pre-commit hook runs cargo clippy)
cargo fmt --all --check                          rc=0
cargo clippy -p khive-pack-tool --all-targets    rc=0   -D warnings
cargo test -p khive-pack-tool --test verbs       16 passed 0 failed
cargo test -p khive-pack-tool --test grant_pins  20 passed 0 failed
cargo test -p khive-pack-tool (lib + remaining)  rc=0
```

Five mutations, each predicted in writing before it ran, applied to the source and reverted with a sha256 check (`restore sha256 match: yes`, working tree clean) after each:

| mutation | predicted | observed |
|---|---|---|
| M1 restore the 500-row grant cap | `an_active_grant_survives_five_hundred_newer_rows` and `an_active_grant_still_allows_after_five_hundred_later_grants` fail | `an_active_grant_survives_five_hundred_newer_rows` FAILED (left `"policy"`, right `"grant"`), 19 passed |
| M2 reverse the grant order | `grants_keep_newest_request_precedence_and_namespace` and `of_two_active_grants_the_newer_request_is_cited` fail; `equal_request_times_keep_the_existing_grant_selection` stays green | `grants_keep_newest_request_precedence_and_namespace` FAILED, `equal_request_times...` green, 19 passed |
| M3 match patterns as TEXT again | `grant_matching_keeps_literal_case_sensitive_byte_prefixes` and `a_nul_bearing_prefix_pattern_matches_by_bytes_for_policies_and_grants` fail | `grant_matching_keeps_literal_case_sensitive_byte_prefixes` FAILED, 19 passed |
| M4 rank specificity as TEXT again | only `a_nul_bearing_prefix_pattern_matches_by_bytes_for_policies_and_grants` fails, at the ranked check | grant_pins 20 passed, then `a_nul_bearing...` FAILED (left `"deny"`, right `"allow"`), 15 passed |
| M5 remove the lone-star arm | only `grant_matching_keeps_literal_case_sensitive_byte_prefixes` fails, at the `*`-against-empty arm | `grant_matching_keeps_literal_case_sensitive_byte_prefixes` FAILED (left `false`, right `true`), 19 passed |

Stated precisely, because the runs are narrower than the predictions: the mutation command runs both test binaries in one invocation without `--no-fail-fast`, so it stops at the first binary that fails. In M1, M2, M3 and M5 that was `grant_pins`, so the `verbs` half of each prediction was never executed and is unobserved rather than refuted. M4 is the one arm where `grant_pins` passed, so both binaries ran and the full prediction was observed. A re-run with `--no-fail-fast` follows and the result is posted here.

After the last mutation the tree was restored and re-run green: `grant_pins` 20 passed, `verbs` 16 passed.

The branch is 5 commits behind `main` at gate time. `git merge-tree --write-tree` against `origin/main` is clean, and those 5 commits touch no file this pull request touches, so the local result stands for the merge as well. The hosted checks test the merge ref regardless.


ohdearquant:

The completing mutation run promised in the description has finished. The body discloses that the first run used `cargo test -p <pkg> --test verbs --test grant_pins` with no `--no-fail-fast`, so cargo stopped at the first failing binary and four of the five arms never executed their second half. This run adds the flag, so both binaries run in every arm.

Predictions are unchanged from the ones written before the first run. Each mutation is applied alone, the source is restored and re-hashed between arms (`restore sha256 match: yes`, tree clean after each).

| mutation | predicted | observed, both binaries |
| --- | --- | --- |
| M1 — grant cap restored | exactly `an_active_grant_still_allows_after_five_hundred_later_grants` and `an_active_grant_survives_five_hundred_newer_rows` | both FAILED, one in each binary; 19+1 and 15+1 |
| M2 — selection order reversed | exactly `of_two_active_grants_the_newer_request_is_cited` and `grants_keep_newest_request_precedence_and_namespace`, with `equal_request_times_keep_the_existing_grant_selection` staying green | both FAILED; the tie-break arm passed |
| M3 — text predicate instead of byte | exactly `a_nul_bearing_prefix_pattern_matches_by_bytes_for_policies_and_grants` at its first deny check and `grant_matching_keeps_literal_case_sensitive_byte_prefixes` at a nul arm | both FAILED |
| M4 — text rank instead of byte rank | only `a_nul_bearing_prefix_pattern_matches_by_bytes_for_policies_and_grants`, at the ranked check | that one FAILED; the other binary 20 passed, 0 failed |
| M5 — lone-star arm removed | only `grant_matching_keeps_literal_case_sensitive_byte_prefixes`, at the `(*, empty value)` arm | that one FAILED at `grant_pins.rs:933` (`left: false, right: true`); the other binary 16 passed, 0 failed |

Restored tree: 20 passed and 16 passed, 0 failed.

The distinction M4 and M5 carry is the one worth stating. M3 and M4 differ only in which half of the byte-wise handling is reverted, and they redden different checks inside the same test, so a fix that got the predicate right and the rank wrong is visible. M5 reverts only the lone-star arm and reddens exactly one test at exactly one case, which is what makes the claim about `substr` of a zero-length blob a measured one rather than a reading of the SQL.

No code changed between the two runs; only the test command did.

Issue #2367, opened 2026-09-02
Coverage measurement hangs in memory recall until the reader checkout timeout, so the ratchet never runs


ohdearquant:
## Summary

The advisory coverage measurement (`cargo llvm-cov --workspace`) started failing on the hosted Linux runner on 2026-09-02. Three consecutive runs from about 20:10 UTC either exceeded the job budget (cut at 30:19 twice) or, on #2365, ended at 19:27 with one memory integration test failing:

```
test_weighted_fusion_multi_model_text_not_zeroed
recall with weighted fusion and 2 vector models:
DeadlineExceeded { operation: "memory.recall", budget_ms: 30000, elapsed_ms: 300010 }
```

The two cancelled runs show a second memory test in the same state before the cut (`prune_excludes_pruned_memory_via_default_fusion_recall`, running for more than 60 seconds). The 300 seconds is `KHIVE_CHECKOUT_TIMEOUT_SECS`, which this job sets to widen the writer-checkout contention window under instrumentation. Runs of the same tree earlier the same day passed in about 21 minutes, and the failing heads do not share a code change (one of them differs from its passing predecessor only in Python files), so this is an intermittent contention hang surfacing under the instrumented build, not the workspace outgrowing its budget.

With #2365 a failed measurement is reported as unavailable instead of cancelling the job, so the merge gate is no longer blocked by it. The cost is that the coverage ratchet is skipped whenever this happens, and it happened on every run of the afternoon.

## What would restore the signal

- Find why a memory recall on an in-memory runtime can wait the full reader checkout timeout under `cargo test` parallelism, and make that wait bounded by the recall deadline rather than by the checkout timeout (the deadline fired at 30 s of budget but reported 300 s elapsed).
- Once the hang is understood, drop the 300-second override in this job back toward the plain-run value so a contention stall fails fast instead of eating the job budget.
- The ratchet's own run on the resulting measurement is the landing evidence.

PR #2889, opened 2026-09-17
fix(ci): scope the instrumented checkout margin to the test that needs it

ohdearquant:

Part of #2367, and **not** a closure: the measurement still times out with this change in, so the
issue stays open. The comparison that shows what it does and does not fix is in a comment below.

4 files, 1 commit.

## The problem

The coverage job sets an environment variable for the whole job:

```yaml
env:
  KHIVE_CHECKOUT_TIMEOUT_SECS: "300"
```

That variable configures the reader pool, so it applies to every test binary in the run, not to the
one test it was raised for. Under an instrumented run with real contention, unrelated tests could
each wait up to 300 seconds for a checkout instead of the usual 5. The job has a 28 minute budget,
so the budget goes into waiting rather than measuring, and coverage comes back unavailable.

## The change

The job now sets `KHIVE_TEST_REPLAY_CHECKOUT_TIMEOUT_SECS=300`, which has exactly one consumer: the
replay test from #705. That test computes its timeout as the maximum of three things, so nothing it
previously relied on is lost:

- its own 120 second plain-run floor,
- the new test-specific value, if set,
- any explicit ambient `KHIVE_CHECKOUT_TIMEOUT_SECS` a caller set deliberately.

Its serial tags and the RAII restoration of the ambient override are unchanged. Every other test
binary is back to the default five second checkout.

One comment in `pool.rs` said a defaults test clears the pool environment because "CI jobs set
KHIVE_CHECKOUT_TIMEOUT_SECS". After this change no CI job does. The clearing is still needed, for a
better reason that the comment now gives: a sibling test in the same process sets that variable
around its own body.

## Verification

The replay test passes locally with the new variable set and the global override absent:

```
KHIVE_TEST_REPLAY_CHECKOUT_TIMEOUT_SECS=300 cargo test -p khive-mcp --lib \
  pending_events::tests::replayable_action_dispatches_without_failure_at_trigger_time -- --exact
test result: ok. 1 passed; 0 failed
```

That arm proves the test still runs with its margin.

The acceptance I originally stated here was this PR's own coverage job producing an available
measurement and a real ratchet result. **It did not.** The job timed out at 28 minutes and the
ratchet skipped. Measured against a run without this change, the 336-second stall is gone and 15
more test binaries complete in the same wall-clock, so the mechanism is real and this removes it,
but the measurement needs roughly 38 minutes against a 28 minute budget and one five-minute stall
was never going to close that. Details and the caveat on that comparison are in a comment below.


ohdearquant (commented):

ohdearquant:
## The acceptance I stated is not met

I wrote in the description that the acceptance is this PR's own coverage job producing an available
measurement and a real ratchet result, and that a skipped step should be read as this change not
having worked. The ratchet skipped:

```
Coverage measurement (advisory): success   04:09:43Z -> 04:39:48Z
Coverage ratchet:                skipped
##[error]The action 'Compute coverage' has timed out after 28 minutes.
```

So this does not close #2367, and I am removing that claim from the description rather than merging
on it.

## What the change does do, measured

Two pull requests ran the same job within minutes of each other, one with this change and one
without, which is the closest thing to a control available here:

| | gaps ≥ 250s | largest gap | test-result blocks completed | span |
| without this change (#2890) | 1 | **336.6s** | 174 | 20.1 min |
| with this change (#2889) | 0 | 183.7s | **189** | 20.1 min |

A 336-second stall is what a 300-second pool checkout timeout looks like from outside, and it is
gone. In the same 20 minutes of test wall-clock the instrumented run got through 15 more test
binaries. So the mechanism the issue names is real and this removes it.

It is not enough. The measurement needs roughly 38 minutes against a 28 minute budget: about 9
minutes of toolchain and compilation before any test runs, then the instrumented suite. Removing one
five-minute stall does not close a ten-minute shortfall.

**Caveat, stated because the table invites over-reading:** these are two different diffs on two
different runners, n=1 per arm. It is consistent with the mechanism and it is not a controlled A/B.

## Separately: the ratchet is not gating anything

The ratchet is conditioned on `needs.coverage-measurement.outputs.available == 'true'`, and the
measurement job reports success while setting `available` to false when it times out. So the job is
green, the ratchet is skipped, and nothing is red.

The last five CI runs on `main` all read `Coverage measurement=success  Coverage ratchet=skipped`.
The comment above that job says the baseline is raised only by an explicit PR "so every change to
main (the floor included) goes through the gate wall". The wall has not fired on any recent run.
That is not caused by this PR and is worth its own issue: a gate whose skip condition is written by
the thing it gates will read healthy for exactly as long as it is broken.

Issue #1461, opened 2026-07-31
Expose partial ANN-indexing failures to callers


ohdearquant:
## Problem

A write or reindex operation can succeed while one of its indexing stages (full-text search, vector embedding, or snapshot invalidation) fails, leaving those indexes stale or incomplete with no caller-visible signal that a repair is needed.

## Current behavior

Four points in the write and reindex paths currently fail open:

1. **Note creation** (`crates/khive-runtime/src/operations.rs`, `try_create_note`) is best-effort by design: an FTS failure or an embedding/vector failure both warn and continue, and the call still returns success. There is no retry, no repair receipt, and no partial-index status returned to the caller. This path is scoped to ingest-style callers (message ingest, the email channel), not the ordinary entity/note create verb.

2. **Reindexing a note** (`crates/khive-runtime/src/curation.rs`, `reindex_note`) fails closed on the FTS re-upsert, but an embedding/vector failure warns and skips. A caller can see a successful update and fresh lexical search results while the vector index still serves the old value, or none.

3. **Snapshot invalidation during a full reindex** (`crates/kkernel/src/reindex.rs`) fails closed internally on non-missing-table errors, but both of its call sites in the reindex command wrap it in a warning and continue. The command reports successful completion while stale snapshots survive, and the next rebuild trigger never fires, so a later read serves pre-reindex vectors.

4. **No signaling seam** — the note-mutation hook returns no error type, so a post-commit hook failure cannot propagate, and the operation-success result carries no index-generation or staleness field. A log warning is the only artifact, invisible to a caller that needs to make a duplicate-sensitive or freshness-sensitive decision.

Semantic recall already degrades gracefully to lexical-only search when the vector index is unavailable, and that degradation is visible in logs. This cluster means the same invisibility applies to writes: success is reported without the "indexed everywhere" postcondition holding, and nothing downstream can distinguish a fully-indexed record from a lexical-only one.

## Proposed direction (not prescriptive)

- Caller-visible partial-index state on ingest-style creates (the result reports which indexes landed), or a durable repair receipt a strict caller can wait on.
- Treat snapshot-invalidation failure during a full reindex as a command failure, since it is the command's core postcondition, or retry it under a bounded policy.
- A staleness watermark on the read path so degraded serving is observable per record rather than only in logs.

## Acceptance

- A write or reindex operation that partially fails at one indexing stage surfaces that fact to the caller, rather than reporting unqualified success.
[...]

PR #2890, opened 2026-09-17
fix(reindex): a failed Vamana snapshot invalidation fails the run

ohdearquant:

Part of #1461. This closes the snapshot-invalidation arm only; the create, curation and hook arms of
that issue stay open, so it does not auto-close.

2 files, 1 commit.

## The problem

After reindex rewrites the graph substrate it invalidates the namespace Vamana snapshots, so that
search stops answering out of vectors that no longer match the rows behind them. When that
invalidation failed, the error went out at warn level and the run reported success:

```rust
if let Err(e) = invalidate_vamana_snapshots(&rt, &ns_str).await {
    tracing::warn!(error = %e, "failed to invalidate Vamana snapshots after reindex");
}
```

The snapshots stayed stale and the exit code said the reindex was complete. Anything downstream that
keys on the exit code, which is the normal way to sequence a reindex, had no way to find out.

## The change

The failure sets `vamana_snapshot_invalidation_failed` on the report. That feeds `has_failures()`,
which already drives the existing finish and CLI semantics, so the behaviour follows the same rules
as every other failure source: a normal run exits nonzero, and `--best-effort` keeps both its warning
and its zero exit.

The human report gains a line that says the snapshots may be stale while prior writes remain
committed, which is the accurate description of the state the run leaves behind. The
machine-readable report is serialized from the same struct, so it carries the field without a second
code path to keep in sync.

Unchanged on purpose: a missing table remains a no-op, and the independent best-effort active-memory
snapshot deletion and completion-epoch bump keep their own flag.

## Tests

Three arms, and they divide the work deliberately:

| arm | what it pins |
| --- | --- |
| `snapshot_invalidation_failure_is_reported_and_controls_exit` | the report semantics, from a hand-built report |
| `run_reindex_snapshot_failure_fails_closed_and_preserves_committed_work` | the wiring, driven through `run_reindex` with a DELETE trigger that rejects |
| `run_reindex_snapshot_missing_table_remains_successful` | the no-op path, plus a same-fixture retry |

The middle arm also asserts the committed FTS rows and the completion epoch are still there, because
"fails closed" has to mean the run reports the failure without discarding work that did land.

Mutation check. Removing the one line that sets the flag:

```
snapshot_invalidation_failure_is_reported_and_controls_exit ............ ok
run_reindex_snapshot_failure_fails_closed_and_preserves_committed_work . FAILED
run_reindex_snapshot_missing_table_remains_successful .................. ok
```

Only the arm that goes through `run_reindex` dies, which is the right result: the other two do not
touch the production assignment, and an arm that reddened there would be testing its own fixture.

Local run of the whole module: 54 passed, 0 failed.


ohdearquant (commented):

PR #2891, opened 2026-09-17
fix(db): run a borrowed reader lease under the request deadline

### LINKED ISSUE #2360: ANN rotation watcher: cover the daemon-shutdown exit path with a direct test
The rotation watcher started by a writable warm in both the knowledge and memory ANN packs exits on two conditions: its weak state reference fails to upgrade, or the daemon shutdown token is cancelled. The pack-level lifecycle tests exercise the first exit (drop the state, advance past a tick) but not the second, so the shutdown branch can regress without a failing test.

The shutdown token is a process-wide singleton, and cancelling it inside a test would leak into every other test in the same binary. A direct test therefore needs the watcher loop to take its token as a parameter (production passes the daemon token, tests pass a local one), or an equivalent injection seam.

Acceptance: a test in each pack that starts the watcher against a live state, cancels a locally owned token, and asserts the tracked task count returns to its baseline without advancing time.

### LINKED ISSUE #2363: Reader lease: run query_row under the request deadline and hide the prepared statement from the mapper
`ReaderGuard::query_row` (`crates/khive-db/src/pool.rs`) is the only public accessor on a pooled reader lease. It admits the statement through the read-only classifier, marks the lease dirty, and then calls `Connection::query_row` directly. Two gaps remain on that surface.

**1. The query does not run under the request deadline.** Pooled `SqlReader` reads enter `run_interruptible_read` (`crates/khive-db/src/sql_bridge.rs`), so request cancellation and the deadline apply to them. `query_row` bypasses that path: a caller holding a lease can run an admitted `SELECT` that neither cancellation nor the deadline interrupts. Either route `query_row` through the same interruptible wrapper, or make it crate-private and expose only bounded typed inspection.

**2. The row mapper receives a `rusqlite::Row`, which is `AsRef<rusqlite::Statement>`.** In the pinned rusqlite (0.40.1) the immutable statement borrow exposes metadata only (parameter and column names and counts, expanded SQL, status counters, `is_explain`, `readonly`); binding, stepping and execution take `&mut self`, and the connection field is crate-private, so no write path escapes. A khive-owned row view that offers value extraction alone would still make the read-only boundary explicit at the type level instead of relying on rusqlite's receiver discipline.

Acceptance: a test that a cancelled or expired request through `query_row` yields the same cancellation outcome as the pooled reader path, and a mapper argument type owned by `khive-db`.

ohdearquant:

Closes #2363. Closes #2360.

12 files, 1 commit.

## #2363 — a borrowed reader lease ignored the request it was serving

A pooled reader lease taken synchronously has no async waiter, and the waiter is what latched the
request's cancellation signal and its original absolute deadline. So a cancelled request kept
stepping to completion, and an already-expired deadline was not noticed until the read returned.

`ReadControl` now polls the request context from inside SQLite's progress handler, and only for the
borrowed lease. The ordinary async path keeps its lock-free predicate and gains no watch polling,
because paying for that on the common path to fix the uncommon one is how a cancellation fix becomes
a throughput regression.

The poll runs at three points: before registration, inside stepping, and again after the read
returns. The third one matters and is not redundant. Mapper and native callback code cannot be
preempted, so if the request stops while that code is running, the only honest thing to do is refuse
the otherwise-successful result afterwards rather than pretend it was observed in time.

## #2363 — the mapper could reach past the row

`query_row` handed the mapper a `rusqlite::Row`, which carries its prepared statement and therefore
its connection. A mapper could step or re-prepare behind the lease's back.

It now receives a `ReaderRow`, which exposes `get` and `get_ref` and nothing else. The guarantee is
type-level, so the test for it is type-level too:

```rust
/// ```compile_fail
/// fn statement(row: &khive_db::ReaderRow<'_, '_>) {
///     let _: &rusqlite::Statement<'_> = row.as_ref();
/// }
/// ```
```

A runtime test cannot observe the absence of an API. A doctest that must fail to compile can.

Two further refusals land on the same path:

- a lease whose cleanup failed is **quarantined** and refuses later queries, rather than serving from
  a connection in an unknown state;
- a query already in progress refuses **re-entry**, because both the params (`ToSql`) and the mapper
  run user code, and either could otherwise replace the connection-global progress handler in the
  middle of the query that installed it.

## #2360 — the watcher's shutdown path had no direct test

The ANN rotation watcher's daemon-shutdown exit was only reached indirectly. It now has a test that
drives that exit path directly, for both the knowledge and memory watchers.

## Verification

Full workspace suite on this exact tree:

```
workspace-format   GREEN
workspace-clippy   GREEN
workspace-test     GREEN   264 blocks, 11361 passed, 0 failed
```

Ten targeted positive arms besides those (reader leases, both watcher paths, six pre-existing reader
regressions, and the `ReaderRow` doctests).

Four mutation controls, each predicted before it ran, each required to fail one named test at one
named location:

| mutation | outcome |
| --- | --- |
| panic in the knowledge watcher shutdown path | failed at `vamana.rs:3507`, 1 failure |
| panic in the memory watcher shutdown path | failed at `ann.rs:3108`, 1 failure |
| bypass the reader scope | failed at `reader_lease_tests.rs:28`, 1 failure |
| remove the re-entrancy check | failed at `reader_lease_tests.rs:366`, 1 failure |

Each control failed exactly its named test and nothing else, which is the part worth stating: a
mutation that reddens the suite broadly proves the suite is sensitive, not that the arm is aimed at
the behaviour it claims.


ohdearquant (commented):

Verdict on head 285cb5d0230bac65137b2ed9b52ee4d88404d308: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2896, opened 2026-09-17
fix(ci): make a missing coverage measurement a red gate, not a skipped one

### LINKED ISSUE #2892: Coverage ratchet never runs: its skip condition is written by the job it gates
> _Issue description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Symptom

The coverage ratchet has not run on any recent change. The last five CI runs on `main`:

```
Coverage measurement=success  Coverage ratchet=skipped
Coverage measurement=success  Coverage ratchet=skipped
Coverage measurement=success  Coverage ratchet=skipped
Coverage measurement=success  Coverage ratchet=skipped
Coverage measurement=success  Coverage ratchet=skipped
```

Three open pull requests measured on the same morning show the same pair.

## Mechanism

The ratchet is conditioned on the job it is supposed to gate:

```yaml
name: Coverage ratchet
needs: coverage-measurement
if: needs.coverage-measurement.outputs.available == 'true'
```

`Compute coverage` has `timeout-minutes: 28`. The measurement currently needs roughly 38 minutes:
about 9 minutes of toolchain install and compilation before any test runs, then the instrumented
suite. When it times out, the job sets `available=false` and still concludes **success**, because
the step is advisory by design. So the measurement is green, the ratchet is skipped, and nothing on
the whole path is red.

The comment above that job states the intent plainly: the baseline is raised only by an explicit PR
that edits `coverage-baseline.txt`, "so every change to main (the floor included) goes through the
gate wall". The wall is not firing.

## Why this is worth a separate issue

A gate whose skip condition is produced by the thing it gates cannot report its own absence. There
is no failing check, no annotation, and the workflow reads exactly as it would if coverage were
healthy — the only visible difference is a job that says `skipped`, which also happens routinely for
jobs that are legitimately not applicable.

## Shape of a fix, not a prescription

Two independent parts, and the first is worth doing even if the second takes a while:

1. **Make the absence loud.** An unavailable measurement should produce something that fails or at
   minimum annotates, rather than a skip that is indistinguishable from a normal one. Whatever is
   chosen, the arm to write first is the one that asserts a timed-out measurement does not read as a
   pass.
2. **Fit the measurement in its budget, or give it a bigger one.** Raising `timeout-minutes` is the
   cheap move and is not obviously wrong for an advisory job, but it should be a decision with a
   number attached rather than a bump until it stops skipping.

Related: #2367, which named one contributor to the budget (a job-wide reader-pool timeout applied to
every test binary). PR #2889 removes that one; measured, it takes out a 336-second stall and gets 15
more test binaries through in the same wall-clock, and the measurement still times out. So the
budget shortfall is larger than that one cause, and this issue is not a duplicate of it.

ohdearquant:

Closes #2892.

Branch diff against `main`: 1 file, 1 commit.

The coverage ratchet has never judged anything, and every run said it was fine. Three layers
each did what they were written to do:

1. `Compute coverage` carries `timeout-minutes: 28` and `continue-on-error: true`. It is cut by
   that budget, and `continue-on-error` turns the step failure into a green job.
2. `available` is `steps.compute_coverage.outcome == 'success'`, so it comes out `false`, and
   `coverage-ratchet` is skipped for want of an input.
3. `CI gate` lists `coverage-ratchet` in its allowed skips, so the skip is forgiven.

Measured on the five most recent completed runs (three pull requests, two main pushes): the
measurement job concluded `success` in 29-30 minutes, `coverage-ratchet` concluded `skipped`,
and `Compute coverage` ran 28m13s on the main push at 06:58Z. The decisive artifact is the next
step: `Report unavailable coverage measurement` has `if: outcome != 'success'`, and the API
shows it executing rather than skipped on those runs. `Coverage ratchet` concluded `skipped` on
5 of 5.

## What changed

- The measurement job fails when no measurement was produced. The reporting step keeps
  `continue-on-error` so the run summary still explains itself, and a new step exits 1.
- `coverage-ratchet` leaves `CI gate`'s allowed-skips list. It skips exactly when the
  measurement did not arrive, which is the state the gate exists to report.
- The step budget goes 28 -> 55 minutes, the job's 40 -> 75. Since no run has ever produced a
  measurement, the time it needs is not known; this pull request's own run is the first
  observation of it, and the number goes in this thread before the merge.
- The job drops "(advisory)" from its name.

## Acceptance

The gate's `jq` predicate, extracted verbatim from the workflow and run over four payloads:

| payload | new predicate | old predicate |
| --- | --- | --- |
| all jobs green | empty | empty |
| measurement green, ratchet skipped (the last five runs) | `coverage-ratchet` | **empty** |
| measurement failed, ratchet skipped | `coverage-measurement coverage-ratchet` | n/a |
| the two remaining allowed skips | empty | empty |

The second row is the finding: the state that has been shipping is now a gate failure and was
not one before.

What this pull request cannot show on its own is the ratchet running green on a main push,
because the ratchet needs a measurement and no measurement has ever completed. The acceptance
that closes the issue is therefore two-part: this run has to produce a measurement inside the
new budget, and the first main push after it has to show `Coverage ratchet` concluding
`success` rather than `skipped`. Both readings go in this thread.


ohdearquant:

Duration numbers, as promised, read from the step records rather than from the job totals.

On this branch, with the raised budget: `Compute coverage` ran `2026-09-17T10:46:45Z` to `2026-09-17T11:16:00Z`, **29m15s**, conclusion `success`. The `Coverage ratchet` job then executed `Check coverage does not regress` and concluded `success`. That is the first run in this series where the comparison step ran at all.

Under the previous 28-minute budget, the five most recent runs all look the same in the API: the measurement job concludes `success`, the unavailable-report step executes, and the ratchet job is listed as skipped, five times out of five. The most recent push to `main` cut `Compute coverage` at **28m13s**.

So the measurement needs about 29 minutes and the budget allowed 28. It was short by roughly one minute, and the cost of that minute was that every run for the past five produced no measurement while still reporting a green job, because a `continue-on-error` step reports `conclusion: success` in the API even when its `outcome` is failure. The new step keyed on `outcome != 'success'` is what turns that back into a red job, and the raised budget is what stops it from firing every time.

PR #2900, opened 2026-09-17
perf(brain): serialize persisted snapshots from a borrowed view

### LINKED ISSUE #1441: brain: reduce full-state copies during persisted mutations
A persisted brain-state mutation snapshots the current state, rebuilds a proposed copy, then creates another snapshot for persistence. Large router gate byte buffers and adapter sets are copied through this path, increasing peak memory use and allocation latency.

## Current behavior

`BrainState::to_snapshot` clones profiles, bindings, router state, and adapter sets, including derived per-profile state (`crates/khive-brain-core/src/brain_state.rs:52-72`); router state owns raw gate-byte vectors and adapter sets own vectors of records (`crates/khive-brain-core/src/brain_state.rs:220-247`). `persist_brain_state_mutation` clones the current state through a snapshot/restore round trip and then snapshots the proposed state again before the database write (`crates/khive-pack-brain/src/persist.rs:349-353`). For a large profile/router configuration, a single mutation can transiently hold the current state, an independently reconstructed proposed state, and a full persistence snapshot at once, raising peak RSS and allocation latency on the mutation path.

## Expected behavior

Preserve atomic commit and rollback behavior while reducing whole-state copying through structural sharing, targeted copy-on-write mutation, or a more direct committed representation.

## Acceptance

- A representative large state shows a predeclared reduction in peak allocation or RSS.
- Snapshot round-trip behavior is unchanged.
- A failed write leaves the in-memory state unchanged.

ohdearquant:

Closes #1441.

## What was wrong

Both durable mutation paths in `khive-pack-brain` built an owned `BrainStateSnapshot`
and then serialized it, so the opaque router byte vectors and the adapter records were
live in two owned copies at the moment the encoder ran: once in the live state, once in
the snapshot built from it. On a state carrying eight one-mebibyte router blobs that is
several megabytes of duplication whose only purpose is to be read once and dropped.

## The change

`BrainState` gains a private `snapshot_view()` returning a serialization-only
`BrainStateSnapshotView<'a>`, and a public `to_snapshot_json()` that encodes it directly:

```rust
pub fn to_snapshot_json(&self) -> Result<String, serde_json::Error> {
    serde_json::to_string(&self.snapshot_view())
}
```

The view borrows exactly the three fields that were being cloned for no reason —
`bindings: &'a [ProfileBinding]`, `router_state: &'a HashMap<..>`,
`adapter_set: &'a HashMap<..>` — and owns the four projections that have to be
materialized anyway. Bindings, router bytes and adapter records now stay borrowed until
synchronous serialization completes, and only the owned JSON string crosses the SQL await
that follows.

`to_snapshot()` keeps its owned return type and is rebuilt from the same view, so its
ownership and wire behaviour are unchanged and both paths cannot drift apart.

**The wire format does not change.** The view declares the same seven fields, in the same
order, with the same names and value types. The `serde(default)` attributes on the owned
struct govern deserialization only, so the encoder emits the same bytes for the same
state.

## Tests

`crates/khive-brain-core/tests/snapshot_allocation.rs` builds eight profiles, eight
one-mebibyte router blobs and 2048 adapter records with live posterior state and
bindings, then measures **requested live allocation growth** across each encode through a
tracking global allocator, requiring the borrowed path to save at least 6 MiB against the
owned one. That measures requested allocation, not resident memory and not elapsed time,
and the assertion is a floor rather than an equality so it does not pin allocator
behaviour.

In `khive-pack-brain/src/persist.rs`, trigger fixtures force the snapshot insert to fail
after the event append and check that the rollback leaves live state unchanged, then
retry and check the committed result — covering both the brain-state and feedback-state
mutation paths.


oceanwaves630 (approved):
Required checks pass at 00b4b013: CI gate and Secret scan (gitleaks), both completed/success.

Snapshot serialization now goes through a borrowed view rather than cloning bindings, router state and the adapter set, and `to_snapshot` keeps its existing owned return by materialising from that view. The added test tracks peak allocation through a global allocator, so the saving is measured rather than asserted.

PR #2901, opened 2026-09-17
fix(runtime): admit one events pool per canonical path

### LINKED ISSUE #2402: events: one events database can be opened writable and read-only in one process, and the raw file opens of either pool run while the other holds SQLite locks
The process-local registry of events-database pools is keyed by `(path, read_only)`. A writable
main runtime and a read-only pack runtime derive the same events path (the multi-backend boot
gives every runtime one canonical events sidecar path), so one physical database can be served by
two pools in one process. Each pool performs a raw file open outside SQLite at construction: the
read-only pool reads the database header to learn whether it is in WAL mode, and the writable pool
runs the sidecar permission hardening before its open. Under POSIX advisory locking, closing a
descriptor on a file releases every lock the process holds on that inode, so whichever pool is
built second can release the locks the first one holds. #2398 fixed this for a single pool by
moving the hardening before the open and by making the post-open check `lstat`-only; the two-pool
case is unchanged.

Direction: key the registry by canonical path only and serve both modes from one entry (one
coordinated pool that hands out read-only and writable handles), and perform the raw opens once,
under the registry lock, before any pool for that path exists. Reading the header through an
already-open SQLite connection (`PRAGMA journal_mode`) removes the read-only pool's raw open
entirely.

Acceptance: a test that builds a writable and a read-only pool for one events path in both orders,
holds a write transaction on the first, and asserts that building the second neither releases the
held locks (`F_GETLK` from a second process reports the lock still held) nor blocks. The
cross-process lock test from #2398 stays green.

ohdearquant:

Closes #2402.

## What was wrong

The process-wide events backend registry was keyed by the path as given, paired
with the access mode. Both halves of that key were wrong, and both are reachable
from ordinary callers.

**The path as given.** A relative path, a symlinked parent and the absolute path
were three distinct entries for one database, so one file could back several pools
at once. SQLite takes its locks per connection, so the second pool does not see the
first one's write lock.

**The access mode.** A read-only open and a writable open of the same file were
also distinct entries, so the second open missed the live pool and went on to the
writable preflight: a `create_new` open, then a chmod of the `-wal` and `-shm`
sidecars belonging to a connection that was open at the time. The first connection
lost the lock it was holding.

## The change

`BackendMap` becomes `HashMap<PathBuf, (bool, Arc<StorageBackend>)>` — one entry per
canonicalized path, with the access mode carried inside the entry rather than in the
key.

- A database that does not exist yet canonicalizes through its parent and rejoins
  the file name, so it is keyed identically before and after it is created.
- An open that asks for the mode the entry does not hold is refused with
  `InvalidInput`, naming both modes, instead of being served a second pool over the
  first.
- The symlink refusal and the directory-trust check now run on every open rather
  than only the first. They are stat-only, and the path can change underneath a
  cached entry. The raw-file steps stay behind the admission check, which is what
  keeps them off a live pool.

## Tests

`crates/khive-runtime/tests/events_registry_keeps_sqlite_locks.rs`:

- `writable_then_read_only_keeps_sqlite_locks` and
  `read_only_then_writable_keeps_sqlite_locks` cover both orderings of the
  mixed-mode open; each asserts the first connection still holds its lock.
- `concurrent_compatible_openers_reuse_one_canonical_entry` has several callers race
  on aliases of one path and requires that they all receive the same `Arc`.
- `read_only_initialization_preserves_missing_and_frozen_files` pins that read-only
  initialization still leaves a missing or frozen file untouched.


oceanwaves630 (approved):
Required checks pass at c49e3549: CI gate and Secret scan (gitleaks), both completed/success.

The registry now admits one entry per canonical path and returns an explicit error when a second open asks for a different access mode, instead of handing out a second pool with different guarantees for the same file. The added test inspects the held SQLite locks from a separate process, which is the only way to observe them.

PR #2902, opened 2026-09-17
fix(config): close the [actor] table against unknown keys

### LINKED ISSUE #2899: [actor] is a policy table but is not closed against unknown keys

`KhiveConfig`'s own doc comment states the rule this configuration surface follows:

> Unknown top-level keys are silently ignored by serde for forward
> compatibility. The `[gate]`, `[brain]`, and `[telemetry]` tables are closed with
> `deny_unknown_fields` so a misspelled policy key always fails startup.

`[actor]` is a policy table and is not on that list. `ActorConfig` derives
`Deserialize` without `deny_unknown_fields`, directly above `GateSectionConfig`,
which has it and carries a comment giving the reason: misspelled or future keys
should fail startup instead of being silently ignored at an authorization boundary.

Both keys `[actor]` carries are authorization input:

- `visible_namespaces` widens the default multi-record read scope to
  `['local'] union visible_namespaces` (ADR-007 Rev 4 Rule 3b).
- `allowed_outbound_namespaces` is the sender-side permission for cross-namespace
  delivery, deny-all when absent.

Two consequences today:

1. A misspelling of either key parses, starts, and silently applies the default.
   For `allowed_outbound_namespaces` that default is deny-all, so the symptom is
   delivery refused against a config that reads as if it permits it; for
   `visible_namespaces` the default read scope quietly stays narrow.

2. A key belonging to `[gate]` that is written under `[actor]` is discarded with no
   diagnostic. `grant_unattributed = false` placed one table too high leaves
   anonymous admission at whatever it already was, while the file on disk says
   otherwise. That is the expensive direction: the gate table refuses to ignore the
   key, and the actor table immediately above it ignores it in silence.

Proposal: add `deny_unknown_fields` to `ActorConfig`, and add `[actor]` to the closed
list in the `KhiveConfig` doc comment so the stated rule and the code agree. Scope is
those two edits plus a test that an unknown key under `[actor]` fails to load, with a
known-good config as the control in the same test.

Compatibility: every `[actor]` key present in the configuration files on the machine
where this was found is one of `id`, `display_name`, `visible_namespaces` and
`allowed_outbound_namespaces`, so the change rejects nothing that parses today.

ohdearquant:

Closes #2899.

One file, three changes.

## `ActorConfig` gains `deny_unknown_fields`

`KhiveConfig`'s doc comment states the rule this configuration surface follows: unknown
top-level keys are ignored for forward compatibility, and the policy tables are closed so
a misspelled policy key always fails startup. `[actor]` is a policy table and was not on
that list, even though `GateSectionConfig` sits directly below it with the attribute and a
comment giving the reason.

Both keys `[actor]` carries are authorization input — `visible_namespaces` widens the
default multi-record read scope (ADR-007 Rev 4 Rule 3b), and
`allowed_outbound_namespaces` is the sender-side permission for cross-namespace delivery,
deny-all when absent. A misspelling of either parsed, started, and silently applied the
default. A `[gate]` key written one table too high was discarded with no diagnostic, which
leaves anonymous admission at whatever it already was while the file on disk says
otherwise.

## The doc comment now names `[actor]` among the closed tables

So the stated rule and the code agree. The three tables it already named were each checked
at source rather than taken from the sentence: `[gate]` and `[brain]` carry the attribute
directly, and `[telemetry]` gets it through `RawTelemetryConfig`, which
`TelemetryConfig` deserializes via `serde(try_from)`.

## Test

`unknown_actor_key_fails_to_load` writes `grant_unattributed = false` under `[actor]`,
requires the load to fail, and asserts the error names the offending key. Its control is
in the same test and runs first: the identical table carrying only supported keys must
still load, so the refusal is attributable to the unknown key rather than to the fixture.

## Compatibility

This makes a previously accepted config fail startup, so the population was enumerated
before proposing it. Every `[actor]` key present across the 1134 configuration files on
the machine where this was found is one of `id`, `display_name`, `visible_namespaces` and
`allowed_outbound_namespaces`. The only other key written under `[actor]` anywhere in this
repository is `writable_namespaces`, which appears solely in ADR-059 — withdrawn, and
superseded by ADR-007 Rev 2.

The full `khive-runtime` lib suite runs 1742 passed / 0 failed / 9 ignored with the table
closed, so no existing fixture depended on the table being open.


oceanwaves630 (approved):
Required checks pass at 899c9df6: CI gate and Secret scan (gitleaks), both completed/success.

Closing the table is safe for existing files: `ActorConfig` declares all four keys it accepts, including `display_name`, so a config written against the documented shape still loads. I checked the `[actor]` tables on a machine running several of these and every key in them is a declared one. The added test carries its own control, so the refusal is attributable to the unknown key rather than to the fixture.

PR #2926, opened 2026-09-18
fix(gtd): validate a priority written through properties and re-derive its salience

### LINKED ISSUE #2925: gtd: a task's priority can only be changed through the unvalidated path, and its derived salience is never updated

`gtd.assign` validates `priority` against `p0..p3`, refuses anything else, and derives the note's
`salience` from it (`priority_to_salience`, which the pack's own doc comment describes as feeding
hybrid search scoring). No gtd verb can change a task's priority afterwards: the pack registers
`gtd.assign`, `gtd.next`, `gtd.complete`, `gtd.tasks` and `gtd.transition`, and none of them takes
one. So re-prioritizing a task means writing `properties.priority` through the generic update, and
that path neither validates the value nor re-derives the salience beside it.

Measured against a running daemon, one task, in order:

```
gtd.assign(title=..., priority="p3")
  -> properties.priority = "p3"
     salience            = 0.25          (the p3 value)

update(id=..., properties={..., "priority": "p0"})
  -> properties.priority = "p0"
     salience            = 0.25          <- still the p3 value

update(id=..., properties={..., "priority": "URGENT!!"})
  -> accepted, version 3
     properties.priority = "URGENT!!"    <- stored verbatim
     salience            = 0.25
```

Two consequences, and the second is the one that hides:

1. **A stored priority can be unqueryable.** `gtd.tasks(priority=...)` validates its filter against
   `p0..p3`, so a task holding `"URGENT!!"` cannot be reached by any priority filter at all. In the
   run above the probe was absent from `gtd.tasks(priority="p0")` (6 rows) and present in
   `gtd.tasks(status="inbox")`, so it is not missing, only invisible to one axis.
2. **The escalation does not reach the ranking.** `salience` is derived state written once at
   create. After a p3 task is escalated to p0 it still ranks at 0.25, and the row reads as
   consistent because both fields are present and well-formed. There is no way to repair it through
   any verb, because the field that needs repairing is the derived one.

`properties.status` is already guarded on this path, and its refusal is the model to follow:

```
invalid input: properties.status is lifecycle-owned and cannot be patched on a task;
use gtd.transition for lifecycle changes or gtd.complete for terminal completion
```

Refusing `priority` the same way would leave no path to change it at all, so the fix is the other
half: validate the value on the property path with the same predicate `gtd.assign` uses, and
re-derive `salience` beside it, in `TaskHook::prepare_note_update` — the one hook every generic
update path already calls. That is the same shape as the `due` / `due_timezone` normalization in
#2924, which is the same defect on a different field: a field with a validating writer and a silent
one, and derived state left standing beside a changed value.

ohdearquant:

1 file, 1 commit. Stacked on #2924, which puts the `due` normalization in the same hook; the two
are the same defect on two fields.

Closes #2925.

`gtd.assign` validates `priority` against `p0..p3`, stores it lowercase, and derives the note's
`salience` from it. The generic property path did none of the three, and no gtd verb can change a
priority afterwards — the pack registers `gtd.assign`, `gtd.next`, `gtd.complete`, `gtd.tasks` and
`gtd.transition`, none of which takes one. So the free-form path is not a second writer, it is the
only one.

Measured against a running daemon before the change, one task, in order:

```
gtd.assign(title=..., priority="p3")
  -> properties.priority = "p3"
     salience            = 0.25

update(id=..., properties={..., "priority": "p0"})
  -> properties.priority = "p0"
     salience            = 0.25          <- still the p3 value

update(id=..., properties={..., "priority": "URGENT!!"})
  -> accepted, version 3, stored verbatim
```

The escalated task was absent from `gtd.tasks(priority="p0")` and present in
`gtd.tasks(status="inbox")`, so it was not missing, only unreachable on one axis — the filter
validates against the same list the writes never did.

Behaviour, stated as rules:

- The level is checked against `p0..p3` and stored in the spelling the create path uses, so a task
  cannot hold a form its own create would never have written.
- `salience` is re-derived beside it. A caller who names a `salience` in the same update keeps
  their value: what the caller states explicitly beats what is derived for them, which is the rule
  the zone resolution beside it already follows.
- `priority: null` lands on 0.5, the salience `gtd.assign` gives a task created without a level,
  rather than leaving the cleared level's value standing as derived state with nothing to derive
  from.
- A non-string, non-null priority is refused rather than stored.
- An update that does not mention the level touches neither field, so a task edited for an
  unrelated reason keeps its ranking.

`properties.status` was already guarded on this path and its refusal is the model the issue
suggested following. Refusing `priority` the same way would leave no path to change it at all, so
this is the other half: validate, and re-derive what the value feeds.

Seven tests, run under mutation rather than only run green:

- Removing the call from `prepare_note_update` reddens the wiring test and nothing else.
- Removing the `salience` derivation reddens four of the remaining six: the escalation, the
  spelling, the clearing fallback, and the wiring test.
- The fifth asserts a `salience` the caller supplied, so it cannot fail when the derivation is
  removed. That is what a must-not-happen arm is: it is the paired arms above that show the
  derivation fires at all.


ohdearquant (commented):

Verdict on head c1a6643415e619a7e040829087bbcb5db195221a: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head e9e0c83efe2fe8dc17d60435a095ea35cb393c28: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2939, opened 2026-09-18
fix(runtime): record the edge a merge drops between its own two operands

### LINKED ISSUE #2934: merge: an edge between the two merged records is hard-deleted with no preimage, no counter and no dry_run prediction

When `merge` rewires an edge and both endpoints resolve to the surviving record, the edge is deleted
with a row delete and the loop continues. No tombstone, no counter, no audit field.

```rust
if new_src == new_tgt {
    if !dry_run {
        conn.execute(
            "DELETE FROM graph_edges WHERE namespace = ?1 AND id = ?2",
            rusqlite::params![&edge.namespace, edge.id.to_string()],
        )?;
    }
    continue;
}
```

`crates/khive-runtime/src/curation.rs:3722-3730` on the note path and `:3062-3070` on the entity
path, the same eight lines twice.

Both endpoints resolve to the survivor in exactly one situation: the edge ran **between the two
records being merged**. So the edges this deletes are the ones that said something about the
relationship between them, and the highest-value members of that set are `supports` and `refutes`.
Merging two records joined by `refutes` destroys the recorded contradiction, permanently, and
returns a success payload that says nothing happened.

Three things make it hard to notice rather than merely unfortunate:

1. **`continue` precedes both counters.** `edges_rewired` and `edges_contract_skipped` are
   incremented further down the loop body, so the deleted edge is counted in neither. The declared
   return shape is `{kept_id, removed_id, edges_rewired, edges_contract_skipped,
   edge_conflict_preimages, properties_merged, tags_unioned, content_appended, dry_run}` — there is
   no field a self-loop deletion could appear in.
2. **`dry_run` cannot predict it.** Under `dry_run` the delete is skipped and the `continue` still
   runs, so a dry run and a real run produce the same summary for this case. The instrument that
   exists to show a caller what a merge will do is blind to the only part of it that is
   unrecoverable.
3. **The audit event does not carry it.** The `note_merged` / `entity_merged` payload is
   `{into_id, from_id, policy, content_strategy, reason}`.

The codebase already treats edge loss during merge as something that needs a preimage. The verb's
own description says so: "Natural-key edge collisions return and audit complete
`edge_conflict_preimages`, including annotations cascaded with the dropped edge." That is the
neighbouring case, handled correctly. This one takes the same kind of loss and records none of it.

Deleting the edge is arguably right — a self-loop `refutes` is not a meaningful row, and keeping it
would assert a record contradicts itself. The defect is that the deletion is silent and the deleted
state is unrecoverable, not that it happens.

Suggested fix, in the order the value lands:

1. Return and audit a preimage for each deleted self-loop edge — id, source, target, relation,
   weight, metadata — through the same mechanism `edge_conflict_preimages` already uses. This alone
   makes the loss recoverable by hand and visible to a caller.
2. Count them, in a named field, so a caller reading only the summary sees a non-zero number.
3. Make `dry_run` report the same field, so the prediction covers the case.
4. Consider refusing a merge whose two operands are joined by `supports` or `refutes` unless
   `force=true`. Two records that explicitly disagree are an odd thing to declare identical, and
   entity merges already have a safety floor (`validate_entity_merge_floor`) while note merges have
   none — that function takes two entities and the note arm of the handler calls no guard at all.
   This step is a behaviour change and worth deciding separately from 1 to 3, which are pure
   additions.

ohdearquant:

Closes #2934.

## What this fixes

`merge` rewires every edge incident on the absorbed record onto the survivor. When both
endpoints of an edge resolve to the survivor, the edge is a self-loop and is deleted. Both
endpoints resolve to the survivor in exactly one situation: the edge ran *between the two
records being merged*. So the set this silently deletes is precisely the set of edges that
recorded something about the relationship between the two operands, and its highest-value
members are `supports` and `refutes`. Merging two notes joined by `refutes` destroyed the
recorded contradiction and returned a summary in which nothing had happened.

Three properties kept it invisible:

- The `continue` runs before both counters, so the edge landed in neither `edges_rewired` nor
  `edges_contract_skipped`.
- The returned `MergeSummary` had no field the deletion could appear in.
- `dry_run` skipped the `DELETE` but still hit the same `continue`, so a dry run and a real run
  produced identical summaries for the one part of a merge that cannot be undone.

## What changed

Each dropped self-loop edge now yields a full row preimage through the same
`edge_row_preimage` mechanism `edge_conflict_preimages` already uses. `MergeSummary` gains
`edges_self_loop_dropped: usize` and `self_loop_edge_preimages: Vec<MergeEdgePreimage>`, both
`#[serde(default)]` so existing deserializers are unaffected. The capture runs on both the
entity path and the note path, and the counts and preimages appear in the audit payload as well
as the return value.

The capture is placed **before** the write gate, not inside it, so a dry run reports the exact
count and the exact rows a real merge will drop. That is the property that makes the dry run
usable as a preflight for an irreversible operation.

Deleting the edge remains correct: a self-loop `refutes` would assert that a record contradicts
itself. What changes is that the loss is counted, recoverable and predicted.

## Scope notes

- **The entity and note summary shapes do not differ.** Both paths write the same two new
  fields on the same `MergeSummary` type; there is no per-path variant and no divergence to
  reconcile.
- **Refusing a `supports`/`refutes` merge without an explicit `force` is deliberately not in
  this PR.** That is a behaviour change to what `merge` accepts, and it belongs to its own
  decision: this change makes the loss visible and recoverable, which is the prerequisite for
  that discussion rather than a substitute for it.

## Verification

`crates/` workspace, `khive-runtime` lib suite:

```
test result: ok. 1747 passed; 0 failed; 9 ignored; 0 measured; 0 filtered out
cargo fmt --check -p khive-runtime -p khive-pack-kg          rc=0
cargo clippy -p khive-runtime -p khive-pack-kg --all-targets -- -D warnings   rc=0
```

The four new tests were then checked against three deliberate mutations of the fix, with the
reddened set stated before each run and reconciled after:

| Mutation | Predicted red | Observed |
| --- | --- | --- |
| Drop both preimage pushes | the 4 new tests | 1743 passed, 4 failed — the 4 predicted |
| Drop both counter increments | the same 4, failing on a different assertion | 1743 passed, 4 failed — same names, assertion line moved as predicted |
| Move the capture back inside the write gate | only the 2 dry-run-equivalence tests | 1745 passed, 2 failed — the 2 predicted |

The restored control returned to `1747 passed; 0 failed` with the file byte-identical to its
pre-mutation state, so each arm is a genuine must-fail control rather than a harness that
failed to apply.

The third mutation is the one worth reading: it is the only arm that separates "the deletion is
recorded" from "a dry run predicts it", and it is the arm that would have stayed green if the
capture had been written inside the `if !dry_run` block, which is where it naturally wants to go.

## Scope, mechanically

`gh pr diff --name-only` = 2 files; `gh pr view --json commits` = 1 commit.


ohdearquant (commented):

PR #2949, opened 2026-09-18
fix(runtime): mask deny_reason before it reaches audit sinks

### LINKED ISSUE #2944: the gate.check audit log serializes deny_reason to the process log unmasked

Every gate check writes its decision to the process log by serializing the whole `AuditEvent` and
emitting it at INFO:

```rust
let audit = AuditEvent::from_check(&gate_req, &decision, self.gate.impl_name());
tracing::info!(
    audit_event = %serde_json::to_string(&audit).unwrap_or_else(|_| "{\"error\":\"serialize\"}".into()),
    "gate.check"
);
```

(`crates/khive-runtime/src/pack.rs`, in the dispatch interceptor around the `self.gate.check(&gate_req)`
call.) `AuditEvent` carries `deny_reason`, and the string goes to the log exactly as the gate
produced it. Nothing on that path masks it.

A gate is a configured extension point (`RuntimeConfig.gate`), so `deny_reason` is
gate-authored text whose contents this crate does not control. A gate that denies *because of*
something in the request has an obvious reason to name what it found, and naming it puts that
value in the log. Observed concretely: a gate that denies `comm.send` for a credential-shaped body
and says so in its reason produces

```
INFO khive_runtime::pack: gate.check audit_event={...,"decision":"deny","deny_reason":"comm.send denied: credential <the credential> in body",...}
```

while the line immediately after it, from the code that consumes the failure, carries the same
value masked. Two sinks, one pass, one of them masked.

This is the same shape as the scheduled-dispatch sinks (#2935) and is fixed the same way: run
`deny_reason` through `secret_gate::bounded_masked_log_text` before it is serialized into the log
line. It is filed separately rather than folded into that change because it sits in a different
crate, fires for every verb rather than for deferred dispatch, and the decision about how much
fidelity an audit line should trade for masking is worth making on its own terms rather than as a
side effect.

**The log is not the only sink.** On a deny, `build_audit_storage_event` serializes the same
`AuditEvent` — `serde_json::to_value(audit)` — straight into the storage `Event`'s payload with
`.with_payload(audit_data)` and appends it. So `deny_reason` reaches a durable, append-only record
by the same unmasked route, which is the higher-value of the two sinks and the harder one to
remediate after the fact.

Scope, stated plainly: this is a defense-in-depth and consistency gap, not a demonstrated leak.
The built-in gate does not echo request content into its reasons. The exposure is proportional to
what a configured gate chooses to put in `deny_reason` — and a gate that denies *because of*
something in the request has an obvious reason to name what it found.

Suggested fix:

1. Mask `deny_reason` in the `gate.check` log line.
2. Mask it on the durable path too. `build_audit_storage_event` re-serializes the whole
   `AuditEvent`, so masking `deny_reason` once on the `AuditEvent` itself — rather than at each
   sink — covers the log line and the stored payload together and cannot drift between them.
3. Test arm: a gate fixture whose `deny_reason` embeds a token the detector recognizes, asserting
   the captured log carries the masked form. Without it the fix is unfalsifiable, since every
   other assertion about the audit line passes whether the masking runs or not.

Found while running the tests for #2935: an end-to-end assertion over the whole captured log
failed on this line rather than on the one under test.

ohdearquant:

Closes #2944.

## The gap

`GateDecision::Deny` carries a `reason` string authored by the `Gate` implementation. Both of
`khive-runtime`'s `AuditEvent::from_check` call sites copied it into the audit event verbatim, and
that event reaches two sinks: the `gate.check` tracing line, and the durable event store. A custom
gate whose reason embeds the value it refused therefore wrote that value, unmasked, to a process log
and to a stored row — a sink whose whole purpose is to be kept and read later.

The runtime's own gate does not write reasons of that shape today, so nothing is leaking now. The
defect is that the surface has no redaction at all, and the reason string is supplied by an
implementation the runtime does not control.

## The change

One private helper, `masked_audit_event`, builds the event and passes `deny_reason` through
`bounded_masked_log_text` — the masker the log path already uses — before returning it. Both
production call sites now go through it: the deferred-audit path and the intercepted-dispatch path.

Two details are worth stating because the obvious fix is wrong on both:

**The masker cannot live inside `AuditEvent`.** `bounded_masked_log_text` is in
`khive-runtime::secret_gate`, and `khive-runtime` depends on `khive-gate`, not the reverse. Masking
inside `AuditEvent::from_check` would require `khive-gate` to reach up into `khive-runtime`, which is
a dependency cycle. So the masking happens in `khive-runtime`, on the constructed event.

**There are two call sites, not one.** Fixing either alone leaves the other writing the same
unmasked string, and the two paths are far enough apart in the file that the second is easy to miss.
The helper's doc comment says so, so the next reader does not have to rediscover it.

The caller-facing `PermissionDenied` error keeps the unmasked reason. That is deliberate: it is
returned to the caller who was refused, and it reaches no log sink. Masking it would degrade the
error message without closing anything.

## Coverage

Four tests, one per (call site × sink):

| test | call site | sink |
|---|---|---|
| `denied_dispatch_masks_secret_shaped_deny_reason_in_stored_event` | deferred audit | stored event row |
| `intercepted_denied_dispatch_masks_secret_shaped_deny_reason_in_stored_event` | intercepted dispatch | stored event row |
| `dispatch_tracing_emits_gate_check_event_with_masked_deny_payload` | deferred audit | `gate.check` tracing line |
| `intercepted_dispatch_tracing_emits_gate_check_event_with_masked_deny_payload` | intercepted dispatch | `gate.check` tracing line |

Each installs a gate whose deny reason embeds a credential-shaped literal, then asserts the captured
payload does not contain it and does contain the mask. Each also carries a non-vacuity assertion, so
an empty capture fails rather than passing for the wrong reason — without it, a test that captured
nothing at all would report green.


ohdearquant:
---

Scope, as the API reports it rather than as the commit list reads: **1 file, 1 commit**.

PR #2952, opened 2026-09-18
feat(runtime): re-run a pack's entity invariant on the generic update path

### LINKED ISSUE #2943: KindHook is never consulted on the entity-update path, so an entity kind's create-time property validation cannot be enforced on update

`KindHook` gives a pack a place to validate its own kind's properties on write. For **note** kinds
both halves exist: `prepare_create` runs on create, and `prepare_note_update` /
`validate_note_update` run on update, dispatched by `find_kind_hook(&note.kind)` at
`crates/khive-runtime/src/pack.rs:3287`, `:3307` and `:3338`.

For **entity** kinds only the create half exists. The entity-update path is
`prepare_guarded_entity_update` in `crates/khive-runtime/src/curation.rs` (reached by both
`update_entity` and `update_entity_if_unchanged`), and it merges `patch.properties` into the
stored JSON through `merge_properties` without consulting `KindHook` at all. A search for
`kind_hook` or `KindHook` across `crates/khive-runtime/src/curation.rs` returns nothing; the same
search against `crates/khive-runtime/src/pack.rs` in the same pass returns ten hits, so the empty
result is the search working rather than the search being broken.

The consequence is not that some pack forgot to write an override. It is that **no entity-kind
hook's create-time property validation can be enforced on update today**, and no override a pack
author writes would be called.

The reachable instance:

`WorkspaceHook::prepare_create` (`crates/khive-pack-workspace/src/hook.rs:17-33`) requires
`properties.schema_version` to be present and an integer, which the pack's own API document states
as its one registration invariant. `workspace` is an entity kind
(`crates/khive-pack-workspace/src/pack.rs:24-25`: `NOTE_KINDS` is empty, `ENTITY_KINDS` carries
it). So:

```
create(kind="entity", entity_kind="workspace", properties={"schema_version": "v2"})  -> refused
update(id=<workspace>,                         properties={"schema_version": "v2"})  -> stored
update(id=<workspace>,                         properties={"schema_version": null})  -> cleared
```

The only per-field guard on that path is `validate_entity_type_for_kind`, which checks the
`entity_type` subtype axis and not `properties.*`. A grep for `schema_version` inside
`crates/khive-pack-workspace/` finds it in the create hook and in two create-only integration
tests; there is no update-path test for it, which is consistent with there being no update path to
test.

Two ways to close it, and the choice is a design call rather than an obvious fix:

1. Call the hook at the seam, in `prepare_guarded_entity_update` itself, on the **merged**
   properties — one call site, after the patch has been applied, so the check sees the record as
   it will be stored rather than the patch in isolation. That function reaches pack-owned
   behaviour the way this file already does everywhere else: through an installed slot on
   `KhiveRuntime`, not a `VerbRegistry` lookup. The precedent is two lines away in the same
   function — `validate_entity_type_for_kind` reads
   `entity_type_validator: Arc<RwLock<Option<EntityTypeValidatorFn>>>`, installed by the transport
   after the registry is built, alongside `edge_rules`, `valid_entity_kinds` and the
   note-mutation hook.

   Pair it with a `KindHook` trait default that re-runs the create-time property validation, so a
   pack author gets update enforcement without writing an override at all — which matters because
   the failure this issue describes is precisely that nobody wrote one. That makes every existing
   entity-kind hook's create contract enforceable and stops the class rather than one instance of
   it.
2. Move the invariant out of the hook and into the storage layer for entity kinds that declare
   required properties, so it does not depend on a hook being called at all.

Option 1 is the smaller change and the one that matches how note kinds already work; option 2 is
the one that does not depend on every future pack author remembering to implement a trait method.
Either way the tri-state patch contract applies: key absent means untouched, explicit `null` means
clear, a value means set — and for a required property, `null` has to be refused rather than
treated as a clear.

Found while enumerating the packs that register a real `kind_hook` (six of eight override sites
register one; two always return `None`) after fixing the note-side instance of this shape in #2938.
The note-side fixes each closed one field on one hook; this is the same shape at the seam, where
it closes every field on every entity-kind hook at once.

ohdearquant:

Closes #2943.

## The gap

`KindHook` let a pack guard its entity kind at create time, through `prepare_create`, and gave it no
way to say anything about an update. The shared entity update path ran the entity-type validator and
the properties merge and then wrote. So a value the owning pack would have refused on create could be
written by updating an existing record instead — the invariant held on one entry path and not the
other.

`khive-pack-workspace` is the concrete case: it requires an integer `properties.schema_version` on
create, and an update could replace it with a string, or drop it, and be accepted.

## The change

A new `KindHook::validate_entity_update`, defaulting to accept, dispatched from
`prepare_guarded_entity_update`.

**Where it runs is the design decision.** It runs *after* the caller's patch has been merged into
the stored properties, so the hook sees the resulting record rather than the raw patch. The
invariant a pack enforces is a claim about the record — "this key must be present and an integer" —
not about what one caller happened to send, and a partial update that omits the key must still be
checked against the merged result. It also runs after tag and entity-type handling, and after the
`changed_fields.is_empty()` early return, so a no-op update does not invoke the hook.

**It is deliberately not a re-run of `prepare_create`.** That was the obvious shape and it is wrong:
a `prepare_create` body may also enforce create-shape requirements — arguments a caller must supply
at create time — that a partial update legitimately omits, so re-running it would reject valid
updates, with an error message written for create. Sharing one predicate between the two methods
keeps both honest without coupling their argument shapes. That is how the note side already works,
where `validate_note_update` implementors share their predicate with their create path.

**Wiring.** `khive-runtime` does not hold a `VerbRegistry` — ownership runs the other way, packs are
constructed from a runtime handle — so `VerbRegistry::entity_kind_hooks` aggregates the
`(entity kind, hook)` pairs for which the owning pack declares the kind, and the transport installs
them on the runtime once the registry is built. Same extension point and timing as the entity-type
validator. The aggregate holds no reference back to the pack or registry that produced it, since
every production `kind_hook()` constructs a fresh stateless hook, so installing it creates no
ownership cycle.

Both transports install it. So does the entity-type backfill, whose apply path builds its own write
runtime and reaches `update_entity_if_unchanged` directly: it is the only production caller of that
function outside `khive-runtime` itself, so without that line its writes would have been the ones
that skipped the new hook.

## Coverage

`khive-pack-workspace` becomes the first implementor: the `schema_version` requirement moves into a
shared `require_integer_schema_version`, which `prepare_create` and `validate_entity_update` both
call, so a value refused on create is refused on update.

Runtime tests use a probe hook registered directly on a bare runtime, so their subject is the
dispatcher rather than any one pack: the hook is reached on a real update, a hook that does not
override `validate_entity_update` does not block the update, and a kind with no registered hook
updates freely. Pack integration tests cover the workspace predicate end to end.

## Scope

10 files, 1 commit. 10 files changed, 569 insertions(+), 16 deletions(-)

## Verification

Full workspace suite at the rebased head (56cee90aa, on top of cfd70dbae):

- `cargo test --workspace` rc 0 — **11487 passed, 0 failed, 32 ignored** across 266 suites

The eight test functions this commit adds were checked by name against the suite output, with the
list derived from the commit diff rather than typed, and all eight report `ok`:

```
compose_registry_installs_entity_kind_hooks_for_backfill_writes
entity_update_dispatches_installed_kind_hook_acceptance
entity_update_dispatches_installed_kind_hook_refusal
entity_update_with_hook_missing_the_default_method_still_succeeds
entity_update_with_no_installed_hook_for_kind_succeeds
every_registered_entity_kind_hook_is_on_the_validated_allowlist
update_workspace_accepts_integer_schema_version
update_workspace_rejects_non_integer_schema_version
```


oceanwaves630 (approved):
Approving.

The generic update path now runs the same pack invariant the create path already ran. What I checked at this head:

- `KindHook::validate_entity_update` defaults to `Ok(())`, so every existing pack keeps its current behaviour and only a pack that opts in gains the check. Nothing silently becomes stricter.
- The dispatch sits after the no-op early return, so an update that changes nothing still does no work and cannot be rejected by an invariant it never had to satisfy before.
- All three install sites are covered, including the entity-type backfill path. That one is easy to miss, because it does not go through either of the two boot paths and would otherwise have been the one place the invariant was absent.
- The workspace pack's two methods now share a single predicate for the integer schema-version requirement instead of carrying separate copies that could drift apart.
- The added tests state the mutation they expect before asserting it, so a hook that stopped firing would redden rather than quietly pass.


Checks at this head: 26 successful, 3 skipped, none failed.

PR #2953, opened 2026-09-18
fix(runtime): split note-update normalization from its sequencer

### LINKED ISSUE #2947: KindHook: validate_note_update_hook has no callers, so a prepare_note_update override silently retires its own validator

`KindHook` gives a pack two places to run note-update validation, and one of them is no longer
reachable. The task hook now enforces its dependency invariant from a second, inline copy, so the
copy that reads like the validator is dead.

## What is unreachable

`VerbRegistry::validate_note_update_hook` (`crates/khive-runtime/src/pack.rs:3300`) has **zero call
sites**. Its own doc comment describes it as "the validation-only compatibility seam for callers
that do not own a mutable request object"; no such caller exists in the workspace.

```
$ grep -rn "validate_note_update_hook" crates --include="*.rs" | wc -l
1                                   # the definition itself
$ grep -rn "prepare_note_update_hook" crates --include="*.rs" | wc -l
5                                   # control: this one has real callers
```

That wrapper is the only caller of the trait method `KindHook::validate_note_update` other than the
default body of `KindHook::prepare_note_update` (`pack.rs:509-519`), which forwards to it.

## Why that matters for one kind in particular

Four hooks override `validate_note_update` and leave `prepare_note_update` at its default, so their
validation still runs through the forwarding default. The task hook is different: it overrides
**both**.

- `crates/khive-pack-gtd/src/hook.rs:352` — `validate_note_update`, body
  `dependency::validate_property_update(...)`
- `crates/khive-pack-gtd/src/hook.rs:363` — `prepare_note_update`, which normalizes coupled fields
  and then calls `dependency::validate_property_update(...)` **directly**, not through
  `self.validate_note_update(...)`

Because the override replaces the forwarding default, and the wrapper that would call the trait
method has no callers, `TaskHook::validate_note_update` cannot be reached on any live path. The
invariant is still enforced today, by the duplicate call in `prepare_note_update`. So this is not a
correctness bug right now.

It is a durability bug. The invariant has two copies and only one of them can run:

- Reading the hook to answer "how are task note updates validated" leads to the dead override first,
  since that is the method with the matching name.
- A future change to the dependency check that lands in the unreachable copy compiles, passes review,
  and changes nothing.
- A test written against `validate_note_update` exercises code the product never calls, and reads
  as coverage for the path that actually runs.

The same trap is open for any kind that later overrides `prepare_note_update`: the override silently
takes its `validate_note_update` out of service, with no error and no warning.

## Suggested shape

Either give the seam a caller, or collapse the duplication:

1. Delete `validate_note_update_hook` and have a `prepare_note_update` override call
   `self.validate_note_update(...)` as its last step, so there is one copy of each invariant and the
   default and the override reach it the same way. The task hook's inline call becomes that call.
2. Or keep the wrapper and give it the caller its doc describes, so the trait method is reachable
   without a mutable request object.

Whichever way, a test that fails when a `prepare_note_update` override omits its validation step
would close the class rather than this instance.

ohdearquant:

Closes #2947.

## The gap

`KindHook` had two note-update methods and the relationship between them was a trap.
`prepare_note_update` was a provided method whose default body forwarded to
`validate_note_update`. A pack that needed to normalize coupled fields overrode
`prepare_note_update` — and in doing so silently replaced the forwarding, taking its own validator
out of service with no compile error and no warning.

`VerbRegistry::validate_note_update_hook`, which would have called the trait method directly, had
**zero call sites**. So an override removed the only remaining path to `validate_note_update` for
that kind.

The population, from `grep -rn 'fn validate_note_update\|fn prepare_note_update' crates --include='*.rs'`:

| pack | overrides | state before this change |
|---|---|---|
| `khive-pack-gtd` | both | `validate_note_update` unreachable; the dependency invariant survived only because `prepare_note_update` repeated the check inline. Two copies, one reachable. |
| `khive-pack-memory` | `prepare_note_update` only | latent — no validator today, so nothing lost; a validator added there later would be dead on arrival. |
| `khive-pack-code`, `khive-pack-comm`, `khive-pack-git` (×2) | `validate_note_update` only | fine, reached through the forwarding default. |

## Why neither obvious fix was taken

Giving the dead wrapper a caller, or having the gtd override call `self.validate_note_update(...)`
as its last step, both fix this instance and leave the next `prepare_note_update` override free to
drop validation exactly the same way.

Instead, normalization and sequencing are separated:

- **`normalize_note_update`** is new, provided, defaults to a no-op, and is the method a
  kind-owning pack overrides for coupled-field work.
- **`prepare_note_update`** becomes the sequencer: `normalize_note_update`, then
  `validate_note_update`, unconditionally.
- `validate_note_update_hook` is deleted.

The gtd hook's body moves to `normalize_note_update` minus its trailing duplicate validation call,
so `validate_note_update` — the surviving single copy of the dependency invariant — is reachable
again. `khive-pack-memory` and the two test-module hooks are mechanical renames with unchanged
bodies.

**The guarantee is weaker than it reads, and the doc comment says so.** Rust does not prevent an
override of a provided method, so this removes a footgun by naming; it is not an invariant enforced
by the type system. What enforces it is the coverage below, plus the fact that after this change no
pack in the tree overrides the sequencer at all.

## Coverage

Two tests whose subject is the dispatcher rather than any one pack. A probe hook overrides
`normalize_note_update` (moving a caller-supplied top-level field into `properties`) and
`validate_note_update` (refusing based on what it finds there), and never overrides
`prepare_note_update`. The value the validator inspects does not exist in `properties` until the
normalizer puts it there, so the assertion that the validator *saw* it proves the ordering, not just
that both ran.

One arm ends in the refusal, one in acceptance; both assert that normalization landed in `args`
either way, and both assert each method ran exactly once. Each carries the mutation that reddens
it: revert the sequencer to its pre-change body and the normalizer never runs, so the validator
observes `None` where it expects a value.

The existing gtd regression
(`khive-pack-gtd/tests/dependencies.rs`, `generic_update_rejects_direct_and_multihop_property_cycles`)
already reaches the dependency invariant through `prepare_note_update_hook` via the generic `update`
verb, and pins the reported instance unchanged.

## API note

`KindHook` is a public trait and a provided method is renamed, so this is a breaking change for any
out-of-tree implementor. There are none in this workspace.

## Scope

8 files, 4 commits. 8 files changed, 306 insertions(+), 26 deletions(-)

Five Rust files, one per-crate design document, and two ADRs. The ADRs ride with the code rather
than following it because they describe the seam this changes: leaving them for later means an
accepted document states something false about main for as long as the gap lasts.

## Verification at the rebased head

Gated twice. The second run is the one that counts: the first ran on a base that predates
023e03bbc, and both that commit and this one change
`crates/khive-runtime/src/pack.rs`, so a suite on the older base is a strict subset of what the
merge ref exercises.

At the rebased head `36172aed2`, on top of `023e03bbc`:

- `cargo test --workspace` rc 0 — **11489 passed, 0 failed, 32 ignored** across 266 suites

The rebase was checked for faithfulness rather than assumed from a clean exit: for each of the 5
changed paths, the set of added and removed lines is identical before and after, `pack.rs` included
at 279 changed lines. A clean three-way merge can otherwise keep both far-apart edits or drop one
without conflicting.

The two arms this commit adds, with the list derived from the commit diff rather than typed:

```
prepare_note_update_hook_runs_normalize_before_validate_and_validate_can_accept
prepare_note_update_hook_runs_normalize_before_validate_and_validate_can_refuse
```

Both report `ok`, as does the existing property-cycle regression that exercises the migrated
override: `generic_update_rejects_direct_and_multihop_property_cycles`.

## Two corrections made on this branch

**The branch had deleted `VerbRegistry::validate_note_update_hook`.** That was out of scope and it
is restored. The method's only statement is a call to `KindHook::validate_note_update`, which this
branch does not touch, so keeping it costs nothing; removing it broke a public entry point for
callers that hold no mutable request, and ADR-017 states the registry retains it. Its doc comment
now says what reaching a hook through it actually does: runs the validator alone, without
`normalize_note_update`. That is the point of the seam, and the reason a caller that can supply a
mutable request should not use it.

**Three ADR statements were left false by the rename, and are corrected.** ADR-017's trait sketch,
ADR-017's prose instructing pack authors to override the exact method they now must not, and
ADR-019's hook sketch. A fourth, ADR-019's file-summary line, surfaced by re-deriving mentions of
the renamed method under `docs/` after the first three were fixed. A code-anchored change does not
see the documents describing it, so that sweep has to be run deliberately rather than left to
turn up on its own.

That re-derivation was not complete, and an earlier version of this description said it was. Two
further references survived it, because they are prose that does not live under `docs/`: the
per-crate `crates/khive-pack-gtd/docs/design.md`, and the module doc comment at the top of
`crates/khive-pack-gtd/src/hook.rs`. Both name the new method now. The population a rename has to
clear is every place the method is described in prose, which is not the same set as every file under
one directory, so the sweep now runs over the whole tree, with a mention of the new method required
in the same pass as the absence of the old one.

ADR-017 gains a dated sequencing amendment in the style that section already uses: what moved, why,
and the limit that remains. Rust does not prevent a pack from overriding the sequencing method too,
so the ordering guarantee rests on naming and on test coverage rather than on the type system. The
amendment states that removing that limit is the accepted direction rather than an open question,
and cites #2956, where the sequencing moves off the trait entirely so a pack cannot express an
ordering at all. That move is #2956's change, not this one.

Re-gated at `5cf1b7ce2`: fmt 0, test 0 (**11489 passed, 0 failed**, 266 suites), clippy 0, both
sequencing arms `ok`, the gtd property-cycle regression `ok`. The final commit is markdown only,
and that is asserted rather than eyeballed: zero non-markdown paths differ between that gated head
and this one, with a paired control requiring at least one markdown path so the check cannot pass
on an empty commit.

## Mutation notes recorded declaratively

The two sequencing tests carry a note describing the mutation that reddens them. Those notes were
written in the imperative — "revert this, then that happens" — and are now written as a record of
what was run and what reddened.

The content is unchanged. The mood is, because a checked-in tree should not carry text that reads as
an instruction to modify it: prose in a public repository is parsed by tools as well as people, and
imperative text is the shape that invites being acted on rather than read.

## Verification at the final head

- `cargo test --workspace` rc 0 — **11489 passed, 0 failed** across 266 suites

Named arms, each checked by exact name:

```
prepare_note_update_hook_runs_normalize_before_validate_and_validate_can_accept   ok
prepare_note_update_hook_runs_normalize_before_validate_and_validate_can_refuse   ok
generic_update_rejects_direct_and_multihop_property_cycles                        ok
```

The gate also refuses if either stale prose reference returns, or if the must-match control finds no
mention of the new method — so a future rename cannot quietly reopen the gap this closed.


ohdearquant (commented):

Verdict on head 36172aed2a27857fc66f2e43ab6a264a39eb9a53: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Read at e233434e, including the delta over the head I read first. That delta is prose in three files and nothing else: two references renamed from `prepare_note_update` to `normalize_note_update`, and two test doc comments rewritten from an imperative procedure into the result that was actually observed.

The rename is the accurate one. `prepare_note_update` is now the sequencing method on the runtime trait: it calls `normalize_note_update` and then `validate_note_update`, in that order. The gtd hook overrides the two extension points and not the sequencer, so a pack that supplies a normalizer gets it run before validation without having to remember to chain the validator itself. That is what makes the refusal reachable from the update path rather than only from create.

Verified rather than taken from the description: the sequencing is at `crates/khive-runtime/src/pack.rs`, the gtd hook defines `normalize_note_update` and `validate_note_update` and no `prepare_note_update`, and the two tests named after that ordering exercise both the refusing and the accepting case.

ohdearquant (commented):

PR #2954, opened 2026-09-18
fix(code): check finding evidence on the update path, not only at create

### LINKED ISSUE #2941: finding evidence must be an array at create and is unchecked on the update path

A `finding` record's `evidence` property must be a JSON array at creation
(`crates/khive-pack-code/src/hook.rs:122-128`, an inline `if !v.is_array()` in
`FindingHook::prepare_create`). The generic property-update path does not check it.

`FindingHook::validate_note_update` (`crates/khive-pack-code/src/hook.rs:168-201`) loops the
shared `FINDING_ENUMS` table only — `kind_status`, `severity`, `confidence`. `evidence` appears
nowhere in that function. So:

```
create(kind="note", note_kind="finding", properties={"evidence": 42})   -> refused
update(id=<finding>,                     properties={"evidence": 42})   -> stored verbatim
```

The update seam is reached by the ordinary generic `update` verb, with no kind-specific gating:
`crates/khive-pack-kg/src/handlers/update.rs:302` calls `prepare_note_update_hook`, which resolves
the hook by `find_kind_hook(&note.kind)` at `crates/khive-runtime/src/pack.rs:3287`. That is the
same chain the `commit`/`issue`/`pull_request` kinds use.

Why it is worth fixing rather than noting: the row stays present and well-formed either way, so no
shape check, schema check or round-trip catches it. Every reader of `evidence` written against the
create-time contract — anything that iterates it, counts it, or indexes into it — is reading a
field whose guarantee silently does not hold for records that have ever been edited.

Suggested fix: hoist the array check into a named predicate and call it from both
`prepare_create` and `validate_note_update`, the same way the enum fields already share
`validate_finding_enum`. The tri-state patch contract applies: key absent means untouched, an
explicit `null` means clear, a value means set — so the update-path check keys on presence, not on
truthiness.

Test worth having: one arm that updates `evidence` to a non-array and asserts the refusal. Without
it the fix is unfalsifiable, because every other assertion about the record passes whether the
check runs or not.

Found while looking for other instances of the shape fixed in #2930 — which covered exactly the
three enum fields on this hook and did not cover this one.

ohdearquant:

Closes #2941.

A `finding`'s `evidence` property must be a JSON array at creation and was accepted unchecked on
the update path:

```
create(kind="note", note_kind="finding", properties={"evidence": 42})   -> refused
update(id=<finding>,                     properties={"evidence": 42})   -> stored verbatim
```

## Why the two paths disagreed

`FindingHook::prepare_create` checks the key with an inline `is_array` arm.
`FindingHook::validate_note_update` loops the shared `FINDING_ENUMS` table — `kind_status`,
`severity`, `confidence` — and `evidence` is not in that table, so the update seam never looked at
it. The generic `update` verb reaches that validator with no kind-specific gating, so an ordinary
edit could leave a finding holding a value no create could have written.

## The rule applied

The check now runs on both paths under the tri-state patch contract the enum fields already follow:

| key state in the patch | behaviour |
|---|---|
| absent | stored value left alone |
| present, `null` | cleared — allowed |
| present, non-null | must be an array, same as create |

Clearing is allowed because `evidence` is optional at create, so a cleared finding is a state the
create path can also produce. That is the same reasoning the file already applies to `severity` and
`confidence`, and the reason `kind_status` is refused instead of cleared: create defaults it, so
every finding carries one.

## Arms

Exactly one of these reddens with the fix reverted, which is what makes the fixture isolating rather
than merely green:

| arm | with the fix reverted |
|---|---|
| `create_finding_rejects_non_array_evidence` | passes — control, must not move |
| `update_finding_rejects_non_array_evidence` | **fails** — the defect |
| `update_finding_accepts_array_evidence` | passes |
| `update_finding_clears_evidence_on_null` | passes |

`prepare_create`'s array check had no test of its own before this change, so the control arm is new
coverage rather than a restatement. The crate does already test `evidence` elsewhere:
`ingest_rejects_invalid_evidence_shape` covers `ingest_findings_json` refusing an evidence *entry*
that is neither a string nor an object. Different function, different error type, and it never
reaches the hook.

## One adjacent gap this does not close

Noting it because the ingest test above makes it visible. `ingest_findings_json` validates each
evidence *entry* (string or object); `prepare_create` and now `validate_note_update` only require
that `evidence` is an array. So `{"evidence": ["src/lib.rs:42", 42]}` is accepted by both verb paths
and rejected by ingest. Bringing the entry rule to the verb paths is a larger change with its own
compatibility question for existing rows, so it is deliberately out of scope here rather than
overlooked.

## Scope

2 files, 2 commits. 2 files changed, 127 insertions(+), 8 deletions(-)

## Verification

At `d1bb6095f`, rebased onto `023e03bbc`:

- `cargo test --workspace` rc 0 — **11491 passed, 0 failed** across 266 suites

The rebase was checked rather than assumed from a clean exit: for each changed path the set of added
and removed lines is identical before and after (15 and 75 lines). A clean three-way merge can
otherwise keep both far-apart edits or silently drop one.

The four arms this commit adds, with the list derived from the commit diff rather than typed, each
required to appear with `ok`:

```
create_finding_rejects_non_array_evidence
update_finding_accepts_array_evidence
update_finding_clears_evidence_on_null
update_finding_rejects_non_array_evidence
```

All four ran and passed.

Mutation check, predictions written before running: reverting only `hook.rs` to its state on `main`
reddens exactly one arm, `update_finding_rejects_non_array_evidence`, whose panic prints the
rejected value. That is the defect itself rather than a proxy for it, and the create-path control
stays green through the revert, so the two arms cannot both pass on one mechanism.

## A correction made on this branch

The update check originally admitted an explicit `null`, reasoning that `evidence` is optional at
create, so clearing it lands in a state create can also produce. **That reasoning was wrong about
what a null does, and the measurement settles it.** Updating with `properties: {"evidence": null}`
leaves the stored row as:

```
{"evidence": Null, "kind_status": "open"}
```

The generic property merge stores the null rather than removing the key. So admitting one persists
`properties.evidence: null` — a shape no create call can produce, since create refuses any present
`evidence` that is not an array.

Both write paths now route through one `validate_finding_evidence` holding the same rule: present
means array, and a null is present. That also removes the duplicated block and its duplicated error
string. The file's `FINDING_ENUMS` table already carries the argument for doing it this way, in a
doc comment about two writers of one rule drifting apart; `evidence` was the field left outside that
pattern, and it drifted exactly as described.

### The test that should have caught it could not

It asserted:

```rust
updated["properties"]["evidence"].is_null()
```

under the message "evidence must be gone". Indexing a missing key in `serde_json` yields
`Value::Null`, so that form reads true whether the key was removed or stored as a null — the two
states the test existed to tell apart. It is replaced by an arm asserting the refusal, and asserting
that the refused update left the row unchanged rather than half-applied.

The same assertion shape sits in a pre-existing arm for the optional closed-set fields, and the
behaviour there is the same: clearing `severity` with a null stores `"severity": Null`. That is not
this change — it is #2957, with both measurements, because fixing it alters update semantics for
fields this branch does not touch and the contract question there has two coherent answers.

## Verification at the final head

- `cargo test --workspace` rc 0 — **11491 passed, 0 failed** across 266 suites

The four evidence arms, each checked by exact name:

```
create_finding_rejects_non_array_evidence   ok
update_finding_rejects_non_array_evidence   ok
update_finding_accepts_array_evidence       ok
update_finding_rejects_null_evidence        ok
```

The gate also asserts structurally that both write paths still reach one validator — one definition
and two call sites — so they cannot quietly stop sharing it the way they did before.


ohdearquant (commented):

Verdict on head d1bb6095f251c043af309b10598c7e99ca8a84ca: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

oceanwaves630 (approved):
Verified that `validate_finding_evidence` is reached from both write paths: `prepare_create` and `validate_note_update`. A null or malformed `evidence` value is now refused on create and on update alike, through one function rather than two copies of the check.

I also enumerated the other validated properties instead of reading the one field, to see whether the asymmetry was a class. `kind_status` already had its own null refusal on the update path, the enum-valued keys ran on both sides before this change, and title/content are create-only by construction. `evidence` was the only property validated on create and not on update, so this closes the set.

PR #2955, opened 2026-09-18
fix(runtime): give audit retry exhaustion its own terminal reason

### LINKED ISSUE #2950: audit: retry exhaustion and a hard store fault both report StoreFailure, so the cause is unrecoverable from the error

`AuditTerminalReason::StoreFailure` is produced by two paths that mean opposite things, and the
reason code is the only thing a caller gets to tell them apart.

## The two producers

`crates/khive-runtime/src/audit_batch.rs:511` — the classifier's fallback arm, for a store error
that is genuinely not retryable:

```rust
_ => RetryDecision::Terminal(AuditTerminalReason::StoreFailure),
```

`crates/khive-runtime/src/audit_batch.rs:577` — the commit loop, when a **retryable** error runs out
of attempts:

```rust
RetryDecision::Retry => {
    return GenerationResult::Failed(AuditTerminalReason::StoreFailure)
}
```

So a transient condition that the classifier deliberately marked `Retry` is reported, on exhaustion,
under the same terminal reason as a hard storage fault.

## Why that matters here specifically

The arm immediately above the fallback is explicit about which transient condition it has in mind
(`audit_batch.rs:492-498`):

> Transient availability conditions, not judgments on the batch. The events-daemon forwarding lane
> (ADR-170) reports an unreachable or stalled daemon as `Pool`/`Timeout` [...] Both are exactly what
> the configured bounded retries exist for — treating them as terminal would abandon a generation on
> the first blip of a daemon restart.

The retries sized for that condition are, at `audit_batch.rs:287-288`:

```rust
max_commit_attempts: std::num::NonZeroU8::new(3).unwrap(),
retry_backoff: Duration::from_millis(20),
```

Three attempts with two backoff sleeps, so roughly **40 ms** of tolerance. `AuditBatchConfig` is
not overridden anywhere outside tests (`pack.rs:659-660, 845-851` take `Option<_>` and fall back to
`Default`), so that is what a default deployment runs.

A restart of the events daemon measured here took about **10 seconds** from signal to the
supervisor's replacement being up, which is around 250x the retry budget. The comment names daemon
restarts as the thing the retries protect against; the budget does not cover one.

## Observed

A `comm.mark_read` issued inside such a restart window returned a domain-committed result with an
audit obligation failure whose reason was `store_failure`. The rows did flip to read, so
`fold_audit_obligation` behaved as designed (`pack.rs:4900-4913`) — the problem is not the envelope,
it is that `store_failure` is also what a real storage fault reports, so the cause is unrecoverable
from the error.

I cannot state which arm produced that particular failure, and that is the point of the report: with
a single reason code covering both, neither an operator nor an automated retry policy can tell a
restart window from a storage fault.

## Two separable changes

1. **Give retry exhaustion its own terminal reason** carrying the transient classification that was
   retried (`Pool` / `Timeout` / writer-task state). This is the part that affects callers: it makes
   the two causes distinguishable without changing any tolerance.
2. **Size the tolerance against the condition the comment names**, or narrow the comment to say
   restarts are out of scope. 40 ms does not survive a daemon restart, so as written the code
   documents an intent it does not implement.

(1) is worth doing on its own even if (2) is declined, since a longer budget still terminates
eventually and still reports the wrong reason when it does.

ohdearquant:

Closes #2950, part 1 of the two separable changes that issue lists.

## The defect

`AuditTerminalReason::StoreFailure` had two producers that mean opposite things, and the reason code
is the only thing a caller gets to tell them apart.

| producer | what it means |
| --- | --- |
| `classify_store_error`'s fallback arm | the store error is not retryable |
| `run_generation`'s commit loop, on exhaustion | the classifier judged the error retryable, and every configured attempt still failed |

The arm directly above the fallback is explicit about which conditions it has in mind — write-queue
and writer-task pressure, and the `Pool` / `Timeout` shapes the forwarding lane reports for an
unreachable or stalled daemon. Those are exactly the conditions the bounded retries exist for. On
exhaustion they were reported under the same code as a hard storage fault.

## The change

One new variant, `RetryExhausted`, wire code `retry_exhausted`, returned from the exhaustion arm.

Nothing else moves. No tolerance, deadline, retry count or backoff changes, and no classifier
decision changes. `StoreFailure` keeps its meaning and its wire code; its doc comment is narrowed to
what it actually covers, including the case a reader would get wrong — a generation whose earlier
attempts failed retryably and whose next one returns a non-retryable error reports `StoreFailure`,
because the attempt that decided the outcome is the non-retryable one.

A unit variant rather than one carrying the retried classification. The enum has exactly one
exhaustive match in the tree, `AuditObligationFailure::wire_code`, so this costs one arm. Carrying
the classification would require `RetryDecision::Retry` itself to carry it, touching every arm of
`classify_store_error` plus a new supporting enum, in order to expose detail no caller asked for.
It also follows the enum's own precedent: `ResolutionDeadlineExpired` is a bare variant kept
distinct from `AdmissionDeadlineExpired` for exactly this purpose.

## Reach

Every production reference to the enum was enumerated rather than sampled:

```
grep -rn --include='*.rs' 'AuditTerminalReason::' crates/
```

- `error.rs` `wire_code` — the only exhaustive match. One arm added.
- `pack.rs` obligation-degrade branch — equality tests against two specific variants, not an
  exhaustive match. `RetryExhausted` falls through it the same way `StoreFailure` did, which is the
  correct behaviour: it is not an admission-pressure reason.
- Every other reference constructs one specific variant or names one in a doc comment.

## Arms

The existing exhaustion test asserted `StoreFailure`, and tracing its fixture shows why: its store
fails with `WriterTaskRequestFailed { TransactionRolledBack }`, which the classifier returns `Retry`
for, so the default three attempts exhaust and it lands on the exhaustion arm. It was the only test
pinning that behaviour, so it now asserts `RetryExhausted`, with the reason written into the
assertion message.

That alone would leave the other side unpinned, so a second test covers it: a store error with no
explicit classifier arm still reports `StoreFailure`, and the same test asserts the store was called
exactly once. The two arms therefore cannot both pass on a single mechanism — one requires retries to
have run and one requires that they did not.

## No specification change

ADR-133 sets the wire `code` to "the terminal reason (`admission_deadline_expired`, `store_failure`,
and the other `AuditTerminalReason` values in snake case)". That is a reference to the enum rather
than a closed list, so a new variant renders under the existing contract and needs no amendment.

## Out of scope, deliberately

Part 2 of the issue — sizing the retry budget against the condition the comment names — is left
open. The two changes are separable, and this one changes no behaviour a deployment depends on: the
same generations fail, under a more precise name.

## Scope

3 files, 1 commit. 3 files changed, 84 insertions(+), 4 deletions(-)

## Verification

At `3a0d95504`, rebased onto `023e03bbc`:

- `cargo test --workspace` rc 0 — **11488 passed, 0 failed** across 266 suites

The rebase was checked rather than assumed from a clean exit: for each changed path the set of added
and removed lines is identical before and after.

Both arms report `ok`. The added one is derived from the commit diff rather than typed:

```
d8_non_retryable_store_error_still_reports_store_failure
```

The repointed one is named separately, because an instrument that derives its population from added
test functions is structurally blind to a test whose assertion changed:

```
d8_exhausted_commit_retries_set_the_lifetime_degraded_flag
```


ohdearquant (commented):

oceanwaves630 (approved):
Approving.

What I checked at this head, reading the source at the ref rather than the diff alone:

- The split is well defined by the classifier and not by the call site. `classify_store_error` returns `Retry` only for the write-queue, writer-task, pool, timeout and writer-request-state shapes, and its fallback arm returns `Terminal(StoreFailure)`. So the new reason means exactly "the classifier said retry on every attempt and the attempts ran out", and the old one keeps everything the classifier judged hopeless.
- One return value changed, at the exhaustion branch. The `Terminal(reason)` branch is untouched, so a non-retryable error arriving on a later attempt after earlier retryable ones still reports the old reason. That is what the new doc comment on `StoreFailure` claims, and the loop does it.
- I enumerated every site in the workspace that reads the old reason or its wire string, to check that nothing silently stops matching. Outside the two production sites in `error.rs` (the wire mapping, updated here, and a `from_store` constructor on a path that never runs the batch loop), every remaining occurrence is a test fixture. No production consumer branches on the old reason to decide behaviour, so no behaviour is lost by the rename on this one path.
- `append_events_idempotent` is idempotent by contract, which is what makes the retryable set safe to retry at all, including the ambiguous writer-request state. The new reason's advice to wait and try again rests on that.
- The added test arms the non-retryable shape through a distinct counter rather than reusing the retryable one, and asserts the call count is 1, so it pins that a non-retryable classification is not retried at all. That assertion is what makes the arm about classification rather than about the returned value.

Two non-blocking follow-ups, neither of which should hold this up:

- ADR-133 enumerates the terminal reasons by name in two places and describes that set's behaviour, and neither enumeration now mentions the new reason. The behaviour is unchanged, so the code is right and the document is incomplete, which is the direction that decays quietly.
- The doc comment describes a mixed sequence, retryable attempts followed by a terminal error, and no test pins that sequence. The two arms here cover retryable-throughout and terminal-on-the-first-attempt.

One API note: the enum carries no `#[non_exhaustive]` and its own documentation calls itself exhaustive, so adding a variant is source-breaking for any out-of-tree exhaustive match. In-tree is covered, which the green build shows. Worth a changelog line if this crate is published.

Checks at this head: 22 successful, 7 skipped, none failed.

PR #2961, opened 2026-09-18
fix(ci): exempt the secret gate's test constants by value and path, not by fingerprint

ohdearquant:

## Scope

3 files, 2 commits. 3 files changed, 121 insertions(+), 6 deletions(-)

Two scanner configuration files and the test that fixes their behaviour in place. No source file
changes, and in particular the fixture itself is untouched: the constant that trips the scanner is
the input a test needs, so the answer is to describe it correctly to the scanner rather than to
weaken the test.

## The failure

The scheduled full-history run of `Secret scan (gitleaks)` on `main` has failed every day since
2026-09-12. Pull-request runs are green, and that is by design: they scan the change range, so the
finding sits outside what they look at. The scheduled run takes no range and scans every ref, which
makes it the only job in the matrix that sees the whole repository.

```
09-18  failure    09-15  failure    09-12  failure
09-17  failure    09-14  failure    09-11  success
09-16  failure    09-13  failure
```

The finding is `572aec015…:crates/khive-runtime/src/secret_gate.rs:generic-api-key:3963`, a
synthetic constant that a secret-gate regression test formats into an api-key-shaped string and
asserts the gate refuses. The value is credential-shaped on purpose: a test that the gate blocks a
credential needs something the gate will block.

## Why the exemption expired

The exemption that covered this fixture was a fingerprint in `.gitleaksignore`, which is
`commit:path:rule:line`. Two of those four coordinates change whenever the file is edited at all.
The fixture moved from line 3338 to line 3963 and arrived under a new commit, so the entry stopped
matching — not because anything about the value changed, but because the file around it did.
`secret_gate.rs` is over 8000 lines and is edited often, so the same expiry is due again on roughly
every future edit.

The inline `// gitleaks:allow` marker that now sits beside the fixture fails from the other
direction. A history scan reads each blob as it was, and the blob that carries the finding predates
the marker. A marker at the tip cannot exempt a commit that has already happened.

## What this changes

`.gitleaks.toml` already carried the shape that survives an edit: the cache-lock fixtures are exempt
by value and path, described in `.gitleaksignore` as covering them "independent of their commit and
line number", with an arm in `scripts/tests/test_gitleaks_config.py` proving they survive a line
move under a new commit. One fixture had that treatment and the other did not.

The secret gate's two constants now get the same allowlist: two exact values, anchored, at one exact
path. The fingerprint they no longer need is removed, and the comment in `.gitleaksignore` now says
what a fingerprint costs, so the next person reaching for one reads the trade-off first.

The exemption is deliberately narrow, and the tests are what make that checkable rather than
asserted:

| arm | what it fixes in place |
| --- | --- |
| exempt at the gate path | the two constants pass where they live |
| not exempt at three other paths | including a prefixed and a suffixed spelling of the same path |
| superstrings and a reversed sibling still reported | the exemption is two values, not a file |
| survives a line move under a new commit | with a positive control proving the scanner re-attributes the values to the new commit when the exemption is removed |
| each exempt value still occurs in the file it names | a dead exemption fails loudly instead of quietly widening what the scanner skips |

## Verification

**The gate's own command, run over the same commit scope with only the config and ignore file
differing.** 

Two runs three minutes apart over the repository's own refs, the second differing only in which
config and ignore file it was handed:

```
BEFORE  6634 commits scanned   leaks found: 7
AFTER   6636 commits scanned   leaks found: 4
```

The three that disappear are exactly the three at `crates/khive-runtime/src/secret_gate.rs`,
including the one the scheduled job reports. The four that remain are findings that job does not
report: three in `ADR-101-kg-changeset-model.md`, and one at `core/crates/khive-runtime/src/secret_gate.rs`,
a path this file had before a directory move. That last one is left deliberately — the sibling
allowlist is anchored to an exact path and the test suite asserts that prefixed spellings are not
exempt, so widening the anchor to cover a path the file no longer has would trade a known
constraint for an unknown one.

The two runs differ by two commits because `main` advanced between them; the finding sets are
otherwise identical apart from the three removed.

**The arms are load-bearing, not decorative.** With the allowlist removed and everything else
untouched, three of the five new arms fail and two pass. The split was written down before the run:
the two that pass are the two whose subject is not the allowlist — the neighbouring-value arm, whose
values were never exempt, and the staleness arm, which reads the source file rather than the config.
The configuration was restored from a snapshot and the restore verified by sha identity, because the
working tree held the real change at the time.

**The suite**: `python3 -m unittest scripts.tests.test_gitleaks_config -v` — 12 tests, all pass, the
7 pre-existing arms included.

**The arms are version-independent, and that took a second commit.** The first version of two of
them asserted how many findings the scanner reports and at which lines, which is a claim about the
scanner rather than about the exemption. It differs by version: under 8.30.1 both synthetic
constants are reported at every non-exempt path, under the 8.21.2 the workflow installs only one of
them is, and one superstring neighbour is not reported at all. The job's own step failed where the
host had passed:

```
test_gate_constants_are_not_exempt_at_other_paths          AssertionError: 3 != 6
test_neighbouring_values_at_the_gate_path_are_not_exempt   lines {1, 3}, expected {1, 2, 3}
```

Both arms now scan twice in the same run — once with the default rules and no exemption, to learn
what this scanner reports at all, then with this repository's config — and compare fingerprint sets.
That states the claim that matters, which is that the exemption removes the exempt path's findings
and nothing else, and it holds under either version. Checked under both. The comparison is by set
rather than by list because the report is not deterministically ordered: an intermediate version
compared lists and failed one run in three.

## Out of scope, named rather than silently skipped

Two fingerprints remain in `.gitleaksignore`, both for a hyphenated phrase in a heading in
`docs/adr/ADR-101-kg-changeset-model.md`. They have the same decay, and a different remedy: the
value there is prose, so the fix is either a wording change or an allowlist scoped to a document
rather than to a test fixture. That is a separate change and it is not urgent, because a document's
headings are edited far less often than an 8000-line source file.


ohdearquant (commented):

oceanwaves630 (approved):
Exempting these constants by value and path rather than by fingerprint is the part that makes this hold, and it is worth saying why, because the two approaches look equivalent until the scanner moves.

I measured each constant on its own against both the version installed here and the version this repo pins, rather than reading the report's `Secret` field, which is redacted and so carries no attribution. The first constant is detected by both versions as `generic-api-key`. The second is detected only by the newer one; the pinned version does not flag it at all. So the set of findings these values produce is a function of the scanner version.

A fingerprint is computed from the finding, which means it inherits that version dependence: the exemption silently stops matching when the scanner changes what it reports, and it fails open rather than loudly. A value-and-path exemption is computed from the input instead, so it keeps identifying the same two lines across both versions. That is the difference between the old form and this one, and it is the reason the local result and the CI result could disagree while both were honest.

One consequence worth recording for whoever next edits this: the install step is job-level, so the scheduled run resolves the same pinned version as the pull-request run. An exemption written against a newer scanner's behaviour would therefore be inert in both, not just in one.

PR #2964, opened 2026-09-18
fix(memory): refuse shared creation of kind=memory in the pack that owns it

### LINKED ISSUE #2684: kg `create` of a `memory` or `task` note skips the defaults the specialised verbs apply, so the same record reads differently by path

ohdearquant:

Closes #2684.

## The gap

`create(kind="note", note_kind="memory", …)` stored a row with `salience`, `decay_factor` and
`properties.memory_type` all unset, while `memory.recall` rendered that same row as episodic at the
default salience. One record, two readings: what storage holds and what the verb that exists to read
it reports.

The cause is written down in the tree, as a comment asserting an invariant nothing enforced.
`MemoryHook::prepare_create` was an explicit no-op whose body was two lines of prose:

```rust
// `memory.remember` owns the create path and validates there. This hook exists for the
// update path; adding create-side checks here would duplicate that verb's contract.
```

That is a true statement about intent and a false one about reach. Nothing stopped a caller from
reaching the kind through shared creation, and the hook that said the create path was owned
elsewhere was the one place that could have said otherwise.

## Where the refusal goes, and why that is the decision

The refusal is now that hook's body. It is not in the generic pack.

This is ADR-021's decision rather than a preference, and the amendment states it in the negative,
which is why it is worth quoting:

> The memory pack owns admission through the existing ADR-017 `KindHook::prepare_create` interface.
> KG does not hard-code memory, redirect requests, or enforce a general prohibition on kinds with
> specialized verbs.

Two things follow that a kind-name arm in the generic handler does not give you.

**One hook covers both admitting paths.** Shared `create` calls it at
`crates/khive-pack-kg/src/handlers/create.rs:381`, after parameter validation and before any note or
provenance edge is written. A `stream.batch` member of the kind calls the same hook during
preparation at `crates/khive-runtime/src/streams.rs:1126`. So one edit closes both, and the
`stream.batch` refusal is a preparation failure: no preceding sibling in that batch commits.

**The capability question disappears instead of being solved.** A refusal must never name a verb the
caller cannot dispatch. A hook exists only when the pack that owns the kind is registered, so a
pack that declares a `memory` note kind without the memory pack has no hook, is admitted, and is
never handed a message pointing at `memory.remember`. That is the same discrimination as a registry
probe, with nothing to keep in sync.

`memory.remember` itself is unaffected, and this was checked rather than assumed: it writes through
`create_note_with_decay_for_embedding_model`, a runtime primitive, and `prepare_create` does not
appear anywhere in `crates/khive-runtime/src/runtime.rs`. A refusal in the hook cannot refuse the
verb it directs callers to.

## Why the first commit is documentation

ADR-021's creation-admission amendment and ADR-007's reciprocal qualification were introduced as
proposals in #2724 on 2026-09-14, and the amendment says in its own text that its implementation
requires acceptance before it merges. So this branch accepts them first and implements second, in
that order.

The flip is not a one-line status edit, because the headings carry the anchors: dropping
"(proposed)" renames them and eight references across the two files point at them. Four of those
references are sentences whose truth changes at acceptance ("superseded only on acceptance", "on
acceptance", "requires acceptance before its implementation merges"), so they are rewritten as
claims rather than repointed. Anchor integrity was checked directly — every same-file fragment and
both cross-file links resolve to a heading that derives them — because the repository's ADR
reference lint validates titles and does not read fragments at all. That blindness is filed as
#2966; it is not fixed here.

## What this PR does not close

ADR-021 excuses two public creation routes by contract: standalone `stream.append`, which admits any
registered note kind, and approved `propose`/`review` changesets containing `AddNote`. Neither
invokes the shared create hook, so after this change they are the remaining writers that can produce
a memory row with the derived fields unset.

That residual is asserted rather than described. A `stream.append` memory row and an `AddNote`
memory row each land with `salience` and `decay_factor` unset, and recall renders them at the
effective defaults — a test that passes today and pins a known difference, so it is a number and not
an inference. #2967 carries the open question; no fix rides here.

Existing rows are not repaired, which is also the ADR's decision: absent fields are neither
backfilled nor filled on read.

## Coverage

Thirteen arms: two pin the placement, ten are the amendment's own acceptance list, and one was
already there.

**The placement.**

- `generic_create_refuses_the_memory_kind_and_names_its_writer` — the refusal itself, with an
  ordinary `observation` note created in the same request shape as the control, and
  `memory.remember` then shown to still write a row whose three fields are all set. Closing an entry
  point is only correct if the remaining one still holds the property.
- `generic_create_accepts_the_memory_kind_when_its_owning_pack_is_not_registered`, in
  `khive-pack-kg` — the over-reach direction. A fixture pack declares a `memory` note kind and
  registers no hook, so the kind is admitted and no caller is handed a refusal naming a verb it
  cannot dispatch. It asserts `find_kind_hook("memory").is_none()` first, so it cannot silently stop
  exercising that branch.
- `search_kind_memory_routes_to_note_substrate_via_registry` — not new, and not about creation. It
  is the arm that reddened when an earlier version of this change keyed the refusal on the kind name
  alone, which is why it belongs in this list.

**The amendment's acceptance list**, one arm per clause:

- Both generic spellings refuse: `create(kind="memory", …)` as well as
  `create(kind="note", note_kind="memory", …)`.
- A refusal changes nothing — no note, no provenance edge, and an existing key holder still resolves
  to the same id after a refused keyed request aimed at its key.
- Keyed, explicit-namespace and self-supplied-default requests refuse too. The third matters most: a
  caller who passes `salience`, `decay_factor` and `memory_type` itself is still refused, because the
  refusal is about the contract rather than about which fields the request happens to carry.
- A `stream.batch` member of the kind aborts the batch during preparation, in atomic **and**
  per-member mode, before a sibling append ordered ahead of it commits.
- The specialized writer is visible three ways: in raw storage with its fields set, through generic
  `get`, and in a non-empty `memory.recall`.
- Identical keyed replay preserves the original record.
- Updating an existing memory still works, because the hook is create-only; and another pack's kind
  with its own hook still creates through generic `create`.
- The residual is measured rather than described: standalone `stream.append` and an approved
  `AddNote` changeset both admit the kind and leave `salience` and `decay_factor` unset, while recall
  renders that same row at the effective defaults. That arm is expected to pass — it pins a
  divergence that exists rather than guarding against one, its comment says so in as many words, and
  #2967 carries the question of whether it should ever close.

**Every counting instrument in these arms carries a control inside the same test.** The no-mutation
arm proves the note count moves under a known-positive `create` and the edge count moves under a
known-positive `link`, before asserting that neither moves under the refusal. The second of those is
deliberately the narrower claim it looks like: `create` has no edge-writing parameter at all, so that
assertion pins an untouched graph read from a counter shown to be live, not a write it could have
caught. Each batch arm first commits an identical batch without the refused member, because a read of
a stream nothing ever wrote returns zero entries too, and an assertion of zero would pass whether or
not the sibling was rolled back.

## Scope

7 files, 6 commits, +831 −48 against `main`.

`crates/khive-pack-kg/src/handlers/create.rs` is not one of them. That is the placement argument as
a measurement rather than as a sentence: the first commit added a kind-name arm to the generic
handler and the third removed it, so the end state leaves the generic pack untouched by this change.

Two of the seven files are the ADR acceptances and the §4 contract fix, one is the hook that now
carries the refusal, one is a dev-dependency the coexistence arm needs, one is the lockfile, and two
are test files.

## Verification

The workspace suite, formatter and linter on the branch as it stands, which is `main` merged in
rather than a bare branch, because the hosted checks run on the synthetic merge ref and a local run
on the unmerged branch is a strict subset of it.

```
cargo fmt --all -- --check          rc 0
cargo test --workspace --no-fail-fast   11506 passed, 0 failed, 266 suites reported a result
cargo clippy --workspace --all-targets -- -D warnings   rc 0, 0 errors
```

Three predicates beyond the totals, because a green total does not say that a named arm ran — a
rename or a filter typo reads as success:

- each of the twelve new arms printed its own `... ok` line: **12 of 12**, checked by exact name.
- the workspace meta-test that refuses pack tests reaching the config-ledger seam without
  `#[serial_test::serial(config_ledger)]`: **1 of 1**. Two of the new arms call `memory.recall` and
  reach that seam; both carry the attribute.
- **266 suites reported a result.** This one is not decoration. Two earlier runs of this same gate
  stopped at suite 16 and suite 51 of 266 on wall-clock budgets in `khive-db` and `khive-mounts`,
  neither of which this branch touches, because `cargo test` stops at the first failing suite. The
  pass count looked healthy both times while most of the tree never executed. `--no-fail-fast` plus
  a suite count is what distinguishes a complete run from a truncated one; the timing arms
  themselves are #2971 and both passed here on an unloaded machine.

The hook's placement is also asserted structurally rather than by reading the diff: the refusal is
present in the memory pack, the update hook sits on `normalize_note_update` with zero references to
the removed `prepare_note_update` seam, and `crates/khive-pack-kg/src/handlers/create.rs` does not
appear in the branch's diff against `main` at all.


ohdearquant (commented):

Verdict on head 1bc392e44e3c0ea5893a66e6ebbfaad873af988d: REQUEST-CHANGES, 2 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head e4df69468ef968d75ebbd0889c4db5cabb4286c4: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #2965, opened 2026-09-18
fix(kg): refuse kind=channel_health in generic create, naming the verb that owns it

### LINKED ISSUE #2963: generic create mints an unreachable channel_health row

`comm.heartbeat` identifies a `channel_health` row by a **deterministic** UUID:
`heartbeat_note_id(namespace, channel_kind, channel_slug)` is `Uuid::new_v5` over a key built from
those three parts, and the row is written with `insert_note_if_absent` / `replace_note_if_unchanged`
against that id. That is how one channel keeps exactly one health row and how a later heartbeat
updates it in place.

Generic `create(kind="note", note_kind="channel_health", …)` mints an ordinary random id instead, and
`comm`'s `kind_hook` returns a hook only for `message`, so nothing on the generic path knows about
the identity rule.

The result is worse than an under-validated row. It is a row that no heartbeat will ever find, no
`comm.health()` projection will ever reconcile with the real one, and no later write will ever
update — an orphan that looks like a health record and reports a channel's state forever at the
value it was created with. Two rows for one channel is a state the identity scheme exists to make
impossible.

## Why this is filed separately from the other generic-create gaps

The `memory` case was a row missing its defaults, and the `session` case is a row missing its
validation. Both are recoverable by reading the row and fixing it. This one produces a record whose
identity is wrong, which is not visible by reading the record: it looks complete, and the damage is
that a different, correct row exists or will exist beside it.

## Acceptance

Generic `create` cannot mint a `channel_health` row with an id the heartbeat path would not have
chosen. Either the kind is refused there, naming `comm.heartbeat` — the shape `scheduled_event`,
`edge` and `proposal` already have in that same match — or the identity derivation moves somewhere
both paths reach. A test creates a channel's health row through the generic path and then runs a
heartbeat for that same channel, asserting there is exactly one row afterwards.

ohdearquant:

Closes #2963.

## The gap

`comm.heartbeat` addresses a channel's health row rather than searching for it. `heartbeat_note_id`
derives a v5 UUID from the namespace, channel kind and slug, and the row is written at that id
through `insert_note_if_absent` / `replace_note_if_unchanged`. That derivation is the whole identity
scheme: it is how one channel keeps exactly one health row, and how a heartbeat five minutes later
updates the row the first one wrote instead of adding a second.

Shared `create(kind="note", note_kind="channel_health", …)` accepted the kind and minted an ordinary
random id, and a `stream.batch` write member naming the kind did the same. Nothing on either path
knows about the derivation: comm answered `kind_hook` only for `message`, so the kind reached the
shared write path with no owner between the caller and the store.

**This is not the same defect as a row with unset fields.** A row missing its defaults or its
validation is wrong and readable: you can fetch it, see the gap and repair it. This one is wrong in
its *identity*, which is invisible from the record — it has content, a kind and a timestamp, and it
looks complete. No heartbeat will ever find it and no later write will update it.

And it is not dormant. `comm.health` reads its channel list by querying notes filtered on
`kind: "channel_health"`, so an orphan row is not merely unreachable — it is **reported**, as an
additional channel, carrying whatever poll interval and outcome it was created with, permanently.
Two rows for one channel is precisely the state the derived id exists to make impossible, and the
projection that would reveal the duplicate is the one that renders it as fact.

## The change

`ChannelHealthHook::prepare_create` refuses the kind and names `comm.heartbeat`.

The refusal lives on the owning pack's hook rather than in the generic create handler, and that
placement is the decision rather than an implementation preference:

- The hook is where both admitting paths converge. Shared `create` calls it before it writes the
  note or its provenance edge, and `stream.batch` calls it while preparing a write member, so one
  refusal answers both entry points and a batch aborts before any sibling commits. A kind-name arm
  in the generic handler covers one path and has to be duplicated for the other — which is what the
  neighbouring `scheduled_event` refusal does today, once in `create` and once in the batch member
  parser.
- The generic pack does not know this kind. A hook exists only when the pack that owns the kind is
  registered, so the refusal can never reach a caller who has no `comm.heartbeat` to dispatch. The
  hook's existence *is* that capability check; nothing has to ask the registry whether the verb is
  present.

`comm.heartbeat` writes through the note store primitives rather than shared `create`, so it is
unaffected by the refusal. That is asserted rather than assumed: the arm below writes two heartbeats
for one channel after the refusal and reads exactly one row back.

The issue's other option — move the identity derivation somewhere both paths reach — is not
available here. A `KindHook` gets `&mut Value` of the create arguments, and `CreateParams` has no
`id` field, so a hook cannot tell the create path which id to write. Its `key` field is a uniqueness
key, not an identity: it makes a second row a `key_conflict` rather than making the two rows one, and
the UUID is still minted by the store. So there is no id for both writers to agree on until `create`
grows a way to accept one. The source comment records that as the falsifier for this refusal.

## Coverage

Two arms in comm's integration suite, one per admitting path:

`generic_create_refuses_the_channel_health_kind_and_names_its_writer`

- It asserts the pack registers a hook for the kind first, so it cannot silently stop exercising the
  refusing branch if the fixture changes.
- An ordinary `observation` note created in the same request shape is the control, so the refusal is
  attributable to the kind rather than to the request.
- The refusal must contain `comm.heartbeat`, because a refusal that does not name the working path
  leaves the caller nowhere to go.
- Then the invariant the refusal exists to protect: two heartbeats for one channel, and
  `comm.health()` reports exactly one channel row. Closing an entry point is only correct if the
  remaining one still holds the property, so that is asserted rather than assumed.

`stream_batch_refuses_a_channel_health_write_member_before_any_sibling_commits`

- An ordinary keyed write through the same batch shape is the control.
- A batch whose first member is an ordinary keyed note and whose second names `channel_health` is
  refused naming `comm.heartbeat`, and the first member is then shown not to have committed — the
  refusal happens in preparation, not as a per-member conflict.
- `comm.health()` reports no channels afterwards, so the batch path created no row.

**What is not covered.** The over-reach direction — that a pack declaring this kind *without* comm
loaded gets no refusal — has no arm here, because this suite's fixture registers comm. The
equivalent property is exercised for a different kind in `khive-pack-kg`, where a fixture pack
declares a note kind and registers no hook for it.

Standalone `stream.append` is a third creation route. It accepts a registered `note_kind` and dispatches
no pack hook, so it still admits this kind, as the amendment states (item 5) and as ADR-021 states for
`memory`. That is deliberately outside this change; #2974 records it as its own row, with the options.

## Scope

3 files, 6 commits, +263 −6 against `main` (two of the commits merge current `main` in; the second after #2973 landed under this branch).

- `crates/khive-pack-comm/src/pack.rs` (+60 −6) — `ChannelHealthHook`, registered for the kind
  beside the existing `MessageHook`; the validator's doc now says where `channel_health` is answered.
- `crates/khive-pack-comm/tests/integration.rs` (+160) — the two arms above.
- `docs/adr/ADR-017-pack-standard.md` (+43) — the amendment recording that a pack may refuse shared
  creation of a kind it owns through `prepare_create`, and why the generic handler is the wrong place.

`crates/khive-pack-kg/src/handlers/create.rs` is not in the diff. The first commit on this branch put
a kind-name arm and a registry probe there; the third removed both, so the generic pack ends
untouched.

## Verification

The workspace suite, formatter and linter at the current head, which has `main` merged in rather than being a bare branch, because the hosted checks run on the synthetic merge ref and a
local run on the unmerged branch is a strict subset of it.

```
cargo fmt --all -- --check                               rc 0
cargo test --workspace --no-fail-fast                    11510 passed, 0 failed, 266 suites reported a result
cargo clippy --workspace --all-targets -- -D warnings    rc 0, 0 errors
```

Predicates beyond the totals, each checked by exact test name, because a green total does not say
that a named arm ran:

- the two new arms, and the two pre-existing arms that stand as controls (`comm.heartbeat` still
  writes one row per channel; the pack still declares the kind): **4 of 4** printed their own `ok`.
- **266 suites reported a result.** `cargo test` stops at the first failing suite, so without
  `--no-fail-fast` and a suite count a truncated run and a complete one report the same way.

The first full run of this gate reddened one of the two new arms on the arm's own read: it took
`items` from a `list` response, which answers with `notes`, so its "sibling not committed" assertion
compared `None` to `Some(0)` and could not pass on any code. The property held (the page was empty).
The last commit reads the right key and adds the control that arm lacked: the same `list` over the
control batch's key must find exactly one note, so an empty page for the sibling means the sibling
was not committed rather than that the filter matched nothing.

The shape is also asserted structurally rather than read off the diff: the refusal is present in the
comm pack; the generic create handler carries no `has_verb` probe and exactly its three pre-existing
refusal arms; `crates/khive-pack-kg/src/handlers/create.rs` does not appear in the branch's diff
against `main`; and the comm arm asserts the hook is registered before it exercises the refusal.


ohdearquant (commented):

ohdearquant (commented):

Verdict on head d5520902467975228385eeb39111048b4ae44c47: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant (commented):

Verdict on head a5a2abe257d075d0670f2a66b5f89fefc444d6eb: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):

PR #2973, opened 2026-09-18
refactor(runtime): move note-update sequencing off KindHook onto the registry

### LINKED ISSUE #2956: runtime: move note-update sequencing off KindHook so a pack cannot override the order

`KindHook::prepare_note_update` sequences a note update: it runs `normalize_note_update`, then
`validate_note_update`. It is a trait method with a default body, so a pack can override it — and an
override replaces the whole sequence, including the validator.

That is not hypothetical. A pack overrode it to normalize coupled fields and thereby retired its own
`validate_note_update`, which kept working only because the override repeated the validation call
inline. Two copies of one invariant, with the reachable copy being the one nobody was looking at.

Splitting normalization into its own method fixed the immediate case and moved the override point,
but the guarantee is still only naming and test coverage. A trait default method is an extension
point by construction; documentation saying "this one is not" is a claim the language does not back.

## What should change

Move the sequencing off the trait.

- `prepare_note_update` leaves `KindHook`.
- The registry, or a free function in `khive-runtime`, calls `normalize_note_update` and then
  `validate_note_update` in that order.
- Packs implement the two halves and cannot express an ordering at all, so they cannot get it wrong.

## Scope notes

- `VerbRegistry::prepare_note_update_hook` already exists and is what canonical and atomic CRUD call,
  so the dispatch site is in place; the change is about where the ordering lives.
- `VerbRegistry::validate_note_update_hook` stays as the validation-only entry point for callers that
  hold no mutable request.
- This is a breaking change for any out-of-tree `KindHook` implementor that overrides
  `prepare_note_update`. There are none in this workspace.
- The documents describing the current shape are `docs/adr/ADR-017-pack-standard.md` and
  `docs/adr/ADR-019-gtd-pack.md`; both need amending with the move.

ohdearquant:

Closes #2956.

## What moves

`KindHook::prepare_note_update` is removed. `VerbRegistry::prepare_note_update_hook` now calls
`normalize_note_update` and then `validate_note_update` itself:

```rust
crate::curation::normalize_note_update_tags(args)?;
if let Some(hook) = self.find_kind_hook(&note.kind) {
    hook.normalize_note_update(runtime, token, note, args).await?;
    let properties = args.get("properties").filter(|value| !value.is_null());
    hook.validate_note_update(runtime, token, note, properties).await?;
}
```

Nothing about what runs, or in what order, changes. What changes is who can change it.

## Why a sequencer on the trait was the defect

The removed method sequenced two other methods on the same trait, and an override of a sequencer
replaces the whole sequence. So a pack that overrode it in order to normalize its coupled fields
retired its own validator in the same stroke, silently — the override compiles, the pack's tests of
the validator still pass when they call it directly, and the only thing that stops is the call the
generic update path makes.

That is not hypothetical. The task hook overrode it, and its dependency-cycle check remained
reachable only because the override happened to repeat the validator call inline. The earlier fix
split normalization into its own method and documented which method a pack should override, leaving
the sequencer on the trait with a note that Rust does not prevent overriding it either, so the
guarantee rests on naming and on test coverage of the sequence.

Naming and coverage are a description of what a type system does. With no sequencing method on the
trait, a pack implements the two halves and cannot express an order at all, so there is no sequence
to replace and no way to retire the validator by overriding something else. The ordering becomes a
property of the single dispatch site every generic update path already goes through.

**This is a breaking change** for an out-of-tree implementor that overrode the removed method. None
exists in this workspace. An implementor that did override it was expressing exactly the thing this
change removes, so there is no migration to offer beyond splitting the override into the two halves.

## The doctest pair, and why it is a pair

A `compile_fail` doctest on the trait pins the removal: an impl overriding the removed method must
not compile.

On its own that arm is close to worthless, and the reason is worth stating because the failure is
invisible. **A `compile_fail` doctest passes when the code fails to compile for any reason at all**
— a renamed type, a stale import, a typo in a path. Such an arm goes on reporting `ok` forever while
testing nothing, and nothing about its output says which.

So it is paired with a control that must compile: structurally identical, same imports, same
signatures, overriding the two halves a pack is meant to implement instead of the removed method.
The two arms differ in exactly one thing, which is the property under test. If the control ever
stops compiling, the arm beside it has stopped being about the method and the suite says so.

That is not hypothetical either: **the control failed on its first run.** Both arms had omitted the
trait's two required methods and its `Debug` bound, so the `compile_fail` arm was passing for a
missing impl rather than for the removed method, and nothing in a green `compile_fail` result could
have said so. Both arms now implement what the trait requires, and the surviving arm is pinned by
mutation: deleting the override of the removed method from the `compile_fail` block makes that
snippet compile, and the arm turns red.

## Tests

Two arms in `khive-runtime` already covered the sequence through the dispatcher; they are unchanged
in substance and their module header now says the probe overrides the only two halves that exist.
One of them carries a recorded mutation result: with the sequence reduced to a single
`validate_note_update` call, the normalizer does not run, its marker never reaches `properties`, the
validator observes `None` where it expects `Some(true)`, and the call returns `Ok` where the arm
expects `Err`.

The two task-hook arms that pin this pack's normalization called the hook method directly, and
**they could not simply be renamed onto the new method.** Their own comments say what they are for:
the dispatch site every generic update path calls must run the normalizer, or the function is
correct and unreachable. Since that site is now the registry, pointing them at
`KindHook::normalize_note_update` would have left them asserting only that the function does what it
does — which is the unreachable state the comment names as the thing to rule out. They go through
`VerbRegistry::prepare_note_update_hook` on a registry holding the pack, which is the path a caller
takes.

## Documentation

ADR-017 gains an amendment recording the move, which supersedes the earlier sequencing paragraph
where the two disagree; its trait sketch drops the removed method, leaving the two halves a pack
implements. The method-level docs on both halves now point at the registry for the ordering rather
than at a sibling method.

ADR-019 described twice when the task hook's validator runs, both times as "through the trait's
sequencing". Those two clauses now name the registry's. They are the only statements in the ADR
set that name the trait as the sequencer.

## Scope

5 files, 2 commits, +181 −64 against `main`, which the branch sits directly on at its current head.

- `crates/khive-runtime/src/pack.rs` (+123 −41) — the method removed, the registry sequencing the two
  halves, the doctest pair, and both halves' docs repointed at the registry.
- `crates/khive-pack-gtd/src/hook.rs` (+31 −7) — the two tests moved onto the registry, and the
  helper that builds one holding the pack and the pack it requires.
- `docs/adr/ADR-017-pack-standard.md` (+13 −14) — the amendment, and the trait sketch losing the
  removed method.
- `docs/adr/ADR-019-gtd-pack.md` (+2 −2) — the two clauses that named the trait as the sequencer.
- `CHANGELOG.md` (+12) — the removal recorded under Unreleased, with the migration for an
  out-of-tree implementor and why there is no compatibility path.

## Verification

The workspace suite, formatter and linter at the code commit, which sits directly on current
`main`, so the local run covers the same tree the hosted checks build on the merge ref. The second
commit touches only `CHANGELOG.md`.

```
cargo fmt --all -- --check                               rc 0
cargo test --workspace --no-fail-fast                    11508 passed, 0 failed, 266 suites reported a result
cargo clippy --workspace --all-targets -- -D warnings    rc 0, 0 errors
```

Predicates beyond the totals, each checked by exact test name, because a green total does not say
that a named arm ran:

- the four sequencing arms (the two in `khive-runtime` through the dispatcher, the two task-hook
  arms through the registry) each printed its own `ok`: **4 of 4**.
- the `KindHook` doctest pair, the `compile_fail` arm and its must-compile control: **2 of 2**.
  `khive-runtime`'s doctest count moved from 10 to 12.
- **266 suites reported a result.** `cargo test` stops at the first failing suite, so without
  `--no-fail-fast` and a suite count a truncated run and a complete one report the same way.

The shape is also asserted structurally rather than read off the diff: the registry sequencer is
defined once; the only `fn prepare_note_update(` left in the tree is the one inside the
`compile_fail` doctest; `normalize_note_update` is called from one place, the registry sequencer;
and `validate_note_update` from two, the sequencer and the validation-only
`validate_note_update_hook` seam that predates this change.


ohdearquant (commented):

Verdict on head 9e440264dc27b322462e10b54d5be1e497b1c6c7: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
Recorded the removal in CHANGELOG.md under Unreleased / Removed at 546fb10a: what was removed, the two methods an implementor moves its logic into, and why there is no compatibility path (an override of the sequencer is the behaviour being removed, so a shim would reintroduce it). The trait's public surface otherwise carries no deprecation window in this repository; the entry is the record.

ohdearquant (commented):

oceanwaves630 (approved):
Approving on my own read at 546fb10a, under the operator instruction 5702882e which names this head.

What I checked: the automated review at this head is clean with zero blocking findings; the delta from the previous head is the CHANGELOG entry that was the outstanding request; hosted CI at this head is 23 passing and 6 skipped with no failures; the merge state is clean against the base.

One open issue is minor and documentary: the ADR describes the normalize-then-validate ordering as a type-system property, while the ordering is in fact enforced by the registry hook. It is recorded to ride a later documentation change and is not a condition of this approval.

This is a reading, not a stamp: the evidence above is re-derivable from the PR and its checks.

PR #2978, opened 2026-09-20
fix(kg): project the update echo like every other note-returning verb

### LINKED ISSUE #2977: update() echo of a task note reports the note lifecycle as its top-level status, unlike get/create/list

`update(id=<task>, properties={...})` returns a note record whose top-level `status` is the note
lifecycle value (`active`) while `properties.status` carries the task state (`next`). Every other
verb that returns the same note (`create`, `get`, `list`, `restore`) passes the record through
`remap_note_status`, which lifts the task state to the top level and adds `display_name` and the
lifted `tags`; `update` does not.

Where: `crates/khive-pack-kg/src/handlers/update.rs`, the `KindSpec::Note` branch of
`handle_update` builds its response with `normalize_entity_timestamps(to_json(&note)?)` alone.
The restore branch a few hundred lines below wraps the same expression in `remap_note_status`, as do
`create.rs`, `get.rs`, `list.rs` and `note_list.rs`.

Observed (main 96c04f9a): after `update(id="<task in next>", properties={"transition_note": "..."})`
with no status field, the echo reads `"status": "active"` at the top level beside
`"properties": {"status": "next", ...}`; `get` on the same id reads `"status": "next"` and
`"lifecycle": "active"`.

Why it matters: the echo of a write is what a caller reads back instead of re-querying, and here it
reports a lifecycle transition that was not requested and did not happen. It reads in the direction
that looks like a successful state change. A caller checking their own write spends a round chasing
a status write that never occurred.

Fix: apply `remap_note_status` on the update echo, matching the other four verbs, with a test that
updates a task in `next` through the generic verb and asserts the echo's top-level `status` equals
the `get` read of the same row.

ohdearquant:

Closes #2977.

## The gap

The generic `update` verb returned a note's raw serialization as its echo. Every other verb that
returns a note record (`create`, `get`, `list`, `restore`) passes it through `remap_note_status`,
which lifts a task's `properties.status` to the top level, adds `display_name`, and lifts
`properties.tags`. So updating a task in `next` echoed `"status": "active"` (the note lifecycle)
beside `"properties": {"status": "next"}`, while `get` on the same row read `"status": "next"`.
The echo of a write is what a caller reads instead of re-querying, and this one reported a
lifecycle transition that was neither requested nor made.

## The change

`handle_update`'s note branch wraps its response in `remap_note_status`, the same call the restore
branch of the same file and the other four verbs already make. No other behaviour changes: the
`unchanged` marker and the embedding-truncation warning are added on top as before.

## Coverage

`generic_update_echo_reports_the_task_state_that_get_reports` (task pack integration suite):
assigns a task, transitions it to `next`, updates it through generic `update` with a
`transition_note` property, and asserts the echo's top-level `status` and `display_name` equal the
`get` read of the same row, with the read's `status == "next"` as the control. Reverting the
one-line change turns the arm red (echo `active` vs read `next`).

## Scope

Branch diff against main: 2 files, 1 commit (numbers from the API at open time).


oceanwaves630 (approved):
Read at ae8f6259 (two files). The update handler's note branch now passes its echo through the same projection the other note-returning verbs use, so a task's top-level status and display name in the echo match what a subsequent get returns. The integration test drives a task to next, updates a property through the generic verb, and asserts the echo equals the read with the read's own state as the control.

PR #3005, opened 2026-09-21
comm: seek the outbound recipient index in the outbox scan

### LINKED ISSUE #1859: outbox loop: pending email notes silently starve out of the newest-first selection window
**Defect.** The outbound email delivery loop selects candidate notes with a fixed-size, newest-first page. A pending email note that fails delivery long enough (for any reason — see #1853 for one such reason) is silently pushed out of the selection window by newer outbound traffic and is never scanned again. Retry log lines stop, no terminal state is recorded, and the note simply disappears from the loop's view while remaining live, eligible, and undelivered in the store.

**Mechanism** (`crates/khive-mcp/src/serve.rs`, `channel_outbox_loop`): the per-tick selection is

```json
{"kind": "message", "direction": "outbound", "delivered": false, "limit": 200}
```

with results ordered newest-first, and the `email:` recipient-prefix filter applied per-note *after* the page is fetched. Two properties make the window shrink monotonically past any stalled note:

1. The pool matched by `direction=outbound, delivered=false` includes every locally-addressed outbound message (actor-to-actor messages are never marked delivered), so the population grows without bound as messaging traffic accumulates. Email-addressed notes are a small minority of the page.
2. `limit: 200` newest-first means the scan covers only the 200 most recent outbound notes. Once 200 newer outbound notes exist, an older pending email note is unreachable by the loop — permanently, since the pool never shrinks.

**Observed live.** A pending outbound email note (created 12:49:53Z) was skipped with a claim-failure WARN every ~6s tick (the #1853 defect) until 17:03:15Z, then the WARN lines ceased — while the note remained live, undelivered, unclaimed, and eligible (read back directly: no `external_id`, no `delivered` flag, `updated_at` unchanged since creation). At verification time the 200th-newest row of the loop's exact selection was created 12:56:26Z: the window cutoff had moved past the stalled note's creation time. Roughly 200 outbound notes in ~5 hours of ordinary traffic was enough to starve it.

**Consequences.** (a) Any email note that cannot deliver within the traffic-dependent window (~hours under ordinary load, minutes under heavy load) is stranded forever with zero signal — the observable symptom (retry log lines stopping) reads as recovery. (b) The failure interacts badly with #1853's acceptance criterion (c): a visible failed/parked state is unreachable for a note the loop can no longer see. (c) Selection cost grows with total message volume rather than pending email volume.

**Acceptance criterion.** (a) Selection is scoped to the actual delivery population — email-addressed, undelivered notes — server-side (filter or dedicated queue/index), so unrelated local messaging traffic cannot displace pending email work. (b) Coverage is unbounded in age: a pending email note is scanned every tick regardless of how many newer outbound notes exist; a test creates a pending email note, inserts more than the page size of newer outbound local notes, and asserts the loop still selects the email note. (c) If any bounded page remains, overflow is loud: the loop logs when the candidate population exceeds what it scanned.

ohdearquant:

The channel delivery loops' outbox scan (`list_undelivered_outbound_messages`) applied the channel prefix on `to_actor` in Rust, after fetching an unordered page of up to 10,000 rows matching the pending predicate. Every actor-to-actor outbound message satisfies that predicate forever (nothing marks those delivered), so on a store with more than 10,000 such rows the page fills with rows the channel can never use and the channel's own rows are never reached. Outbound email stopped leaving the box; no log line, no delivery state on the rows.

## What changes

- **`FilterOp::TextStartsWithIndexed`** (khive-storage, khive-db): a prefix predicate rendered as an index-seekable half-open range over the plain `json_extract` expression (`expr >= prefix AND expr < next(prefix)`, `next` = last code point incremented, trailing `char::MAX` dropped, surrogate gap skipped). Non-text values sort outside the text range and never match; an empty prefix admits every text value.
- **`idx_comm_message_outbound_recipient`** (comm pack schema plan): `(namespace, kind, direction, to_actor, created_at DESC, id ASC) WHERE deleted_at IS NULL`, so the scan is a seek on direction plus the recipient range.
- **The scan** puts the channel prefix into the statement and orders `created_at DESC, id ASC` in SQL under the scan cap, so the cap bounds only the channel's own pending rows and a future predicate miss still surfaces the newest rows first. The Rust prefix check and the in-memory sort are gone; the `next_attempt_at` deadline check stays in Rust.

## Tests

- `list_undelivered_outbound_messages_email_row_past_the_scan_window` (khive-runtime): 10,050 pending actor-to-actor rows, then one older email row with a later rowid and a recipient that collates after the fillers, so it sorts past the cap in every candidate index order. Red against the pre-fix code (assertion `left: []`, `right: [<email row id>]`, the four pre-existing scan tests green in the same run), green with the fix. `list_undelivered_outbound_messages_matches_legacy_predicate` (under the cap) stays as the control.
- `text_starts_with_indexed_matches_prefix_only` and `text_prefix_upper_bound_increments_the_last_code_point` (khive-db): both neighbours of the prefix rejected, missing/null/numeric fields never match, empty prefix admits text only.
- `comm_filter_plan_tests`: `outbox_filter_plan_seeks_the_outbound_recipient_index` asserts the plan seeks the new index; `outbox_filter_work_is_bounded_by_the_channels_own_rows` asserts 10,000 foreign pending outbound rows change neither the ids nor the VM step count. The former no-temporary-b-tree assertion is replaced: ordering after a range seek needs a sort, and that sort covers only the channel's own pending rows.

## Verification

Workspace gate on the head commit: `cargo fmt --check`, `cargo check --workspace --all-targets`, `cargo test --workspace`, `cargo clippy --workspace --all-targets -D warnings`.

Result at the fix commit: fmt, check and clippy exit 0; the workspace test run reported 10378 passed, 1 failed, 266 suites. The one failure was the pre-analyzed upgrade fixture in `comm_filter_plan_tests`, which pins the comm index catalog at four entries; the second commit counts the new index (five) and the `comm_filter_plan_tests` module re-run reports 11 passed, 0 failed.

Regression witness: the new runtime test is red on the pre-fix code (`list_undelivered_outbound_messages_email_row_past_the_scan_window ... FAILED`, assertion `left: []`, `right: [<email row id>]`, the four pre-existing scan tests green in that run) and green with the fix.

Files: 7 · Commits: 2

Closes #1859


oceanwaves630 (approved):
Reviewed the scan change and the regression test; approving.

